PR #13, opened 2015-07-24
do not allow to set Request.ID different than Operation.ID

SergeyKanzhelev:
1. Make Request.ID a proxy for Context.Operation.ID
2. Add CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,39 @@
+# Changelog
+
+This changelog will be used to generate documentation on [release notes page](http://azure.microsoft.com/en-us/documentation/articles/app-insights-release-notes-dotnet/).
+
+## Version vNext
+
+- TrackDependency will produce valid JSON when not all required fields were specified.
+- Redundant property ```RequestTelemetry.ID``` is now just a proxy for ```RequestTelemetry.Operation.Id```.
+
+## Version 1.2
+- First version shipped from github
+
+## Version 1.1
+
+- SDK now introduces new telemetry type ```DependencyTelemetry``` which contains information about dependency call from application
+- New method ```TelemetryClient.TrackDependency``` allows to send information about dependency calls from application
+
+## Version 0.17
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,36 @@
+# Changelog
+
+This changelog will be used to generate documentation on [release notes page](http://azure.microsoft.com/en-us/documentation/articles/app-insights-release-notes-dotnet/).
+
+## Version 1.2
+
+- TrackDependency will produce valid JSON when not all required fields were specified.
+- Redundant property ```RequestTelemetry.ID``` is now just a proxy for ```RequestTelemetry.Operation.Id```.
```
abaranch: Will your changes be part of 1.3?

SergeyKanzhelev: I should have spelled it out as vNext so we can put the version later. Will fix it

src/Core/Managed/Core.Net46/Extensibility/Implementation/External/OperationContextData.cs
```diff
@@ -57,38 +57,17 @@ public string Name
...
-            get { return this.tags.GetTagValueOrNull(ContextTagKeys.Keys.OperationParentId); }
-            set { this.tags.SetStringValueOrRemove(ContextTagKeys.Keys.OperationParentId, value); }
-        }
-
-        public string RootId
-        {
-            get { return this.tags.GetTagValueOrNull(ContextTagKeys.Keys.OperationRootId); }
-            set { this.tags.SetStringValueOrRemove(ContextTagKeys.Keys.OperationRootId, value); }
-        }
-
```
JakubOleksy: These are breaking changes...so only 2.0?

SergeyKanzhelev: these properties were not public. I reverted this change anyway as those files are autogenerated.

JakubOleksy: I see. So the public contract did not change? We can change it, but if we do, we need to go to 2.0.

abaranch:
:shipit:

JakubOleksy:
:shipit:

PR #15, opened 2015-08-05
Support PCL 5.0 and Net46

MeniZalzman:
Adding the support for UWP flavor in core
Adding the support for Net46 flavor in core


SergeyKanzhelev:
In this PR you have many projects that overrides OutputPath like this:

```
<PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|ARM'">  
  <OutputPath>bin\ARM\Release\</OutputPath> 
```

This may prevent files from being placed into output directory so you wouldn't find them in build drop and cannot run tests. Can you please clean those up.

SergeyKanzhelev:
Mentioning @MeniZalzman and @izikl to make sure you'll see this comment

PR #17, opened 2015-08-10
Add user context auth id field

Assaf-Neufeld:


src/Core/Managed/Net45/Extensibility/Implementation/External/UserContextData.cs
```diff
@@ -77,6 +77,15 @@ public string StoreRegion
         {
             get { return this.tags.GetTagValueOrNull(ContextTagKeys.Keys.UserStoreRegion); }
             set { this.tags.SetStringValueOrRemove(ContextTagKeys.Keys.UserStoreRegion, value); }
+        }
+
+        /// <summary>
+        /// Gets or sets the AuthUserId of an application-defined account associated with the user.
+        /// </summary>
+        public string AuthUserId
```
SergeyKanzhelev: Name is bad. @JoshuaWeber, can you please help with the name here?

SergeyKanzhelev:
there are also UWP and 4.6 projects recently merged by @MeniZalzman . Make sure you've updated them as well

src/Core/Managed/Net45/Extensibility/Implementation/External/UserContextData.cs
Assaf-Neufeld: This is the name of the field in the BOND schema. I can't/shouldnt modify it here

SergeyKanzhelev: This is a public class we expose to customers and we need to make sure that names are user friendly and well-explained. It doesn't matter what are the names of the actual schema fields

Assaf-Neufeld: This is the generated class.
I missed updating the public "UserContext" class. I've updated the PR with the additional change.  I named the property "AuthenticatedUserId" - does that sound good to you?

SergeyKanzhelev: Yes. Makes sense now

abaranch:
:shipit:

PR #18, opened 2015-08-12
Create an OperationTelemetry base abstract class for RequestTelemetry…

SergeyKanzhelev:
… and DependencyTelemetry

### DIFF
--- a/src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
+++ b/src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
@@ -16,12 +16,13 @@
     /// passing an instance of the <see cref="RequestTelemetry"/> class to the <see cref="TelemetryClient.TrackRequest(RequestTelemetry)"/> 
     /// method.
     /// </remarks>
-    public sealed class RequestTelemetry : ITelemetry, ISupportProperties
+    public sealed class RequestTelemetry : OperationTelemetry, ITelemetry, ISupportProperties
     {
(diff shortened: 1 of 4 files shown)


src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
```diff
@@ -35,7 +36,7 @@ public RequestTelemetry()
             // Initialize required fields
             this.Id = WeakConcurrentRandom.Instance.Next().ToString(CultureInfo.InvariantCulture);
             this.ResponseCode = "200";
-            this.Success = true;
+            this.Success = null;
```
abaranch: We either need to remove ResponseCode initialization or return true back

src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// Gets the object that contains contextual information about the application at the time when it handled the request.
+        /// </summary>
+        public abstract TelemetryContext Context { get; }
+
+        /// <summary>
+        /// Gets or sets the value that defines absolute order of the telemetry item.
+        /// </summary>
+        public abstract string Sequence { get; set; }
```
abaranch: Isn't that deprecated property?..

abaranch:
Approved with suggestions

JakubOleksy:
:shipit:

PR #20, opened 2015-08-13
Add Cloud Context and deprate RoleName and RoleInstance on device

SergeyKanzhelev:
This is a beginning of Cloud context that will be extended in future. Long wished moving server properties out of Device.


abaranch:
:shipit:

JakubOleksy:
:shipit:

PR #21, opened 2015-08-13
Initialize method on TelemetryClient

SergeyKanzhelev:
Make Initialize method on TelemetryClient public. We are using Initialize method on multiple occasions when we need to initialize item on a thread, different from one that Track will be called from. For instance we are doing it for DependencyTelemetry. We are also doing it for UnhandledException logic when we use a separate channel to ensure delivery of unhandled exception.

In general having this method public simplifies our life without confusing people or making any difficulties for developers to understand the code.

### DIFF
--- a/src/Core/Managed/Shared/TelemetryClient.cs
+++ b/src/Core/Managed/Shared/TelemetryClient.cs
@@ -337,35 +337,11 @@
                     return;
                 }
 
-                var telemetryWithProperties = telemetry as ISupportProperties;
-                if (telemetryWithProperties != null)
-                {
-                    if (this.Channel.DeveloperMode.HasValue && this.Channel.DeveloperMode.Value)
-                    {
-                        if (!telemetryWithProperties.Properties.ContainsKey("DeveloperMode"))
-                        {
-                            telemetryWithProperties.Properties.Add("DeveloperMode", "true");
-                        }
-                    }
+                this.Initialize(telemetry);
 
-                    Utils.CopyDictionary(this.Context.Properties, telemetryWithProperties.Properties);
-                }
-
-                telemetry.Context.Initialize(this.Context, instrumentationKey);
-                foreach (ITelemetryInitializer initializer in this.configuration.TelemetryInitializers)
+                if (string.IsNullOrEmpty(telemetry.Context.InstrumentationKey))
                 {
-                    try
-                    {
-                        initializer.Initialize(telemetry);
-                    }
-                    catch (Exception exception)
-                    {
-                        CoreEventSource.Log.LogError(string.Format(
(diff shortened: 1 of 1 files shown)


kartang:
Looks good!

upendras:
:shipit:

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -325,56 +325,65 @@ public void Track(ITelemetry telemetry)
...
-                    Utils.CopyDictionary(this.Context.Properties, telemetryWithProperties.Properties);
-                }
+            if (string.IsNullOrEmpty(instrumentationKey))
+            {
+                instrumentationKey = this.configuration.InstrumentationKey;
+            }
+
+            if (string.IsNullOrEmpty(instrumentationKey))
+            {
+                return;
```
abaranch: I wonder if we can add error message here as well

abaranch:
I love it :shipit:

src/Core/Managed/Shared/TelemetryClient.cs
SergeyKanzhelev: in windows store case it is a normal flow to not have instrumentation key

JakubOleksy: Why does this have to be on TelemetryClient? Wouldn't it make sent to put it something else so as to not clutter the public surface? (You can pass a TelemetryClient in as a parameter)

SergeyKanzhelev: ```TelemetryClient``` has a private ``TelemetryConfiguration```. We do not want to expose it as it will create confusion - many clients shares the same telemetry configuration object and changing one you wouldn't expect that another will be changed as well.

Now, when you want to build extension for TelemetryClient - we found out that we need to have access to telemetry initializers related to this client. So instead of exposing configuration for telemetry client we propose to expose this ```Initialize``` method.

For example , we are building Operation API. We want to build it as an extension for TelemetryClient. For operation API we need to initializer operation on the thread where operation was started. So we need to call initializers. If we will put operations inside the Core package - it will not be a problem. But before we'll do it - this is a nice solution for the problem.

JoshuaWeber:
So what does the Initialize method do? It triggers the TelemetryInitializers?

SergeyKanzhelev:
@JoshuaWeber, yes it Initializes telemetry item by reading all ```'TelemetryClient.Context``` properties and applying all telemetry initializers

JakubOleksy:
:shipit: 

I think we need to figure out the right architecture of this going forward, but fine for now.

SergeyKanzhelev:
@JoshuaWeber , I'll merge it for now and than we can discuss it offline.

PR #23, opened 2015-08-16
Add ISupportSampling and implement it on Telemetry types that support it.

upendras:


upendras:
@SergeyKanzhelev, @vitalyf007, @JakubOleksy please have a look.

src/Core/Managed/Shared/Extensibility/Implementation/Telemetry.cs
```diff
@@ -9,6 +9,14 @@ internal static class Telemetry
         public static void WriteEnvelopeProperties(this ITelemetry telemetry, IJsonWriter json)
         {
             json.WriteProperty("time", telemetry.Timestamp);
+
+            var samplingSupportingTelemetry = telemetry as ISupportSampling;
+
+            if ((samplingSupportingTelemetry != null) && (samplingSupportingTelemetry.SamplingPercentage < 100.0 - 1.0E-12))
```
SergeyKanzhelev: - check that it is more than 0, discard if it less than zero

src/Core/Managed/Shared/Constants.cs
```diff
@@ -18,5 +18,10 @@ internal class Constants
         internal const string EventSourceGroupTraitKey = "ETW_GROUP";
 
         internal const int MaxExceptionCountToSave = 10;
+
+        /// <summary>		
+        /// Sampling percentage to use if none is specified.		
+        /// </summary>		
+        internal const double DefaultSamplingPercentage = 100.0;
```
SergeyKanzhelev: other constants doesn't have comment. Can omit it here as well.

SergeyKanzhelev:
:shipit: with the change of condition in serialization

vitalyf007:
Looks good to me.

JakubOleksy:
:shipit:

PR #29, opened 2015-08-18
Make ISupportSampling public and add unit tests for SamplingPercentage Serialization

upendras:
Since Internals are not visible to other SDKs which need to make use of SamplingPercentage property, making ISupportSampling public.

@SergeyKanzhelev, @vitalyf007, @JakubOleksy  please have a look.


JakubOleksy:
:shipit:

vitalyf007:
I know its too late but this opens SamplingPercentage property settable by the customer. What do you think customer would expect to happen to an item where they set SamplingPercentage=10?

Issue #26, opened 2015-08-17
Allow to report property's values longer than 1024 symbols


SergeyKanzhelev:
Data platform should be capable to accept property values up to 8 * 1024 symbols. However SDK sanitize values with the limit of 1024 symbols. See [Property.cs#L67](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/1648ecd5bf32cc151557d15cbb0886cb86e84270/src/Core/Managed/Shared/Extensibility/Implementation/Property.cs#L67)

PR #32, opened 2015-08-23
increase property allowed value length

SergeyKanzhelev:
Fixing #26.


SergeyKanzhelev:
@JakubOleksy @izikl , can you pls take a look? We need this increased property length. May it be an issue for any other channel for devices?

JakubOleksy:
:shipit: 

Looks ok to me. I assume the endpoint will accept this length?

SergeyKanzhelev:
Yes, looked at the code and actually tested it =).

PR #37, opened 2015-08-25
Remove UWP & .Net46 projects from build.

upendras:


SergeyKanzhelev:
We need to remember to restore UWP in nuspec of pre-release SDK

PR #38, opened 2015-09-01
If iKey is provided only in telemetry item context this item is dropped

### LINKED ISSUE #34: If iKey is provided only in telemetry item context this item is dropped
TelemetryClient checks iKey in its own Context and configuration. If it is empty in both places item is dropped. But telemetry item may have instrumentation key defined on the item level which should be checked as well.

SergeyKanzhelev:
Fix #34

### DIFF
--- a/src/Core/Managed/Shared/TelemetryClient.cs
+++ b/src/Core/Managed/Shared/TelemetryClient.cs
@@ -334,6 +334,11 @@
 
                 if (string.IsNullOrEmpty(instrumentationKey))
                 {
+                    instrumentationKey = telemetry.Context.InstrumentationKey;
+                }
+
+                if (string.IsNullOrEmpty(instrumentationKey))
+                {
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -325,18 +325,6 @@ public void Track(ITelemetry telemetry)
...
-                string instrumentationKey = this.Context.InstrumentationKey;
-
-                if (string.IsNullOrEmpty(instrumentationKey))
-                {
-                    instrumentationKey = this.configuration.InstrumentationKey;
-                }
-
-                if (string.IsNullOrEmpty(instrumentationKey))
-                {
-                    return;
```
abaranch: Is that done this way instead of checking telemetry item because we assume that IKey can be set by telemetry initializer?

SergeyKanzhelev: Exactly

abaranch: Can we still have this check before passing item to the channel?

abaranch:
:shipit:

PR #42, opened 2015-09-24
Features/docker support

Yonisha:
Adding support for Dockerized apps.
The Docker context is commonly used for monitoring and diagnostics.

When application is running within a Docker container, telemetries lose their host context, i.e. device ID, which is represented with an integer within a Docker container.
Moreover, the developer cannot know the Docker context in which the telemetry was sent: host machine, image name, container name and container ID.


SergeyKanzhelev:
@Yonisha, Some feedback

- please use develop branch, not in master
- we do not include platform-specific initializers into core package. You'll need to move it to WindowsServer package
- Prefer telemetry initializers to [context initializers](http://apmtips.com/blog/2015/06/09/do-not-use-context-initializers/)

SergeyKanzhelev:
Rejecting this pull request. Please address my feedback.

Issue #45, opened 2015-09-28
When telemetry channel null - throw user friendly error


SergeyKanzhelev:
Error described [here](https://github.com/Microsoft/ApplicationInsights-Home/issues/23) should be more informative:

Hello,
 after updating Application-Insights.Web from v1.1 to 1.2 we get the following error several times at runtime in our Azure cloud Service (others too http://stackoverflow.com/questions/31993394/application-insights):

[NullReferenceException]: Object reference not set to an instance of an object.
 at Microsoft.ApplicationInsights.TelemetryClient.Track(ITelemetry telemetry)
 at Microsoft.ApplicationInsights.TelemetryClient.TrackException(ExceptionTelemetry telemetry)
 at Microsoft.ApplicationInsights.TelemetryClient.TrackException(Exception exception, IDictionary 2 properties, IDictionary 2 metrics)
 at System.Web.Http.ExceptionHandling.ExceptionLogger.LogAsync(ExceptionLoggerContext context, CancellationToken cancellationToken)
 at System.Web.Http.ExceptionHandling.CompositeExceptionLogger.LogAsync(ExceptionLoggerContext context, CancellationToken cancellationToken)
 at System.Web.Http.HttpServer.d__0.MoveNext()
 --- End of stack trace from previous location where exception was thrown ---
 at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
 at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
 at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
 at System.Web.Http.WebHost.HttpControllerHandler.d__0.MoveNext()
 --- End of stack trace from previous location where exception was thrown ---
 at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
 at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
 at System.Web.TaskAsyncHelper.EndTask(IAsyncResult ar)
 at System.Web.HttpApplication.CallHandlerExecutionStep.System.Web.HttpApplication.IExecutionStep.Execute()
 at System.Web.HttpApplication.ExecuteStep(IExecutionStep step, Boolean& completedSynchronously)

How can we avoid this?

PR #51, opened 2015-09-30
friendly message when channel is null

SergeyKanzhelev:
fixing #45

### DIFF
--- a/src/Core/Managed/Shared/TelemetryClient.cs
+++ b/src/Core/Managed/Shared/TelemetryClient.cs
@@ -334,6 +334,11 @@
 
                 telemetry.Sanitize();
 
+                if (this.Channel == null)
+                {
+                    throw new InvalidOperationException("Telemetry channel should be configured for telemetry client before tracking telemetry.");
+                }
+
                 this.Channel.Send(telemetry);
 
                 if (System.Diagnostics.Debugger.IsAttached)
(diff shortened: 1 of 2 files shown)


src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -334,6 +334,11 @@ public void Track(ITelemetry telemetry)
 
                 telemetry.Sanitize();
 
+                if (this.Channel == null)
```
abaranch: Why this is not the first check we do after IsEnabled?

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -360,7 +365,7 @@ public void Initialize(ITelemetry telemetry)
             var telemetryWithProperties = telemetry as ISupportProperties;
             if (telemetryWithProperties != null)
             {
-                if (this.Channel.DeveloperMode.HasValue && this.Channel.DeveloperMode.Value)
+                if ((this.Channel != null) && (this.Channel.DeveloperMode.HasValue && this.Channel.DeveloperMode.Value))
```
abaranch: Why would we try to enrich data that we will drop? Let's remove this check and throw before we get here

SergeyKanzhelev:
The reason doing so is to enable VS integration in future when we will debug output it before throwing an exception.

abaranch:
:shipit:

PR #53, opened 2015-10-01
Sergkanz/operations

SergeyKanzhelev:
I'm adding this properties to start using them in web NuGet for end-to-end correlation. Marked as non browseable as UI not fully support it yet.


abaranch:
:shipit:

PR #58, opened 2015-10-07
remove RequestTelemetry.Id

SergeyKanzhelev:
Removing ```RequestTelemetry.Id```. Instead ```RequestTelemetry.Context.Operation.Id``` will be used. Also automatic id generation was turned off as long term we will decide whether to use CorrelationVector or Id depending on the context.


SergeyKanzhelev:
cc @abaranch

abaranch:
:shipit:

Issue #61, opened 2015-10-08
Max event name is 512, not 1024


SergeyKanzhelev:
Need to change Sanitize method for EventTelemetry so event with the name longer than 512 symbols wouldn't be rejected by endpoint

PR #64, opened 2015-10-09
set limit for event name to be 512

SergeyKanzhelev:
Fixing https://github.com/Microsoft/ApplicationInsights-dotnet/issues/61


abaranch:
:shipit:

PR #65, opened 2015-10-12
first version of call context based operation

SergeyKanzhelev:
Implemented CallContext-based operation tracking API. This API will only be available in FW40, FW 4.5 and FW 4.6 and not currently available for WRT, WP8 and PCL. Next iteration will bring [AsyncLocal](https://msdn.microsoft.com/en-us/library/dn906268.aspx)-based implementation

### DIFF
--- a/src/Core/Managed/Operation.CC.Shared/Extensibility/Implementation/CallContextBasedOperationCorrelationTelemetryInitializer.cs
+++ b/src/Core/Managed/Operation.CC.Shared/Extensibility/Implementation/CallContextBasedOperationCorrelationTelemetryInitializer.cs
@@ -1,13 +1,12 @@
-﻿namespace Microsoft.ApplicationInsights.Operation
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation
 {
-    using Extensibility.Implementation;
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Extensibility;
 
     /// <summary>
     /// Telemetry initializer that populates OperationContext for the telemetry item based on context stored in CallContext.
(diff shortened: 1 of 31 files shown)


SergeyKanzhelev:
CC: @abaranch, @vitalyf007, @JoshuaWeber, @BogdanBe

src/Core/Managed/Operation.CC.Shared/Extensibility/Implementation/CallContextBasedOperationCorrelationTelemetryInitializer.cs
```diff
@@ -0,0 +1,45 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation
+{
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.Extensibility;
+
+    /// <summary>
+    /// Telemetry initializer that populates OperationContext for the telemetry item based on context stored in CallContext.
+    /// </summary>
+    internal class CallContextBasedOperationCorrelationTelemetryInitializer : ITelemetryInitializer
```
abaranch: Do you need it to be internal?

abaranch:
:shipit:

src/Core/Managed/Operation.CC.Shared/Extensibility/Implementation/CallContextBasedOperationCorrelationTelemetryInitializer.cs
SergeyKanzhelev: I have a public extension method to add this telemetry initializer. It's good feedback however and I'll fix it in the next PR

PR #76, opened 2015-10-20
Fix application lockups when calling into app insights from Task sche…

mousetraps:
…duled

on UI thread
- It is currently possible for the long-running InMemoryTransmitter.Runner
  task to be executed (and thereby block during WaitOne) on the UI thread when
  called by a task scheduled on the UI thread. This fix guarantees that
  InMemoryTransmitter.Runner never blocks the main thread by explicitly
  specifying TaskScheduler.Default when calling StartNew.

We recently ran into this issue in Node.js Tools for VS, and managed to work around it by using TaskCreationOptions.HideScheduler, but it would be great to fix it at the source. 
* https://github.com/Microsoft/nodejstools/issues/427#issuecomment-147409676

And here's a blog post that provides a little more context on the issue:
http://blogs.msdn.com/b/pfxteam/archive/2012/09/22/new-taskcreationoptions-and-taskcontinuationoptions-in-net-4-5.aspx

### DIFF
--- a/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
+++ b/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
@@ -44,7 +44,7 @@ namespace Microsoft.ApplicationInsights.Channel
             this.buffer.OnFull = this.OnBufferFull;
 
             // Starting the Runner
-            Task.Factory.StartNew(this.Runner, TaskCreationOptions.LongRunning)
+            Task.Factory.StartNew(this.Runner, CancellationToken.None, TaskCreationOptions.LongRunning, TaskScheduler.Default)
                 .ContinueWith(
                     task => 
                     {


SergeyKanzhelev:
cc @JakubOleksy, @mihailsm 

Thanks @mousetraps! Can you please re-send this PR for develop branch.

Have you tried PersistenceChannel for nodejstools? You can find info about persistence channel on my [blog](http://apmtips.com/blog/2015/09/03/more-telemetry-channels/) and on [azure docs\(https://azure.microsoft.com/documentation/articles/app-insights-windows-desktop/#persistence-channel). I see calls to ```StartNew``` there, but it looks like all of them are not called directly from UI thread.

mousetraps:
Just resent from develop branch. In this case, it wasn't actually stemming from our app insights calls, but rather from one in a VS SCC component that we have no control over. We are indeed using PersistenceChannel for NTVS.

PR #82, opened 2015-10-21
Adaptive sampling on Sdk

vitalyf007:


src/TelemetryChannels/ServerTelemetryChannel/Shared/TelemetryChannelBuilderExtensions.cs
```diff
@@ -20,5 +20,41 @@ public static void UseSampling(this TelemetryProcessorChainBuilder builder, doub
...
+        /// <summary>
+        /// Adds <see cref="AdaptiveSamplingTelemetryProcessor"/> to the <see cref="TelemetryProcessorChainBuilder" />
+        /// </summary>
+        /// <param name="builder">Instance of <see cref="TelemetryProcessorChainBuilder"/></param>
+        /// <param name="settings">Set of settings applicable to dynamic sampling percentage algorithm.</param>
+        /// <param name="callback">Callback invoked every time sampling percentage evaluation occurs.</param>
+        /// <return>Instance of <see cref="TelemetryProcessorChainBuilder"/>.</return>
+        public static TelemetryProcessorChainBuilder UseAdaptiveSampling(
+            this TelemetryProcessorChainBuilder builder, 
+            SamplingPercentageEstimatorSettings settings,
```
SergeyKanzhelev: Instead of exposing this class we can recommend to write code like this:

``` csharp
.Use((next) =>
{
  return new AdaptiveSamplingTelemetryProcessor (next){
    Setting1 = 25;
    Setting2 = 25;
    Setting3 = 25;
  }
}
```
This extra class duplicates all the properties processor should expose as public parameters anyway. It also doesn't simplify configuration code a bit.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
```diff
@@ -0,0 +1,228 @@
...
+    /// <param name="currentSamplingPercentage">Current sampling percentage that was used by the algorithm.</param>
+    /// <param name="newSamplingPercentage">Suggested new sampling percentage that will allow to keep desired telemetry item generation rate given the volume of items states the same.</param>
+    /// <param name="isSamplingPercentageChanged">A value indicating whether new sampling percentage will be applied by dynamic sampling algorithm. New sampling percentage may not be immediately applied in case it was recently changed.</param>
+    /// <param name="settings">Dynamic sampling algorithm settings.</param>
+    public delegate void AdaptiveSamplingPercentageEvaluatedCallback(
+        double afterSamplingTelemetryItemRatePerSecond,
+        double currentSamplingPercentage,
+        double newSamplingPercentage,
+        bool isSamplingPercentageChanged,
+        SamplingPercentageEstimatorSettings settings);
```
SergeyKanzhelev: let's pass "this" instead - meaning telemetry processor istance

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
```diff
@@ -0,0 +1,228 @@
...
+    /// <summary>
+    /// Represents a method that is invoked every time sampling percentage is evaluated
+    /// by the dynamic sampling algorithm.
+    /// </summary>
+    /// <param name="afterSamplingTelemetryItemRatePerSecond">Rate of telemetry items generated by this instance of the application after current sampling percentage was applied.</param>
+    /// <param name="currentSamplingPercentage">Current sampling percentage that was used by the algorithm.</param>
+    /// <param name="newSamplingPercentage">Suggested new sampling percentage that will allow to keep desired telemetry item generation rate given the volume of items states the same.</param>
+    /// <param name="isSamplingPercentageChanged">A value indicating whether new sampling percentage will be applied by dynamic sampling algorithm. New sampling percentage may not be immediately applied in case it was recently changed.</param>
+    /// <param name="settings">Dynamic sampling algorithm settings.</param>
+    public delegate void AdaptiveSamplingPercentageEvaluatedCallback(
```
SergeyKanzhelev: As I mentioned in the issue comment I wonder can this callback be replaced with tracing

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
```diff
@@ -0,0 +1,228 @@
...
+
+        /// <summary>
+        /// Evaluation timer.
+        /// </summary>
+        private Timer evaluationTimer;
+
+        /// <summary>
+        /// Current evaluation timeout.
+        /// </summary>
+        private int evaluationIntervalMs;
```
SergeyKanzhelev: Use TimeSpan

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorSettings.cs
```diff
@@ -0,0 +1,195 @@
...
+            {
+                return (int)Math.Ceiling(100 / AdjustSamplingPercentage(this.MinSamplingPercentage));
+            }
+        }
+
+        /// <summary>
+        /// Gets effective sampling percentage evaluation interval
+        /// adjusted in case user makes an error while setting a value.
+        /// </summary>
+        internal int EffectiveEvaluationIntervalSeconds
```
SergeyKanzhelev: Use TimeSpan - than you don't need to specify "InSeconds" in the name

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,262 @@
...
+
+        /// <summary>
+        /// Gets or sets exponential moving average ratio (factor) applied
+        /// during calculation of rate of telemetry items produced by the application.
+        /// </summary>
+        public double MovingAverageRatio
+        {
+            get
+            {
+                return this.estimatorSettings.MovingAverageRatio;
```
SergeyKanzhelev: when customer sets this you need to re-initialize ```ExponentialMovingAverageCounter```. When SDK reads configuration file - constructor will be called and construct everything with the standard settings. Only after that property will be overridden, but all the classes you've already initialized will not receive updated value

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,262 @@
...
+        {
+            this.Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Disposes the object.
+        /// </summary>
+        /// <param name="disposing">True if disposing.</param>
+        protected virtual void Dispose(bool disposing)
```
SergeyKanzhelev: Use ```disposing``` in code. Alternatively seal the class and do not create this protected method at all.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ExponentialMovingAverageCounter.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    /// <summary>
+    /// Exponential moving average counter.
+    /// </summary>
+    internal class ExponentialMovingAverageCounter
+    {
+        /// <summary>
+        /// Exponential coefficient (must be between 0 and 1).
+        /// </summary>
+        private readonly double coefficient;
```
SergeyKanzhelev: maybe call it weight?

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
vitalyf007: Changed to re-initialize counter if coefficient changed. Thanks!

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
vitalyf007: Fixed.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ExponentialMovingAverageCounter.cs
vitalyf007: It is called "coefficient" in math literature, generally. Including Wikipedia article I referred to in the docs. So, I'll stick to "coefficient".

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorSettings.cs
vitalyf007: Changed in the "settings" class. Since it is not generally used in config file (I was informed there are problems in deserializing complex types like TimeSpan or DateTime, etc), I left telemetry processor parameters as "...seconds"

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
vitalyf007: I'd like to keep it as callback. More structured, simple to use and allows tracing, metric extraction, etc

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
vitalyf007: Telemetry processor (this) has a lot wider public surface. Since I'd like to keep "settings" class, passing that and not "this" makes more sense I think...

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
vitalyf007: Fixed. Thanks.

src/TelemetryChannels/ServerTelemetryChannel/Shared/TelemetryChannelBuilderExtensions.cs
vitalyf007: I'm concerned that this reduces discoverability... UseAdaptiveSampling(settings) is easily discoverable. Discovering a class and a way to "use" it is much much harder.

Test/ServerTelemetryChannel.Test/Shared.Tests/Helpers/Utils.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Microsoft.ApplicationInsights.WindowsServer.Channel.Helpers
+{
+    using System;
+
+    public static class Utils
+    {
+        internal static bool EqualsWithPrecision(this double value1, double value2, double precision)
```
upendras: nit - We can use XUnit.Assert.Equal() overload that takes in a precision instead.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ExponentialMovingAverageCounter.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <param name="coefficient">Exponential coefficient.</param>
+        public ExponentialMovingAverageCounter(double coefficient)
+        {
+            this.Coefficient = coefficient;
+        }
+
+        /// <summary>
+        /// Gets exponential coefficient (must be between 0 and 1).
+        /// </summary>
+        public double Coefficient { get; private set; }
```
upendras: should we check that it's b/w 0 and 1?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
upendras: I agree that we should send settings as it has lesser public surface area. I am wondering though, if we should send a copy of settings instead of the working copy else, what stops a user from modifying some of the settings? Do we want to stop that?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
```diff
@@ -0,0 +1,244 @@
...
+        /// Callback for sampling percentage evaluation timer.
+        /// </summary>
+        /// <param name="state">Timer state.</param>
+        private void EstimateSamplingPercentage(object state)
+        {
+            // get observed after-sampling eps
+            double observedEps = this.itemCount.StartNewInterval() / this.evaluationInterval.TotalSeconds;
+
+            // we see events post sampling, so get pre-sampling eps
+            double beforeSamplingEps = observedEps * this.currenSamplingRate;
```
upendras: would it be better to be more accurate by actually measuring eps pre-sampling?

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,265 @@
...
+            SamplingPercentageEstimatorSettings settings,
+            AdaptiveSamplingPercentageEvaluatedCallback callback,
+            ITelemetryProcessor next)
+        {
+            this.estimatorSettings = settings;
+            this.evaluationCallback = callback;
+
+            // make estimatortelemetry processor  work after sampling was done
+            this.estimatorProcessor = new SamplingPercentageEstimatorTelemetryProcessor(settings, this.SamplingPercentageChanged, next);
+            this.samplingProcessor = new SamplingTelemetryProcessor(this.estimatorProcessor);
```
SergeyKanzhelev: Let's use DI to inject the sampling processor here. As you mentioned that one of the reason to have two sampling processors - one simple and one Adaptive is to be able to replace implementation of sampling rate calculation, but keep "adaptive" part of it intact. Maybe ```SaplingTelemetryProcessor``` should sign up on sampling interval change callback or something...

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
```diff
@@ -0,0 +1,265 @@
...
+            {
+                this.estimatorSettings.MaxSamplingPercentage = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets duration of the sampling percentage evaluation 
+        /// interval in seconds.
+        /// </summary>
+        public int EvaluationIntervalSeconds
```
SergeyKanzhelev: How difficult would it be to implement reading TimeSpan from config? I'd prefer to have it strongly typed.

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
SergeyKanzhelev: Can you please push update with this code?

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
SergeyKanzhelev: Also if you'll use DI or some sort of explicit instantiation of ```SamplingTelemetryProcessor``` - you wouldn't need ```InitialSamplingPercentage``` property any longer

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        /// <summary>
+        /// Callback for sampling percentage evaluation timer.
+        /// </summary>
+        /// <param name="state">Timer state.</param>
+        private void EstimateSamplingPercentage(object state)
+        {
+            // get observed after-sampling eps
+            double observedEps = this.itemCount.StartNewInterval() / this.evaluationInterval.TotalSeconds;
```
SergeyKanzhelev: Woulds timer guarantee that time elapsed is exactly as you requested? Should we maintain our "own" timer?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
```diff
@@ -0,0 +1,244 @@
...
+
+            // see if evaluation interval was changed and apply change
+            if (this.evaluationInterval != this.settings.EffectiveEvaluationInterval)
+            {
+                this.evaluationInterval = this.settings.EffectiveEvaluationInterval;
+                this.evaluationTimer.Change(this.evaluationInterval, this.evaluationInterval);
+            }
+
+            // check to see if sampling rate needs changes
+            bool samplingPercentageChangeNeeded = suggestedSamplingRate != this.currenSamplingRate;
```
SergeyKanzhelev: should you compare with precision?

SergeyKanzhelev:
CC: @abaranch

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        /// <summary>
+        /// Processes telemetry item.
+        /// </summary>
+        /// <param name="item">Telemetry item to process.</param>
+        public void Process(ITelemetry item)
+        {
+            // increment post-samplin telemetry item counter
+            this.itemCount.Increment();
```
SergeyKanzhelev: You should only increment if item implements ```ISupportSampling```, right?

Test/ServerTelemetryChannel.Test/Shared.Tests/Helpers/Utils.cs
vitalyf007: Cool. Switched to XUnit and removed custom implementation.

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
vitalyf007: I think yet another extensibility point would be excessive... TelemetryProcessor is the "concept" we already use...

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
vitalyf007: Don't know. I do not think it'd be too hard. Outside of the "realm" of sampling though...

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
vitalyf007: I did. The check is made in the timer callback.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ExponentialMovingAverageCounter.cs
vitalyf007: And if the check fails we'd do what? Throwing isn't a good idea (failure just in the monitoring code potentially preventing app from executing entirely). If it becomes (reported) to be a problem we can add some checking (say, in debug code only or something).

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
vitalyf007: This is a convenient place to change settings if needed. I intentionally pass settings by ref...

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
vitalyf007: Originally I had it like that. But then I changed it. My ultimate goal is to reduce telemetry stream overall (including items subject to sampling and not). So, I count all items but sampling percentage I set affects only subject-to-sampling items and that is the way I want it...

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
vitalyf007: I was thinking about it.... I think it is fine as is. The interval would be in seconds usually and timer precision (plus overhead of callback execution) is few ms tops. So, it will be well within approximation boundaries...

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
vitalyf007: We could do that but would need tree modules (measure pre-sampling, post-sampling andsampling itself). Besides, there is a complication of subject-to-sampling and not-subject-to-sampling types of telemetry. I think the code serves it purpose as is...

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/SamplingPercentageEstimatorTelemetryProcessor.cs
vitalyf007: Sampling rates are INTs expressing sampling percentage in form of 1/N (take 1 of every N items). So, they can be compared w/o precision...

src/TelemetryChannels/ServerTelemetryChannel/Shared/AdaptiveSamplingTelemetryProcessor.cs
SergeyKanzhelev: I checked - it is already supported; Just do something like this:

``` xml
  <Add Type="ConsoleApplication2.AAA, ConsoleApplication2" >
    <Duration>00:00:05.123</Duration>
  </Add>
```
Code that reads configuration is here: https://github.com/Microsoft/ApplicationInsights-dotnet/blob/9f4c715fa703753c2e6107afde79051447018a02/src/Core/Managed/Net40/Extensibility/Implementation/TelemetryConfigurationFactory.cs#L297

SergeyKanzhelev:
:shipit: Thanks @vitalyf007

upendras:
:shipit:

PR #89, opened 2015-10-27
Changes for Web SDK blackfield

abaranch:
For .Net 4.0 and .Net 4.5 write to VS debug output window even if iKey is not configured.

### DIFF
--- a/Global.props
+++ b/Global.props
@@ -109,14 +109,14 @@
       Update for every public release. 
     -->
     <SemanticVersionMajor>1</SemanticVersionMajor>
-    <SemanticVersionMinor>2</SemanticVersionMinor>
+    <SemanticVersionMinor>3</SemanticVersionMinor>
     <SemanticVersionPatch>0</SemanticVersionPatch>
 
     <!-- 
       Date when Semantic Version was changed. 
       Update for every public release.
     -->
-    <SemanticVersionDate>2015-7-17</SemanticVersionDate>
+    <SemanticVersionDate>2015-10-27</SemanticVersionDate>
 
     <!-- 
       Pre-release version is used to distinguish internally built NuGet packages.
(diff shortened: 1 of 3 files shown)


src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -314,6 +313,85 @@ public void TrackDependency(DependencyTelemetry telemetry)
             this.Track(telemetry);
         }
 
+        //// Preserve original version of 'Track' with TelemetryInitializer being called after iKey check for Devices SDK till we completely move to a different SDK.
+#if NET40 || NET45
```
SergeyKanzhelev: isn't it too much logic duplication under this "if"?

SergeyKanzhelev: it is hard to understand what's the difference between those two versions of ```Track```. Can you please switch to ``if``` inside the Track method itself?

Global.props
```diff
@@ -109,14 +109,14 @@
       Update for every public release. 
     -->
     <SemanticVersionMajor>1</SemanticVersionMajor>
-    <SemanticVersionMinor>2</SemanticVersionMinor>
+    <SemanticVersionMinor>3</SemanticVersionMinor>
```
avidhanju: Will we have to change the SDK packages for our VSIX?

src/Core/Managed/Shared/TelemetryClient.cs
abaranch: In 2.0-beta we will have only 1 method as it is now in master just with one line moved upper. Here because we cannot touch devices part it will be even harder to understand if we try to use "ifs". In old version we had: 1. check iKey and return 2. Call initializers 3. Call Send 4. Write to debug. In the new version we want 1. Call Initializers 2. Write to debug 3. Check iKey and return 4. Call send.

Global.props
abaranch: We will definitely need to update the service to pick the latest version. 1.2.0 was the latest stable version that is already shipped and we cannot ship new nuget package with the same version.

abaranch: I actually plan to change to 1.2.3 because we have only bug fixes and no new features.

PR #94, opened 2015-10-29
Root name is checked for null after initialize and updated from telemetry name

karthiktangirala1:
…etry name


abaranch:
:shipit:

Issue #47, opened 2015-09-28
Support end-to-end cross-tier correlation


SergeyKanzhelev:
cc: @abaranch, @briancr-ms, @BogdanBe, @vitalyf007 

Stable version of Application Insights SDK doesn't support hierarchical correlation today. Correlation is implemented using two fields: `RequestTelemetry.ID` and `ITelemetry.Context.Operaiton.Id`. When these fields match - we show them as related in UI.

Cross-tier and hierarchical correlation requires more fields. Typical scenario for this type of correlation that we want to support is for every page view event (collected by JavaScript SDK) we show as "children" all ajax calls run on the page. For every ajax call we show corresponding server request collected by .NET server SDK. Every server request may have children operations like "business logic execution" or "rendering" which in turn may have calls to external dependencies as children activities. So we can troubleshoot page loading sloweness and show the root cause when sloweness was caused by database issue.

We propose the following terminology:
- _Transaction Name_ - name of the first/topmost/root operation that initiated entire chain of execution.
- _Transaction ID_ - correspondingly unique ID of operation described above.
- *Parent Operation ID" - unique id of the operation that was active at the moment when telemetry was reported
- _Operation Name_ - name of the telemetry item if it is request telemetry or dependency telemetry
- _Operation ID_ - unique id of the telemetry item if it is request telemetry or dependency telemetry

**_Note:**_ We propose to use term _Transaction_ as oppose to _Root Operation_ as we believe that transaction is easier to understand and explain.

Label for corresponding telemetry items on request details blade will be updated to this:
"All traces/exceptions/dependencies for this <strike>request</strike> transaction".

We can use the following http header names to propagate correlation information:
- `x-ms-request-id` for parent operation ID
- `x-ms-transaction-id` for transaction ID
- `x-ms-transaction-name` for transaction name

Going forward In order to support proper end-to-end cross-tier events correlation we want to start using concept called "correlation vector". The idea behind correlation vector is very similar to the new EventSource correlation and causation feature in FW 4.6. Here is [Vance Morrison blog post](http://blogs.msdn.com/b/vancem/archive/2015/09/14/exploring-eventsource-activity-correlation-and-causation-features.aspx) explaining this feature.

Correlation vector is a string consist of base64-encoded GUID and a dot-delimited vector of integer values: HyCFaiQoBkyEp0L3.1.2 

It support two method: `Increment` that will increment the last integer in a vector and `Extend` that will add `.0` to the end of the vector.

[...]

PR #95, opened 2015-10-30
Return RequestTelemetry.Id and remove RootID

SergeyKanzhelev:
Addressing https://github.com/Microsoft/ApplicationInsights-dotnet/issues/47
```DependencyTelemetry``` will not send it's own ```id``` to the backend as we don't use it and don't have in schema now. Will be working on adding it to the schema


kartang:
Looks good!

PR #99, opened 2015-11-05
Add Operation telemetry initializer by default

SergeyKanzhelev:
1. Remove TimeStamp telemetry initializer. Allow to do:
``` csharp
(new TelemetryClient(new TelemetryConfiguration())).TrackTrace("test")
```
that will actually send event that will be correct from DC standpoint.

2. Use ```Clock``` instead of ```OperationWatch```

3. Add operation context by default


SergeyKanzhelev:
cc @abaranch @kartang

abaranch:
:shipit:

kartang:
Looks good!

PR #105, opened 2015-11-10
Initialize modules only from Active configuration

abaranch:
- Initialize modules collection only when active configuration is accessed and do not initialize it for CreateDefault (Otherwise if application uses both autocollection with ai.config and code instrumentation, telemetry modules are initialized with first configuration accessed that may lead to sending telemetry to the wrong iKey)
- Remove dependency on modules singleton from configuration factory 
- Add unit tests for telemetry modules

### DIFF
--- a/src/Core/Managed/Net40/Extensibility/Implementation/TelemetryConfigurationFactory.cs
+++ b/src/Core/Managed/Net40/Extensibility/Implementation/TelemetryConfigurationFactory.cs
@@ -7,20 +7,24 @@
     using System.IO;
     using System.Linq;
     using System.Reflection;
+    using System.Threading;
     using System.Xml.Linq;
+
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.ApplicationInsights.Extensibility.Implementation.Platform;
 
     internal class TelemetryConfigurationFactory
     {
+        protected string configurationFileContent;
+
         private const string AddElementName = "Add";
         private const string TypeAttributeName = "Type";
         private static readonly MethodInfo LoadInstancesDefinition = typeof(TelemetryConfigurationFactory).GetRuntimeMethods().First(m => m.Name == "LoadInstances");
         private static readonly XNamespace XmlNamespace = "http://schemas.microsoft.com/ApplicationInsights/2013/Settings";
 
         private static TelemetryConfigurationFactory instance;
-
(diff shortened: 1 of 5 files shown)


src/Core/Managed/Net40/Extensibility/Implementation/TelemetryConfigurationFactory.cs
```diff
@@ -7,20 +7,24 @@
...
+    using System.Threading;
     using System.Xml.Linq;
+
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.ApplicationInsights.Extensibility.Implementation.Platform;
 
     internal class TelemetryConfigurationFactory
     {
+        protected string configurationFileContent;
```
SergeyKanzhelev: let's not keep file content in memory for the entire life of the program.

src/Core/Managed/Net40/Extensibility/Implementation/TelemetryConfigurationFactory.cs
```diff
@@ -53,11 +57,11 @@ public virtual void Initialize(TelemetryConfiguration configuration)
 #endif
 
             // Load customizations from the ApplicationsInsights.config file
-            string text = PlatformSingleton.Current.ReadConfigurationXml();
-            if (!string.IsNullOrEmpty(text))
+            LazyInitializer.EnsureInitialized(ref this.configurationFileContent, PlatformSingleton.Current.ReadConfigurationXml);
```
SergeyKanzhelev: today CreateDefault will re-read configuration file if it was changed. With your change it's no longer the case.

src/Core/Managed/Net40/Extensibility/Implementation/TelemetryConfigurationFactory.cs
```diff
@@ -44,7 +48,7 @@ public static TelemetryConfigurationFactory Instance
             set { instance = value; }
         }
 
-        public virtual void Initialize(TelemetryConfiguration configuration)
+        public virtual void Initialize(TelemetryConfiguration configuration, IList<ITelemetryModule> modules)
```
SergeyKanzhelev: Can we pass TelemetryModules class instead of using ```IList``` directly? We can change implementation of it later this way

SergeyKanzhelev:
:shipit:

Issue #112, opened 2015-11-11
Telemetry sanitisation should happen in the channel


abaranch:
Now item is sanitized after TelemetryInitializers but before TelemetryProcessors. But TelemetryProcessors may enrich the data so that it will exceed size limit or add some strings that need to be encoded.

PR #114, opened 2015-11-12
Move Sanitize to JsonSerializer

abaranch:
Fixing issue #112 
Moved Sanitize to JsonSerializer instead of all the channels by 2 reasons:
- All channels use same Serializer
- Serialization happens on a different thread that will improve performance


SergeyKanzhelev:
Looks good. Thanks! :shipit:

PR #116, opened 2015-11-16
Track every telemetry item as a rich payload ETW

iusafaro:
Track every telemetry item as a rich payload ETW - for integration with ServiceProfiler

### DIFF
--- a/src/Core/Managed/Net46/Extensibility/Implementation/RichPayloadEventSource.cs
+++ b/src/Core/Managed/Net46/Extensibility/Implementation/RichPayloadEventSource.cs
@@ -20,7 +20,7 @@
         internal readonly EventSource eventSource;
 
         /// <summary>Event provider name.</summary>
-        private const string EventProviderName = "Microsoft-ApplicationInsights";
+        private const string EventProviderName = "Microsoft-ApplicationInsights-Data";
 
         /// <summary>
         /// Initializes a new instance of the RichPayloadEventSource class.
@@ -29,9 +29,7 @@
         {
             this.eventSource = new EventSource(
                EventProviderName,
-               EventSourceSettings.EtwSelfDescribingEventFormat,
-               "ETW_GROUP",
-               "{E3AFA6F4-0FEA-4757-AFFF-80B5CA685BB8}");
+               EventSourceSettings.EtwSelfDescribingEventFormat);
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/Core/Managed/Net46/Extensibility/Implementation/RichPayloadEventSource.cs
```diff
@@ -0,0 +1,268 @@
...
+        /// <summary>
+        /// Initializes a new instance of the RichPayloadEventSource class.
+        /// </summary>
+        public RichPayloadEventSource()
+        {
+            this.eventSource = new EventSource(
+               EventProviderName,
+               EventSourceSettings.EtwSelfDescribingEventFormat,
+               "ETW_GROUP",
+               "{E3AFA6F4-0FEA-4757-AFFF-80B5CA685BB8}");
```
SergeyKanzhelev: where is this constant from? Consider moving it to https://github.com/Microsoft/ApplicationInsights-dotnet/blob/271edd5fd8b9c7d84a9bd72448ec14f9b2145259/src/Core/Managed/Shared/Constants.cs

Also I'm not sure that we need to specify ETW_GROUP - I vaguely remember it was used for UTC. @izikl, can you please advice here?

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -324,6 +324,11 @@ public void Track(ITelemetry telemetry)
                     telemetry.Timestamp = Clock.Instance.Time;
                 }
 
+#if NET46
+                // logs rich payload ETW event for any partners to process it
+                RichPayloadEventSource.Log.Process(telemetry);
```
SergeyKanzhelev: do it after processors as they can modify the telemetry item

src/Core/Managed/Net46/Extensibility/Implementation/RichPayloadEventSource.cs
```diff
@@ -0,0 +1,268 @@
...
+    internal sealed class RichPayloadEventSource : IDisposable
+    {
+        /// <summary>RichPayloadEventSource instance.</summary>
+        public static readonly RichPayloadEventSource Log = new RichPayloadEventSource();
+        
+        /// <summary>Event source.</summary>
+        internal readonly EventSource eventSource;
+
+        /// <summary>Event provider name.</summary>
+        private const string EventProviderName = "Microsoft-ApplicationInsights";
```
abaranch: Is this the name for this event source? Can we use "Microsoft-ApplicationInsights-RichPayload"?

SergeyKanzhelev:
Consider adding test that you can modify elements in initializers and will report modified element in etw.

SergeyKanzhelev:
Besides :shipit:

PR #128, opened 2015-11-26
Move ResponseCode and Success population to Sanitize

abaranch:
Move ResponseCode and Success population to Sanitize so in telemetry initializers we can distinguish whether these fields were populated by customer's code

### DIFF
--- a/src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
+++ b/src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
@@ -35,9 +35,6 @@
             this.Data = new RequestData();
             this.context = new TelemetryContext(this.Data.properties, new Dictionary<string, string>());
             this.Id = Convert.ToBase64String(BitConverter.GetBytes(WeakConcurrentRandom.Instance.Next()));
-
-            this.ResponseCode = "200";
-            this.Success = true;
         }
 
         /// <summary>
@@ -123,10 +120,8 @@
                 {
                     return this.Data.success;
                 }
-                else
-                {
-                    return null;
-                }
+
+                return null;
             }
 
             set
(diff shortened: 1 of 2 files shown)


src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
```diff
@@ -222,7 +217,17 @@ void ITelemetry.Sanitize()
             this.Data.id = this.Data.id.SanitizeName();
             this.Data.id = Utils.PopulateRequiredStringValue(this.Data.id, "id", typeof(RequestTelemetry).FullName);
 
-            this.ResponseCode = Utils.PopulateRequiredStringValue(this.ResponseCode, "responseCode", typeof(RequestTelemetry).FullName);
+            // Required field
+            if (string.IsNullOrEmpty(this.ResponseCode))
```
AlexBulankou: What are the scenarios when response code is null or empty?

src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
```diff
@@ -222,7 +217,17 @@ void ITelemetry.Sanitize()
...
+            {
+                this.ResponseCode = "200";
+            }
+
+            // Required field
+            if (!this.Success.HasValue)
+            {
+                this.Success = true;
```
AlexBulankou: Should success be based on response code (>=400)

src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
abaranch: It is not initialized by default. If it is not autogenerated item you can use parameterless constructor and forget to initialize StatusCode. But as it is a required field we initialize it is sanitize. (Before we were setting it in parameterless constructor and now we move initialization in Sanitize method that is called before Send. This way in Web SDK we may check if request status code was initialized from customer's code and do not override it)

src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
abaranch: Good point. They can be initialized separately.

src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
```diff
@@ -35,9 +35,6 @@ public RequestTelemetry()
             this.Data = new RequestData();
             this.context = new TelemetryContext(this.Data.properties, new Dictionary<string, string>());
             this.Id = Convert.ToBase64String(BitConverter.GetBytes(WeakConcurrentRandom.Instance.Next()));
```
SergeyKanzhelev: should we do the same with ID?

abaranch: No
1. we never try to override it or need to check if it ws changed in customer's code
2. we need it for correlation on early stages

SergeyKanzhelev:
:shipit:

PR #138, opened 2015-12-11
Updated DependencyTelemetry to expose new property ResultCode

madeleinefletcher:


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This changelog will be used to generate documentation on [release notes page](ht
 ## Version 2.0.0-beta4
 
 - UseSampling and UseAdaptiveSampling extensions were moved to Microsoft.ApplicationInsights.Extensibility
+- Updated ```DependencyTelemetry``` to have new property ```ResultCode```
```
abaranch: And Id

src/Core/Managed/Shared/DataContracts/DependencyTelemetry.cs
```diff
@@ -201,6 +214,7 @@ void ITelemetry.Sanitize()
         {
             this.Name = this.Name.SanitizeName();
             this.Name = Utils.PopulateRequiredStringValue(this.Name, "name", typeof(DependencyTelemetry).FullName);
+            this.Id = this.Id.SanitizeName();
```
abaranch: Id needs to be initialized in the constructor the same way it is done for RequestTelemetry.Id. ResultCode needs to be Sanitized.

madeleinefletcher: What should the max length for ResultCode be when we sanitize? And do we need to sanitize Id?

PR #152, opened 2016-01-12
Changed telemetry channel to not retry on 402 response from API

madeleinefletcher:


SergeyKanzhelev:
Is there retry on client side errors like DNS resolution? Do we retry on 502? 
CC: @upendras 
@lukekim , what other status codes should we expect from DC?

madeleinefletcher:
As far as I can tell the only retries are on too many requests: 429 and 439.

upendras:
We do retries for 500 class errors, but that is in a different error handing policy. The throttling policy only deals with 402,429 and 439 and the retry is guided by server hint in retry-after header.

SergeyKanzhelev:
Ok, makes sense. :shipit:

PR #156, opened 2016-01-14
Updated generations to dotNet5.4

kartang:


SergeyKanzhelev:
@upendras I'm actually confused - do you know whether we need to use `netstandard` moniker now instead of `dotnet`? See https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/standard-platform.md

@kartang for asp.net 5 we are still using PCL binaries. Do you know how hard will it be to start compiling code with the proper target?

kartang:
@SergeyKanzhelev For this pull request, I just got the 5.4 working. Once we have a complete end-to-end set up ready, I will work on compiling the code with the respective target.

SergeyKanzhelev:
@kartang , makes sense.

PR #161, opened 2016-01-22
Changes required to move signed build to v2

madeleinefletcher:
Added microbuild packages and nuget script for v2 builds, so we can build directly from github


VSOPushNugetPackages.cmd
```diff
@@ -0,0 +1,25 @@
+setlocal
```
SergeyKanzhelev: You don't need this script with microbuild, right?

SergeyKanzhelev:
:shipit:

VSOPushNugetPackages.cmd
madeleinefletcher: Yes, this is so we can publish the nuget packages to myget automatically during the signed build. I added basically a copy of PushNugetPackages.cmd, because the original uses TFS build variables that don't work with V2. I kept the old one for now so we don't break the existing builds. we can remove it when we switch over completely.

SergeyKanzhelev: makes sense. Thanks for explanaition

Issue #153, opened 2016-01-12
Telemetry should be written to debug output after Telemetry Processors are run


madeleinefletcher:
Currently, telemetry is written to debug output before telemetry processors are executed, so it doesn't reflect items that were modified or filtered out by the processors. Debug output should be written after telemetry processors are executed and not print items that were filtered out. However, if there is no instrumentation key, debug output should still be written (as telemetry processors are not executed in that case).

PR #168, opened 2016-01-26
Changing debug output so it writes telemetry as it is sent

madeleinefletcher:
Fix for this bug: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/153


AlexBulankou:
We always know from the place we are calling new utility method whether it will write the message as configured or not configured. So perhaps for the sake of code clarity we should just make isConfigured a parameter to the utility function?
Otherwise :shipit:

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -293,11 +293,10 @@ public void Track(ITelemetry telemetry)
             if (this.IsEnabled())
             {
                 this.Initialize(telemetry);
-                
-                this.WriteTelemetryToDebugOutput(telemetry);
```
SergeyKanzhelev: Even if telemetry was sampled Visual Studio diagnostics hub will still need to show this telemetry item. Perhaps we can somehow indicate that telemetry was sampled out by setting sampleRate to zero. 

So I'd recommend just to move this call couple lines to the bottom right after the call to processors

madeleinefletcher: We should make a decision what direction we want to go here. I was under the impression we didn't want to show items that were filtered out by the processors. Do we always want to show them? Or should we only show them for sampling processor?

AlexBulankou: I remember we discussed that we would like to set an "IN" flag on a telemetry item before processing and then clear it if the item made it all the way to transmission. This way we know that the items with this flag remaining are the ones that are filtered or sampled out, so we can have additional prefix before them in the output window. 
This obviously only applies to the telemetry items with ikey present, when ikey is not present, all of them will appear in the output.

SergeyKanzhelev: It is an interesting discussion. How to distinguish between important telemetry that was aggregated and sampled and non-important telemetry or trash that user has chosen to filter away. If let's say I filtered out Verbose traces for non-failed requests. Does it mean Visual Studio should not display it? Does it mean customer do not care about this telemetry?

Also if we will decide that there is a way to distinguish bad telemetry - we need to apply the same logic of not writing it to this call: ```RichPayloadEventSource.Log.Process(telemetry);```

SergeyKanzhelev: + @dimazaid

AlexBulankou: I am thinking of output window is a diagnostics tool for AI. I would like to see what telemetry items were filtered out (by any processor). So it would be super helpful if I can see all filtered telemetry items as
[Filtered out - not sent to server]: {telemetry item}

madeleinefletcher: Ok, check out this new implementation.

dimazaid: "If let's say I filtered out Verbose traces for non-failed requests." how do you do this today?

AlexBulankou:
I like the new implementation - great to have info about all telemetry, even those items filtered out, in the output window.

src/Core/Managed/Shared/TelemetryClient.cs
SergeyKanzhelev: @dimazaid - you can write telemetry processor to filter out data. This PR introduces the new prefix in output window that Diagnostics Hub should ideally pick up and visualize somehow

SergeyKanzhelev:
So now we are tracing every telemetry item and introduced Boolean at every telemetry item just for the case when it is working in Visual Studio. I'd propose to trace "sent" from Processor as you did before instead if we need to indicate filtered telemetry items instead

SergeyKanzhelev:
If we need to indicate it at all.

madeleinefletcher:
I'm confused by what you mean. If I don't have the "sent" flag, either I'd have to record some telemetry items twice (once before processing and once when they're sent), or I wouldn't be able to trace the ones that were filtered out at all...

SergeyKanzhelev:
I mean just trace everything that has instrumentation key after processors. If we want to know which one were not filtered we can output string "sent" from Processor - the place where you planned to move debug output before

madeleinefletcher:
Ok, I updated it to use a ConditionalWeakTable...will that work? Otherwise, I will just have to print the sent telemetry items twice, once when they're sent and once after so that we don't exclude filtered out ones

src/Core/Managed/Shared/Extensibility/Implementation/TransmissionProcessor.cs
```diff
@@ -41,6 +39,8 @@ public void Process(ITelemetry item)
             }
 
             this.configuration.TelemetryChannel.Send(item);
+
+            item.Context.ExtraTelemetryProperties.Add(item, new InternalTelemetryProperties { Sent = true });
```
SergeyKanzhelev: let's do it only when debug output is ON.

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -418,13 +419,30 @@ public void Flush()
...
-        private void WriteTelemetryToDebugOutput(ITelemetry telemetry)
+        private void WriteTelemetryToDebugOutput(ITelemetry telemetry, bool isConfigured)
         {
             if (this.debugOutput.IsAttached() && this.debugOutput.IsLogging())
             {
-                string prefix = string.IsNullOrEmpty(telemetry.Context.InstrumentationKey) ?
-                    "Application Insights Telemetry (unconfigured): " :
-                    "Application Insights Telemetry: ";
+                InternalTelemetryProperties properties = null;
+                Utils.ExtraTelemetryProperties.TryGetValue(telemetry, out properties);
```
SergeyKanzhelev: Let's remove it after we read it once. Just in case customer wants to reuse the same item later since we do not force them not to do it, I know it is not thread safe, but it's better than nothing =)

SergeyKanzhelev:
Looks good. :shipit:

src/Core/Managed/Shared/TelemetryClient.cs
madeleinefletcher: Done

abaranch:
This solution will not work. Customer (or we in the future) may write async telemetry processors that buffer items or just return before item reached transmitter than flag will not be set.

abaranch:
Also deleting item from non tread safe collection without lock will just crash at some point...

madeleinefletcher:
As far as thread safety, I thought ConditionalWeakTable was... "Instances of the ConditionalWeakTable<TKey, TValue> class are thread safe. They do not require callers to do any additional locking." as per the msdn docs. https://msdn.microsoft.com/en-us/library/dd287757(v=vs.110).aspx

madeleinefletcher:
If the buffer resides in core, we can set the flag when the item is added to the buffer right? Here are a few options for an implementation that would work in this case:

1) Have buffer set the flag when you add the item. This might require making the conditional table public, so if you write your own transmission processor or buffer, you can set the flag correctly.
2) Print sent items twice. You'd print every item before the telemetry processors, then if it's sent, print again in the transmission processor. While the debug output would be more confusing, this would prevent the case of all items being marked filtered out if you write your own transmission processor and would not require a flag.

Which would you prefer? Also, can you think of any other ways to do this?

abaranch:
I misread Sergey's comment about thread safety.
As we discussed offline let's print from each TelemetryProcessor so we also know which one filtered item out as you proposed.

madeleinefletcher:
See PR https://github.com/Microsoft/ApplicationInsights-dotnet/pull/180

Issue #55, opened 2015-10-02
Configuration reading should be resilient and easy to troubleshoot


SergeyKanzhelev:
When one of the types wasn't found - all other types will not be read by Application Insights SDK. See 
https://github.com/Microsoft/ApplicationInsights-Home/issues/23 for details.

Configuration reading should be more resilient and extra tracing should be added.

shanec-:
I was facing the same issue as mschmeller in https://github.com/Microsoft/ApplicationInsights-Home/issues/23. And like him, I too was overriding the instrumentation key with one stored in the `web.config` via the `Global.asax`. Everything worked fine locally but bombs out when it was deployed to azure.

In my case, it turned out that Azure did not like the use of `WebConfigurationManager` in this context. I switched my code from:

`Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration.Active.InstrumentationKey = System.Web.Configuration.WebConfigurationManager.AppSettings["MyInstrumentationKey"];` 

to 

`Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration.Active.InstrumentationKey = System.Configuration.ConfigurationManager.AppSettings["MyInstrumentationKey"]; `

and everything seems to work as expected. 

I hope this helps.

SergeyKanzhelev:
Are you sure this issue is related to the wrongly set instrumentation key? My guess would be that you still have some bad telemetry initializers in configuration file. If the code in the first snippet fails you'll not initialize ```InstrumentationKey```, but ```TelemetryChannel``` will be not null. So you should not get ```NullReferenceException```.

alancameronwills:
Same problem with mismatched property names. In TelemetryProcessors, if there's a property referenced in the .config that doesn't exist in the code, the SDK silently fails to send any telemetry at all.

SergeyKanzhelev:
Another problem happens when unit test project doesn't have reference to some of the modules like `DependencyCollector` and the call to `new TelemetryClient()` fails with `TypeLoadException`.

PR #173, opened 2016-02-01
Partial configuration loading

abaranch:
In case if configuration file is broken we will try to load as much as possible. Addressing: #55


abaranch:
@SergeyKanzhelev

abaranch:
@SergeyKanzhelev , what do you think if we keep throwing exception in DeveloperMode and have proposed changes only for release version?

SergeyKanzhelev:
Looks good for me. I don't think we need to throw in developer mode as HttpModule in Web will catch exception anyway. It will just make it very cryptic and hard to troubleshoot

PR #180, opened 2016-02-05
Allowing debug output to be written by telemetry processors

madeleinefletcher:


src/Core/Managed/Shared/Extensibility/Implementation/DebugOutput.cs
```diff
@@ -8,11 +8,41 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
...
+    /// </summary>
+    public class DebugOutput : IDebugOutput
     {
-        public void WriteLine(string message)
+        /// <summary>
+        /// Write the specified <see cref="ITelemetry"/> item to debug output.
+        /// </summary>
+        /// <param name="telemetry">Item to write.</param>
+        /// <param name="filteredBy">If specified, indicates the telemetry item was filtered out and not sent to the API.</param>
+        public void WriteTelemetry(ITelemetry telemetry, string filteredBy = null)
```
madeleinefletcher: I'm open to suggestions on other places to put this method. It does need to be publicly exposed though, so the telemetry processors can access it.

src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
```diff
@@ -65,6 +66,7 @@ public void Process(ITelemetry item)
                         TelemetryChannelEventSource.Log.ItemSampledOut(item.ToString());
                     }
 
+                    new DebugOutput().WriteTelemetry(item, this.GetType().FullName);
```
madeleinefletcher: Should writing debug output be static?

src/Core/Managed/Shared/Extensibility/Implementation/DebugOutput.cs
```diff
@@ -8,11 +8,41 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
...
+            {
+                string prefix = "Application Insights Telemetry: ";
+                if (string.IsNullOrEmpty(telemetry.Context.InstrumentationKey))
+                {
+                    prefix = "Application Insights Telemetry (unconfigured): ";
+                }
+
+                if (!string.IsNullOrEmpty(filteredBy))
+                {
+                    prefix = "Application Insights Telemetry (filtered by " + filteredBy + "): ";
```
madeleinefletcher: What do you think of this prefix? I can provide a regex so the debug output can be filtered for this.

madeleinefletcher:
What do you think of this implementation?

src/Core/Managed/Shared/Extensibility/Implementation/TelemetryDebugWriter.cs
```diff
@@ -0,0 +1,78 @@
...
+    using System;
+    using System.Diagnostics;
+    using Channel;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Platform;
+
+    /// <summary>
+    /// Writes telemetry items to debug output.
+    /// </summary>
+    public class TelemetryDebugWriter : IDebugOutput
```
SergeyKanzhelev: You are right, we don't need interface anymore. Also make class static

src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
```diff
@@ -65,6 +66,7 @@ public void Process(ITelemetry item)
                         TelemetryChannelEventSource.Log.ItemSampledOut(item.ToString());
                     }
 
+                    TelemetryDebugWriter.WriteTelemetry(item, this.GetType().FullName);
```
SergeyKanzhelev: Does `this.GetType().FullName` includes namespaces? I'd say we only need name to make output more compact

SergeyKanzhelev:
:shipit: with two comments

src/Core/Managed/Shared/Extensibility/Implementation/TelemetryDebugWriter.cs
madeleinefletcher: The are a lot of unit tests that actually test whether items are written to debug output. So I ended up reaching a compromise where the publicly exposed method is static, but we keep the interface for our unit tests.

src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
madeleinefletcher: Ok, makes sense.

src/Core/Managed/Shared/Extensibility/Implementation/TelemetryDebugWriter.cs
SergeyKanzhelev: ok. maybe time will come and we will clean it up =)

PR #185, opened 2016-02-09
Stop adding custom properties to ApplicationInsights internal traces

abaranch:
#181


SergeyKanzhelev:
thanks, :shipit: . Can you also mark these traces with "SyntheticSource=SDKTelemetry"

PR #191, opened 2016-02-12
Rename to rc1

abaranch:


SergeyKanzhelev:
Don't change semantic version as file version depend on it and we want it to increment all the time

Issue #194, opened 2016-02-18
Exception: Method not found: 'Void System.Diagnostics.Tracing.EventSource.WriteEvent(Int32, Int32, System.String)'.


abaranch:
Framework 4.6
Worker roles crashes with:
_AppDomain Unhandled Exception for role WorkerRole1_IN_0
Exception: Method not found: 'Void System.Diagnostics.Tracing.EventSource.WriteEvent(Int32, Int32, System.String)'.
   at Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.CoreEventSource.DiagnoisticsEventThrottlingSchedulerTimerWasCreated(Int32 intervalInMilliseconds, String appDomainName)
   at Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.DiagnoisticsEventThrottlingScheduler.ScheduleToRunEve_

Failure happens in diagnostics module. Web SDK catches this exception but there would be no internal telemetry available. Core SDK starting from the next SDK version would add Diagnostics module by default. In case there is no global catch in the worker role where it sets instrumentation key, it crashes.
Note: Azure does not support 4.6 now but I ignored the warning and tried to deploy anyway. ApplicationInsights becomes first point of failure.
This should also happen if 4.6 application just uses Core SDK and initializes everything programmatically.

Workaround: http://stackoverflow.com/questions/32097492/method-not-found-void-system-diagnostics-tracing-frameworkeventsource-beginget 

@alancameronwills We need to add this to some troubleshooting page or known issues list
@SergeyKanzhelev We can check method signatures that we use and try to exclude (int, int, string) if that is possible. Though I do not know if that is the only problematic signature.

SergeyKanzhelev:
So you build 4.6 app and deploy to the environment with no 4.6 available? I'm surprised it is the only thing that fails in this case =) 

Why would we crash the application? Can we just catch this exception?

abaranch:
1. We cannot put try/catches around all places that we use EventSource. 2. I did not notice warning about cloud apps and 4.5; nothing prevents you from deploying it, moreover Web Role seemed to worked perfectly fine. When you see this failure that may not be obvious what happened.

alancameronwills:
I guess all we can say is that if you come across this problem, please note that you can't deploy a 4.6 app to Azure (yet). It might look like an AI problem, but you'd likely see the issue elsewhere in your app anyway.

I don't understand your link to the workaround, which is talking about WP8.1.

[...]

PR #197, opened 2016-02-23
Fix EventSource failure #194

abaranch:
#194
Pull request to master because we want to include this bug in the stable version.


abaranch:
Build is failing because CodeCode is not in master yet.

SergeyKanzhelev:
:shipit:

Issue #102, opened 2015-11-09
Do not allow to send a NaN and +- infinity as a value of measurments


SergeyKanzhelev:
DC wouldn't accept measurements when it is set to NaN.

...On second though maybe we need to make DC accept it as it's unclear how to tell that the metric is bad... It's better to send it than to throw exception or send zero instead.

SergeyKanzhelev:
Converted to the bug as Breeze do not allow NaN and Infinity any longer. Need to send zeroes instead. @abaranch can we take it for stable 2.0?

abaranch:
Sure

PR #198, opened 2016-02-23
Disallow Nan, +-Infinity measurements

abaranch:
#102 
Pull request to master because we want to include this bug in the stable version.

### DIFF
--- a/src/Core/Managed/Shared/Extensibility/Implementation/Property.cs
+++ b/src/Core/Managed/Shared/Extensibility/Implementation/Property.cs
@@ -140,14 +140,8 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
                     dictionary.Remove(entry.Key);
 
                     string sanitizedKey = SanitizeKey(entry.Key, dictionary);
-
-                    // Disallow Nan and Infinity since Breeze does not accept it
-                    double sanitizeValue = entry.Value;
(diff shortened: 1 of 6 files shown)


abaranch:
Build is failing because CodeCode is not in master yet.

src/Core/Managed/Shared/Extensibility/Implementation/Property.cs
```diff
@@ -141,8 +141,15 @@ public static void SanitizeMeasurements(this IDictionary<string, double> diction
 
                     string sanitizedKey = SanitizeKey(entry.Key, dictionary);
 
+                    // Disallow Nan and Infinity since Breeze does not accept it
+                    double sanitizeValue = entry.Value;
```
SergeyKanzhelev: do we have any other double values anywhere? I believe Metric may have it and Measurements colleciton. Worth checking maybe we have anything else that potentially can become NaN

abaranch: We have it also in Metric and in Sampling. I updated the first one and I do mot think we need to update the second.

SergeyKanzhelev:
:shipit: I'd suggest to sanitize `Sampling` anyway. It may already be sanitized by `>0` and '<100' though

abaranch:
We already check sampling: https://github.com/Microsoft/ApplicationInsights-dotnet/blob/master/src/Core/Managed/Shared/Extensibility/Implementation/Telemetry.cs#L17

PR #205, opened 2016-03-05
Update generations to support netstandard

kartang:
1) dnxcore50 is updated to netstandardapp1.5-> In reference to the following links:
https://github.com/NuGet/Home/issues/1956


kartang:
/ @davidfowl

upendras:
:shipit:

src/Core/NuGet/Core.NuGet-Beta/Package.nuspec
```diff
@@ -24,7 +24,7 @@
       </group>
       <group targetFramework="net46">
       </group>
-      <group targetFramework="dnxcore50">
+      <group targetFramework=".NETStandardApp,Version=v1.5">
         <dependency id="System.Collections" version="4.0.10-beta-22816" />
```
SergeyKanzhelev: As we shipping stable version of NuGet all references should be changed to the latest stable version of dependencies. No betas.

src/Core/NuGet/Core.NuGet-Beta/Package.nuspec
```diff
@@ -24,7 +24,7 @@
       </group>
       <group targetFramework="net46">
       </group>
-      <group targetFramework="dnxcore50">
+      <group targetFramework=".NETStandardApp,Version=v1.5">
```
SergeyKanzhelev: Why `netstandardapp`, not `netstandard`? It should work on any platform and cannot be run by itself.

src/Core/NuGet/Core.NuGet-Beta/Package.nuspec
kartang: Will change that. Thanks!

SergeyKanzhelev: Also as we making it stable - remove this definition from repo: https://github.com/Microsoft/ApplicationInsights-dotnet/tree/master/src/Core/NuGet/Core

src/Core/NuGet/Core/Package.nuspec
```diff
@@ -24,14 +24,40 @@
       </group>
       <group targetFramework="net46">
       </group>
+      <group targetFramework=".NETStandard,Version=v1.5">
```
SergeyKanzhelev: So you changed to NetStandard now. @davidfowl - this is the main question. Should we use netstandard or netstandardAPP here?

SergeyKanzhelev:
:shipit:

PR #215, opened 2016-03-31
Support partial success response code (206)

abaranch:


Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/ErrorHandlingTransmissionPolicyTest.cs
```diff
@@ -97,6 +100,249 @@ public void ResumesTransmissionSenderAfterPauseDuration()
...
+                    OnEnqueue = t => { enqueuedTransmissions.Add(t); }
+                };
+
+                var policy = new ErrorHandlingTransmissionPolicy();
+                policy.Initialize(transmitter);
+
+                var items = new List<ITelemetry> { new EventTelemetry(), new EventTelemetry() };
+                Transmission transmission = new Transmission(new Uri("http://uri"), items, "type", "encoding");
+
+                string breezeResponse =
```
dnduffy: Other than the breezeResponse text this test method is identical to the one above and could be combined into a helper method that is passed the differing text.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Shared.projitems
```diff
@@ -17,6 +17,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\ApplicationFolderProvider.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\ApplicationLifecycleTransmissionPolicy.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Implementation\ApplicationStoppingEventArgs.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Implementation\BreezeResponse.cs" />
```
AlexBulankou: Should we not reference Breeze which is internal codename in public sources? Perhaps call it something neutral, e.g. DataCollector?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ErrorHandlingTransmissionPolicy.cs
```diff
@@ -63,47 +75,141 @@ private void HandleTransmissionSentEvent(object sender, TransmissionProcessedEve
...
                 }
+            }
+        }
 
+        private IDictionary<int, string> ParsePartialSuccessResponse(Transmission initialTransmission, string response)
+        {
+            BreezeResponse breezeResponse;
+            try
+            {
+                breezeResponse = new JavaScriptSerializer().Deserialize<BreezeResponse>(response);
```
AlexBulankou: store a static instance of JavaScriptSerializer as variable to reuse every time instead of create a new instance for every call?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ErrorHandlingTransmissionPolicy.cs
```diff
@@ -63,47 +75,141 @@ private void HandleTransmissionSentEvent(object sender, TransmissionProcessedEve
...
+                    .Deserialize(initialTransmission.Content)
+                    .Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+
+                newTransmissions = new Dictionary<int, string>();
+
+                foreach (var error in breezeResponse.Errors)
+                {
+                    if (error != null)
+                    {
+                        if (error.Index >= items.Length)
```
AlexBulankou: || error.Index < 0 ?

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
```diff
@@ -54,6 +54,51 @@ public static byte[] Serialize(IEnumerable<ITelemetry> telemetryItems, bool comp
...
+        }
+
+        /// <summary>
+        /// Deserializes and decompress the telemetry items into a JSON string.
+        /// </summary>
+        /// <param name="telemetryItemsData">Serialized telemetry items.</param>
+        /// <param name="compress">Should deserialization also perform decompression.</param>
+        /// <returns>Telemetry items serialized as a string.</returns>
+        public static string Deserialize(byte[] telemetryItemsData, bool compress = true)
+        {
```
AlexBulankou: Do we need unit tests to cover specifically ConvertToByArray and Deserialize?

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
```diff
@@ -54,6 +54,51 @@ public static byte[] Serialize(IEnumerable<ITelemetry> telemetryItems, bool comp
...
+        [SuppressMessage("Microsoft.Usage", "CA2202:Do not dispose objects multiple times", Justification = "Disposing a MemoryStream multiple times is harmless.")]
+        public static byte[] ConvertToByteArray(string telemetryItems, bool compress = true)
+        {
+            if (string.IsNullOrEmpty(telemetryItems))
+            {
+                throw new ArgumentNullException("telemetryItems");
+            }
+
+            var memoryStream = new MemoryStream();
+            using (Stream compressedStream = compress ? CreateCompressedStream(memoryStream) : memoryStream)
```
AlexBulankou: why not use using(... during memoryStream creation?

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
```diff
@@ -54,6 +54,51 @@ public static byte[] Serialize(IEnumerable<ITelemetry> telemetryItems, bool comp
...
+        /// </summary>
+        /// <param name="telemetryItemsData">Serialized telemetry items.</param>
+        /// <param name="compress">Should deserialization also perform decompression.</param>
+        /// <returns>Telemetry items serialized as a string.</returns>
+        public static string Deserialize(byte[] telemetryItemsData, bool compress = true)
+        {
+            var memoryStream = new MemoryStream(telemetryItemsData);
+            
+            using (Stream decompressedStream = new GZipStream(memoryStream, CompressionMode.Decompress))
+            {
```
AlexBulankou: use CreateCompressedStream instead of new GZipStream(... ?

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
```diff
@@ -54,6 +54,51 @@ public static byte[] Serialize(IEnumerable<ITelemetry> telemetryItems, bool comp
...
 
         /// <summary>
+        /// Converts serialized telemetry items to a byte array.
+        /// </summary>
+        /// <param name="telemetryItems">Serialized telemetry items.</param>
+        /// <param name="compress">Should serialization also perform compression.</param>
+        /// <returns>The compressed and serialized telemetry items.</returns>
+        [SuppressMessage("Microsoft.Usage", "CA2202:Do not dispose objects multiple times", Justification = "Disposing a MemoryStream multiple times is harmless.")]
+        public static byte[] ConvertToByteArray(string telemetryItems, bool compress = true)
+        {
```
AlexBulankou: Do we need ConvertToByArray - how is it different from Serialize method we already have?

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/ErrorHandlingTransmissionPolicyTest.cs
abaranch: I think it would be more difficult to understand what test is doing if I add helper methods, parameters, start using string.format and add more elements in "items" depending on the parameter...

src/TelemetryChannels/ServerTelemetryChannel/Shared/Shared.projitems
abaranch: Class is internal, but fine, I can rename it.

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
abaranch: Because I do not have list of telemetry items, I have a string and I need it as byte array to create a new Tranmission

AlexBulankou: got it, makes sense

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/BreezeResponse.cs
```diff
@@ -0,0 +1,30 @@
...
+        public int ItemsReceived { get; set; }
+
+        [DataMember(Name = "itemsAccepted")]
+        public int ItemsAccepted { get; set; }
+
+        [DataMember(Name = "errors")]
+        public Error[] Errors { get; set; }
+
+        [DataContract]
+        public class Error
```
dnduffy: Should we follow the convention of putting public classes in their own file?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryChannelEventSource.cs
```diff
@@ -30,6 +30,24 @@ public bool IsVerboseEnabled
             }
         }
 
+        [Event(5, Message = "Backend response {1} was not parsed. Some items may be dropped: {0}.", Level = EventLevel.Warning)]
+        public void BreezeResponseWasNotParsedWarning(string exception, string response, string appDomainName = "Incorrect")
+        {
+            this.WriteEvent(5, exception ?? string.Empty, response ?? string.Empty, this.ApplicationName);
```
dnduffy: What do the "magic numbers" 5, 6, and 7 represent?  Are there constants for these with descriptive names?

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
abaranch: I need Decompress, not Compress parameter that is used in that method. I would actually remove this helper method. As I understand we had it because we were supporting not compressed streams before which is not the case any more.

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
abaranch: Gzip steam owns MemoryStream and disposes it: http://stackoverflow.com/questions/8182598/correct-idisposable-implementation-for-this-code

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/BreezeResponse.cs
abaranch: I did not get it... It is in its own file and class matches file name.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryChannelEventSource.cs
abaranch: They are unique event ids. EventsSource is our unified mechanizm for tracing internal telemetry and there are certain rules how you do that. One is that you need to provide event is and you cannot use consts in the attribete so you just use same number in the attribute and inside the method.

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
abaranch: Ok, we actually use uncompressed stream in couple unit tests so I would leave existing code as is.

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
abaranch: Added

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/BreezeResponse.cs
dnduffy: The outer class BreezeResponse has the matching file name but there is an inner class named Error which is public.

dnduffy:
:shipit:

AlexBulankou:
:shipit:

PR #217, opened 2016-04-08
Return Throttling support back

abaranch:
1. 429 is supported by backend now; 439 will be supported in near future (as statuscode for the whole package). So I return ThrottlingTransmissionPolicy back that I cut with previous checkin
2. Before in ThrottlingTransmissionPolicy  we used "Retry-After" header and if it is not provided dropped items. Now we will use exponential retry policy.
3. Same check for Retry-After header before exponential retry is added for 206 and logic is moved from ErrorTransmissionPolicy to a new PartialSuccessTransmissionPolicy
4. Most tests that you see as cut actually were moved  to a different class because logic also moved (like from ErrorTransmissionPolicy to a new PartialSuccessTransmissionPolicy or GetBackoffTime to TransmissionPolicy)


src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionPolicy.cs
```diff
@@ -46,6 +53,45 @@ public virtual void Initialize(Transmitter transmitter)
...
+                double delayInSeconds;
+
+                if (this.ConsecutiveErrors <= 1)
+                {
+                    delayInSeconds = SlotDelayInSeconds;
+                }
+                else
+                {
+                    double backOffSlot = (Math.Pow(2, this.ConsecutiveErrors) - 1) / 2;
+                    var backOffDelay = this.random.Next(1, (int)Math.Min(backOffSlot * SlotDelayInSeconds, int.MaxValue));
```
AlexBulankou: What does this give us:
(int)Math.Min(backOffSlot * SlotDelayInSeconds, int.MaxValue)
why not just backOffSlot * SlotDelayInSeconds
and also why do we need to randomize it?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/PartialSuccessTransmissionPolicy.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                        TelemetryChannelEventSource.Log.ItemRejectedByEndpointWarning(error.Message);
+
+                        if (error.StatusCode == ResponseStatusCodes.RequestTimeout ||
+                            error.StatusCode == ResponseStatusCodes.ServiceUnavailable ||
+                            error.StatusCode == ResponseStatusCodes.InternalServerError ||
+                            error.StatusCode == ResponseStatusCodes.ResponseCodeTooManyRequests ||
+                            error.StatusCode == ResponseStatusCodes.ResponseCodeTooManyRequestsOverExtendedTime)
+                        {
+                            if (string.IsNullOrEmpty(newTransmissions))
```
AlexBulankou: Looks like you are not initializing newTransmissions anywhere so it will always remain null. Did you forget to call ParsePartialSuccessResponse?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ResponseStatusCodes.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation
+{
+    internal class ResponseStatusCodes
+    {
+        public const int RequestTimeout = 408;
+        public const int ResponseCodeTooManyRequests = 429;
+        public const int ResponseCodeTooManyRequestsOverExtendedTime = 439;
+        public const int InternalServerError = 500;
+        public const int ServiceUnavailable = 503;
+    }
```
AlexBulankou: Add 206 to this list as well?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/PartialSuccessTransmissionPolicy.cs
```diff
@@ -0,0 +1,153 @@
...
+
+#if NET45
+    using TaskEx = System.Threading.Tasks.Task;
+#endif
+
+    internal class PartialSuccessTransmissionPolicy : TransmissionPolicy, IDisposable
+    {
+        private readonly JavaScriptSerializer serializer = new JavaScriptSerializer();
+
+        private TaskTimer pauseTimer = new TaskTimer { Delay = TimeSpan.FromSeconds(SlotDelayInSeconds) };
```
AlexBulankou: We don't want pauseTimer to be in the base TransmissionPolicy? Looks like it is used in multiple implementations of TransmissionPolicy

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ThrottlingTransmissionPolicy.cs
```diff
@@ -12,19 +12,7 @@
...
-        /// be paused before attempting to resume transmission after a network error is detected.
-        /// </summary>
-        public TimeSpan PauseDuration
-        {
-            get { return this.pauseTimer.Delay; }
-        }
+        private readonly TaskTimer pauseTimer = new TaskTimer { Delay = TimeSpan.FromSeconds(SlotDelayInSeconds) };
 
         public void Dispose()
         {
```
AlexBulankou: what about disposable pattern? You know, like it is done in the other policy
```
public void Dispose()
        {
             this.Dispose(true);
             GC.SuppressFinalize(this);
         }


        private void Dispose(bool disposing)
         {
             if (disposing)
             {
                 if (this.pauseTimer != null)
                 {
                     this.pauseTimer.Dispose();
                     this.pauseTimer = null;
                 }
             }
         }
```

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/PartialSuccessTransmissionPolicy.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        public override void Initialize(Transmitter transmitter)
+        {
+            base.Initialize(transmitter);
+            transmitter.TransmissionSent += this.HandleTransmissionSentEvent;
```
AlexBulankou: Looks like every policy is subscribing for transmitter.TransmissionSent. Why not move it base class and make HandleTransmissionSentEvent method protected virtual?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionPolicy.cs
abaranch: I did not change this algorithm. See comment above. Please read explanation from the comment, it is implemented accordingly.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ResponseStatusCodes.cs
abaranch: Resolved

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/PartialSuccessTransmissionPolicy.cs
abaranch: In half. Base class does not need it and I do not want to create and dispose it if I do not need it; it is an expensive resource

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ThrottlingTransmissionPolicy.cs
abaranch: Resolved

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/PartialSuccessTransmissionPolicy.cs
abaranch: Not all, only those that I touched which is half.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionPolicy.cs
AlexBulankou: Thanks, got it, trying to avoid all clients retrying at the same time DC failure periods.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/PartialSuccessTransmissionPolicy.cs
AlexBulankou: OK, make sense

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/PartialSuccessTransmissionPolicy.cs
AlexBulankou: OK

AlexBulankou:
:shipit:

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TransmissionPolicyTest.cs
```diff
@@ -98,6 +102,76 @@ public void MaxStorageCapacityCanBeSetByPolicy()
...
+                Assert.Equal(TimeSpan.FromSeconds(10), delay);
+            }
+
+            [TestMethod]
+            public void UpperBoundOfDelayIsMaxDelay()
+            {
+                var policy = new TestableTransmissionPolicy();
+                policy.ConsecutiveErrors = int.MaxValue;
+                TimeSpan delay = policy.GetBackOffTime(new WebHeaderCollection());
+                Assert.InRange(delay, TimeSpan.FromSeconds(10), TimeSpan.FromSeconds(3600));
```
dnduffy: How is being in the range proving the upper bound?

dnduffy:
:shipit:

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TransmissionPolicyTest.cs
abaranch: We check that it is not more than this interval. Since we use Random for generation it is best we can do.

PR #218, opened 2016-04-08
Add Sdk version by default to all telemetry. Delete SdkVersionTelemet…

abaranch:
…ryInitializer so it cannot be removed by customer which would result in not adding the prefix.


AlexBulankou:
:shipit:

PR #219, opened 2016-04-12
IndexOutOfRangeException in PortalQueueSender

abaranch:
Resolving https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/69

### DIFF
--- a/Test/CoreSDK.Test/Shared/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModuleTest.cs
+++ b/Test/CoreSDK.Test/Shared/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModuleTest.cs
@@ -8,7 +8,10 @@
 #if NET40
     using Microsoft.Diagnostics.Tracing;
 #endif
+    using System.Threading;
+    using System.Threading.Tasks;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Assert = Xunit.Assert;
 
     [TestClass]
     public class DiagnosticsTelemetryModuleTest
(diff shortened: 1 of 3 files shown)


Test/CoreSDK.Test/Shared/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModuleTest.cs
```diff
@@ -53,20 +56,47 @@ public void TestDiagnosticsModuleSetSeverity()
...
         [TestMethod]
         public void TestDiagnosticModuleDoesNotThrowIfInitailizedTwice()
         {
-            DiagnosticsTelemetryModule module = new DiagnosticsTelemetryModule();
-            module.Initialize(TelemetryConfiguration.Active);
-            module.Initialize(TelemetryConfiguration.Active);
+            using (DiagnosticsTelemetryModule module = new DiagnosticsTelemetryModule())
+            {
+                module.Initialize(new TelemetryConfiguration());
+                module.Initialize(new TelemetryConfiguration());
```
dnduffy: Why two of the same line?

Test/CoreSDK.Test/Shared/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModuleTest.cs
```diff
@@ -53,20 +56,47 @@ public void TestDiagnosticsModuleSetSeverity()
...
+                        queueSender.Send(new TraceEvent());
+                        Thread.Sleep(1);
+                    }
+                }, cancellationTokenSource.Token);
+
+
+                Assert.DoesNotThrow(() => module.Initialize(new TelemetryConfiguration()));
+
+                cancellationTokenSource.Cancel();
+                cancellationTokenSource.Dispose();
```
dnduffy: How about wrapping the cancellationTokenSource in a using that would call Dispose automatically?

Test/CoreSDK.Test/Shared/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModuleTest.cs
abaranch: Because test checks that Initialize would not throw if called twice. In the test I changed only TC.Active on empty configuration. It is better since recently we started adding same module by defaut in TC.Active

Test/CoreSDK.Test/Shared/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModuleTest.cs
```diff
@@ -53,20 +56,47 @@ public void TestDiagnosticsModuleSetSeverity()
...
+
+        [TestMethod]
+        public void DiagnosticModuleDoesNotThrowIfQueueSenderContinuesRecieveEvents()
+        {
+            using (DiagnosticsTelemetryModule module = new DiagnosticsTelemetryModule())
+            {
+                var queueSender = module.Senders.OfType<PortalDiagnosticsQueueSender>().First();
+
+                CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+                TaskEx.Run(() =>
```
dnduffy: I think a race condition can occur here, how do you know the task thread will actually get run by the time you get to the call to Cancel?

dnduffy:
:shipit:

PR #232, opened 2016-05-24
Add additional verbose channel tracing

abaranch:


Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/ErrorHandlingTransmissionPolicyTest.cs
```diff
@@ -219,7 +285,7 @@ private class TestableErrorHandlingTransmissionPolicy : ErrorHandlingTransmissio
         {
             public TimeSpan BackOffTime { get; set; }
 
-            public override TimeSpan GetBackOffTime(NameValueCollection headers = null)
+            protected override TimeSpan GetBackOffTime(NameValueCollection headers = null)
```
dnduffy: Why is this method protected but the plain property above is public?  And why have both?

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -14,6 +14,15 @@ internal sealed class CoreEventSource : EventSource
 
         private readonly ApplicationNameProvider nameProvider = new ApplicationNameProvider();
 
+        public bool IsVerboseEnabled
```
dnduffy: Do you want helpers for the other levels too?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/PartialSuccessTransmissionPolicy.cs
```diff
@@ -14,9 +14,9 @@
 
     internal class PartialSuccessTransmissionPolicy : TransmissionPolicy, IDisposable
     {
-        private readonly JavaScriptSerializer serializer = new JavaScriptSerializer();
+        private TaskTimer pauseTimer = new TaskTimer { Delay = TimeSpan.FromSeconds(TransmissionPolicyHelpers.SlotDelayInSeconds) };
 
-        private TaskTimer pauseTimer = new TaskTimer { Delay = TimeSpan.FromSeconds(SlotDelayInSeconds) };
+        public int ConsecutiveErrors { get; set; }
```
dnduffy: I was just reading the retry code earlier today and I have a curiosity question; is the ConsecutiveErrors count effectively per transmission or is the policy instance reused for multiple transmissions?

dnduffy:
:shipit:

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/ErrorHandlingTransmissionPolicyTest.cs
abaranch: That is a test class that allows to override Backoff time in tests. Method is protected  because that is what we have in the ErrorHandlingTransmissionPolicy class. public property is public so we can easily set what we want in unit tests 

---
In reply to: [64498867](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/232#discussion_r64498867) [](ancestors = 64498867)

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
abaranch: We never had a need for them. IsVeroseEnabled is used when additional tracing has performance implications. 

---
In reply to: [64499033](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/232#discussion_r64499033) [](ancestors = 64499033)

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/PartialSuccessTransmissionPolicy.cs
abaranch: It is reused. It keeps the count of consecutive errors so it can increase delay internal because we use exponential backoff. 

---
In reply to: [64499637](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/232#discussion_r64499637) [](ancestors = 64499637)

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/ErrorHandlingTransmissionPolicyTest.cs
dnduffy: Since this is a test class it may not be an issue but from a design perspective the property seems to circumvent the helper method.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/PartialSuccessTransmissionPolicy.cs
dnduffy: What happens when several unrelated retries occur over time, does it keep getting longer of a delay even if they are unrelated and occur far apart?

abaranch: Yes, it will get longer even if they are unrelated.

---
In reply to: [64651855](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/232#discussion_r64651855) [](ancestors = 64651855)

PR #234, opened 2016-06-01
Change version format from x.x.x.x to x.x.x-x

abaranch:


kartang:
We might also think about using the nuget versions as the release versions instead of build numbers. Build numbers could be used internally, as opposed to released packages?

kartang:
:shipit:

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -432,18 +432,22 @@ public void Flush()
...
+            
 #else
-            return VersionPrefix + typeof(TelemetryClient).GetTypeInfo().Assembly.GetCustomAttributes<AssemblyFileVersionAttribute>()
+            string versionStr = typeof(TelemetryClient).GetTypeInfo().Assembly.GetCustomAttributes<AssemblyFileVersionAttribute>()
                     .First()
                     .Version;
 #endif
+
+            Version version = new Version(versionStr);
+            return VersionPrefix + version.ToString(3) + "-" + version.Revision;
```
dnduffy: Why hyphen revision?
The version.ToString(3) should give you major.minor.revision so this will concatenate -revision on again.

abaranch:
Since we need it in run time we would either need to hardcode it and do not forget to update on each release or add some complicated logic to parse and substitute it during compilation. I do not think it worth the effort.

---
In reply to: [223155781](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/234#issuecomment-223155781) [](ancestors = 223155781)

src/Core/Managed/Shared/TelemetryClient.cs
abaranch: When we have version 2.1.0.25000 than 25000 is a revision and version.ToString(3) gives 2.1.0. That is verified by unit test.

---
In reply to: [65466084](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/234#discussion_r65466084) [](ancestors = 65466084)

dnduffy: My apologies for the mistake, I was thinking that .NET used Revision like Patch from semver.org and that it was Build which was the high incrementing number and not the other way around.

PR #239, opened 2016-06-06
Add check against Validate endpoint

abaranch:


Test/CoreSDK.Test/Shared/TelemetryClientTest.cs
```diff
@@ -786,7 +791,70 @@ public void AllTelemetryIsSentWithDefaultSamplingRate()
...
+            EventTelemetry telemetry1 = new EventTelemetry();
+            MetricTelemetry telemetry2 = new MetricTelemetry();
+            DependencyTelemetry telemetry3 = new DependencyTelemetry();
+            ExceptionTelemetry telemetry4 = new ExceptionTelemetry();
+            MetricTelemetry telemetry5 = new MetricTelemetry();
+            PageViewTelemetry telemetry6 = new PageViewTelemetry();
+            PerformanceCounterTelemetry telemetry7 = new PerformanceCounterTelemetry();
+            RequestTelemetry telemetry8 = new RequestTelemetry();
+            SessionStateTelemetry telemetry9 = new SessionStateTelemetry();
+            TraceTelemetry telemetry10 = new TraceTelemetry();
```
dnduffy: nitpick: I would love to see clearer names than 1 through 10, even if it is just the camel case type name.

dnduffy:
:shipit:

Test/CoreSDK.Test/Shared/TelemetryClientTest.cs
abaranch: I think for this unit test it is fine.

---
In reply to: [65974024](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/239#discussion_r65974024) [](ancestors = 65974024)

PR #241, opened 2016-06-08
Switched to using reflection to create the desired condition for testing an expected exception.

dnduffy:
Switched to using reflection to create the desired condition for testing an expected exception.

### DIFF
--- a/Test/CoreSDK.Test/Shared/Channel/TransmissionTest.cs
+++ b/Test/CoreSDK.Test/Shared/Channel/TransmissionTest.cs
@@ -159,12 +159,7 @@
             {
                 AsyncTest.Run(async () =>
                 {
-                    var request = new StubWebRequest();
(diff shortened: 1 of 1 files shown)


Test/CoreSDK.Test/Shared/Channel/TransmissionTest.cs
```diff
@@ -160,11 +161,12 @@ public void ThrowsInvalidOperationExceptionWhenTransmissionIsAlreadySending()
                 {
                     var request = new StubWebRequest();
                     request.OnBeginGetRequestStream = (callback, state) => TaskEx.Delay(TimeSpan.FromMilliseconds(10)).AsAsyncResult(callback, request);
```
abaranch: Do you need a delay now? Do you need to override this method at all?

abaranch:
:shipit:

Test/CoreSDK.Test/Shared/Channel/TransmissionTest.cs
dnduffy: Good catch, I have removed the "dead code".

Issue #223, opened 2016-04-26
Make sampling be configurable by telemetry types


abaranch:
We would like to add a possibility to exclude certain telemetry types from sampling.

**Proposal**:
Make sampling telemetry processors (both Fixed rate and Adaptive) configurable.
We would like to add a configuration setting that indicates what telemetry types would be sampled.

**Defaults**:
By default all telemetry items except MetricTelemetry would be sampled (as before).

**Example:**

```
<Add Type="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.AdaptiveSamplingTelemetryProcessor, Microsoft.AI.ServerTelemetryChannel">
    <SampledTelemetry>Dependency;Exception;PageView;PerformanceCounter;Request</SampledTelemetry>
</Add>
```

```
<Add Type="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.SamplingTelemetryProcessor, Microsoft.AI.ServerTelemetryChannel">
    <SampledTelemetry>Dependency;Exception;PageView;PerformanceCounter;Request</SampledTelemetry>
</Add>
```

**SampledTelemetry possible values** (these are envelope type names except for dependency):

- Dependency

- Event

- Exception

- Metric

- PageView

- PerformanceCounter

- Request

- Message

Note that with proposed changes you can add multiple sampling telemetry processors and configure samling settings separately for different telemetry types. For example you can use fixed rate sampling for traces and adaptive sampling for everything else.

kartang:
Will there be possible correlation scenario where we miss the correlation between different event types, as we no more sample at user ID/operation id level, which assures that all telemetry for a particular operation is either in or out?

abaranch:
We do not change sampling algorithm. If item needs to be sampled we still use user, operation and session as before.

AlexBulankou:
Instead of `Message`, I would call it `Trace`, I think this is more known with customers.
Also just wanted to include the full snippet of how adaptive sampling telemetry processor will look, including MaxTelemetryItemsPerSecond property

```
<Add Type="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.AdaptiveSamplingTelemetryProcessor, Microsoft.AI.ServerTelemetryChannel">
      <MaxTelemetryItemsPerSecond>5</MaxTelemetryItemsPerSecond>      
      <SampledTelemetry>Dependency;Exception;PageView;PerformanceCounter;Request </SampledTelemetry>
[...]

PR #243, opened 2016-06-16
Make sampling be configurable by telemetry types

abaranch:
#223

### DIFF
--- a/Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryProcessorChainBuilderExtensionsTest.cs
+++ b/Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryProcessorChainBuilderExtensionsTest.cs
@@ -30,6 +30,12 @@
         }
 
         [TestMethod]
+        public void UseSamplingWithExcluedTypesParameterThrowsArgumentNullExceptionBuilderIsNull()
+        {
+            Assert.Throws<ArgumentNullException>(() => TelemetryProcessorChainBuilderExtensions.UseSampling(null, 5, "request"));
+        }
+
+        [TestMethod]
         public void UseSamplingSetsAddsSamplingProcessorToTheChainWithCorrectExcludedTypes()
         {
             var tc = new TelemetryConfiguration { TelemetryChannel = new StubTelemetryChannel() };
@@ -56,6 +62,12 @@
             channelBuilder.Build();
 
             Assert.IsType<AdaptiveSamplingTelemetryProcessor>(tc.TelemetryProcessorChain.FirstTelemetryProcessor);
+        }
+
+        [TestMethod]
+        public void UseAdaptiveSamplingWithExcludedTypesThrowsArgumentNullExceptionBuilderIsNull()
+        {
+            Assert.Throws<ArgumentNullException>(() => TelemetryProcessorChainBuilderExtensions.UseAdaptiveSampling(null, "request"));
         }
 
         [TestMethod]
--- a/Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryProcessorChainBuilderExtensionsTest.cs
+++ b/Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryProcessorChainBuilderExtensionsTest.cs
@@ -1,0 +1,124 @@
+﻿namespace Microsoft.ApplicationInsights.WindowsServer.Channel.Implementation
+{
+    using System;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Web.TestFramework;
+    using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    
+    using Assert = Xunit.Assert;
+
+    [TestClass]
+    public class TelemetryProcessorChainBuilderExtensionsTest
+    {
+        [TestMethod]
+        public void UseSamplingThrowsArgumentNullExceptionBuilderIsNull()
+        {
+            Assert.Throws<ArgumentNullException>(() => TelemetryProcessorChainBuilderExtensions.UseSampling(null, 5));
+        }
+
+        [TestMethod]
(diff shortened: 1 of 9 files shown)


Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryProcessorChainBuilderExtensionsTest.cs
```diff
@@ -0,0 +1,172 @@
...
+            AdaptiveSamplingPercentageEvaluatedCallback callback = (second, percentage, samplingPercentage, changed, estimatorSettings) => { };
+
+            var tc = new TelemetryConfiguration { TelemetryChannel = new StubTelemetryChannel() };
+            var channelBuilder = new TelemetryProcessorChainBuilder(tc);
+            channelBuilder.UseAdaptiveSampling(settings, callback, "request");
+            channelBuilder.Build();
+
+            Assert.Equal(13, ((AdaptiveSamplingTelemetryProcessor)tc.TelemetryProcessorChain.FirstTelemetryProcessor).MaxSamplingPercentage);
+            Assert.Equal("request", ((AdaptiveSamplingTelemetryProcessor)tc.TelemetryProcessorChain.FirstTelemetryProcessor).ExcludedTypes);
+        }
```
dnduffy: It looks like there is some overlap between tests.

Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
```diff
@@ -64,79 +65,188 @@ public void TelemetryItemHasSamplingPercentageSet()
...
         
         [TestMethod]
         public void PerformanceCounterTelemetryIsNotSubjectToSampling()
         {
             TelemetryTypeDoesNotSupportSampling(
-                (telemetryProcessors) => telemetryProcessors.Process(new PerformanceCounterTelemetry("category", "counter", "instance", 1.0)));
+                telemetryProcessors =>
+                {
+                    telemetryProcessors.Process(new PerformanceCounterTelemetry("category", "counter", "instance", 1.0));
+                    return 1;
```
dnduffy: Just curious, what are the "return 1" statements for?

src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
```diff
@@ -26,8 +41,71 @@ public SamplingTelemetryProcessor(ITelemetryProcessor next)
...
+                        {
+                            newExcludedTypesHashSet.Add(typeof(RequestTelemetry));
+                            continue;
+                        }
+
+                        if (string.Compare(item, TraceTelemetryName, StringComparison.OrdinalIgnoreCase) == 0)
+                        {
+                            newExcludedTypesHashSet.Add(typeof(TraceTelemetry));
+                            continue;
+                        }
```
dnduffy: The above might be shorter and easier to read as a Dictionary<string, Type> lookup.

src/TelemetryChannels/ServerTelemetryChannel/Shared/TelemetryChannelBuilderExtensions.cs
```diff
@@ -42,7 +90,33 @@ public static TelemetryProcessorChainBuilder UseAdaptiveSampling(this TelemetryP
...
+        }
+
+        /// <summary>
+        /// Adds <see cref="AdaptiveSamplingTelemetryProcessor"/> to the <see cref="TelemetryProcessorChainBuilder" />
+        /// </summary>
+        /// <param name="builder">Instance of <see cref="TelemetryProcessorChainBuilder"/></param>
+        /// <param name="maxTelemetryItemsPerSecond">Maximum number of telemetry items to be generated on this application instance.</param>
+        /// <param name="excludedTypes">Semicolon separated list of types that should not be sampled.</param>
+        /// <return>Instance of <see cref="TelemetryProcessorChainBuilder"/>.</return>
+        public static TelemetryProcessorChainBuilder UseAdaptiveSampling(this TelemetryProcessorChainBuilder builder, double maxTelemetryItemsPerSecond, string excludedTypes)
```
dnduffy: Given the number of overloads needing to be added it might be time to consider having a single signature with a simple config object (e.g. property bag).

SergeyKanzhelev:
:shipit:

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryProcessorChainBuilderExtensionsTest.cs
abaranch: They test different extension methods

---
In reply to: [67446623](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/243#discussion_r67446623) [](ancestors = 67446623)

Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
abaranch: I did not want to change helper methods much. 1 means that 1 item was processed. In assert before there was a hardcoded expected number, now I use this return value and in some other tests call "Process" N times and return N.

---
In reply to: [67446947](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/243#discussion_r67446947) [](ancestors = 67446947)

dnduffy:
:shipit:

src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
abaranch: I do not need to deal with duplicates and it would work faster on lookup since I do not need to do ToString

---
In reply to: [67455785](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/243#discussion_r67455785) [](ancestors = 67455785)

dnduffy: There is no need for you to call ToString, you are already comparing "item" to a type name string and that is all the dictionary would do internally.  The type is the looked up value and would never need to be converted to a string, same as it is now.

PR #248, opened 2016-07-05
Add diagnostics event for backoff logic.

abaranch:
- Refactor server channel policies that use backoff logic to share same timer and counter of errors (Move logic to BackoffLogicManager that belongs to Transmittor)
- Post diagnostic message if backoff interval is more that 30 min and when backoff logic is disabled
TODO in the next pull request: Make 30 min interval overridable.


Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
```diff
@@ -0,0 +1,248 @@
...
+
+    public class BackoffLogicManagerTest
+    {
+        [TestClass]
+        public class GetBackendResponse
+        {
+            public void ReturnNullIfArgumentIsNull()
+            {
+                var manager = new BackoffLogicManager(TimeSpan.Zero);
+                Assert.Null(manager.GetBackendResponse(null));
```
dnduffy: These appear to be tests but they are not marked with the [TestMethod] attribute, why is that?

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
```diff
@@ -0,0 +1,248 @@
...
+    using Microsoft.ApplicationInsights.Web.TestFramework;
+    using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Helpers;
+    using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation;
+    
+    using Assert = Xunit.Assert;
+
+    public class BackoffLogicManagerTest
+    {
+        [TestClass]
+        public class GetBackendResponse
```
dnduffy: Is it necessary to have the nested classes or could they go into their own files?

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
```diff
@@ -0,0 +1,248 @@
...
+                string content = BackendResponseHelper.CreateBackendResponse(100, 1, new[] {"206"}, 84);
+
+                var manager = new BackoffLogicManager(TimeSpan.Zero);
+                var backendResponse = manager.GetBackendResponse(content);
+
+                Assert.Equal(1, backendResponse.ItemsAccepted);
+                Assert.Equal(100, backendResponse.ItemsReceived);
+                Assert.Equal(1, backendResponse.Errors.Length);
+                Assert.Equal(84, backendResponse.Errors[0].Index);
+                Assert.Equal(206, backendResponse.Errors[0].StatusCode);
```
dnduffy: I realize this is a test but I don't think the backend will return an error with a value of 206, only the overall request status code would be 206.

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
```diff
@@ -0,0 +1,248 @@
...
+            public void BackendResponseIsReturnedForCorrectContent()
+            {
+                string content = BackendResponseHelper.CreateBackendResponse(100, 1, new[] {"206"}, 84);
+
+                var manager = new BackoffLogicManager(TimeSpan.Zero);
+                var backendResponse = manager.GetBackendResponse(content);
+
+                Assert.Equal(1, backendResponse.ItemsAccepted);
+                Assert.Equal(100, backendResponse.ItemsReceived);
+                Assert.Equal(1, backendResponse.Errors.Length);
```
dnduffy: I think this might be confusing or misleading for someone else reading the code in future because if 100 items are sent and only 1 is accepted then wouldn't the length of the errors be 99?

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
```diff
@@ -0,0 +1,248 @@
...
+                manager.ScheduleRestore(string.Empty, () => null);
+                Assert.Equal(TimeSpan.FromSeconds(10), manager.CurrentDelay);
+            }
+
+            [TestMethod]
+            public void UpperBoundOfDelayIsMaxDelay()
+            {
+                var manager = new BackoffLogicManager(TimeSpan.Zero) { ConsecutiveErrors = int.MaxValue };
+                manager.ScheduleRestore(string.Empty, () => null);
+                Assert.InRange(manager.CurrentDelay, TimeSpan.FromSeconds(10), TimeSpan.FromSeconds(3600));
```
dnduffy: nitpick: you could make constants for the "magic" numbers.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryChannelEventSource.cs
```diff
@@ -30,6 +30,26 @@ public bool IsVerboseEnabled
             }
         }
 
+        // Verbosity is Error - so it is always sent to portal; Keyword is Diagnostics so throttling is not applied.
+        [Event(1,
+            Message = "It is a diagnostics message. Backoff logic was disabled. Transmission will be resolved.",
```
dnduffy: nitpick: If this is a message that will be in the logs for us to trouble shoot with I might make it more concise: "Diagnostic message: backoff logic disabled, transmission will be resolved."...
Also, what is "resolved" meant to imply, is transmission is currently not resolving to an endpoint?

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
abaranch: We have a convention to create one test class for one product class and name of test class is name of product test class + "Test". Nested classes help organize what is being tested. For example for BackoffLogicManagerTest@GetBackendResponse tests all scenarios involving  BackoffLogicManager.GetBackendResponse 

---
In reply to: [69619157](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/248#discussion_r69619157) [](ancestors = 69619157)

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
abaranch: Nice catch :) Those tests were moved here from TransmissionPolicyHelperTests and I did not change them.  So it was missing for quite some time.

---
In reply to: [69618955](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/248#discussion_r69618955) [](ancestors = 69618955)

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
abaranch: For SDK it does not matter

---
In reply to: [69619459](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/248#discussion_r69619459) [](ancestors = 69619459)

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
abaranch: SDK should be able to handle this. If we know only about 1 error out of 99, we should retry only that item

---
In reply to: [69619689](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/248#discussion_r69619689) [](ancestors = 69619689)

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
abaranch: I would not want to test if product and test code uses same constant. If we change 3600 in product code I want someone to change this test so that he is doing this consciously. 

---
In reply to: [69628722](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/248#discussion_r69628722) [](ancestors = 69628722)

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryChannelEventSource.cs
abaranch: It means that we were not sending data and now we disable backoff logic and resolving sending data on the regular interval.

---
In reply to: [69644269](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/248#discussion_r69644269) [](ancestors = 69644269)

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
dnduffy: Yes of course, I was just thinking it might be confusing for future readers that the magic numbers don't match real counts from the backend.

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/BackoffLogicManagerTest.cs
dnduffy: Sorry, I didn't mean to suggest using the same constants in both as that would not catch the scenario you mention.  I agree that changes need to be made conscientiously.  I was thinking independent constants but it was just a nitpick.

dnduffy:
:shipit:

PR #250, opened 2016-07-06
Do not update ConsecutiveErrors often

abaranch:


src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/BackoffLogicManager.cs
```diff
@@ -21,19 +21,59 @@ internal class BackoffLogicManager : IDisposable
...
 
         /// <summary>
         /// Gets or sets the number of consecutive errors SDK transmitter got so far while sending telemetry to backend.
         /// </summary>
-        public int ConsecutiveErrors { get; set; }
+        public int ConsecutiveErrors
+        {
+            get { return this.consecutiveErrors; }
+
+            set
```
dnduffy: Exposing the counter and opening it to an arbitrary set value feels "odd".  I would expect that the internal counter would not be exposed in the programming API, even with a lock, and that any classes which consume the BackoffLogicManager would at most be given a "reset" method along with an "increment" or "report error" method which would internally just do an atomic increment or lock as needed.
Sorry, I meant to comment on the usage in the other PR.

abaranch: This code was moved form policies to manager without changing. I would not want to refactor it more since that would require rewriting all unit tests that we have now.

dnduffy: I see how it changed/moved, and I don't want to be annoying, but the change is such that it now looks architecturally weird to have what used to be a local field still being accessed that way even though now it is a property in another class; the encapsulation doesn't match the usage anymore.

dnduffy:
:shipit:

PR #254, opened 2016-07-19
fix WritePropertyDoubleWritesDoubleValueWithoutQuotationMark test

ki0ki0:
fix WritePropertyDoubleWritesDoubleValueWithoutQuotationMark test for some cultures

Some cultures uses comma as decimal symbol.
Test ignored current culture only for one of parameters.


SergeyKanzhelev:
Thanks!

Issue #256, opened 2016-07-21
ExcludedTypes for sampling doesn't include linked telemetry 


hallatore:
When a data point is excluded from sampling it's linked data points are not.

We are currently testing it on a site which has sample set to 5% and we are excluding exceptions. This feature is ideal to keep the cost down while providing good logging of exceptions, and I think it will be a good configuration for many.

Current behavior: I get every exception, but not always the linked request/dependencies/etc.
Expected behavior: Get every exception and it's linked request/dependencies/etc.

Ref #243

hallatore:
The issue I'm describing is very much dependent on Exception telemetry. It makes no sense if Dependency is the excluded telemetry.

The problem still remains, but maybe it could be resolved differently?

A good solution might be a way to tell the sampling to not sample an telemetry item. Then we can make a initializer that sets this value for stuff we don't want to sample. This would also have made it easy to create the ExcludeTypes behavior as an initializer.

abaranch:
I did not quite got your question. So let me try to explain how it currently works.
If you did not provide ExludedTypes SamplingProcessor will use common sampling rate for all telemetry items. If one telemetry in determined to be "in" all related items are "in", if "out" - than "out".
Now we got a requirement from customers that they want either to have different sampling rates for different types or all events of certain types not sampled. So let's say we exclude exceptions from sampling. That means that you get all your exceptions. But other types are still sampled with default rates. Which means that some exception will be correlated with requests and some not, just because requests are sampled and exceptions are not. 
It feels that your would like to have a different scenario that does not fit in this enhancement: if request failed and it has exception, I want it "in". Currently it is not supported. For this scenario I think this may be a workaround: http://apmtips.com/blog/2015/05/06/diy-data-sampling/

hallatore:
@abaranch The intial idea was not correct as it would only work in my specific scenario.

I've created a pull request #257 that addresses the needed feature. Moving discussion over to #257

PR #257, opened 2016-07-21
Added ability to manually skip sampling for individual telemetry items

hallatore:
This change makes it is possible to create initializers that skip the sampling for individual telemetry items.

Main reason for this change is to make it possible to skip request sampling on items with exceptions while still maintaining a strong sampling rate globally.

ref #256

### DIFF
--- a/Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
+++ b/Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
@@ -63,20 +63,34 @@
         }
 
         [TestMethod]
-        public void TelemetryItemIsSkippedBySampling()
+        public void TelemetryItemSamplingIsSkipped()
         {
             var sentTelemetry = new List<ITelemetry>();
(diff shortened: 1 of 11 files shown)


SergeyKanzhelev:
Thanks for PR. Can you please update it with the following:

1. PR should go to `develop`, not `master`
2. No need to have another field in the interface. Instead - can you please set `SamplingPercentage` to 100% and telemetry processor to not override it if it's set to 100%

This should solve your problem

hallatore:
@SergeyKanzhelev The default SamplingPercentage is already 100, so checking that would currently always override sampling.

SamplingPercentage is just used in these two places, so I think changing it to `double?` would be preferable if we want to avoid adding another field.
- https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Core/Managed/Shared/Extensibility/Implementation/Telemetry.cs#L16-L21
- https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs#L126

PR #258, opened 2016-07-21
Changed SamplingPercentag to be double? to support overriding samplers

hallatore:
With this change the sampler won't sample items that has it's SamplePercentage already set.

This fixes two issues:
- Allows for multiple samplers as a already touched item would be igored by the next.
- Allows for setting the sample percentage of a telemetry manually for scenarios where you'd want to override an item's sampling.

Ref #257 @SergeyKanzhelev

### DIFF
--- a/Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
+++ b/Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
@@ -63,7 +63,7 @@
         }
 
         [TestMethod]
-        public void TelemetryItemSamplingIsSkipped()
+        public void TelemetryItemSamplingIsSkippedWhenSetByUser()
         {
             var sentTelemetry = new List<ITelemetry>();
             var processor = new SamplingTelemetryProcessor(new StubTelemetryProcessor(null) { OnProcess = t => sentTelemetry.Add(t) })
@@ -76,21 +76,6 @@
             processor.Process(requestTelemetry);
 
             Assert.Equal(1, sentTelemetry.Count);
-        }
-
-        [TestMethod]
-        public void TelemetryItemSamplingIsNotSkippedByDefault()
-        {
-            var sentTelemetry = new List<ITelemetry>();
-            var processor = new SamplingTelemetryProcessor(new StubTelemetryProcessor(null) { OnProcess = t => sentTelemetry.Add(t) })
-            {
-                SamplingPercentage = 0
-            };
-
-            var requestTelemetry = new RequestTelemetry();
-            processor.Process(requestTelemetry);
-
-            Assert.Equal(0, sentTelemetry.Count);
         }
 
         [TestMethod]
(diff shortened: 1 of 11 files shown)


SergeyKanzhelev:
CC: @vitalyf007 - FYI

Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
```diff
@@ -61,7 +61,38 @@ public void TelemetryItemHasSamplingPercentageSet()
 
             Assert.Equal(20, ((ISupportSampling)sentTelemetry[0]).SamplingPercentage);
         }
-        
+
+        [TestMethod]
+        public void TelemetryItemSamplingIsSkipped()
```
SergeyKanzhelev: Please make a name more specific. `TelemetryItemSamplingIsSkippedWhenSetByUser`

SergeyKanzhelev:
Thanks again for PR. I like it better with nullable field.

hallatore:
This change also enables multiple samplers to sample more refined for those who need that. Only issue right now is that you would need to exclude all types except the one you want that sampler to do.

Example under would sample 50 % of exceptions and 10% on everything else.
```
<Add Type="...AdaptiveSamplingTelemetryProcessor">
    <SamplingPercentage>50</SamplingPercentage>
    <ExcludeTypes>Dependency;Event;PageView;Request;Trace</ExcludeTypes>
</Add>
<Add Type="...AdaptiveSamplingTelemetryProcessor">
    <SamplingPercentage>10</SamplingPercentage>
</Add>
```

Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
hallatore: Fixed.

SergeyKanzhelev:
Can you please elaborate on why this change enables the syntax with multiple sampling processors and it didn't work before?

hallatore:
@SergeyKanzhelev The sampler will ignore an item where SamplerPercentage is not null. So if the first sampler only works on exceptions, then all exceptions will be skipped by the next sampler.

This is the line that makes this chaining scenario possible. https://github.com/hallatore/ApplicationInsights-dotnet/blob/a82aa2d44b6b6a57040e460af4f4e610322414fb/src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs#L123

hallatore:
It also makes it less error prone as this example under show. With the old code it would only report 10 % of dependencies but tell app insights that number was 50%. So you would se only 20% of what you actually had.

```
<Add Type="...AdaptiveSamplingTelemetryProcessor">
    <SamplingPercentage>20</SamplingPercentage>
    <ExcludeTypes>Event;PageView;Request;Trace</ExcludeTypes>
</Add>
<Add Type="...AdaptiveSamplingTelemetryProcessor">
    <SamplingPercentage>50</SamplingPercentage>
    <ExcludeTypes>Exception</ExcludeTypes>
</Add>
```
Notice how `Dependency` is missing from both and therefore sampled twice with old code. With these changes it would be sampled once and correctly reported with 20% sample rate.

SergeyKanzhelev:
yeah. For exception scenario - consider using http context extension method ` var requestTelemetry = ctx.GetRequestTelemetry();` to take correlated request and mark it's `SamplingPercentage`

PR #260, opened 2016-07-22
Added helper method to insert TelemetryProcessors at the start of the chain

hallatore:
With this we can add TelemetryProcessors at the start of the chain. This is useful if you inject processors by code and want your processor to run first.

### DIFF
--- a/src/Core/Managed/Shared/Extensibility/Implementation/TelemetryProcessorChainBuilder.cs
+++ b/src/Core/Managed/Shared/Extensibility/Implementation/TelemetryProcessorChainBuilder.cs
@@ -40,6 +40,18 @@
         }
 
         /// <summary>
+        /// Uses given factory to add TelemetryProcessor at the start of the chain of processors. The processors
+        /// in the chain will be invoked in the same order in which they are chained.
+        /// </summary>
+        /// <param name="telemetryProcessorFactory">A delegate that returns a <see cref="ITelemetryProcessor"/>
+        /// , given the next <see cref="ITelemetryProcessor"/> in the call chain.</param>
+        public TelemetryProcessorChainBuilder UseFirst(Func<ITelemetryProcessor, ITelemetryProcessor> telemetryProcessorFactory)
+        {
+            this.factories.Insert(0, telemetryProcessorFactory);
+            return this;
+        }
+
+        /// <summary>
         /// Builds the chain of linked <see cref="ITelemetryProcessor" /> instances and sets the same in configuration object passed.
         /// A special telemetry processor for handling Transmission is always appended as the last
         /// processor in the chain.


hallatore:
Example usage scenario where Dependency telemetry will be sampled by 1% regardless of what other samplers exist in the config.

```
TelemetryConfiguration.Active.TelemetryProcessorChainBuilder.UseFirst((tp) => new SamplingTelemetryProcessor(tp)
{
    SamplingPercentage = 1,
    IncludeTypes = "Dependency"
});

TelemetryConfiguration.Active.TelemetryProcessorChainBuilder.Build();
```

SergeyKanzhelev:
We've been discussing this scenario and I wonder - why do you have half of processors initialization in configuration and half in code? If you move all initializers into code or into configuration file you will not have this problem.

hallatore:
I'm testing out to standardize all our apps on Application Insights. To help with that I've made a nuget package which initializes most stuff so that the setup becomes trivial for individual projects. Ignore heartbeats from lb, ensure exception tracking, etc.

An example is that I want to set Dependency sampling to 1%, but in order to ensure that this sampler needs to go before the ones in the config.

So the problem is that I need to inject some processors before the others, but I don't want every project to have to do this manually. Specially since we are talking about many many projects in the long run, and I want them to be able to just update our nuget package and be done with it.

SergeyKanzhelev:
This is exactly the reason we didn't want to introduce this method. It will be very hard to find out the actual order. That's why we are using this trick with live metrics telemetry processor which also wants to be the first one: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/7c86689810be38a8a8a412c0720a4f2614d7207d/Src/PerformanceCollector/NuGet/ApplicationInsights.config.install.xdt#L29

Alternatively your NuGet package can uninstall `ApplicationInsights.config` and perform initialization in code.

Mixing two approaches is error prone and hard to troubleshoot.


offtopic: Is it azure lb probe you are filtering? What header you rely on?

hallatore:
Since you have to .Build() the processors. Why isn't the list just exposed? Is there a reason for hiding it like it's currently is? I would like a programmatic way of manipulating the list, but I can of course use config transforms if there is no other option. 

offtopic:
Currently the packages does a couple of stuff and adds some helpers normal stuff.

- We use LB's or deployment slots (app services) to deploy new releases without downtime. One of the deployment steps is priming the site with some web tests. These tests naturally take long because they start the app, and they also create noise in the statistics because of this. So one processor filters out these requests.
- Added initialiser that adds version to every telemetry
- Added initialiser that adds referer to every request telemetry
- Added processor that blocks SQL dependency from being tracked. (Will switch out with 1% Dependency sampler once that is released)

The point is that as we gain experience in using App insights we add wanted features to our package and the projects just need to update that package and have all the good stuff enabled. 

Since we are planning on using it so widely cutting cost is also a huge factor. So I'm planning on testing out tracking only 1-5% sampling percentage while still tracking 100% of exceptions and their related telemetry. And with my other pull requests I think this will work out very nicely.

dnduffy:
I think we should discuss the code vs. config approaches briefly since .NET Core favors the code approach vs. the older ASP.NET config approach.  While mixing the two might be error prone for an individual app we can take steps to make the sequencing clear for debugging and I think we should look for a clean way to support both models since both have their benefits and detractions.

hallatore:
Since the factories list needs to be built before taking effect. What is the potential downside of just making the list public?

The code is still pretty clear. Whatever is in that list gets chained together when .Build() is invoked.

SergeyKanzhelev:
You have the access to the collection here: https://github.com/Microsoft/ApplicationInsights-dotnet/blob/master/src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs#L138 

Will this work?

1. var builder = TelemetryProcessorChainBuilder(config)
2. var processors = config.TelemetryProcessors();
3. builder.Use(MyFirstProcessor);
4. foreach(var p in processors) builder.Use(p);
5. builder.Use(MyLastProcessor);
6. builder.Build();

Is it too confusing API?

hallatore:
Hmm. I'm not exactly sure I understand what you are saying. Factories is `List<Func<ITelemetryProcessor, ITelemetryProcessor>>` but TelemetryProcessors is ReadOnlyCollection<ITelemetryProcessor>.

And at this point the chain has already been buildt once. I don't see the need to create a new one and override the already initialized one. Would't this possible be more error prone?

hallatore:
I think I see your point. But then we will lose whatever logic was in the `Func<ITelemetryProcessor, ITelemetryProcessor>` part of factories. So if anyone does anything special in that one then that would be lost.

But on the sake of that, why is that func really there? It seems be some something that should have been internal and not exposed like that.

SergeyKanzhelev:
Chain will already be built once anyway after config fully loaded. 

But you are correct - code will be a little bit more involved. And yes, it is better to have a list of functions to not loose the logic.

hallatore:
Without changing the code too much I think making factories field public is the best option. It would allow for advanced manipulation for those who want/need that, while 99% will use .Use() and the config.

SergeyKanzhelev:
So you have options:
1. Insert your configuration into `ApplicationInsights.config`
2. Move entire configuration into code
3. We expose somehow the list of functions for manipulation

Are we in agreement that exposing `UseFirst` is not a very good approach? And current API is bad.

I'd really love if you can evaluate the first and second approach. This will allow to control which of default configuration you need and which you don't. For instance, you may not need to use telemetry initializer that reads application version from `BuildInfo.config` (see [here](http://apmtips.com/blog/2015/06/18/application-versioning-semantic-or-automatic/)) and maybe one that reads Azure Cloud Instance Name.

Exposing and recommending the way to rebuild the chain with the "implied" logic is not something we want to do. I'd be happy to understand the reasoning of the mixed approach better.

hallatore:
For my specific scenario I can always go for option 1. Or I can even go for the approach you said since I know none of the func's do anything special. But I don't think they are ideal.

But as @dnduffy said, a programmatic approach is becoming more and more the way to go.

And I don't think exposing and recommending is the same thing here. This is not something you recommend, it's an option for those who need for it.

I agree that UseFirst is not a very good approach as it lands somewhat in the middle. It looks a lot like Use, and it's also not guaranteed that it actually will go first if it's used more than once. 
That's why I think exposing the factories field is a better options.

And are we hiding the factories field because most people should't touch it, or because none should touch it? Since it's only used by Build I think it's a safe field to expose.

hallatore:
I would argue that this is the right approach: https://github.com/hallatore/ApplicationInsights-dotnet/commit/2a5309a29f4016423b2baddf196b973a3f92ee4e

Btw: Initialisers are already exposed the same way in TelemetryConfiguration.cs.

SergeyKanzhelev:
Sorry for delayed response. 

Considering that our typical pattern is to define configuration file first and then do something like this

``` csharp
TelemetryConfiguration.Active.InstrumentationKey = "Foo";
TelemetryConfiguration.Active.TelemetryInitializers.Add(new MyInitializer());
```

to update the configuration - I'd agree with you. So making the factories list public sounds consistent with how the rest of SDK works. 

However today once you'll rebuild a chain - Live Metrics will stop working. @tokaplan may have plans to fix it going forward. I'd suggest to use configuration file modification for now until we'll fix live metrics.

Please comment if configuration approach is completely broken for you.

SergeyKanzhelev:
I'm closing this PR for now.

PR #262, opened 2016-07-26
Remove processors tag if it is empty

abaranch:
#246


src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
```diff
@@ -45,9 +45,11 @@ public static byte[] Serialize(IEnumerable<ITelemetry> telemetryItems, bool comp
         {
             var memoryStream = new MemoryStream();
             using (Stream compressedStream = compress ? CreateCompressedStream(memoryStream) : memoryStream)
-            using (var streamWriter = new StreamWriter(compressedStream, TransmissionEncoding))
             {
-                SeializeToStream(telemetryItems, streamWriter);
+                using (var streamWriter = new StreamWriter(compressedStream, TransmissionEncoding))
```
dnduffy: The previous sequential usings has the same semantic meaning as this nesting syntax so this change is not needed.

dnduffy:
:shipit:

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
abaranch: I know. It is hard to read for me. I'm used to having {} in place.

Issue #216, opened 2016-03-31
Provide a timeout option for Flush()


DanSeefeldt:
## Provide a timeout option for Flush()

We'd like the ability to specify a timeout for `Flush()` which would cancel the operation and return if the timeout was exceeded.  Our scenario is a short running .NET Core application with telemetry being uploaded at the end of the app by calling `Flush()` before exiting.  In some cases, network latency is causing the app to hang before exiting.  We'd rather lose the telemetry data than delay the app.

### Issue Details:

When running our app, we're seeing reasonable return times for the `Flush()` command in almost all cases, with the network connected and without.  The failing case for us is when we're running our app in a Docker container and the host's network is disconnected.  In this case, the app acts as if it's connected but cannot reach the network resource, resulting in a 40 second delay while the network call times out.

AlexBulankou:
Thanks Dan! We'll collect feedback on how typical your case is to see if we need to make it part of the API. As a workaround would doing something like this help:
```
            Thread thread = new Thread(() =>
            {
              telemetryClient.Flush();
            });
            thread.Start();
            if (!thread.Join(1000*5)) // wait 5 seconds
            {
                thread.Abort();
            }
```

daniel-kun:
I'd give a +1 here. It's not a must-have-feature, but it is something that a) I expect any API to IO operations to provide and b) I have wished in the past, but then it was not important enough to put effort into (especially testing).

LakshanF:
+1 for the feature since the .NET Core telemetry scenario is on a critical CLI tool where time lags are noticeable.

PR #264, opened 2016-07-27
Allow to override Timeout in Flush. Start 2.2-beta2.

abaranch:
#216

### DIFF
--- a/Test/CoreSDK.Test/Shared/Channel/InMemoryChannelTest.cs
+++ b/Test/CoreSDK.Test/Shared/Channel/InMemoryChannelTest.cs
@@ -4,6 +4,17 @@
     using System.Collections.Generic;
     using System.Linq;
     using Microsoft.ApplicationInsights.TestFramework;
+
+#if !NET40
+    using System.Diagnostics.Tracing;
+#endif
+
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+
+#if NET40
+    using Microsoft.Diagnostics.Tracing;
+#endif
+
     using Microsoft.VisualStudio.TestTools.UnitTesting;
 
     using Assert = Xunit.Assert;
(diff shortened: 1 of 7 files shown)


Test/CoreSDK.Test/Shared/Channel/InMemoryChannelTest.cs
```diff
@@ -4,6 +4,17 @@
...
+
+#if !NET40
+    using System.Diagnostics.Tracing;
+#endif
+
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+
+#if NET40
+    using Microsoft.Diagnostics.Tracing;
+#endif
```
dnduffy: Can this #if be merged into an #else or #elif with the one above?

src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
```diff
@@ -94,7 +94,7 @@ private void Runner()
                 while (this.enabled)
                 {
                     // Pulling all items from the buffer and sending as one transmissiton.
-                    this.DequeueAndSend();
+                    this.DequeueAndSend(timeout: default(TimeSpan)); // use default timeout
```
dnduffy: Won't the default of TimeSpan be zero?
At first I read the comment as meaning that the system has some non-zero timeout value it uses by default which is different than what the code does so it might be clearer to add to the comment so it reads "// Use default timeout of 0."

dnduffy:
:shipit:

src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
abaranch: It is actually using the default timeout which is not zero. If we pass default, it uses some predefined value on Transmission level. I can make comment more clear.

Test/CoreSDK.Test/Shared/Channel/InMemoryChannelTest.cs
abaranch: No, stylecop required namespaces be sorted

src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
dnduffy: The value of default(TimeSpan) is 0 (or TimeSpan.Zero which is still 0).

Issue #263, opened 2016-07-27
When required sring field is not provided substitute it with a short string


abaranch:
Currently is response code is not provided we send responseCode: responseCodeisarequiredfieldforMicrosoft.ApplicationInsights.DataContracts.RequestTelemetry which is very long. Substitute it on "n/a".
Same applies to all other required string fields.

shklar:
I totally agree :)
The unreasonably long string currently shown looks weird and takes up too much space,

PR #265, opened 2016-07-28
Empty required fields are replaced with "n/a", properties with required

abaranch:
#263 
Required empty property value is replaced with "n/a"
Required empty property name is replaced with "required" (since we may have multiple such properties, so that we add numbers in the end and "n/a1" "n/a2" would look strange)


dnduffy:
:shipit:

Issue #255, opened 2016-07-20
Add TrackAvailability to SDK


ericmattingly:
Would like to add a new API call to the SDK "TrackAvailability" which will allow us to inject availability data into App Insights.  While this is currently possible by calling the REST API directly, I want to expose this through the SDK instead.  We understand that it may not render in the Azure Portal but we will be able to access it through App Analytics.  This will help fill the gap we have of using App Insights for all up availability tracking due to missing features in Availability monitoring out of box.

PR #266, opened 2016-07-29
Add TrackAvailability to SDK

ericmattingly:
#255

### DIFF
--- a/Test/CoreSDK.Test/Shared/DataContracts/AvailabilityTelemetryTest.cs
+++ b/Test/CoreSDK.Test/Shared/DataContracts/AvailabilityTelemetryTest.cs
@@ -66,15 +66,7 @@
         [TestMethod]
         public void SanitizeWillAddDefaultFieldsForAvailability()
         {
-            AvailabilityTelemetry telemetry = new AvailabilityTelemetry
-            {
-                Message = "Test Message",
-                RunLocation = "Test Location",
-                TestName = "Test Name",
-                TestTimeStamp = DateTimeOffset.Now,
-                Duration = TimeSpan.FromSeconds(30),
-                Success = true
-            };
+            AvailabilityTelemetry telemetry = CreateAvailabilityTelemetry();
 
             ((ITelemetry)telemetry).Sanitize();
 
@@ -84,15 +76,7 @@
         [TestMethod]
         public void SanitizeWillLeaveCustomMessageForAvailability()
         {
-            AvailabilityTelemetry telemetry = new AvailabilityTelemetry
-            {
-                Message = "Test Message",
-                RunLocation = "Test Location",
-                TestName = "Test Name",
-                TestTimeStamp = DateTimeOffset.Now,
-                Duration = TimeSpan.FromSeconds(30),
-                Success = true
-            };
+            AvailabilityTelemetry telemetry = CreateAvailabilityTelemetry();
 
             ((ITelemetry)telemetry).Sanitize();
 
@@ -102,14 +86,9 @@
         [TestMethod]
         public void SanitizeWillSetFailedMessageForAvailability()
         {
-            AvailabilityTelemetry telemetry = new AvailabilityTelemetry
-            {
-                RunLocation = "Test Location",
-                TestName = "Test Name",
-                TestTimeStamp = DateTimeOffset.Now,
-                Duration = TimeSpan.FromSeconds(30),
-                Success = false
-            };
+            AvailabilityTelemetry telemetry = CreateAvailabilityTelemetry();
+            telemetry.Message = string.Empty;
+            telemetry.Success = false;
 
             ((ITelemetry)telemetry).Sanitize();
 
@@ -119,14 +98,8 @@
         [TestMethod]
         public void SanitizeWillSetPassedMessageForAvailability()
         {
-            AvailabilityTelemetry telemetry = new AvailabilityTelemetry
-            {
-                RunLocation = "Test Location",
-                TestName = "Test Name",
-                TestTimeStamp = DateTimeOffset.Now,
-                Duration = TimeSpan.FromSeconds(30),
-                Success = true
-            };
+            AvailabilityTelemetry telemetry = CreateAvailabilityTelemetry();
+            telemetry.Message = string.Empty;
 
             ((ITelemetry)telemetry).Sanitize();
 
--- a/Test/CoreSDK.Test/Shared/DataContracts/AvailabilityTelemetryTest.cs
+++ b/Test/CoreSDK.Test/Shared/DataContracts/AvailabilityTelemetryTest.cs
@@ -1,0 +1,177 @@
+﻿namespace Microsoft.ApplicationInsights.DataContracts
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.External;
+    using Microsoft.ApplicationInsights.TestFramework;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Assert = Xunit.Assert;
+    using DataPlatformModel = Developer.Analytics.DataCollection.Model.v2;
+
+    [TestClass]
+    public class AvailabilityTelemetryTest
+    {
+        [TestMethod]
+        public void AvailabilityTelemetrySerializesToJson()
+        {
+            AvailabilityTelemetry expected = this.CreateAvailabilityTelemetry();
+            var item = TelemetryItemTestHelper.SerializeDeserializeTelemetryItem<AvailabilityTelemetry, DataPlatformModel.AvailabilityData>(expected);
+
+            Assert.Equal(expected.Timestamp, item.Time);
+            Assert.Equal(expected.Sequence, item.Seq);
+            Assert.Equal(expected.Context.InstrumentationKey, item.IKey);
+            Assert.Equal(expected.Context.Tags.ToArray(), item.Tags.ToArray());
+            Assert.Equal(typeof(DataPlatformModel.AvailabilityData).Name, item.Data.BaseType);
+
+            Assert.Equal(expected.Duration, item.Data.BaseData.Duration);
+            Assert.Equal(expected.Message, item.Data.BaseData.Message);
+            Assert.Equal(expected.Success, (item.Data.BaseData.Result == DataPlatformModel.TestResult.Pass) ? true : false);
+            Assert.Equal(expected.RunLocation, item.Data.BaseData.RunLocation);
+            Assert.Equal(expected.TestName, item.Data.BaseData.TestName);
+            Assert.Equal(expected.Id.ToString(), item.Data.BaseData.TestRunId);
+            Assert.Equal(expected.TestTimeStamp, item.Data.BaseData.TestTimeStamp);
+
+            Assert.Equal(expected.Properties.ToArray(), item.Data.BaseData.Properties.ToArray());
+        }
+
+        [TestMethod]
+        public void SanitizeWillTrimAppropriateFieldsForAvailability()
+        {
+            AvailabilityTelemetry telemetry = new AvailabilityTelemetry();
+            telemetry.Message = new string('Z', Property.MaxAvailabilityMessageLength + 1);
+            telemetry.RunLocation = new string('Y', Property.MaxRunLocationLength + 1);
+            telemetry.TestName = new string('D', Property.MaxTestNameLength + 1);
+            telemetry.Properties.Add(new string('X', Property.MaxDictionaryNameLength) + 'X', new string('X', Property.MaxValueLength + 1));
+            telemetry.Properties.Add(new string('X', Property.MaxDictionaryNameLength) + 'Y', new string('X', Property.MaxValueLength + 1));
+
+            ((ITelemetry)telemetry).Sanitize();
+
+            Assert.Equal(new string('Z', Property.MaxAvailabilityMessageLength), telemetry.Message);
+            Assert.Equal(new string('Y', Property.MaxRunLocationLength), telemetry.RunLocation);
+            Assert.Equal(new string('D', Property.MaxTestNameLength), telemetry.TestName);
+
+            Assert.Equal(3, telemetry.Properties.Count); //AvailabilityTelemetry sanitize already sets one property which is why this is 3 instead of 2
+            Assert.Equal(new string('X', Property.MaxDictionaryNameLength), telemetry.Properties.Keys.ToArray()[0]);
+            Assert.Equal(new string('X', Property.MaxValueLength), telemetry.Properties.Values.ToArray()[0]);
+            Assert.Equal(new string('X', Property.MaxDictionaryNameLength - 3) + "001", telemetry.Properties.Keys.ToArray()[1]);
+            Assert.Equal(new string('X', Property.MaxValueLength), telemetry.Properties.Values.ToArray()[1]);
+
+            Assert.Same(telemetry.Properties, telemetry.Properties);
+        }
+
+        [TestMethod]
+        public void SanitizeWillAddDefaultFieldsForAvailability()
+        {
+            AvailabilityTelemetry telemetry = new AvailabilityTelemetry
+            {
+                Message = "Test Message",
+                RunLocation = "Test Location",
+                TestName = "Test Name",
+                TestTimeStamp = DateTimeOffset.Now,
+                Duration = TimeSpan.FromSeconds(30),
+                Success = true
+            };
+
+            ((ITelemetry)telemetry).Sanitize();
+
+            Assert.Equal(telemetry.Properties["FullTestResultAvailable"], "false");
+        }
+
+        [TestMethod]
+        public void SanitizeWillLeaveCustomMessageForAvailability()
+        {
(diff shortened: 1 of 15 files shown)


src/Core/Managed/Shared/Utils.cs
```diff
@@ -74,6 +74,20 @@ public static TimeSpan ValidateDuration(string value)
             return interval;
         }
 
+        /// <summary>
+        /// Returns min DateTimeOffset value if not a valid DateTimeOffset.
+        /// </summary>
+        public static DateTimeOffset ValidateDateTimeOffset(string value)
+        {
+            DateTimeOffset timestamp;
+            if (!DateTimeOffset.TryParse(value, CultureInfo.InvariantCulture, DateTimeStyles.AssumeLocal, out timestamp))
```
abaranch: >AssumeLocal [](start = 93, length = 11)

Why?

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
```diff
@@ -0,0 +1,170 @@
...
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.External;
+    using System.Globalization;
+
+    /// <summary>
+    /// Telemetry type used for availability test results.
+    /// Contains a time and message and optionally some additional metadata.
+    /// </summary>
+    public sealed class AvailabilityTelemetry : ITelemetry, ISupportProperties //TODO: Should this implement IOperation like RequestTelemetry?
```
abaranch: >TODO [](start = 81, length = 4)

SergeyK?

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AvailabilityTelemetry"/> class with empty properties.
+        /// </summary>
+        public AvailabilityTelemetry(string testName, DateTimeOffset testTimeStamp, TimeSpan duration, string runLocation, bool success, string message = null)
+        {
+            this.Data = new AvailabilityData();
+            this.context = new TelemetryContext(this.Data.properties);            
+            this.Data.testTimeStamp = testTimeStamp.ToString("o", CultureInfo.InvariantCulture);
```
abaranch: >testTimeStamp [](start = 22, length = 13)

I do not think we need a separate timestamp. That would be confusing and inconsistent with other types. Do you have scenarios in UI where you need both timestamps and they are expected to be different? I do not expect customers to guess that we have a separate field just for this type when they create queries in AI.

Test/CoreSDK.Test/Shared/DataContracts/AvailabilityTelemetryTest.cs
```diff
@@ -0,0 +1,177 @@
...
+        public void SanitizeWillSetFailedMessageForAvailability()
+        {
+            AvailabilityTelemetry telemetry = new AvailabilityTelemetry
+            {
+                RunLocation = "Test Location",
+                TestName = "Test Name",
+                TestTimeStamp = DateTimeOffset.Now,
+                Duration = TimeSpan.FromSeconds(30),
+                Success = false
+            };
```
dnduffy: It might be worth considering turning these mostly identical object creation blocks in each test into a one liner helper method call to make the code more succinct.

src/Core/Managed/Shared/Utils.cs
ericmattingly: Probably a question for @MeniZalzman This was coped directly from RequestTelemetry, I made the assumption that validating offsets in AvailabilityTelemetry would need to be the same.

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
ericmattingly: There are scenarios where someone may be computing availability with an external system (Gomez/Keynote/etc) and write a “connector” to send them to AI.  In this case the difference between when the telemetry item is submitted and when availability was computed could be very different.  The UI should be showing when the availability was measured, not when it was sent to AI.  The REST endpoint for AI currently requires a testDateTime as well.

Test/CoreSDK.Test/Shared/TelemetryClientTest.cs
```diff
@@ -369,6 +369,41 @@ public void TrackDependencySendsGivenDependencyTelemetryToTelemetryChannel()
 
         #endregion
 
+        #region TrackAvailability
+
+        [TestMethod]
+        public void TrackAvailabilitySendsAvailabilityTelemetryWithGivenNameRunlocationRunidDurationResultandMessageToTelemetryChannel()
```
dnduffy: nitpick: Capitalize "Id" and "And" in the test method name.

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
```diff
@@ -0,0 +1,170 @@
...
+            this.Data.runLocation = runLocation;
+            this.Data.message = message;           
+        }        
+
+        /// <summary>
+        /// Gets or sets the test run id guid
+        /// </summary>
+        public Guid Id
+        {
+            get { return Guid.Parse(this.Data.testRunId); }
```
abaranch: >Parse [](start = 30, length = 5)

CoreSDK never throws exceptions. It substitutes it with default values. E.g. in ip address we check that it is ip in v4 format.

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        /// <summary>
+        /// Gets or sets availability test duration.
+        /// </summary>
+        public TimeSpan Duration
+        {
+            get
+            {
+                //TODO: Note: this ends up logging an error messsage for Request Telemetry if it fails, Both PageView and Request telemetry use this to validate so PV already has the bug.
```
abaranch: >TODO [](start = 18, length = 4)

Is that smth you plan to do or is it a question?..

abaranch:
> namespace Microsoft.ApplicationInsights.DataContracts

I wonder if you tried to compile in release. I have a filling that Stylecop will complain in couple places. But I may be wrong.

---
Refers to: src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs:1 in 6875562. [](commit_id = 6875562304c82e7fccccb1ca35968f727a319028, deletion_comment = False)

src/Core/Managed/Net46/Extensibility/Implementation/RichPayloadEventSource.cs
```diff
@@ -259,6 +274,11 @@ public sealed class Keywords
             /// Keyword for session state.
             /// </summary>
             public const EventKeywords SessionState = (EventKeywords)0x100;
+
+            /// <summary>
+            /// Keyword for availability.
+            /// </summary>
+            public const EventKeywords Availability = (EventKeywords)0x120;
```
dnduffy: Should be 0x200.
(Because 0x120 would be SessionState + Metrics and not a new value)

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
dnduffy: I don't see IOperation on RequestTelemetry.

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
dnduffy: Something doesn't sound right about this TODO comment; if there is already a bug then it should be fixed and not used as justification for another bug.

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
```diff
@@ -0,0 +1,170 @@
...
+                return Utils.ValidateDuration(this.Data.duration);
+            }
+            set
+            {
+                this.Data.duration = value.ToString();
```
dnduffy: Why is it validated on Get instead of on Set?  It seems like what is set should be validated so that get does not need to validate because only valid data can be set.

src/Core/Managed/Shared/Utils.cs
abaranch: Would you agree it should be AssumeUniversal?

---
In reply to: [72832994](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/266#discussion_r72832994) [](ancestors = 72832994)

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
ericmattingly: Correct, OperationTelemetry which is what Dependency uses (was wrong on both somehow there).  @SergeyKanzhelev what do you think? Should I extend that?

ericmattingly:
You are correct on StyleCop complaining, I need to clean up multiple spots

src/Core/Managed/Shared/Utils.cs
ericmattingly: This is used for validating TestTimeStamp and we don't know if the caller is passing us a UTC offset or not.  There is precedent in the SDK set since RequestTelemetry is already using AssumeLocal for their start/end-time and timestamp.  Is there a better way to enforce passing a timezone denoted timespan when its assigned?

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
ericmattingly: This is the pattern followed by both PageView and RequestTelemetry for duration.  I agree that it should not be necessary. @SergeyKanzhelev can you shed some light on it or can edit PV and Request as well?

ericmattingly:
Passing StyleCop with latest commit

Test/CoreSDK.Test/Shared/DataContracts/AvailabilityTelemetryTest.cs
ericmattingly: Fixed in latest commit

Test/CoreSDK.Test/Shared/TelemetryClientTest.cs
ericmattingly: Fixed in latest commit

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
ericmattingly: Fixed in latest commit

src/Core/Managed/Net46/Extensibility/Implementation/RichPayloadEventSource.cs
ericmattingly: Fixed in latest commit

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
ericmattingly: If its agreed in the other comment that there is no need to validate on the get accessor, than this line can be removed from here/PV/request also removing the bug. @SergeyKanzhelev

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
dnduffy: I've filed an enhancement to cover this more generally, you may ignore it for now.
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/269

dnduffy:
My opinion is that validation should occur on set methods rather than get methods.
:shipit:

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
SergeyKanzhelev: I think it should implement operation telemetry interfaces. The idea may be to associate availability result with the dependency telemetry objects representing every step in multi-step test

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
```diff
@@ -0,0 +1,170 @@
...
+        private readonly TelemetryContext context;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AvailabilityTelemetry"/> class with empty properties.
+        /// </summary>
+        public AvailabilityTelemetry()
+        {
+            this.Data = new AvailabilityData();
+            this.context = new TelemetryContext(this.Data.properties);
+            this.Data.testRunId = Guid.NewGuid().ToString();
```
SergeyKanzhelev: Please use the same ID generator as RequestTelemetry

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
SergeyKanzhelev: #Reopen
It is fine. You still can report test time using Timestamp. This is how we design our data model now and we will get rid of startTime for RequestTelemetry

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
SergeyKanzhelev: Let's not force it to be a guid. Make it aligned with other identifiers we have

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
SergeyKanzhelev: Also should it be AvailabilityResultTelemetry? Just Availability sounds unclear

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
```diff
@@ -0,0 +1,170 @@
...
+        public DateTimeOffset TestTimeStamp
+        {
+            get { return Utils.ValidateDateTimeOffset(this.Data.testTimeStamp); }
+            set { this.Data.testTimeStamp = value.ToString("o", CultureInfo.InvariantCulture); }
+        }
+
+        /// <summary>
+        /// Gets or sets the test name.
+        /// </summary>
+        public string TestName
```
SergeyKanzhelev: Consider making it just "Name".

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
```diff
@@ -613,6 +618,43 @@ private static void SerializePerformanceCounter(PerformanceCounterTelemetry perf
...
+                writer.WritePropertyName("baseData");
+                {
+                    writer.WriteStartObject();
+
+                    writer.WriteProperty("ver", availabilityTelemetry.Data.ver);
+                    writer.WriteProperty("testRunId", availabilityTelemetry.Data.testRunId);
+                    writer.WriteProperty("testName", availabilityTelemetry.Data.testName);
+                    writer.WriteProperty("testTimeStamp", availabilityTelemetry.TestTimeStamp);
+                    writer.WriteProperty("duration", availabilityTelemetry.Duration);
+                    writer.WriteProperty("result", availabilityTelemetry.Data.result.ToString());
```
SergeyKanzhelev: why do you need ToString here?

src/Core/Managed/Shared/Utils.cs
abaranch: Since it is a server side SDK I would assume that everything needs to be in UTC. So I would change that in RequestTelemetry as well.

---
In reply to: [72862498](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/266#discussion_r72862498) [](ancestors = 72862498)

SergeyKanzhelev: Please fix it

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
ericmattingly: Result is an enum (TestResult)

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
ericmattingly: Changed

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
ericmattingly: Fixed

src/Core/Managed/Shared/Utils.cs
ericmattingly: Fixed

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
ericmattingly: @SoubhagyaDash what do you think?  I have no preference.

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
ericmattingly: opened another issue to track this change #276

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
SergeyKanzhelev: I think we can address those all together with the #276.

Issue #175, opened 2016-02-01
Remove BCL dependency


abaranch:

daniel-kun:
I'd second this suggestion, since the BCL dependency is quite a bugger when you want to use AppInsights in a C++/CLR project.

dnduffy:
https://github.com/Microsoft/ApplicationInsights-dotnet/pull/267

PR #267, opened 2016-08-01
Issue 175, no BCL references for .Net40

ki0ki0:
BCL dependencies has been removed from base API project.
TelemetryChannel still uses BCL.

[Issue 175](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/175)

### DIFF
--- a/Test/CoreSDK.Test/Net40/TaskExTests.cs
+++ b/Test/CoreSDK.Test/Net40/TaskExTests.cs
@@ -1,0 +1,227 @@
+using System.Linq;
+
+namespace Microsoft.ApplicationInsights
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Assert = Xunit.Assert;
+
+    [TestClass]
+    public class TaskExTests
+    {
+        /// <summary>
+        /// Tests the scenario if RethrowIsFaulted doesn't throw exception for not faulted task.
+        /// </summary>
+        [TestMethod]
+        public void RethrowIfFaultedDoesntThrowIfNoExceptionOccured()
+        {
+            Task task = Task.Factory.StartNew(() => { });
+
+            task.Wait();
+
+            Assert.DoesNotThrow(() => task.RethrowIfFaulted());
+        }
+
+        /// <summary>
+        /// Tests the scenario if RethrowIsFaulted throws exception for not completed task.
+        /// </summary>
+        [TestMethod]
+        public void RethrowIfFaultedThrowsIfTaskIsNotCompleted()
+        {
+            Task task = TaskEx.Delay(TimeSpan.FromMilliseconds(100));
+
+            Assert.Throws<ArgumentException>(() => task.RethrowIfFaulted());
+        }
+
+        /// <summary>
+        /// Tests the scenario if RethrowIsFaulted throws exception for faulted task.
+        /// </summary>
+        [TestMethod]
+        public void RethrowIfFaultedThrowsIfExceptionOccured()
+        {
+            Task task = Task.Factory.StartNew(() => { throw new Exception(); });
+
+            try
+            {
+                task.Wait();
+            }
+            catch (Exception)
+            {
+                // ignore
+            }
+
+            Assert.Throws<AggregateException>(() => task.RethrowIfFaulted());
+        }
+
+        /// <summary>
+        /// Tests the scenario if Delay with zero timeout completes the task.
+        /// </summary>
+        [TestMethod]
+        public void DelayWithZeroTimeoutCompletesTask()
+        {
+            Task task = TaskEx.Delay(TimeSpan.Zero);
+            task.Wait();
+
+            Assert.Equal(task.Status, TaskStatus.RanToCompletion);
+        }
+
+        /// <summary>
+        /// Tests the scenario if Delay with timeout completes task after timeout has elapsed.
+        /// </summary>
+        [TestMethod]
+        public void DelayWaitsAtLeastTimeout()
+        {
+            TimeSpan timeout = TimeSpan.FromMilliseconds(50);
+            DateTime time = DateTime.UtcNow;
+            Task task = TaskEx.Delay(timeout);
+            
+            task.Wait();
+            TimeSpan spent = DateTime.UtcNow - time;
+
+            Assert.True(spent >= timeout);
+        }
+
+        /// <summary>
+        /// Tests the scenario if Delay turns the task in Canceled state after CancellationTokenSource is signaled.
+        /// </summary>
+        [TestMethod]
+        public void DelayWhithCancelationCancelsTask()
+        {
+            TimeSpan timeout = TimeSpan.FromSeconds(1);
+            CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+            Task task = TaskEx.Delay(timeout, cancellationTokenSource.Token);
+
+            cancellationTokenSource.Cancel();
+
+            Assert.True(task.IsCanceled);
+        }
+
+        /// <summary>
+        /// Tests the scenario if Delay task throws on wait after cancelation.
+        /// </summary>
+        [TestMethod]
+        public void DelayWhichIsCanceledThrows()
+        {
+            TimeSpan timeout = TimeSpan.FromSeconds(1);
+            CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+            Task task = TaskEx.Delay(timeout, cancellationTokenSource.Token);
+
+            cancellationTokenSource.Cancel();
+
+            AggregateException aggregateException = Assert.Throws<AggregateException>(() => task.Wait());
+            Assert.True(aggregateException.InnerExceptions.Single() is TaskCanceledException);
+        }
+
+        /// <summary>
+        /// Tests the scenario if Delay task completes after cancelation without waiting for timeout.
+        /// </summary>
+        [TestMethod]
+        public void DelayWith1SecondAndCanceledDoesntWaitForASecond()
+        {
+            TimeSpan timeout = TimeSpan.FromSeconds(1);
+            CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+            Task task = TaskEx.Delay(timeout, cancellationTokenSource.Token);
+
+            DateTime time = DateTime.UtcNow;
+            TimeSpan cancelTimeout = TimeSpan.FromMilliseconds(50);
+            TaskEx.Delay(cancelTimeout).ContinueWith(task1 => cancellationTokenSource.Cancel());
+            try
+            {
+                task.Wait();
+            }
+            catch (Exception)
+            {
+                // ignore
+            }
+
+            TimeSpan spent = DateTime.UtcNow - time;
+
+            Assert.True(spent >= cancelTimeout && spent < timeout);
+        }
+
+        /// <summary>
+        /// Tests the scenario if FromResult sets provided object as task result.
+        /// </summary>
+        [TestMethod]
+        public void FromResultReturnsInputAsResult()
+        {
+            object result = new object();
+            Task<object> task = TaskEx.FromResult<object>(result);
+
+            task.Wait();
+
+            Assert.True(result == task.Result);
+        }
+
+        /// <summary>
+        /// Tests the scenario if WhenAny throws exception for empty tasks array.
+        /// </summary>
+        [TestMethod]
+        public void WhenAnyThrowsForNoTasks()
+        {
+            Assert.Throws<ArgumentException>(() => { TaskEx.WhenAny(); });
+        }
+
+        /// <summary>
+        /// Tests the scenario if WhenAny returns first task if it has completed faster.
+        /// </summary>
+        [TestMethod]
+        public void WhenAnyReturnsFirstCompletedTask()
+        {
+            Task task1 = TaskEx.Delay(TimeSpan.FromMilliseconds(50));
+            Task task2 = TaskEx.Delay(TimeSpan.FromMilliseconds(100));
+
+            Task<Task> completedTask = TaskEx.WhenAny(task1, task2);
+            completedTask.Wait();
+
+            Assert.True(task1 == completedTask.Result);
+        }
+
+        /// <summary>
+        /// Tests the scenario if WhenAny returns second task if it has completed faster.
+        /// </summary>
+        [TestMethod]
+        public void WhenAnyReturnsFirstCompletedTask2()
+        {
+            Task task1 = TaskEx.Delay(TimeSpan.FromMilliseconds(100));
+            Task task2 = TaskEx.Delay(TimeSpan.FromMilliseconds(50));
+
+            Task<Task> completedTask = TaskEx.WhenAny(task1, task2);
+            completedTask.Wait();
+
+            Assert.True(task2 == completedTask.Result);
+        }
+
+        /// <summary>
+        /// Tests the scenario if WhenAny returns first completed task even if it has been faulted.
+        /// </summary>
+        [TestMethod]
+        public void WhenAnyReturnsFirstCompletedTaskEvenOnException()
+        {
+            Task task1 = TaskEx.Delay(TimeSpan.FromMilliseconds(100));
+            Task task2 = TaskEx.Delay(TimeSpan.FromMilliseconds(50)).ContinueWith(task => { throw new Exception(); });
+
+            Task<Task> completedTask = TaskEx.WhenAny(task1, task2);
+            completedTask.Wait();
+
+            Assert.True(task2 == completedTask.Result);
+        }
+
+        /// <summary>
+        /// Tests the scenario if WhenAny returns first completed task even if it has been canceled.
+        /// </summary>
+        [TestMethod]
+        public void WhenAnyReturnsFirstCompletedTaskEvenOnCanceled()
+        {
+            Task task1 = TaskEx.Delay(TimeSpan.FromMilliseconds(100));
+            Task task2 = TaskEx.Delay(TimeSpan.FromMilliseconds(50), new CancellationTokenSource(TimeSpan.FromMilliseconds(10)).Token);
+
+            Task<Task> completedTask = TaskEx.WhenAny(task1, task2);
+            completedTask.Wait();
+
+            Assert.True(task2 == completedTask.Result);
+        }
+    }
+}
--- a/src/Core/Managed/Net40/TaskEx.cs
+++ b/src/Core/Managed/Net40/TaskEx.cs
@@ -1,0 +1,105 @@
+﻿namespace Microsoft.ApplicationInsights
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    internal static class TaskEx
+    {
+        /// <summary>
+        /// Check and rethrow exception for failed task.
+        /// </summary>
+        /// <param name="task">Task to check.</param>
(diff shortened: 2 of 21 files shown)


abaranch:
Thank you for your contribution. We will review it soon.

Test/CoreSDK.Test/Net40/TaskExTests.cs
```diff
@@ -0,0 +1,173 @@
...
+    using Assert = Xunit.Assert;
+
+    [TestClass]
+    public class TaskExTests
+    {
+
+        [TestMethod]
+        public void RethrowIfFailedDoesntThrowIfNoExceptionOccured()
+        {
+            var task = Task.Factory.StartNew(() => { });
```
dnduffy: nitpick: I know this is a hotly debated topic online but I respectfully ask that you use concrete types instead of var as it increases readability of the code (especially outside of Visual Studio where there is no IntelliSense).

Test/CoreSDK.Test/Net40/TaskExTests.cs
```diff
@@ -0,0 +1,173 @@
+namespace Microsoft.ApplicationInsights
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Assert = Xunit.Assert;
+
+    [TestClass]
```
dnduffy: Documentation comments and general StyleCop compliance would be really appreciated.  Thank-you. :)

Test/CoreSDK.Test/Net40/TaskExTests.cs
```diff
@@ -0,0 +1,173 @@
...
+        [TestMethod]
+        public void DelayWhichIsCanceledThrows()
+        {
+            var timeout = TimeSpan.FromSeconds(1);
+            var cancellationTokenSource = new CancellationTokenSource();
+            var task = TaskEx.Delay(timeout, cancellationTokenSource.Token);
+
+            cancellationTokenSource.Cancel();
+
+            Assert.Throws<AggregateException>(() => task.Wait());
```
dnduffy: To be safe, the test should look into the AggregateException and make sure the cancellation is indeed the reason.

Test/CoreSDK.Test/Net40/TaskExTests.cs
```diff
@@ -0,0 +1,173 @@
...
+            {
+                task.Wait();
+            }
+            catch (Exception)
+            {
+                // ignore
+            }
+            var spent = DateTime.UtcNow - time;
+
+            Assert.True(spent >= cancelTimeout && spent < timeout);
```
dnduffy: It's probably safe enough for the test but just FYI this is technically a form of race condition that could fail in an edge case situation of high CPU load.

Test/CoreSDK.Test/Net40/TaskExTests.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            task.Wait();
+
+            Assert.DoesNotThrow(() => task.RethrowIfFailed());
+        }
+
+        [TestMethod]
+        public void RethrowIfFailedThrowsIfTaskIsNotCompleted()
+        {
+            var task = TaskEx.Delay(TimeSpan.FromMilliseconds(100));
```
dnduffy: In later files you have replaced TaskEx with Task so why use TaskEx here?

Test/CoreSDK.Test/Net40/TaskExTests.cs
macrogreg: Usually "var" is OK as long as the type is clear from the *same* line without specific knowledge of the code (and in complex Linq expressions).
var x = new MyType(); // this is OK
var y = MyFunc(); // Not OK since it requires knowing what MyFunc returns.
this particular one is on the border since Task.Factory.StartNew is "almost" like new task :)

SergeyKanzhelev: I agree with @macrogreg - we are using `var` widely in the code already.

src/Core/Managed/Net40/TaskEx.cs
```diff
@@ -0,0 +1,85 @@
...
+            Timer timer = null;
+            timer = new Timer(
+                state =>
+                    {
+                        timer.Dispose();
+                        tcs.TrySetResult(null);
+                    }, 
+                null, 
+                timeout, 
+                TimeSpan.FromMilliseconds(-1));
```
macrogreg: Please use "Timeout.Infinite"

src/Core/Managed/Net40/TaskEx.cs
```diff
@@ -0,0 +1,85 @@
...
+        public static Task<Task> WhenAny(params Task[] tasks)
+        {
+            if (tasks.Length == 0)
+            {
+                throw new ArgumentException("The tasks argument contains no tasks");
+            }
+
+            var taskCompletionSource = new TaskCompletionSource<Task>();
+
+            foreach (var task in tasks)
```
macrogreg: Micro-optimization: use a for loop to avoid an enumerator object creation.

src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
```diff
@@ -133,18 +133,22 @@ private void DequeueAndSend(TimeSpan timeout)
...
+#else
+                return Task.FromResult<object>(null);
+#endif
             }
 
             byte[] data = JsonSerializer.Serialize(telemetryItems);
             var transmission = new Transmission(this.endpointAddress, data, "application/x-json-stream", JsonSerializer.CompressionType, timeout);
 
-            await transmission.SendAsync().ConfigureAwait(false);
+            return transmission.SendAsync();
```
macrogreg: This is behavior change: the original await cause any exceptions that occurred in the async part of  SendAsync to be thrown here. Now they will instead be thrown when the Task returned by SendAsync is awaited. Can this be a problem?

src/Core/Managed/Shared/Channel/Transmission.cs
```diff
@@ -148,27 +153,39 @@ public string Id
...
-                                
-                Task completedTask = await TaskEx.WhenAny(timeoutTask, sendTask).ConfigureAwait(false);
-                if (completedTask == timeoutTask)
+                Task timeoutTask = TaskEx.Delay(this.Timeout).ContinueWith(task =>
                 {
-                    request.Abort(); // And force the sendTask to throw WebException.
-                }
+                    if (!sendTask.IsCompleted)
+                    {
+                        request.Abort(); // And force the sendTask to throw WebException.
```
macrogreg: There is a race condition here on sendTask completing after the check. I do not recall what happens if Abort is called on a finished request. If nothing, this is benign. To avoid this you can use a synchronous continuation to the send task and null out the request reference or set an interlocked flag. And yes - the issue existed before the change. :)

src/Core/Managed/Shared/Channel/Transmission.cs
```diff
@@ -131,14 +131,19 @@ public string Id
...
+#else
         public virtual async Task<HttpWebResponseWrapper> SendAsync()
+#endif
         {
             if (Interlocked.CompareExchange(ref this.isSending, 1, 0) != 0)
             {
                 throw new InvalidOperationException("SendAsync is already in progress.");
             }
-
+#if !NET40
```
macrogreg: due to the complex #if , this method is hard to read and understand. I would prefer to see a global (method-level) #if and then two clean implementations for 4.0 and not 4.0.

src/Core/Managed/Shared/Channel/Transmission.cs
```diff
@@ -148,27 +153,39 @@ public string Id
...
-                    request.Abort(); // And force the sendTask to throw WebException.
-                }
+                    if (!sendTask.IsCompleted)
+                    {
+                        request.Abort(); // And force the sendTask to throw WebException.
+                    }
+                });
+
+#if NET40
+                return TaskEx.WhenAny(timeoutTask, sendTask).ContinueWith(
```
macrogreg: Use synchronous continuation to unset the flag.

src/Core/Managed/Shared/Channel/Transmission.cs
```diff
@@ -221,6 +238,38 @@ protected virtual WebRequest CreateRequest(Uri address)
...
         }
 
+#if NET40
+        private Task<HttpWebResponseWrapper> SendRequestAsync(WebRequest request)
+        {
+            return Task.Factory.FromAsync(request.BeginGetRequestStream, request.EndGetRequestStream, null)
+                .ContinueWith(
+                    task =>
+                    {
+                        var requestStream = task.Result;
```
macrogreg: Here, the type is not clear from the same line. Please do not use VAR.

src/Core/Managed/Shared/Channel/Transmission.cs
```diff
@@ -221,6 +238,38 @@ protected virtual WebRequest CreateRequest(Uri address)
...
+                        return Task.Factory.FromAsync(
+                            (callback, o) => requestStream.BeginWrite(Content, 0, Content.Length, callback, o),
+                            requestStream.EndWrite, 
+                            null).ContinueWith(
+                                task1 =>
```
dnduffy: To avoid confusion I would rename task1 to something else such as innerTask.

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
```diff
@@ -77,7 +81,33 @@ public void Start(Func<Task> elapsed)
                             // It should return Task.FromResult but just in case we check for null if someone returned null
                             if (task != null)
                             {
+#if !NET40
                                 await task.ConfigureAwait(false);
+#else
+                                task.ContinueWith(
+                                    task1 =>
```
dnduffy: Please rename task1 to avoid confusion for future readers.

Test/CoreSDK.Test/Net40/TaskExTests.cs
dnduffy: Like I said, I know it is a hotly debated topic.  I am however very firmly on the side of readability and the framework design guidelines that Microsoft used to publish online and later in book form.

src/Core/Managed/Shared/Channel/Transmission.cs
```diff
@@ -221,6 +238,38 @@ protected virtual WebRequest CreateRequest(Uri address)
             return request;
         }
 
+#if NET40
+        private Task<HttpWebResponseWrapper> SendRequestAsync(WebRequest request)
+        {
+            return Task.Factory.FromAsync(request.BeginGetRequestStream, request.EndGetRequestStream, null)
+                .ContinueWith(
+                    task =>
```
macrogreg: can we call this "getRequestStreamTask"?

src/Core/Managed/Shared/Channel/Transmission.cs
macrogreg: a better name would be "writeTask"

src/Core/Managed/Net40/TaskEx.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    internal static class TaskEx
+    {
+        public static void RethrowIfFailed(this Task task)
+        {
+            if (!task.IsCompleted)
```
macrogreg: The method is throw *ifFaied* so if the task is not competed, it has not failed and nothing should happen.

macrogreg: Also, if you want to emulate the behavior of await, yo need to unwrap the aggregate exception.

src/Core/Managed/Shared/Channel/Transmission.cs
```diff
@@ -221,6 +238,38 @@ protected virtual WebRequest CreateRequest(Uri address)
...
+                            requestStream.EndWrite, 
+                            null).ContinueWith(
+                                task1 =>
+                                {
+                                    requestStream.Dispose();
+                                    task1.RethrowIfFailed();
+                                });
+                    }).Unwrap().ContinueWith(task =>
+                    {
+                        task.RethrowIfFailed();
```
macrogreg: You know that the task is completed. You could get the exception by asking for the result and unwrapping the aggregate (to be like await). Maybe adjust the implementation of RethrowIfFailed?

src/Core/Managed/Shared/Channel/Transmission.cs
```diff
@@ -221,6 +238,38 @@ protected virtual WebRequest CreateRequest(Uri address)
...
+                        return Task.Factory.FromAsync(
+                            request.BeginGetResponse, request.EndGetResponse, null);
+                    }).Unwrap().ContinueWith(task =>
+                    {
+                        using (var response = task.Result)
```
macrogreg: What is the type of response? Please do not use VAR here

src/Core/Managed/Shared/Channel/Transmission.cs
```diff
@@ -221,6 +238,38 @@ protected virtual WebRequest CreateRequest(Uri address)
...
+                        {
+                            return CheckResponse(response);
+                        }
+                    });
+        }
+#else
```
macrogreg: Please add a comment to say that "else" means "not 4.0"

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
```diff
@@ -66,7 +66,11 @@ public void Start(Func<Task> elapsed)
 
             TaskEx.Delay(this.Delay, newTokenSource.Token)
                 .ContinueWith(
-                    async previousTask => 
+#if !NET40
+                async previousTask =>
+#else
+                    previousTask =>
```
macrogreg: Re line 78 below:
This is why VAR is so bad. I am reading that (unchanged) bit of code and I need to stop and think about the type returned by elapsed(). I have to look around and rely on correctly used conventions for a correct guess. :)

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
```diff
@@ -77,7 +81,33 @@ public void Start(Func<Task> elapsed)
...
                             {
+#if !NET40
                                 await task.ConfigureAwait(false);
+#else
+                                task.ContinueWith(
+                                    task1 =>
+                                    {
+                                        try
+                                        {
+                                            task1.Wait();
```
macrogreg: you use Wait to hause exceptions to propagate. Elsewhere you use RethrowIf.... for that. Why the difference?

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
```diff
@@ -66,7 +66,11 @@ public void Start(Func<Task> elapsed)
 
             TaskEx.Delay(this.Delay, newTokenSource.Token)
                 .ContinueWith(
-                    async previousTask => 
+#if !NET40
```
macrogreg: I have reservations about this entire method's existence, but the translation seems correct. :)

Test/CoreSDK.Test/Net40/TaskExTests.cs
abaranch: You could run build in release. Stylecop is automatically checked.

---
In reply to: [73396854](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/267#discussion_r73396854) [](ancestors = 73396854)

src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
abaranch: We should not change the behavior.

---
In reply to: [73466495](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/267#discussion_r73466495) [](ancestors = 73466495)

SergeyKanzhelev:
@ki0ki0 - are you still interested in completing this PR? There are some comments that needs to be addressed before we can merge.

Test/CoreSDK.Test/Net40/TaskExTests.cs
ki0ki0: I've replaced all var where type is not clear from assignment

Test/CoreSDK.Test/Net40/TaskExTests.cs
ki0ki0: fixed

src/Core/Managed/Net40/TaskEx.cs
ki0ki0: fixed

src/Core/Managed/Net40/TaskEx.cs
ki0ki0: I've replaced implantation with ContinueWhenAny

src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
ki0ki0: The only usage of this private method is private method DequeueAndSend witch handles all exceptions.
So there are no changes in behavior.

src/Core/Managed/Shared/Channel/Transmission.cs
ki0ki0: Abort on completed request doesn't have any effects,

src/Core/Managed/Shared/Channel/Transmission.cs
ki0ki0: fixed

src/Core/Managed/Shared/Channel/Transmission.cs
ki0ki0: fixed

src/Core/Managed/Shared/Channel/Transmission.cs
ki0ki0: fixed

src/Core/Managed/Shared/Channel/Transmission.cs
ki0ki0: fixed

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
ki0ki0: fixed

src/Core/Managed/Shared/Channel/Transmission.cs
ki0ki0: Renamed to GetResponseAsync. This method is implemented as async method for NET45 so I didn't remove Async suffix.

src/Core/Managed/Shared/Channel/Transmission.cs
ki0ki0: Unwrapping of AggregateException for Net40 is tricky. I don't think it is a good idea.
IMHO RethrowIfFaulted is better option than checking result of task just to get an exception.

src/Core/Managed/Shared/Channel/Transmission.cs
ki0ki0: fixed

src/Core/Managed/Shared/Channel/Transmission.cs
ki0ki0: fixed

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
ki0ki0: fixed

ki0ki0:
Sorry, I had no time to finish it.
I've fixed all issues and rebased on current "develop" position.

PR #275, opened 2016-08-08
Iteration 2. Use bond compiler instead of custom-built DataPlatform nuget packages

SergeyKanzhelev:
Let's use bond compiler instead of custom-built DataPlatform nuget packages.

Minimal semantical changes in tests - bond compiler initialize strings with "", not null. Also bond compiler do not generate DateTime and TimeSpan. Accommodated for this in the test code. Most changes however is to change names from pascal case to camel case.


Schema/PublicSchema/PerformanceCounterData.bond
```diff
@@ -0,0 +1,33 @@
...
+namespace AI
+
+[Description("Instance of Performance Counter represents data for a specific Windows Performance Counter.")]
+struct PerformanceCounterData
+    : Domain
+{
+    [Name("version")]
+    10: required int32 	 ver = 2;
+    
+    20: required string 	 categoryName;
```
dnduffy: No descriptions on these properties?

Test/CoreSDK.Test/ApplicationInsightsTypes/app.config
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-9.0.0.0" newVersion="9.0.0.0" />
```
dnduffy: Do the JSON binding redirects need to be 9.0.1?

Schema/PublicSchema/PerformanceCounterData.bond
SergeyKanzhelev: those will be deprecated. I generated them so deprecation will be a separate PR that easier to review

Test/CoreSDK.Test/ApplicationInsightsTypes/app.config
SergeyKanzhelev: thanks removed from all test projects

Test/CoreSDK.Test/Shared/Channel/ITelemetryTest.cs
```diff
@@ -168,7 +169,7 @@ private void SerializeWritesSequenceAsExpectedByEndpoint()
             
             TelemetryItem<TEndpointData> actual = TelemetryItemTestHelper.SerializeDeserializeTelemetryItem<TTelemetry, TEndpointData>(expected);
             
-            Assert.AreEqual(expected.Sequence, actual.Seq);
+            Assert.AreEqual(expected.Sequence, actual.seq);
```
SergeyKanzhelev: those are bond-generated as oppose to custom-generated

dnduffy: Is there an option we can pass to get proper looking property casing back? (or are these just fields instead of properties?)

SergeyKanzhelev: nope. It is just an internal representation so shouldn't be a problem

PR #282, opened 2016-08-12
Added rules and enabled code analysis on build.

dnduffy:


Test/CoreSDK.Test/Net45/app.config
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
```
abaranch: >Json [](start = 43, length = 4)

Why do we need binding redirects for unit tests?

abaranch:
:shipit:

Test/CoreSDK.Test/Net45/app.config
dnduffy: I don't think we should but VS stuck that in there automatically for some reason.  I can manually remove it.

PR #284, opened 2016-08-15
deprecate perfromance counters type

SergeyKanzhelev:
Performance counter type got deprecated


Test/CoreSDK.Test/Shared/DataContracts/PerformanceCounterTelemetryTest.cs
```diff
@@ -8,6 +8,8 @@
     [TestClass]
     public class PerformanceCounterTelemetryTest
     {
+#pragma warning disable 618
```
dnduffy: Why is this pragma turned on and off so often?

SergeyKanzhelev: I marked class as `Obsolete` and test code wouldn't compile if you keep reference on those classes

dnduffy:
:shipit:

Issue #288, opened 2016-08-19
Make properties collection of types thread-safe


SergeyKanzhelev:
Today properties collection of telemetry types is not thread safe: https://github.com/Microsoft/ApplicationInsights-dotnet/blob/publicSchemaUpdate/src/Core/Managed/Shared/Extensibility/Implementation/External/RequestData_types.cs#L74

Need to make sure that actual interface object uses thread safe properties collection (https://github.com/Microsoft/ApplicationInsights-dotnet/blob/master/src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs#L155

PR #289, opened 2016-08-19
properties should support concurrency

SergeyKanzhelev:
#288


abaranch:
:shipit:

dnduffy:
It works, I sure hope we don't forget about it and have someone accidentally change it in future.
Can you add a unit test which asserts that the type of the IDictionary provided in key types is ConcurrentDictionary as expected to be safe?
:shipit:

PR #291, opened 2016-08-22
Merge from develop

SergeyKanzhelev:


src/Core/Managed/Shared/Extensibility/Implementation/Property.cs
```diff
@@ -23,6 +23,9 @@ internal static class Property
         public const int MaxMessageLength = 32768;
         public const int MaxUrlLength = 2048;
         public const int MaxCommandNameLength = 2 * 1024;
+        public const int MaxTestNameLength = 1024;
+        public const int MaxRunLocationLength = 2024;
```
dnduffy: Just wondering if this was supposed to be 2048 and why the one two lines up isn't just written as 2048?

dnduffy:
:shipit:

PR #292, opened 2016-08-22
added measurements to availability and dependency, renamed dependency…

SergeyKanzhelev:
1. added `measurements` to availability and dependency
2. updated availability schema
3. renamed `dependencyTypeName` to `type`
4. deprecated `handledAt`


Test/CoreSDK.Test/Shared/DataContracts/DependencyTelemetryTest.cs
```diff
@@ -38,9 +38,9 @@ public void SerializeWritesNullValuesAsExpectedByEndpoint()
             DependencyTelemetry original = new DependencyTelemetry();
             original.Name = null;
             original.Data = null;
-            original.DependencyTypeName = null;
+            original.Type = null;
             original.Success = null;
-            original.DependencyTypeName = null;
+            original.Type = null;
```
dnduffy: Why is there a duplicate of this line two lines up?

dnduffy:
:shipit:

Test/CoreSDK.Test/Shared/DataContracts/DependencyTelemetryTest.cs
SergeyKanzhelev: thanks. fixed. There is a duplication as there used to be setting of both - `DependencyKind` and `DependencyType`. Now it should only be `Type`

SergeyKanzhelev:
Tests failing as Breeze do not support some of those changes yet. Working on it

Issue #293, opened 2016-08-22
Support RichpayloadEventSource in .Net 4.5 package


northtyphoon:
I have a .Net 4.5 project and add AI package. I deploy the project to Azure Website which runs on .Net 4.6. 
To utilize RichpayloadEvent in .Net 4.6, I need to re-target my project to 4.6, delete the old AI package, re-add 4.6 AI package and republish the project. 
Ideally I expect AI .Net 4.5 package can forward-support RichpayloadEvent so that I can directly use the feature as long as the runtime is 4.6 and above.

PR #294, opened 2016-08-22
Enable RichPayloadEventSource in Net45 package.

northtyphoon:
#293 The change is to dynamically enable RichPayloadEventSource in Net45 package. As EventDataAttribute is only available in 4.6, the code uses anonymous type for all telemetry data.

"[Write<T>](https://msdn.microsoft.com/en-us/library/dn823293(v=vs.110).aspx) requires T to be an anonymous type or marked with the EventDataAttribute attribute. In addition, the properties of T must be a supported property type; either a simple framework type such as Int16, Int32, Int64, String, Guid, DateTime or DateTimeOffset, or an array of primitive types."

@SergeyKanzhelev @abaranch

### DIFF
--- a/Test/CoreSDK.Test/Net46/Extensibility/Implementation/RichPayloadEventSourceTest.cs
+++ b/Test/CoreSDK.Test/Net46/Extensibility/Implementation/RichPayloadEventSourceTest.cs
@@ -207,25 +207,25 @@
             var actualProperties = actualEventPayload.Keys.Select(k => new { Key = k, Value = actualEventPayload[k] });
 
             Assert.IsTrue(expectedProperties.Count() == actualProperties.Count());
-            var e1 = expectedProperties.GetEnumerator();
-            var e2 = actualProperties.GetEnumerator();
-            while (e1.MoveNext() && e2.MoveNext())
+            var expectedPropertiesEnumerator = expectedProperties.GetEnumerator();
+            var actualPropertiesEnumerator = actualProperties.GetEnumerator();
+            while (expectedPropertiesEnumerator.MoveNext() && actualPropertiesEnumerator.MoveNext())
             {
-                var expctedProperty = e1.Current;
-                var actualProperty = e2.Current;
-                Assert.AreEqual(e1.Current.Name, e2.Current.Key);
+                var expectedProperty = expectedPropertiesEnumerator.Current;
+                var actualProperty = actualPropertiesEnumerator.Current;
+                Assert.AreEqual(expectedPropertiesEnumerator.Current.Name, actualPropertiesEnumerator.Current.Key);
 
-                if (!expctedProperty.PropertyType.IsValueType
-                    && !expctedProperty.PropertyType.IsPrimitive
-                    && expctedProperty.PropertyType != typeof(string))
+                if (!expectedProperty.PropertyType.IsValueType
+                    && !expectedProperty.PropertyType.IsPrimitive
+                    && expectedProperty.PropertyType != typeof(string))
                 {
-                    if (expctedProperty.PropertyType.IsClass)
+                    if (expectedProperty.PropertyType.IsClass)
                     {
-                        VerifyEventPayload(expctedProperty.PropertyType.GetProperties().AsEnumerable(), (IDictionary<string, object>)actualProperty.Value);
+                        VerifyEventPayload(expectedProperty.PropertyType.GetProperties().AsEnumerable(), (IDictionary<string, object>)actualProperty.Value);
                     }
                     else
                     {
-                        var enumerableType = GetEnumerableType(expctedProperty.PropertyType);
+                        var enumerableType = GetEnumerableType(expectedProperty.PropertyType);
                         Assert.IsNotNull(enumerableType);
                         Assert.IsTrue(actualProperty.Value.GetType().IsArray);
 
--- a/Test/CoreSDK.Test/Net46/Extensibility/Implementation/RichPayloadEventSourceTest.cs
+++ b/Test/CoreSDK.Test/Net46/Extensibility/Implementation/RichPayloadEventSourceTest.cs
@@ -6,7 +6,9 @@
     using System.Linq;
     using Channel;
     using Microsoft.ApplicationInsights.DataContracts;
+
     using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using System.Reflection;
 
     /// <summary>
     /// Tests the rich payload event source tracking.
@@ -23,6 +25,7 @@
             this.DoTracking(
                 RichPayloadEventSource.Keywords.Requests,
                 new RequestTelemetry("TestRequest", DateTimeOffset.Now, TimeSpan.FromMilliseconds(10), "200", true),
+                typeof(External.RequestData),
                 (client, item) => { client.TrackRequest((RequestTelemetry)item); });
         }
 
@@ -35,6 +38,7 @@
             this.DoTracking(
                 RichPayloadEventSource.Keywords.Traces,
                 new TraceTelemetry("TestTrace", SeverityLevel.Information),
+                typeof(External.MessageData),
                 (client, item) => { client.TrackTrace((TraceTelemetry)item); });
         }
 
@@ -47,6 +51,7 @@
             this.DoTracking(
                 RichPayloadEventSource.Keywords.Events,
                 new EventTelemetry("TestEvent"),
+                typeof(External.EventData),
                 (client, item) => { client.TrackEvent((EventTelemetry)item); });
         }
 
@@ -56,9 +61,13 @@
         [TestMethod]
         public void RichPayloadEventSourceExceptionSentTest()
         {
+            var exceptionTelemetry = new ExceptionTelemetry(new SystemException("Test"));
+            exceptionTelemetry.Data.exceptions[0].parsedStack = new External.StackFrame[] { new External.StackFrame() };
+
             this.DoTracking(
                 RichPayloadEventSource.Keywords.Exceptions,
-                new ExceptionTelemetry(new SystemException("Test")),
+                exceptionTelemetry,
+                typeof(External.ExceptionData),
                 (client, item) => { client.TrackException((ExceptionTelemetry)item); });
         }
 
@@ -71,6 +80,7 @@
             this.DoTracking(
                 RichPayloadEventSource.Keywords.Metrics,
                 new MetricTelemetry("TestMetric", 1),
+                typeof(External.MetricData),
                 (client, item) => { client.TrackMetric((MetricTelemetry)item); });
         }
 
@@ -83,6 +93,7 @@
             this.DoTracking(
                 RichPayloadEventSource.Keywords.Dependencies,
                 new DependencyTelemetry("TestDependency", "TestCommand", DateTimeOffset.Now, TimeSpan.Zero, true),
+                typeof(External.RemoteDependencyData),
                 (client, item) => { client.TrackDependency((DependencyTelemetry)item); });
         }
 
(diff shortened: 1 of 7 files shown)


abaranch:
> namespace Microsoft.ApplicationInsights.Extensibility.Implementation

Could you please compile in release? I think stylecop may complain.

---
Refers to: src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.cs:1 in 1f6a40f. [](commit_id = 1f6a40f3b9b036de8f2ad034b311bf792fcb18d2, deletion_comment = False)

abaranch:
Could you please add tests.

abaranch:
:shipit:

Test/CoreSDK.Test/Net46/Extensibility/Implementation/RichPayloadEventSourceTest.cs
```diff
@@ -95,61 +102,116 @@ public void RichPayloadEventSourcePageViewSentTest()
...
-                Assert.IsNotNull(actualEvent.Payload[2]);
+        private static bool IsRunningOnVersion46AndAbove()
+        {
+            string productVersionString = System.Diagnostics.FileVersionInfo.GetVersionInfo(typeof(object).Assembly.Location).ProductVersion;
+
+            Version ver;
+            if (!Version.TryParse(productVersionString, out ver))
+            {
+                // Default is 4.6
+                return true;
```
dnduffy: Assuming v4.6 when parsing fails seems dangerous.
Shouldn't you err on the side of caution and assume that if you can't parse the expected version that it must not be that version and therefore turn off the "feature"?

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.cs
```diff
@@ -0,0 +1,498 @@
...
+                        extendedData = new
+                        {
+                            PartA_iKey = telemetryItem.Context.InstrumentationKey,
+                            PartA_Tags = telemetryItem.Context.Tags,
+                            PartB_PageViewPerfData = new { data.duration, data.measurements, data.name, data.properties, data.url, data.ver }
+                        };
+                        break;
+                    }
+
+                case SessionStateTelemetry.TelemetryName:
```
dnduffy: nitpick: There are a lot of string compare cases in the switch statement, you could build a jump table instead.

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.cs
```diff
@@ -0,0 +1,498 @@
...
+                var pageViewData = new PageViewData();
+                writeMethod = writeGenericMethod.MakeGenericMethod(new
+                {
+                    PartA_iKey = partA_iKey_Value,
+                    PartA_Tags = partA_Tags_Value,
+                    PartB_PageViewPerfData = new { pageViewData.duration, pageViewData.measurements, pageViewData.name, pageViewData.properties, pageViewData.url, pageViewData.ver }
+                }.GetType());
+                this.telemetryHandlers.Add(typeof(PageViewTelemetry), new TelemetryHandler(PageViewTelemetry.TelemetryName, writeMethod, keywords, eventSourceOptions));
+
+                // SessionState
```
dnduffy: It might be more readable and maintainable to break these per type blocks of code into methods.

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.cs
```diff
@@ -0,0 +1,498 @@
...
+                eventSourceOptions = Activator.CreateInstance(eventSourceOptionsType);
+                keywords = Keywords.SessionState;
+                eventSourceOptionsKeywordsProperty.SetValue(eventSourceOptions, keywords);
+                var sessionStateData = new SessionStateData();
+                writeMethod = writeGenericMethod.MakeGenericMethod(new
+                {
+                    PartA_iKey = partA_iKey_Value,
+                    PartA_Tags = partA_Tags_Value,
+                    PartB_SessionStateData = new { sessionStateData.state, sessionStateData.ver }
```
dnduffy: It would definitely be clearer to either use concrete types in place of all the anonymous classes or at least thoroughly document the data fields used, and why, for each "type".  There are so many different anonymous class objects being defined that it is quite verbose and yet not clear for anyone else to read.

dnduffy: P.S. There needs to be some kind of "schema", preferably in class form, that clearly documents what fields, and in what order, are being sent for each type.

Test/CoreSDK.Test/Net46/Extensibility/Implementation/RichPayloadEventSourceTest.cs
northtyphoon: It's unit test and I assume most time it will run on machine with 4.6 runtime and above. I will change to assert.fail if it's more clear.

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.cs
northtyphoon: Unfortunately we have to use anonymous type as 4.5 project doesn't support EventDataAttribute. It's a requirement to emit ETW RichPayload event. I will add documents.

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.cs
```diff
@@ -0,0 +1,498 @@
...
+                                    i.stack,
+                                    i.typeName
+                                }),
+                                data.handledAt,
+                                data.measurements,
+                                data.problemId,
+                                data.properties,
+                                data.severityLevel,
+                                data.ver
+                            }
```
SergeyKanzhelev: If you can move generation of every anonymous class into separate file - we can generate those from the bond schema in future. The similar way we generate data classes today: https://github.com/Microsoft/ApplicationInsights-dotnet/blob/publicSchemaUpdate/Schema/generateSchema.ps1

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -309,7 +309,7 @@ public void Track(ITelemetry telemetry)
                 // invokes the Process in the first processor in the chain
                 this.configuration.TelemetryProcessorChain.Process(telemetry);
 
-#if NET46
+#if NET46 || NET45
```
SergeyKanzhelev: is FW40 difficult to support? Is it because of anonymous classes?

northtyphoon: EventSource is available from 4.5. Is 4.0 important? I may want to look at .Net Core next.

northtyphoon:
Hi, @SergeyKanzhelev @dnduffy @abaranch, thanks for the review. I addressed the comments and updated the PR. Can you please take a look?

northtyphoon:
By the way, I added the validation in unit test to compare the anonymous type and telemetry class recursively. It should be able to catch up the schema change later.

Test/CoreSDK.Test/Net46/Extensibility/Implementation/RichPayloadEventSourceTest.cs
```diff
@@ -95,61 +106,163 @@ public void RichPayloadEventSourcePageViewSentTest()
...
+            }
+        }
+
+        private static void VerifyEventPayload(IEnumerable<PropertyInfo> expectedProperties, IDictionary<string, object> actualEventPayload)
+        {
+            var actualProperties = actualEventPayload.Keys.Select(k => new { Key = k, Value = actualEventPayload[k] });
+
+            Assert.IsTrue(expectedProperties.Count() == actualProperties.Count());
+            var e1 = expectedProperties.GetEnumerator();
+            var e2 = actualProperties.GetEnumerator();
```
dnduffy: Prefer longer meaningful names like "expectedPropertiesEnumerator" instead of "e1" which doesn't mean anything on it's own.

Test/CoreSDK.Test/Net46/Extensibility/Implementation/RichPayloadEventSourceTest.cs
```diff
@@ -95,61 +106,163 @@ public void RichPayloadEventSourcePageViewSentTest()
...
+            while (e1.MoveNext() && e2.MoveNext())
+            {
+                var expctedProperty = e1.Current;
```
dnduffy: Spelling for "expectedProperty".

Test/CoreSDK.Test/Net46/Extensibility/Implementation/RichPayloadEventSourceTest.cs
northtyphoon: Thanks. Will change the variable name. For the var, I don't think "var" in this case impacts understanding the code as long as the variable name provide the essential information. But if you insist it's a coding style in AI Lib I need to follow, I will change it.

Test/CoreSDK.Test/Net46/Extensibility/Implementation/RichPayloadEventSourceTest.cs
northtyphoon: Thanks for catching it.

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.TelemetryHandler.cs
```diff
@@ -0,0 +1,604 @@
...
+            var keywords = Keywords.Requests;
+            eventSourceOptionsKeywordsProperty.SetValue(eventSourceOptions, keywords);
+            var dummyRequestData = new RequestData();
+            var writeMethod = writeGenericMethod.MakeGenericMethod(new
+            {
+                PartA_iKey = this.dummyPartAiKeyValue,
+                PartA_Tags = this.dummyPartATagsValue,
+                PartB_RequestData = new
+                {
+                    // The properties and layout should be the same as RequestData_types.cs
```
dnduffy: That's good, so why not just make PartB_RequestData = new RequestData?

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.TelemetryHandler.cs
```diff
@@ -0,0 +1,604 @@
...
+                // SessionState
+                this.telemetryHandlers.Add(typeof(SessionStateTelemetry), this.CreateHandlerForSessionStateTelemetry(writeGenericMethod, eventSourceOptionsType, eventSourceOptionsKeywordsProperty));
+            }
+            else
+            {
+                CoreEventSource.Log.LogVerbose("Unable to get method: EventSource.Write<T>(String, EventSourceOptions, T)");
+            }
+        }
+
+        private Action<ITelemetry> CreateHandlerForRequestTelemetry(MethodInfo writeGenericMethod, Type eventSourceOptionsType, PropertyInfo eventSourceOptionsKeywordsProperty)
```
dnduffy: Please add the doc comments to the new methods, thank-you. :)

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.TelemetryHandler.cs
```diff
@@ -0,0 +1,604 @@
...
+                        PartA_Tags = telemetryItem.Context.Tags,
+                        PartB_PageViewPerfData = new
+                        {
+                            data.url,
+                            data.duration,
+                            data.ver,
+                            data.name,
+                            data.properties,
+                            data.measurements,
+                        }
```
dnduffy: Can't the declarations for the write method and for the return object be merged and shared by both?

northtyphoon: It will have a performance impact. If we create the method each time we emit the event, 1 million calls take 3000 ms. If we pre-create the method, it takes 300 ms.

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.TelemetryHandler.cs
northtyphoon: That's the reason why we need anonymous type. To use the code like "PartB_RequestData = new RequestData". RequestData needs to have EventDataAttribute but it's only available since .Net 4.6.

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.TelemetryHandler.cs
northtyphoon: I also thought about caching the method for each anonymous type in the fly. I chose current implementation as the logic is simpler and straightforward. The unit test will catch up if anything inconsistent.

dnduffy: Sorry, I didn't mean to imply the method should be recreated each time, maybe I'm just having trouble following the scoping with all the anonymous inline declarations.

dnduffy:
Can we backport the EventDataAttribute, or pick it up from a Microsoft NuGet package, or some other such workaround when the framework is < 4.6?
My main concern is that these anonymous recreations of classes which are generated from the Bond schema files will very easily get out of sync with the real classes when the schema changes.  I worry that this code will become difficult for others to read and maintain in future.
For example, if a new field is added to the schema in the Bond files there is no compile time check to catch that change and require it in this code.

dnduffy:
:shipit:

northtyphoon:
@dnduffy I actually tried porting EventDataAttribute and EventTags before converting everything to anonymous type but it didn't work. An alternative possible solution is using [Microsoft.Diagnostics.Tracing.EventSource](https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.EventSource.Redist) for ETW events in AI 4.5 SDK and below. I need to check with @vancem if the package supports RichPayloadEvents.
To the concern about the schema change, compile will fail if the old property is renamed or removed. unit test will fail if new property is added or new class is defined.

PR #299, opened 2016-09-01
Allow all characters in property names. Change sanitization to optimi…

abaranch:
- removed regex that was used to check that we have only approved characters in property names
- changed sanitation on optimistic approach where we re-add properties only when we actually did sanitation. In 99% cases that should work faster..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This changelog will be used to generate documentation on [release notes page](ht
 
 - InMemoryChannel has a new override for Flush method that accepts timeout.
 - Local storage folder name was changed. That means that when the application stopped, and the application was updated to the new SDK, then the telemetry from the old local folder will not be send.
-- Allow all characters in property names.
+- Allow all characters in property names and measurements names.
 
 ## Version 2.2.0-beta1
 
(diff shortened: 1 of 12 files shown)


SergeyKanzhelev:
Should do the same here: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/master/Src/PerformanceCollector/Shared/PerformanceCollectorModule.cs#L25

abaranch:
That is a different repo. I can do that separately later. I will also be updating JS SDK.

---
In reply to: [244208354](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/299#issuecomment-244208354) [](ancestors = 244208354)

SergeyKanzhelev:
Should you worry about serialization of properties when we have symbols like `\0` or something strange like this?

SergeyKanzhelev:
if not - :shipit:

dnduffy:
:+1:

abaranch:
I tested it with right-to-left strings and UI was fine. I think we do not need to do anything special.

---
In reply to: [244235817](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/299#issuecomment-244235817) [](ancestors = 244235817)

SergeyKanzhelev:
http://www.ietf.org/rfc/rfc4627.txt 

>A string begins and ends with quotation marks.  All Unicode characters may be placed within the quotation marks except for the characters that must be escaped: quotation mark, reverse solidus, and *the control characters (U+0000 through U+001F)*.

I don't see it here: https://github.com/Microsoft/ApplicationInsights-dotnet/blob/1ccd5b466451db1884b242b6a58b3e9401a1fbcf/src/Core/Managed/Shared/Extensibility/Implementation/JsonWriter.cs#L175

SergeyKanzhelev:
Like this: `var str = "\x0000\x0001\x0002\x0003\x0004\x0005 ... \x009F";`

PR #300, opened 2016-09-02
Enable RichPayloadEventSource in Net40 package.

northtyphoon:
Net40 uess Nuget version of EventSource which has already supported RichPayloadEvents. The change simply turns it on.


dnduffy:
:shipit:

PR #304, opened 2016-09-07
Add try/catches to config reading

abaranch:
- Add try/catches to config reading so TelemetryConfiguration.Active does not throw
- Remove couple methods from Platform interface (which needs to be removed all together since now we have only one implementation)

### DIFF
--- a/src/Core/Managed/Net40/Extensibility/Implementation/Platform/PlatformImplementation.cs
+++ b/src/Core/Managed/Net40/Extensibility/Implementation/Platform/PlatformImplementation.cs
@@ -1,25 +1,18 @@
 ﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Platform
 {
     using System;
-    using System.Collections.Generic;
     using System.IO;
+    using System.Security;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
-    using Microsoft.ApplicationInsights.Extensibility.Implementation.External;
     using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
 
     /// <summary>
     /// The .NET 4.0 and 4.5 implementation of the <see cref="IPlatform"/> interface.
     /// </summary>
-    internal class PlatformImplementation : 
-        IPlatform
+    internal class PlatformImplementation : IPlatform
     {
         private IDebugOutput debugOutput = null;
-
-        public IDictionary<string, object> GetApplicationSettings()
-        {
-            throw new NotImplementedException();
-        }
 
         /// <summary>
         /// Returns contents of the ApplicationInsights.config file in the application directory.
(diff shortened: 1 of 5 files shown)


src/Core/Managed/Net40/Extensibility/Implementation/Platform/PlatformImplementation.cs
```diff
@@ -29,32 +22,47 @@ public string ReadConfigurationXml()
...
+            try
             {
-                return File.ReadAllText(configFilePath);
+                // Ensure config file actually exists
+                if (File.Exists(configFilePath))
+                {
+                    return File.ReadAllText(configFilePath);
+                }
+            }
+            catch (FileNotFoundException)
```
dnduffy: How can there be a FileNotFoundException if you just checked File.Exists?

src/Core/Managed/Net40/Extensibility/Implementation/Platform/PlatformImplementation.cs
```diff
@@ -29,32 +22,47 @@ public string ReadConfigurationXml()
...
+                CoreEventSource.Log.ApplicationInsightsConfigNotFoundWarning(configFilePath);
+            }
+            catch (UnauthorizedAccessException)
+            {
+                CoreEventSource.Log.ApplicationInsightsConfigNotFoundWarning(configFilePath);
+            }
+            catch (SecurityException)
+            {
+                CoreEventSource.Log.ApplicationInsightsConfigNotFoundWarning(configFilePath);
             }
```
dnduffy: What about PathTooLongException and NotSupportedException?

src/Core/Managed/Net40/Extensibility/Implementation/Platform/PlatformImplementation.cs
```diff
@@ -29,32 +22,47 @@ public string ReadConfigurationXml()
...
+                // For cases when file was deleted/modified while reading
+                CoreEventSource.Log.ApplicationInsightsConfigNotFoundWarning(configFilePath);
+            }
+            catch (UnauthorizedAccessException)
+            {
+                CoreEventSource.Log.ApplicationInsightsConfigNotFoundWarning(configFilePath);
+            }
+            catch (SecurityException)
+            {
+                CoreEventSource.Log.ApplicationInsightsConfigNotFoundWarning(configFilePath);
```
dnduffy: What about PathTooLongException and NotSupportedException?

src/Core/Managed/Net40/Extensibility/Implementation/Platform/PlatformImplementation.cs
abaranch: Yes, you got it right

---
In reply to: [77934014](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/304#discussion_r77934014) [](ancestors = 77934014)

src/Core/Managed/Net40/Extensibility/Implementation/Platform/PlatformImplementation.cs
abaranch: PathTooLong is IOException. NotSupportedException is thrown when path is in incorrect format - we create path on line 23; it is not possible to have incorrect format.

---
In reply to: [77935073](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/304#discussion_r77935073) [](ancestors = 77935073)

dnduffy:
:+1:

PR #315, opened 2016-09-26
Read iKey from environment variable (APPINSIGHTS_INSTRUMENTATIONKEY)

abaranch:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/136


src/Core/Managed/PCL/Extensibility/Implementation/Platform/PlatformImplementation.cs
```diff
@@ -35,5 +35,10 @@ public IDebugOutput GetDebugOutput()
             
             return this.debugOutput;
         }
+
+        public string GetEnvironmentVariable(string name)
+        {
+            return null;
```
dnduffy: Is there maybe a .NET Core equivalent for getting environment variables we could use?

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -371,6 +373,11 @@ public void Initialize(ITelemetry telemetry)
             if (string.IsNullOrEmpty(instrumentationKey))
             {
                 instrumentationKey = this.configuration.InstrumentationKey;
+
+                if (string.IsNullOrEmpty(instrumentationKey))
+                {
+                    instrumentationKey = PlatformSingleton.Current.GetEnvironmentVariable(InstrumentationKeyWebSitesEnvironmentVariable);
```
SergeyKanzhelev: So this ikey has the least priority? Is it what we want for Web Apps case? In Web Apps you can configure ikey in `AI.config` file and then override it using Web App settings. Not other way around.

Also I'm concerned about performance of it. Maybe just read it once and cache? Or we need to support the scenario when environment variable will be changed later?

src/Core/Managed/PCL/Extensibility/Implementation/Platform/PlatformImplementation.cs
dnduffy: Maybe something in Microsoft.Extensions.Configuration.EnvironmentVariables or Microsoft.Extensions.Configuration could be used.

src/Core/Managed/Shared/TelemetryClient.cs
dnduffy: I agree on caching, discussing the priority ordering, but I don't think it should change during runtime.

src/Core/Managed/PCL/Extensibility/Implementation/Platform/PlatformImplementation.cs
abaranch: Not in PCL. Those are .Net Core specific namespaces. If ever change target framework we can start supporting it.

dnduffy: Good point, I keep thinking of .NET Core as a replacement for PCL but it is completely new and isn't portable backwards (as such).

src/Core/Managed/Shared/TelemetryClient.cs
abaranch: Yes, priority order is correct for web sites (@iusafaro)
Resolved caching.

SergeyKanzhelev: @iusafaro, this priority order doesn't make sense to me. Can you pls elaborate why it is done this way?

abaranch: Per offline discussion we will change the order so it is: code -> environment variable -> config

SergeyKanzhelev:
@iusafaro can you please ensure that we do not need to re-read instrumentation key from the environment variable after Web App swap?

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -370,7 +372,12 @@ public void Initialize(ITelemetry telemetry)
 
             if (string.IsNullOrEmpty(instrumentationKey))
             {
-                instrumentationKey = this.configuration.InstrumentationKey;
+                instrumentationKey = PlatformSingleton.Current.GetEnvironmentVariable(InstrumentationKeyWebSitesEnvironmentVariable);
+                
+                if (string.IsNullOrEmpty(instrumentationKey))
+                {
+                    instrumentationKey = this.configuration.InstrumentationKey;
```
SergeyKanzhelev: So we doing it before initializers have a chance to override it?

abaranch: Yes

SergeyKanzhelev: Thinking about it - I guess it is fine

PR #317, opened 2016-09-28
Sergkanz/level up versions

SergeyKanzhelev:
I use the same version of `Microsoft.ApplicationInsights.dll` in both nuspec. this solves the problem of uneven versions for nugget packages

### DIFF
--- a/src/TelemetryChannels/ServerTelemetryChannel/NuGet/TelemetryChannel.NuGet.csproj
+++ b/src/TelemetryChannels/ServerTelemetryChannel/NuGet/TelemetryChannel.NuGet.csproj
@@ -16,7 +16,7 @@
     <NoWarn>2008</NoWarn>
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\</SolutionDir>
     <PackageSpecFile>$(MSBuildProjectDirectory)\Package.nuspec</PackageSpecFile>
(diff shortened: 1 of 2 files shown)


src/TelemetryChannels/ServerTelemetryChannel/NuGet/TelemetryChannel.NuGet.csproj
```diff
@@ -38,7 +38,7 @@
     </None>
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\..\..\Core\Managed\Net45\Core.Net45.csproj">
+    <ProjectReference Include="..\..\..\Core\Managed\Net40\Core.Net40.csproj">
```
dnduffy: Why are these changing from Net45 to Net40 when the target framework version is still 4.5?
"<TargetFrameworkVersion>v4.5</TargetFrameworkVersion>"

abaranch: That is explain by Sergey in description: Description: I use the same version of `Microsoft.ApplicationInsights.dll` in both nuspec. this solves the problem of uneven versions for nugget packages

---
In reply to: [80967630](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/317#discussion_r80967630) [](ancestors = 80967630)

SergeyKanzhelev: This reference is for build ordering only. Let me remove it

Issue #319, opened 2016-09-30
Telemetry with empty properties collection is dropped


abaranch:
```
            var properties = new Dictionary<string, string>
            {
                {"UserId", ""}
            };
            new TelemetryClient().TrackException(new ApplicationException(), properties);
```

Invalid json, no "," after properties collection.
 

> Application Insights Telemetry: {"name":"Microsoft.ApplicationInsights.XXX.Exception","time":"2016-09-30T23:35:56.1080471Z","iKey":"XXX","tags":{"ai.internal.sdkVersion":"dotnet:2.2.0-31572","ai.operation.id":"/buMs/VxSMI=","ai.operation.name":"GET /Default","ai.operation.parentId":"/buMs/VxSMI=","ai.device.roleInstance":"ZZZ.redmond.corp.microsoft.com"},"data":{"baseType":"ExceptionData","baseData":{"ver":2,"handledAt":"UserCode","properties":{}"exceptions":[{"id":31494208,"typeName":"System.ApplicationException","message":"Error in the application.","hasFullStack":true}]}}}

hallatore:
I think the bug is here: https://github.com/Microsoft/ApplicationInsights-dotnet/blob/37cec526194b833f7cd676f25eafd985dd88d3fa/src/Core/Managed/Shared/Extensibility/Implementation/JsonWriter.cs#L164

It only adds comma to properties with values. But since it strips out null, etc it ends up writing the property but never the comma.

PR #320, opened 2016-10-01
Removed null check as it can break the JsonWriter flow.

hallatore:
We should't remove properties that are null as it might brake the flow of the JsonWriter.

PS: Why was this done in the first place? What could break on the back end with this change?

Ref #319

### DIFF
--- a/src/Core/Managed/Shared/Extensibility/Implementation/JsonWriter.cs
+++ b/src/Core/Managed/Shared/Extensibility/Implementation/JsonWriter.cs
@@ -128,11 +128,6 @@
                 this.WriteStartObject();
                 foreach (KeyValuePair<string, string> item in values)
                 {
-                    if (item.Value == null)
-                    {
-                        continue;
-                    }
-
                     this.WriteProperty(item.Key, item.Value);
                 }
 


abaranch:
Thank you for your contribution. I will check whether we can fix it on the serializer side before accepting the contribution. I think the initial idea was to not send null values since that unnecessarily increases package size. (Also I'm pretty sure some unit tests would fail.)

hallatore:
I get why the code is there, but feel it might be at the wrong place. Right now it looks like a hack since it's placed in the middle of it all.

And it's also not included on this almost identical method. https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/src/Core/Managed/Shared/Extensibility/Implementation/JsonWriter.cs#L108-L121

abaranch:
Sorry, I got pulled away by a different task today.
Do you have a proposal how to fix it but still keep removing null values?..

hallatore:
I would normally say that it makes little sense to trim out the null's if it isn't a big issue. So how does the compressed payload look without this code? How much are we adding?

Easiest way to patch this (if it's a problem) is probably at the `.Count > 0` step at https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/src/Core/Managed/Shared/Extensibility/Implementation/JsonWriter.cs#L110

abaranch:
I do not have numbers right away but since you are saying that fixing it with keeping existing behavior has the same complexity, let's keep the behavior.

abaranch:
I created a pull request that includes both
- to keep existing functionality, change sanitization rules
- your fix in case if in the future we decide that we want to start sending nulls  
Thank you for finding the issue and help with resolving it,.

PR #323, opened 2016-10-06
Metric aggregation public surface changes

vitalyf007:
This PR contains all proposed changes for AI SDK public surface to support metric aggregation.


src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
```diff
@@ -0,0 +1,157 @@
...
+        void ITelemetry.Sanitize()
+        {
+            this.Name = this.Name.SanitizeName();
+            this.Name = Utils.PopulateRequiredStringValue(this.Name, "name", typeof(AggregatedMetricTelemetry).FullName);
+            this.Properties.SanitizeProperties();
+            this.Count = this.Count < 0 ? 0 : this.Count;
+            this.Sum = Utils.SanitizeNanAndInfinity(this.Sum);
+            this.Min = Utils.SanitizeNanAndInfinity(this.Min);
+            this.Max = Utils.SanitizeNanAndInfinity(this.Max);
+            this.StandardDeviation = Utils.SanitizeNanAndInfinity(this.StandardDeviation);
```
dnduffy: Why not sanitize them on input (set) instead of as a separate step?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// Creates single time series data aggregator.
+        /// </summary>
+        /// <param name="metricName">Name of the metric.</param>
+        /// <param name="dimensions">Optional dimensions.</param>
+        /// <returns>Value aggregator for the metric specified.</returns>
+        public MetricAggregator CreateMetricAggregator(string metricName, IDictionary<string, string> dimensions = null)
+        {
+            throw new NotImplementedException();
```
dnduffy: Should I assume that the various "not implemented" methods will be implemented in the future?

Should I also assume that at some point the now marked as obsolete class MetricTelemetry will be deleted in favor of the new AggregatedMetricTelemetry?

vitalyf007: This is a PR for public surface only (from one private branch to the other, not to develop yet). Naturally, "not implemented" methods will be implemented by the time we go to "develop".

As far as obsolete things - we can totally remove them in the next major version. For now they're marked obsolete and hidden from intellisense...

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
vitalyf007: I just copied the way it was done in other types. Are you guys proposing to change it for all types (that would make sense)?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
dnduffy: Okay, thank-you for confirming.

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
dnduffy: I don't have context on the history behind it but if there isn't a strong reason to do it like this I would propose changing it.  Let's ask around in the morning.

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        /// <summary>
+        /// Sanitizes the properties based on constraints.
+        /// </summary>
+        void ITelemetry.Sanitize()
+        {
+            this.Name = this.Name.SanitizeName();
+            this.Name = Utils.PopulateRequiredStringValue(this.Name, "name", typeof(AggregatedMetricTelemetry).FullName);
+            this.Properties.SanitizeProperties();
+            this.Count = this.Count < 0 ? 0 : this.Count;
```
SergeyKanzhelev: should it be 1, not zero?

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
```diff
@@ -0,0 +1,157 @@
...
+            get { return this.Metric.name; }
+            set { this.Metric.name = value; }
+        }
+
+        /// <summary>
+        /// Gets or sets the number of samples for this metric.
+        /// </summary>
+        public int Count
+        {
+            get { return this.Metric.count ?? 0; }
```
SergeyKanzhelev: should the default be 1, not zero?

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
```diff
@@ -236,9 +236,11 @@ private static void SerializeTelemetryItem(ITelemetry telemetryItem, JsonWriter
                 ExceptionTelemetry exceptionTelemetry = telemetryItem as ExceptionTelemetry;
                 SerializeExceptionTelemetry(exceptionTelemetry, jsonWriter);
             }
+#pragma warning disable CS0618
```
SergeyKanzhelev: should you create a serializer for the `AggregatedMetricTelemetry`?

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
```diff
@@ -0,0 +1,157 @@
...
+        /// <remarks>
+        /// Metric statistics provided are assumed to be calculated over a period of time equaling 1 minute.
+        /// </remarks>
+        /// <param name="metricName">Metric name.</param>
+        /// <param name="count">Count of values taken during aggregation interval.</param>
+        /// <param name="sum">Sum of values taken during aggregation interval.</param>
+        /// <param name="min">Minimum value taken during aggregation interval.</param>
+        /// <param name="max">Maximum of values taken during aggregation interval.</param>
+        /// <param name="standardDeviation">Standard deviation of values taken during aggregation interval.</param>
+        public AggregatedMetricTelemetry(
```
SergeyKanzhelev: Can you please expose the constructor without sum, min, max, etc. Mostly to report performance counter values, but also for the back compat

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
```diff
@@ -0,0 +1,157 @@
...
+        /// Sanitizes the properties based on constraints.
+        /// </summary>
+        void ITelemetry.Sanitize()
+        {
+            this.Name = this.Name.SanitizeName();
+            this.Name = Utils.PopulateRequiredStringValue(this.Name, "name", typeof(AggregatedMetricTelemetry).FullName);
+            this.Properties.SanitizeProperties();
+            this.Count = this.Count < 0 ? 0 : this.Count;
+            this.Sum = Utils.SanitizeNanAndInfinity(this.Sum);
+            this.Min = Utils.SanitizeNanAndInfinity(this.Min);
```
SergeyKanzhelev: this call will initialize nullable field. Which is bad as we will need to pass the value over the wire even if we do not want to. Typical example is performance counter value. I do not want to spent bytes over the wire to send the `min` and `max` which are equal to the current value.

SergeyKanzhelev: Also add a unit test that will ensure that not setting `min` and `max` will result to not serializing of those

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
vitalyf007: I was looking more into this and I think I'd have to leave it the was it was coded originally. Sanitization is considered a part of the serialization, actually, and if we change it only for aggregated metric telemetry type, it will look odd and different to all other types. I'd prefer, if you like, to document (via a story) a future effort to clean that up across all telemetry types.

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
vitalyf007: I expose parameter-less ctor that can be used along with property setting to populate whatever properties I like. I long considered adding others or making, say, min/max optional in some way on the ctor that takes all properties, but ruled that out at the end. I want user to explicitly state what they want to do and not omit anything. I think the schema I have right now is best, but we can discuss offline.

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
vitalyf007: Some comment as above - why do you think so?

src/Core/Managed/Shared/Extensibility/Implementation/JsonSerializer.cs
vitalyf007: I did. I was omitted in this commit but will be there for the next one (will be available very shortly)

vitalyf007:
Please note an update on this PR. This is a "nearly" final version. I need to think about a couple of things here and I have tracing not yet done. Otherwise, unless you point out holes and/or problems I propose we consider this as a "we're done" stage and try use it in internal efforts such as Sergey's exception stats feature.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,275 @@
...
+
+    /// <summary>
+    /// Represents a hub for metric aggregation procedures.
+    /// </summary>
+    public sealed class MetricAggregatorManager : IDisposable
+    {
+        /// <summary>
+        /// Name of the property added to aggregation results to indicate duration of the aggregation interval.
+        /// </summary>
+        private static string intervalDurationPropertyName = "IntervalDurationMs";
```
SergeyKanzhelev: Why do we need it? I thought we agreed that we assume 1 minute as a default interval.

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -312,6 +312,42 @@ public void FailedToSend(string msg, string appDomainName = "Incorrect")
...
+            this.WriteEvent(
+                26,
+                ex ?? string.Empty,
+                this.nameProvider.Name);
+        }
+
+        [Event(
+            27,
+            Message = "Failed to invoke metric processor. Exception: {0}.",
+            Level = EventLevel.Error)]
```
SergeyKanzhelev: Level "Error" means that this trace will be sent to the portal as SDK "self-diagnostics" event. To make it more understandable for the user I'd recommend to mark it "User Actionable", trace the processor name and write instruction that "if issue persists - disable this processor"

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,275 @@
...
+        public MetricAggregatorManager(TelemetryClient client)
+        {
+            this.Client = client ?? new TelemetryClient();
+            this.aggregatorDictionary = new ConcurrentDictionary<MetricAggregator, SimpleMetricStatisticsAggregator>();
+
+            this.lastSnapshotStartDateTime = DateTimeOffset.UtcNow;
+
+            this.cancellationSource = new CancellationTokenSource();
+
+            this.snapshotTask = new Task(this.SnapshotRunner, TaskCreationOptions.LongRunning);
```
SergeyKanzhelev: It makes the third thread SDK runs in background. QP configuration, Performance Counters collection and this aggregator. Can you think of any options how to combine this thread with Performance Counters collection thread?

SergeyKanzhelev: It is lower priority comment, we can convert it to "enhancement" issue

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
SergeyKanzhelev: I still think those fields should be optional and we do not need to send them if they are not available. Let's discuss

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -133,6 +134,15 @@ public IList<ITelemetryInitializer> TelemetryInitializers
         }
 
         /// <summary>
+        /// Gets the list of <see cref="IMetricProcessor"/> objects used for custom metric data processing
+        /// before client-side metric aggregation process.
+        /// </summary>
+        public IList<IMetricProcessor> MetricProcessors
```
SergeyKanzhelev: Name `processor` makes it inconsistent with the `TelemetryProcessors` which combines the chain rather than sequential processing. Do you have any other names in mind?

src/Core/Managed/Shared/Extensibility/MetricAggregator.cs
```diff
@@ -0,0 +1,172 @@
...
+
+    /// <summary>
+    /// Represents aggregator for a single time series of a given metric.
+    /// </summary>
+    public class MetricAggregator : IEquatable<MetricAggregator>
+    {
+        /// <summary>
+        /// Aggregator manager for the aggregator.
+        /// </summary>
+        private MetricAggregatorManager manager;
```
SergeyKanzhelev: make readonly

src/Core/Managed/Shared/Extensibility/MetricAggregator.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        /// <summary>
+        /// Metric aggregator id to look for in the aggregator dictionary.
+        /// </summary>
+        private string aggregatorId;
```
SergeyKanzhelev: readonly

src/Core/Managed/Shared/Extensibility/MetricAggregator.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        /// <summary>
+        /// Aggregator hash code.
+        /// </summary>
+        private int hashCode;
```
SergeyKanzhelev: same

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
vitalyf007: ERROR IS CUSTOMER FACING. Get type of the processor. User-actionable. Suggest turning off processor if persists.

src/Core/Managed/Shared/Extensibility/MetricAggregator.cs
vitalyf007: Fixed.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
vitalyf007: I was changed to "processor" from "metric stream listener"to make it more consistent :) I'll leave it as is for now.

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
vitalyf007: Fixed.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
vitalyf007: As discussed, will keep to address the shortened intervals at the start and shutdown times for the application.

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
```diff
@@ -149,7 +149,12 @@ void ITelemetry.Sanitize()
             this.Name = this.Name.SanitizeName();
             this.Name = Utils.PopulateRequiredStringValue(this.Name, "name", typeof(AggregatedMetricTelemetry).FullName);
             this.Properties.SanitizeProperties();
-            this.Count = this.Count < 0 ? 0 : this.Count;
+
+            // note: we set count to 1 if it isn't a postitive integer
+            // thinking that if it is zero (negative case is clearly broken)
+            // that most likely menas somebody set the sum but forgot to set count
```
SergeyKanzhelev: mistype "menas" instead of "means"

vitalyf007: :) Fixed

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
olegsych: Why do we need this constructor? It's unreadable at the point of call. 
```C#
new AggregatedMetricTelemetry("foo", 1, 2, 3, 4, 5);
```

You can make it readable by supplying parameter names, but you can achieve the same effect with property syntax.
```C#
new AggregatedMetricTelemetry("foo", count: 1, sum: 2, min: 3, max: 4, standardDeviation: 5);
new AggregatedMetricTelemetry { Name = "Foo", Count = 1, Sum = 2, Min = 3, Max = 4, StandardDeviation = 5); 
```

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
olegsych: +1. the `Sanitize` method will change it from 0 to 1 so the default value examined through the object will be different than the value sent on the wire..

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
```diff
@@ -0,0 +1,164 @@
...
+        public double Sum
+        {
+            get { return this.Metric.value; }
+            set { this.Metric.value = value; }
+        }
+
+        /// <summary>
+        /// Gets or sets the min value of this metric.
+        /// </summary>
+        public double Min
```
olegsych: Does the non-nullable `double` type indicate that `Count`, `Min`, `Max` and `StandardDeviation` properties are required to have value?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        /// Cancellation token source to allow cancellation of the snapshotting task.
+        /// </summary>
+        private CancellationTokenSource cancellationSource;
+
+        /// <summary>
+        /// Metric aggregator snapshotting task.
+        /// </summary>
+        private Task snapshotTask;
```
olegsych: Why re-implement what the `TaskTimer` does here?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        /// Last time snapshot was initiated.
+        /// </summary>
+        private DateTimeOffset lastSnapshotStartDateTime;
+
+        /// <summary>
+        /// Cancellation token source to allow cancellation of the snapshotting task.
+        /// </summary>
+        private CancellationTokenSource cancellationSource;
```
olegsych: Can any of these fields be `readonly`?

src/Core/Managed/Shared/Extensibility/MetricAggregator.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        /// <summary>
+        /// Forwards value to metric processors.
+        /// </summary>
+        /// <param name="value">Value tracked on time series.</param>
+        private void ForwardToProcessors(double value)
+        {
+            // create a local reference to metric processor collection
+            // if collection changes after that - it will be copied not affecting local reference
+            IList<IMetricProcessor> metricProcessors = this.manager.Client.TelemetryConfiguration.MetricProcessors;
```
olegsych: Ouch. This is a pretty bad violation of the [law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter). From customer viewpoint, this means that in order to test the use of `MetricAggregator` in my code, I'll need to understand a chain of objects 4 levels deep and create it for every test. In practice, I think few people will go through the effort successfully and will decide to wrap the `MetricAggregator` and the `MetricAggregatorManager` in application abstractions as "untestable external dependencies" instead. Is it possible to make `IEnumerable<IMetricProcessor>` a direct dependency of the `MetricAggregator`?

src/Core/Managed/Shared/Extensibility/MetricAggregator.cs
```diff
@@ -0,0 +1,172 @@
...
+    using System.Text;
+    using System.Threading;
+
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+
+    /// <summary>
+    /// Represents aggregator for a single time series of a given metric.
+    /// </summary>
+    public class MetricAggregator : IEquatable<MetricAggregator>
```
olegsych: Are we implementing `IEquatable` in `MetricAggregator` only for the sake of the `MetricAggregatorManager.GetStatisticsAggregator`? It's a lot more code in the `MetricAggregator` and harder to understand in the `MetricAggregatorManager`. Why not simply use `MetricAggregator.aggregatorId` directly?

Test/CoreSDK.Test/Shared/Extensibility/MetricAggregatorManagerTest.cs
```diff
@@ -0,0 +1,606 @@
...
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.TestFramework;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Assert = Xunit.Assert;
+
+    [TestClass]
+    public class MetricAggregatorManagerTest
+    {
+        [TestMethod]
+        public void MetricAggregatorMayBeCreatedForMetricHavingNoDimensions()
```
olegsych: I felt these integration tests were hard to understand for someone who is not already familiar with all 3 classes implementing metric aggregation. Arrange, Act, Assert comments would help. If it's possible at this point, factoring the classes to allow unit testing would help more.

Test/CoreSDK.Test/Shared/Extensibility/MetricAggregatorTest.cs
```diff
@@ -0,0 +1,226 @@
...
+
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.DataContracts;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+    using Microsoft.ApplicationInsights.TestFramework;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Assert = Xunit.Assert;
+
+    [TestClass]
+    public class MetricAggregatorTest
```
olegsych: Are we missing tests for the `IEquatable` implementation?

src/Core/Managed/Shared/Extensibility/Implementation/SimpleMetricStatisticsAggregator.cs
```diff
@@ -0,0 +1,116 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation
+{
+    using System;
+    using System.Threading;
+
+    /// <summary>
+    /// Represents aggregator for a single time series of a given metric.
+    /// </summary>
+    internal class SimpleMetricStatisticsAggregator
```
olegsych: Just a thought: having two aggregator classes, `MetricAggregator` and `SimpleMetricStatisticAggregator` feels confusing. What if they were called `Metric` and `MetricAggregator` respectively? As a potential benefit, this would allow to simplify the `IMetricProcessor` definition to `Track(Metric, double value)`, which would also be more resilient to future changes in the `Metric` class.

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
vitalyf007: The ctor with parameters exists for the case when people aggregated entirely outside our Sdk and just want to send the results to AI. It states explicitly what is expected. If you use parameter-less ctor and property setters, it is not entirely clear what to set to get consistent results. This is somewhat against the purity of the design, but this is where we hit cold reality. We need to capture most common cases of providing metric data and make sure we do not create a large set of concepts. I thought having two concepts combined into one: a) metric that is just count (sum and count) and I do not care about the rest of the properties and b) metric where all properties are interested; would be Ok.

Therefore, people using parameter-less ctor may choose properties they want to set based on what they are going to look at. People using the other ctor know all things that need to be set if you want to go "comprehensive".

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
vitalyf007: Sanitize will set Count to 1 only in case when AggregatedMetricTelemetry was constructed w/o setting "Count" property (or explicitly setting it to <= 0). We consider this a mistake and I initially was setting it to "0" but that may present some problems in the pipeline...

The statement about different value on the wire and through property getter is not unique to this particular case.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
vitalyf007: I wasn't aware of TaskTimer existence. The answer is "almost". TaskTimer fires at regular intervals while perf counter reporting tasks changes interval after each invocation if needed. We can change the TaskTimer, of course, of create a separate class for this one usage...

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,275 @@
...
+            {
+                foreach (KeyValuePair<MetricAggregator, SimpleMetricStatisticsAggregator> aggregatorWithStats in aggregatorSnapshot)
+                {
+                    AggregatedMetricTelemetry aggergatedMetricTelemetry = CreateAggergatedMetricTelemetry(aggregatorWithStats.Key, aggregatorWithStats.Value);
+
+                    aggergatedMetricTelemetry.Properties.Add(intervalDurationPropertyName, ((long)aggregationIntervalDuation.TotalMilliseconds).ToString(CultureInfo.InvariantCulture));
+
+                    if (aggergatedMetricTelemetry != null)
+                    {
+                        this.Client.Track(aggergatedMetricTelemetry);
```
SergeyKanzhelev: I just thought about it - I'd suggest to set an sdkVersion here different from just `dotnet: XX.XX.XX`. Something like `agg-net: XX.XXX.XX` so you'll now how many metrics were collected using this aggregation mechanism.

src/Core/Managed/Shared/Extensibility/MetricAggregator.cs
vitalyf007: This is the same pattern as with regular .Net events. May not be necessary. Lets chat about it...

Test/CoreSDK.Test/Shared/Extensibility/MetricAggregatorManagerTest.cs
vitalyf007: I do not entirely understand. We'll need to chat offline.

Test/CoreSDK.Test/Shared/Extensibility/MetricAggregatorTest.cs
vitalyf007: There is a set of tests in MetricAggregatorManager that tests these indirectly, however, I can (and likely should) add MA equality tests.

src/Core/Managed/Shared/Extensibility/Implementation/SimpleMetricStatisticsAggregator.cs
vitalyf007: I need to think about Metric/MetricAggregator naming...

You're almost right. There are two aggregators, so to speak. MetricAggregator is really a "handle" or a proxy to te actual aggregator. You can hold on to it aslong as you like and operating on it will be proxied to actual aggregator which is SimpleMetricStatisticsAggregator class.
Note that metric data may be aggregated in a number of ways. Extracting/computing simple stats is just a way. You can also do t-digest and some other fun stuff too. Lets chat more about it.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
olegsych: I think `TaskTimer` fires once, once the `TimeSpan` specified by its `Delay` property elapses. You can adjust the `Delay` property from the `elapsed` function and start again with the new delay. With the benefit of perfect hindsight, this is not the most elegant design, but I think it should get the job done. I do think that timer logic is tricky enough to invest in a single robust implementation that's easy to use for aggregation as well.

src/Core/Managed/Shared/Extensibility/MetricAggregator.cs
olegsych: Good point. I'll settle for unit tests then :)

Test/CoreSDK.Test/Shared/Extensibility/MetricAggregatorManagerTest.cs
olegsych: Good idea. For readability, I'd be happy if the tests looked like this. 
```C#
public void TestName()
{
    // Arrange
    ... 

    // Act
    ...

    // Assert
    ...
}
```

Test/CoreSDK.Test/Shared/Extensibility/MetricAggregatorTest.cs
olegsych: 👍

src/Core/Managed/Shared/Extensibility/Implementation/SimpleMetricStatisticsAggregator.cs
olegsych: Ok, so how about calling them `Metric` and `SimpleMetricStatisticAggregator` then?

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
olegsych: So we need a constructor that communicates intent but is unreadable when used. I guess I could argue this point both ways 😏. Let's keep it as is.

src/Core/Managed/Shared/DataContracts/AggregatedMetricTelemetry.cs
vitalyf007: Fixed. Also added unit test.

src/Core/Managed/Shared/Extensibility/MetricAggregator.cs
vitalyf007: I made a fix for this. Technically I'm shifting responsibilities by encapsulating a bit more. Do you see this as acceptable fix (or at least a step in the right direction) or more of a hack?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
vitalyf007: I looked at TaskTimer. I understand what it is doing and it would [almost] work for me if I were to have functionality (in it) to do something if wait is terminated (I need to try to do Flush() to attempt not to lose any telemetry I can salvage). I considered the problem and I think it is too large to be coupled with the change I'm doing.

I suggest separating the area that covers two questions: A) How many timer implementations does one library need? and B) How many utility threads does one library need to run? and tackling it separately. We also have a perfect guy to propose the best solution - Greg. I chatted with him and once he is done with his current thing (should be just a few days) he'll look into this.

Test/CoreSDK.Test/Shared/Extensibility/MetricAggregatorManagerTest.cs
olegsych: Thanks, @vitalyf007! I feel these tests are much easier to read with the arrange/act/assert comments.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        /// <summary>
+        /// Disposes the object.
+        /// </summary>
+        public void Dispose()
+        {
+            this.disposed = true;
+
+            if (this.snapshotTask != null)
```
olegsych: Since `snapshotTask` is initialized in the constructor, can it ever be `null`?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        /// <summary>
+        /// Telemetry client used to track resulting aggregated metrics.
+        /// </summary>
+        private readonly TelemetryClient telemetryClient;
+
+        /// <summary>
+        /// Metric aggregator snapshotting task.
+        /// </summary>
+        private readonly TaskTimer snapshotTask;
```
olegsych: Rename `snapshotTask` to `snapshotTimer`?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,295 @@
...
+            double minutesFromZero = currentTime.Subtract(DateTimeOffset.MinValue).TotalMinutes;
+
+            // we want to wake up exactly at 1 second past minute
+            // to make perceived system latency look smaller
+            var nextWakeTime = DateTimeOffset.MinValue
+                .AddMinutes((long)minutesFromZero)
+                .Add(snapshotFrequency)
+                .AddSeconds(1);
+
+            TimeSpan sleepTime = nextWakeTime - DateTimeOffset.UtcNow;
```
olegsych: Is the use of `DateTimeOffset.UtcNow` instead of the `currentTime` intentional here?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,295 @@
...
+            return this.aggregatorDictionary.GetOrAdd(aggregator, (aid) => { return new SimpleMetricStatisticsAggregator(); });
+        }
+
+        /// <summary>
+        /// Calculates wait time until next snapshot of the aggregators.
+        /// </summary>
+        /// <returns>Wait time.</returns>
+        private static TimeSpan GetWaitTime()
+        {
+            DateTimeOffset currentTime = DateTimeOffset.UtcNow;
```
olegsych: How are we testing the `GetWaitTime` logic? We have the `IClock` interface that can be injected instead of taking a direct dependency on the static `DateTimeOffset.UtcNow`.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,295 @@
...
+                    }
+                });
+        }
+
+        /// <summary>
+        /// Takes snapshot of all active metric aggregators and turns results into metric telemetry.
+        /// </summary>
+        private void Snapshot()
+        {
+            ConcurrentDictionary<MetricAggregator, SimpleMetricStatisticsAggregator> aggregatorSnapshot =
```
olegsych: Convert explicit type to `var` since the type name already appears in the same statement?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        /// <summary>
+        /// Generates telemetry object based on the metric aggregator.
+        /// </summary>
+        /// <param name="aggregator">Metric aggregator.</param>
+        /// <param name="aggregatorStats">Metric aggregator statistics calculated for a period of time.</param>
+        /// <returns>Metric telemetry object resulting from aggregation.</returns>
+        private static AggregatedMetricTelemetry CreateAggergatedMetricTelemetry(MetricAggregator aggregator, SimpleMetricStatisticsAggregator aggregatorStats)
+        {
+            if ((aggregator == null) || (aggregatorStats == null) || (aggregatorStats.Count <= 0))
```
olegsych: Can the `aggregator` or `aggregatorStats` be ever `null` here since they come from `this.aggregatorDictionary`?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,295 @@
...
+                aggregationIntervalDuation = snapshotFrequency;
+            }
+
+            if (aggregatorSnapshot.Count > 0)
+            {
+                foreach (KeyValuePair<MetricAggregator, SimpleMetricStatisticsAggregator> aggregatorWithStats in aggregatorSnapshot)
+                {
+                    AggregatedMetricTelemetry aggergatedMetricTelemetry = CreateAggergatedMetricTelemetry(aggregatorWithStats.Key, aggregatorWithStats.Value);
+
+                    aggergatedMetricTelemetry.Properties.Add(intervalDurationPropertyName, ((long)aggregationIntervalDuation.TotalMilliseconds).ToString(CultureInfo.InvariantCulture));
```
olegsych: The `aggregatedMetricTelemetry` can be `null` here. We should add the `intervalDurationPropertyName` inside the `if` statement below.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        internal SimpleMetricStatisticsAggregator GetStatisticsAggregator(MetricAggregator aggregator)
+        {
+            if (this.disposed)
+            {
+                throw new ObjectDisposedException(nameof(MetricAggregatorManager));
+            }
+
+            if (aggregator == null)
```
olegsych: The `GetStatisticsAggregator` is only called by `MetricAggregator.Track`. Can `aggregator` be ever `null`?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
vitalyf007: Fixed.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
vitalyf007: It does not matter much but yes, it is intentional. I want current time as I hit this line, not few lines earlier.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
vitalyf007: I see your point. I'll complete this PR w/o fixing this but will come over seeking advice on testing timer operations. In Rx world we have virtual time scheduler which we use in these situations...

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
vitalyf007: It is a little iffy, I think... The type is specified indirectly. Do you feel strongly about this?

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
vitalyf007: They cannot be, correct. If favor of failing fast, I'll remove the check.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
vitalyf007: No. Will remove the check.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
```diff
@@ -0,0 +1,295 @@
...
+            return Task.Factory.StartNew(
+                () =>
+                {
+                    try
+                    {
+                        this.Snapshot();
+                    }
+                    catch (Exception ex)
+                    {
+                        CoreEventSource.Log.FailedToSnapshotMetricAggregators(ex.ToString());
```
vitalyf007: I'd rather have a specific message logged here in addition to more generic "timed task failed" message logged by timed task.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
olegsych: @vitalyf007 While I don't have a strong preference, I do prefer making fields readonly over optimizing for double disposal.

src/Core/Managed/Shared/Extensibility/MetricAggregatorManager.cs
olegsych: @vitalyf007 Not strongly, no 😄. I do like to use var when the type name appears in the same statement. I think it makes the code easier to read and helps to emphasize nuances by spelling out the variable type explicitly, when needed. But that's mostly my OCD talking.

src/Core/Managed/Shared/Extensibility/Implementation/SimpleMetricStatisticsAggregator.cs
```diff
@@ -0,0 +1,116 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation
+{
+    using System;
+    using System.Threading;
+
+    /// <summary>
+    /// Represents mechanism to calculate basic statistical parameters of a series of numeric values.
+    /// </summary>
+    internal class SimpleMetricStatisticsAggregator
+    {
```
macrogreg: I have note yet read the whole CR so I may learn this shortly (sorry if that's the case), but what do you use these getters for and how important is it that they give consistent values? You may want to put them under lock too.

src/Core/Managed/Shared/Extensibility/Metric.cs
```diff
@@ -0,0 +1,171 @@
...
+        /// <param name="other">The object to compare with the current object. </param>
+        /// <returns>True if the specified object is equal to the current object; otherwise, false.</returns>
+        public bool Equals(Metric other)
+        {
+            if (other == null)
+            {
+                return false;
+            }
+
+            return string.Compare(this.aggregatorId, other.aggregatorId, StringComparison.Ordinal) == 0;
```
macrogreg: Minor: Implement Equals using Equals (of the aggregatorId)

src/Core/Managed/Shared/Extensibility/Metric.cs
```diff
@@ -0,0 +1,171 @@
...
+        /// <summary>
+        /// Generates id of the aggregator serving time series specified in the parameters.
+        /// </summary>
+        /// <param name="name">Metric name.</param>
+        /// <param name="dimensions">Optional metric dimensions.</param>
+        /// <returns>Aggregator id that can be used to get aggregator.</returns>
+        private static string GetAggregatorId(string name, IDictionary<string, string> dimensions = null)
+        {
+            StringBuilder aggregatorIdBuilder = new StringBuilder(name ?? string.Empty);
+
```
macrogreg: Consider a different string constant than "" here. This might help debugging. Can name actually be empty? If yes, we will confuse it with a null name. Maybe "MetricNameIsNull" or somthing?

src/Core/Managed/Shared/Extensibility/Metric.cs
```diff
@@ -0,0 +1,171 @@
...
+        /// <param name="value">Value tracked on time series.</param>
+        private void ForwardToProcessors(double value)
+        {
+            // create a local reference to metric processor collection
+            // if collection changes after that - it will be copied not affecting local reference
+            IList<IMetricProcessor> metricProcessors = this.manager.MetricProcessors;
+
+            if (metricProcessors != null)
+            {
+                foreach (IMetricProcessor processor in metricProcessors)
```
macrogreg: This is hot path, right?
If so, maybe a For Loop is more performant?

src/Core/Managed/Shared/Extensibility/MetricManager.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// Takes snapshot of all active metric aggregators and turns results into metric telemetry.
+        /// </summary>
+        private void Snapshot()
+        {
+            ConcurrentDictionary<Metric, SimpleMetricStatisticsAggregator> aggregatorSnapshot =
+                Interlocked.Exchange(ref this.metricDictionary, new ConcurrentDictionary<Metric, SimpleMetricStatisticsAggregator>());
+
+            // calculate aggregation interval duration interval
+            TimeSpan aggregationIntervalDuation = DateTimeOffset.UtcNow - this.lastSnapshotStartDateTime;
+            this.lastSnapshotStartDateTime = DateTimeOffset.UtcNow;
```
macrogreg: call DateTimeOffset.UtcNow once and store in a local. Otherwise you have the situation where lastSnapshotTime does not directly correspond to the end of the previous period.

src/Core/Managed/Shared/Extensibility/MetricManager.cs
```diff
@@ -0,0 +1,285 @@
...
+
+            // prevent zero duration for interval
+            if (aggregationIntervalDuation.TotalMilliseconds < 1)
+            {
+                aggregationIntervalDuation = TimeSpan.FromMilliseconds(1);
+            }
+
+            // adjust interval duration to exactly snapshot frequency if it is close (within 1%)
+            double difference = Math.Abs(aggregationIntervalDuation.TotalMilliseconds - snapshotFrequency.TotalMilliseconds);
+
```
macrogreg: How did you pick 1%?

src/Core/Managed/Shared/Extensibility/MetricManager.cs
```diff
@@ -0,0 +1,285 @@
...
+                aggregationIntervalDuation = snapshotFrequency;
+            }
+
+            if (aggregatorSnapshot.Count > 0)
+            {
+                foreach (KeyValuePair<Metric, SimpleMetricStatisticsAggregator> aggregatorWithStats in aggregatorSnapshot)
+                {
+                    AggregatedMetricTelemetry aggergatedMetricTelemetry = CreateAggergatedMetricTelemetry(aggregatorWithStats.Key, aggregatorWithStats.Value);
+
+                    if (aggergatedMetricTelemetry.Count > 0)
```
macrogreg: Use (aggregatorWithStats.Value.Count > 0) and mode the check above the telemetry ctor.

src/Core/Managed/Shared/Extensibility/MetricManager.cs
```diff
@@ -0,0 +1,285 @@
...
+
+#if CORE_PCL || NET45 || NET46
+    using TaskEx = System.Threading.Tasks.Task;
+#endif
+
+    /// <summary>
+    /// Metric factory and controller.
+    /// </summary>
+    public sealed class MetricManager : IDisposable
+    {
```
macrogreg: So I am still reading this, but I really do not like that this guy is disposable.
What are the circumstances where you create more than one MetricManager?
It seems to only be disposable because of the timer. I suggest to look for ways around that. Maybe I can suggest something after I finish reading everything...

macrogreg: I just looked at TaskTimer. It looks like just Task.Delay plus some really weird cancellation logic that looks really suspicious to me. I would not use it. Can we chat to see wht was intended and why directly using Task.Delay is not an option (or is it?)

src/Core/Managed/Shared/Extensibility/MetricManager.cs
```diff
@@ -0,0 +1,285 @@
...
+        /// <summary>
+        /// Name of the property added to aggregation results to indicate duration of the aggregation interval.
+        /// </summary>
+        private static string intervalDurationPropertyName = "IntervalDurationMs";
+
+        /// <summary>
+        /// Reporting frequency.
+        /// </summary>
+        private static TimeSpan snapshotFrequency = TimeSpan.FromMinutes(1);
```
macrogreg: aggregationPeriod would be a better name. frequency is a *rate*, but you are using a time span.

src/Core/Managed/Shared/Extensibility/MetricManager.cs
```diff
@@ -0,0 +1,285 @@
...
+            // to make perceived system latency look smaller
+            var nextWakeTime = DateTimeOffset.MinValue
+                .AddMinutes((long)minutesFromZero)
+                .Add(snapshotFrequency)
+                .AddSeconds(1);
+
+            TimeSpan sleepTime = nextWakeTime - DateTimeOffset.UtcNow;
+
+            // adjust wait time to a bit longer than a minute if the wake up time is within few seconds from now
+            return sleepTime < TimeSpan.FromSeconds(3) ? sleepTime.Add(snapshotFrequency) : sleepTime;
```
macrogreg: You are not adding a minute, but the period. NOt sure what's intended, but please make the comment and code match.

src/Core/Managed/Shared/Extensibility/MetricManager.cs
```diff
@@ -0,0 +1,285 @@
...
+            }
+
+            return this.metricDictionary.GetOrAdd(metric, (m) => { return new SimpleMetricStatisticsAggregator(); });
+        }
+
+        /// <summary>
+        /// Calculates wait time until next snapshot of the aggregators.
+        /// </summary>
+        /// <returns>Wait time.</returns>
+        private static TimeSpan GetWaitTime()
```
macrogreg: This logic is really hard to read, and I think it is also not correct. if your aggregation period is >= 1 minute, your 3 seconds rule will never apply. And is it is small, then you may still end up in the past.
Consider this as an alternative:

`private static DateTimeOffset GetNextSnapshitTime(DateTimeOffset prevTime, TimeSpan aggregationPeriod)
        { 
            DateTimeOffset exactWaitUntilTime = prevTime + aggregationPeriod;

            // Below we have logic that rounds the next snapshot event to a sertain number of seconds past a minute.
            // This would break down if the aggregation period was smaller than a minute. 
            // So, for small aggegation periods we are ot rounding.
            if (aggregationPeriod < TimeSpan.FromMinutes(1))
            {
                return exactWaitUntilTime;
            }

            // We want to snapshot at 1 sec past full minute.
            // Why??
            const int pastMinSecs = 1;
            DateTimeOffset roundedWaitUntilTime = new DateTimeOffset(exactWaitUntilTime.Year, exactWaitUntilTime.Month, exactWaitUntilTime.Day,
                                                                        exactWaitUntilTime.Hour, exactWaitUntilTime.Minute, pastMinSecs,
                                                                        exactWaitUntilTime.Offset);
            // We round to the closest minute boundry (Attention, adjust this condition if NOT (0 <= pastMinSecs < 30) !! )
            if (exactWaitUntilTime.Second > (30 + pastMinSecs))
            {
                roundedWaitUntilTime = roundedWaitUntilTime.AddMinutes(1);
            }

            return roundedWaitUntilTime;
        }`

macrogreg: Sorry, it seems not to do code markup correctly, i hope you can read it.

src/Core/Managed/Shared/Extensibility/Implementation/SimpleMetricStatisticsAggregator.cs
macrogreg: So having read the other bits, the only way this can be a problem is when Track was called right before a Snapshot is made and then lost the race. We would expose the spinlock internally and lock the aggregator while making the snapshot. Because the race is very rare, the lock will be contended very rarely and so should not result in any significant perf decrease.

src/Core/Managed/Shared/Extensibility/MetricManager.cs
```diff
@@ -0,0 +1,285 @@
...
+            if (this.snapshotTimer != null)
+            {
+                this.snapshotTimer.Dispose();
+                this.snapshotTimer = null;
+            }
+
+            this.Flush(true);
+        }
+
+        internal SimpleMetricStatisticsAggregator GetStatisticsAggregator(Metric metric)
```
macrogreg: Under what circumstances is the can the manager be disposed. GetStatisticsAggregator is called from the public Track in metric. Should Track ever throw? And if it may, then is this relationship obvious to the user?

src/Core/Managed/Shared/Extensibility/MetricManager.cs
macrogreg: Looking further, we throw in disposed state for methods, but not properties. Is this right. Most methods, actually, would work fine in the disposed state, the only thing we cannot do is schedule future updates. Can we allow those calls after Dispose()? Can we get rid of IDisposable? I think we should chat. :)

src/Core/Managed/Shared/Extensibility/MetricManager.cs
```diff
@@ -0,0 +1,285 @@
...
+            {
+                CoreEventSource.Log.FailedToFlushMetricAggregators(ex.ToString());
+            }
+        }
+
+        /// <summary>
+        /// Takes a snapshot of aggregators collected by this instance of the manager
+        /// and schedules the next snapshot.
+        /// </summary>
+        private Task SnapshotAndReschedule()
```
macrogreg: Having examined the DelayTask (I am not sure whether I agree with Oleg about using it, but he makes some good points), I think that running is underStartNew is no required. DelayTask calls Task.Delay and then runs the delegate synchronously in respect to the completion of the Delay. Since we are not scheduling the timer again until we finsih the snapshot, there is no reason for us to just threads here.

I would just do

```
return () =>
    {
        // all the shapshot stuff
        return Task.FromResult(1);
    };
```

src/Core/Managed/Shared/Extensibility/MetricManager.cs
macrogreg: So it the period is 1 minute, 1% is 0.6 secs. For 5 mins aggregations, it is 3 secs. I am not sure that this is what you want. Maybe 1% for under period < 10 secs; 0.5 sec for 10 sec < period <= 30 sec; 1 sec for 30 sec < period  <= 3 min; and 3 secs otherwise. This is also random, bu i think it comes closer to what you want. Whaever you choose, please explain it in a comment.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
macrogreg: I agree with Sergey. Suggestion:
type name: IMetricValueTracker
property name: IList<IMetricValueTracker> RawMetricValueTrackers

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -217,6 +221,20 @@ public void TrackMetric(MetricTelemetry telemetry)
...
+        {
+            if (telemetry == null)
+            {
+                telemetry = new AggregatedMetricTelemetry();
+            }
+
+            this.Track(telemetry);
+        }
+
+        /// <summary>
```
macrogreg: I would love to see a flat API here:
TrackAggregatedMetric(string metricName, double value);
TrackAggregatedMetric(string metricName, double value, params string[] dimensions);

src/Core/Managed/Shared/Extensibility/Implementation/SimpleMetricStatisticsAggregator.cs
vitalyf007: Yes, given the read pattern (once right after snapshot) I do not think this is a problem. I do not want to complicate getters with honoring the lock especially considering the fact that they'll be called sequentially.

src/Core/Managed/Shared/Extensibility/Metric.cs
vitalyf007: Done

src/Core/Managed/Shared/Extensibility/Metric.cs
vitalyf007: Will use "n/a" as that is the way Sdk represents other names (such as metric name, event name) during serialization if the name is null.

src/Core/Managed/Shared/Extensibility/Metric.cs
```diff
@@ -0,0 +1,171 @@
...
+        {
+            StringBuilder aggregatorIdBuilder = new StringBuilder(name ?? string.Empty);
+
+            if (dimensions != null)
+            {
+                var sortedDimensions = dimensions.OrderBy((pair) => { return pair.Key; });
+
+                foreach (KeyValuePair<string, string> pair in sortedDimensions)
+                {
+                    aggregatorIdBuilder.AppendFormat(CultureInfo.InvariantCulture, "\n{0}\t{1}", pair.Key ?? string.Empty, pair.Value ?? string.Empty);
```
vitalyf007: They can appear in the name but I do not expect to ever see two real cases of that. Metric names, dimension names and values appear in the UX and generally speaking expected to be symbols and spaces only...

src/Core/Managed/Shared/Extensibility/MetricManager.cs
vitalyf007: Done.

src/Core/Managed/Shared/Extensibility/MetricManager.cs
vitalyf007: Very good point. I do not think Track() should throw... In this particular case it is somewhat bad because one can create manager, create metric, dispose Manager, continue using Metric.Track() expecting things to happen...
On the other had throwing exceptions from monitoring code is considered bad...
Based on further consideration I think I will remove places where I throw "object disposed" exceptions. This will break scenario I spoke of just above, but I think not throwing is better...

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -26,6 +26,7 @@ public sealed class TelemetryConfiguration : IDisposable
         private string instrumentationKey = string.Empty;
         private bool disableTelemetry = false;
         private TelemetryProcessorChainBuilder builder;
+        private SnapshottingList<IMetricProcessor> metricProcessors = new SnapshottingList<IMetricProcessor>();
```
vitalyf007: I'm reusing what's already available. I do not like it very much either but do not intend to fix that particular problem in this effort.

src/Core/Managed/Shared/TelemetryClient.cs
vitalyf007: That was discussed multiple times and we decided not to have these apis in order to treat TelemetryClient as context.

Issue #295, opened 2016-08-22
Do not throw when channel not set


SergeyKanzhelev:
When channel is not set any `TrackFoo` operation throws unhandled exception.

Exception thrown here:
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/1ccd5b466451db1884b242b6a58b3e9401a1fbcf/src/Core/Managed/Shared/Extensibility/Implementation/TransmissionProcessor.cs#L43

Exception is unhandled:
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/e0c93977e9aa05616ecc22c5b11ef39f8cb4587f/src/Core/Managed/Shared/TelemetryClient.cs#L310

Need to:
1. Set channel to default.
2. Make sure `TrackFoo` doesn't throw non-system (like OutOfMemory) exceptions.

abaranch:
Why are you sayin it is not initialized by default? https://github.com/Microsoft/ApplicationInsights-dotnet/blob/0110215df15bca47aa7fc7ee92c46f30d909dead/src/Core/Managed/Net40/Extensibility/Implementation/TelemetryConfigurationFactory.cs#L76

SergeyKanzhelev:
How then you can end up with the exception like customer reported?

abaranch:
They may have created TelemetryConfiguration using constructor instead of CreateDefault. You do not want to put initialization logic in there. Do you have more info on that?

abaranch:
@SergeyKanzhelev Do you agree that we should close the issue or do you propose to initialize channel if TelemetryConfiguration is created using constructor (and not CreateDefault)?

SergeyKanzhelev:
Alternatively we can make it on "Track"

PR #324, opened 2016-10-06
Initialize channel in Track

abaranch:
#295

### DIFF
--- a/src/Core/Managed/Shared/TelemetryClient.cs
+++ b/src/Core/Managed/Shared/TelemetryClient.cs
@@ -351,6 +351,12 @@
                     return;
                 }
 
+                // If someone created configuration from scratch and forgot to initialize channel - use default
+                if (this.configuration.TelemetryChannel == null)
+                {
+                    this.configuration.TelemetryChannel = new InMemoryChannel();
+                }
+
                 // invokes the Process in the first processor in the chain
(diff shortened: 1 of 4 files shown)


src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -351,6 +351,12 @@ public void Track(ITelemetry telemetry)
                     return;
                 }
 
+                // If someone created configuration from scratch and forgot to initialize channel - use default
+                if (this.configuration.TelemetryChannel == null)
+                {
+                    this.configuration.TelemetryChannel = new InMemoryChannel();
+                }
+
```
dnduffy: This works but now they have to pay the extra dereference and comparison check on every call.  It's not much of a cost but it is not free either.  They also won't know they didn't configure it officially since it will silently work with a different channel than maybe they were expecting.  If this is our official position then I am okay with it.

abaranch: We had same initialization logic in CreateDefault and Active: if channel is not provied InMemory channel is used. Our policy is not to crash customer's application and provide defaults. That was not done in case customer decided to create TelemetryConfiguration from scratch so it fixes this behavior. Discussion was in the issue mentioned (#295)

PR #326, opened 2016-10-14
Added additional logging and safety checks plus cleanup.

dnduffy:


src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
```diff
@@ -169,9 +178,18 @@ private static string GetUniqueFileName(string extension)
...
+            Transmission transmission = null;
+            if (file.Exists)
+            {
+                ChangeFileExtension(file, TemporaryFileExtension);
+                transmission = LoadFromTemporaryFile(file, out fileSize);
+                file.Delete();
+            }
+            else
+            {
+                fileSize = 0;
```
SergeyKanzhelev: My understanding that before it would throw if file cannot be read. Will consumer of this function expect `null` as a result?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
```diff
@@ -139,24 +140,32 @@ public virtual Transmission Dequeue()
...
 
-                    throw; // because the process does not have permission to modify this file
+                    throw; // Because the process does not have permission to modify this file.
                 }
-                catch (IOException)
+                catch (IOException ioe)
                 {
-                    continue; // because another thread already loaded this file
+                    TelemetryChannelEventSource.Log.TransmissionFailedToStoreWarning(file.Name, ioe.ToString());
+                    Thread.Sleep(random.Next(1, 100)); // Sleep for random time of 1 to 100 milliseconds to try to avoid future timing conflicts.
```
SergeyKanzhelev: I wonder - if we just return an empty transmission here instead of waiting - will the caller of this method wait before the scheduling of the next read? I know if was already working like this, but ultimately in the scenario we saw with the only one completely broken file - this task will never complete. Maybe we can release task and caller will wait before scheduling it again.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryChannelEventSource.cs
```diff
@@ -417,6 +417,24 @@ public void LastBackendResponseWhenPutToStorage(int statusCode, double currentDe
             this.WriteEvent(58, statusCode, currentDelayInSeconds, this.ApplicationName);
         }
 
+        [Event(59, Message = "Error dequeuing file: {0}. Exception: {1}.", Level = EventLevel.Error)]
+        public void TransmissionStorageDequeueIOError(string fileName, string exception, string appDomainName = "Incorrect")
```
SergeyKanzhelev: I cannot find the use of this message using search in Edge.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
```diff
@@ -139,24 +140,32 @@ public virtual Transmission Dequeue()
...
                     }
 
-                    throw; // because the process does not have permission to modify this file
+                    throw; // Because the process does not have permission to modify this file.
                 }
-                catch (IOException)
+                catch (IOException ioe)
                 {
-                    continue; // because another thread already loaded this file
+                    TelemetryChannelEventSource.Log.TransmissionFailedToStoreWarning(file.Name, ioe.ToString());
```
SergeyKanzhelev: should we do the same `file?.Name ?? string.Empty` as above?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryChannelEventSource.cs
```diff
@@ -417,6 +417,24 @@ public void LastBackendResponseWhenPutToStorage(int statusCode, double currentDe
...
+        {
+            this.WriteEvent(59, fileName, exception, this.ApplicationName);
+        }
+
+        [Event(60, Message = "Unauthorized access dequeuing file: {0}. Exception: {1}.", Level = EventLevel.Warning)]
```
SergeyKanzhelev: Those two messages are warnings. Do you think either of them may be user-actionable? Maybe user can clean up the storages?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
dnduffy: Not necessary because it is below the if (file == null) check.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
dnduffy: Yes, I do not believe that the behavior has changed.  The tests appear to cover this scenario and still pass.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryChannelEventSource.cs
dnduffy: Oops, fixed, I called the wrong message in order to use F12 and forgot to replace it.

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TransmissionStorageTest.cs
```diff
@@ -654,7 +654,7 @@ public void SkipsTransmissionFileBeingCurrentlyLoadedByAnotherThreadAndTriesLoad
                 Assert.NotNull(dequeued);
             }
 
-            [TestMethod, Timeout(50)]
+            [TestMethod, Timeout(150)]
```
dnduffy: I upped the timeout to compensate for the up to 100 ms sleep call in the code under test.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
dnduffy: P.S. The caller does not wait, it runs in a tight loop until there are no more files or no room in the buffer to load them.  The existing logic in the system does not allow it to really stop trying until there are no files physically in the directory.  If we can learn with the additional logic what is happening at the file system level when it fails and gets stuck in a loop then it will give us enough information to try to fix it permanently.

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TransmissionStorageTest.cs
```diff
@@ -529,7 +529,7 @@ public void LoadsTransmissionFromTheOldestTransmissionFileInTheFolder()
 
                 var oldestAddress = new Uri("http://oldest");
                 StubPlatformFile oldestFile = CreateTransmissionFile(oldestAddress);
-                oldestFile.OnGetDateCreated = () => DateTimeOffset.MinValue;
+                oldestFile.OnGetDateCreated = () => DateTimeOffset.Now.AddDays(-1);
```
SergeyKanzhelev: Why this change? Can you please leave a comment in code for the magic number of days

SergeyKanzhelev: Oh, it is just a test. Nevermind

dnduffy: Sure, it is because now there is logic to check if files are older than two days, which means Breeze will reject them, and so test files can't be created with the min value for the creation date or they will be invalid and get deleted right away.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
```diff
@@ -139,24 +140,34 @@ public virtual Transmission Dequeue()
                         return transmission;
                     }
                 }
-                catch (UnauthorizedAccessException)
+                catch (UnauthorizedAccessException uae)
                 {
                     if (file == null)
                     {
-                        return null; // because the process does not have permission to access the folder
+                        TelemetryChannelEventSource.Log.TransmissionStorageDequeueUnauthorizedAccessException(this.folder.Name, uae.ToString());
```
SergeyKanzhelev: why do you output the folder name, not the file name. Didn't we get exception for the specific file access?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
```diff
@@ -139,24 +140,34 @@ public virtual Transmission Dequeue()
...
+                    else
                     {
-                        lastInaccessibleFileName = file.Name;
-                        continue; // because another thread is loading this file right now
+                        // The same file has been inaccessible more than once because the process does not have permission to modify this file.
+                        TelemetryChannelEventSource.Log.TransmissionStorageUnexpectedRetryOfBadFile(name);
                     }
 
-                    throw; // because the process does not have permission to modify this file
+                    Thread.Sleep(random.Next(1, 100)); // Sleep for random time of 1 to 100 milliseconds to try to avoid future timing conflicts.
```
SergeyKanzhelev: you are modifying behavior here. Instead of breaking the loop you are going to the next iteration. Is it what you want?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
```diff
@@ -139,24 +140,34 @@ public virtual Transmission Dequeue()
...
                 }
-                catch (IOException)
+                catch (IOException ioe)
                 {
-                    continue; // because another thread already loaded this file
+                    // This exception can happen when one thread runs out of files to process and reloads the list while another
+                    // thread is still processing a file and has not deleted it yet thus allowing it to get in the list again.
+                    TelemetryChannelEventSource.Log.TransmissionStorageDequeueIOError(file.Name, ioe.ToString());
```
SergeyKanzhelev: wouldn't this happen very often when two processes running simultaneously? I worry on how much noise we can introduce by tracing this as error.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
```diff
@@ -169,9 +180,23 @@ private static string GetUniqueFileName(string extension)
...
         private static Transmission LoadFromTransmissionFile(IPlatformFile file, out long fileSize)
         {
-            ChangeFileExtension(file, TemporaryFileExtension);
-            Transmission transmission = LoadFromTemporaryFile(file, out fileSize);
-            file.Delete();
+            fileSize = 0;
+            Transmission transmission = null;
+            if (file.Exists)
+            {
+                if (file.DateCreated > DateTimeOffset.Now.AddDays(-2))
```
SergeyKanzhelev: please leave a comment for the magic number here

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
```diff
@@ -217,7 +242,7 @@ private IPlatformFile CreateTemporaryFile()
         private IEnumerable<IPlatformFile> GetTransmissionFiles()
         {
             IEnumerable<IPlatformFile> newFiles = this.folder.GetFiles();
-            newFiles = newFiles.Where(f => f.Extension == TransmissionFileExtension);
+            newFiles = newFiles.Where(f => f.Extension == TransmissionFileExtension && !this.badFiles.ContainsKey(f.Name));
```
SergeyKanzhelev: Some sort of resetting the `badFiles` is needed. Maybe just clear the collection every 29 hours?

SergeyKanzhelev: Also `EnsureSizeIsCalculated` will not include bad files now? So the folder can grow infinitely... Maybe those files can be included in size calculation, but not for the transmission.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
dnduffy: In this case there is no file, it is null.
I'm not convinced that this can actually happen but the original logic implied that it can so in keeping with the original logic I'm logging what does exist in accordance with the original comment.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
dnduffy: Yes, the original behavior was throwing an unhandled exception which didn't fix anything (and might have contributed to the failure to send in future we've seen in production).

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
SergeyKanzhelev: got it. thanks for the explanaition

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
SergeyKanzhelev: ok. My biggest concern is that we can take this task forever and never leave the loop, can we?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
dnduffy: Fair points, I did wonder about that...

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
SergeyKanzhelev: ok. if it doesn't happen too often - it's good with me,

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
dnduffy: Done

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
dnduffy: I've changed it to post filter only where the files are actually loaded into the queue so now the size calculation will include the known bad files so the folder can't grow indefinitely.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
dnduffy: I don't believe so because eventually we will run out of files which can be processed and then exit as normal.  The only change in the logic here is to stop throwing what appears to be an unhandled exception that didn't address the underlying issue.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionStorage.cs
dnduffy: I've put a little sleep with a comment before the file list is loaded which should mitigate this exception ever happening (at least for the scenario I was able to repro in testing).

PR #330, opened 2016-10-18
Enable local throttling of telemetry items sent to the backend

OsvaldoRosado:
These changes extend the TransmissionSender to allow it to enforce a throttle on number of telemetry items sent to the backend. Various other files are changed to allow efficient splitting of Transmission objects where possible to help support these changes.

### DIFF
--- a/Test/ServerTelemetryChannel.Test/Shared.Tests/Helpers/StubTelemetrySerializer.cs
+++ b/Test/ServerTelemetryChannel.Test/Shared.Tests/Helpers/StubTelemetrySerializer.cs
@@ -9,7 +9,7 @@
     {
         public Action<IEnumerable<ITelemetry>> OnSerialize = _ => { };
 
-        public override void Serialize(IEnumerable<ITelemetry> items)
+        public override void Serialize(ICollection<ITelemetry> items)
         {
             this.OnSerialize(items);
         }
--- a/Test/ServerTelemetryChannel.Test/TestFramework/Shared/StubTransmission.cs
+++ b/Test/ServerTelemetryChannel.Test/TestFramework/Shared/StubTransmission.cs
@@ -44,7 +44,7 @@
 
         public override Tuple<Transmission, Transmission> Split(Func<int, int> calculateLength)
         {
-            var ret = base.Split(calculateLength);
+            Tuple<Transmission,Transmission> ret = base.Split(calculateLength);
 
             if (ret.Item2 == null)
             {
@@ -87,7 +87,7 @@
             }
             else
             {
-                var compress = this.ContentEncoding == JsonSerializer.CompressionType;
+                bool compress = this.ContentEncoding == JsonSerializer.CompressionType;
                 string[] payloadItems = JsonSerializer
                     .Deserialize(this.Content, compress)
                     .Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
--- a/Test/ServerTelemetryChannel.Test/TestFramework/Shared/StubTransmission.cs
+++ b/Test/ServerTelemetryChannel.Test/TestFramework/Shared/StubTransmission.cs
@@ -1,6 +1,7 @@
 ﻿namespace Microsoft.ApplicationInsights.Web.TestFramework
 {
     using System;
+    using System.Collections.Generic;
     using System.IO;
     using System.Threading.Tasks;
     using Microsoft.ApplicationInsights.Channel;
@@ -26,6 +27,11 @@
         {
         }
 
+        public StubTransmission(ICollection<ITelemetry> telemetry)
+            : base(new Uri("any://uri"), telemetry)
+        {
+        }
+
         public Task SaveAsync(Stream stream)
         {
             return TaskEx.Run(() => this.OnSave(stream));
@@ -34,6 +40,54 @@
         public override Task<HttpWebResponseWrapper> SendAsync()
         {
             return TaskEx.Run(this.OnSend);
+        }
+
+        public override Tuple<Transmission, Transmission> Split(Func<int, int> calculateLength)
+        {
+            var ret = base.Split(calculateLength);
+
+            return Tuple.Create((Transmission)this.Convert(ret.Item1), (Transmission)this.Convert(ret.Item2));
+        }
+
+        private StubTransmission Convert(Transmission transmission)
+        {
+            if (transmission != null)
+            {
+                if (transmission.TelemetryItems == null)
+                {
+                    transmission = new StubTransmission(transmission.Content)
+                    {
+                        OnSave = this.OnSave,
+                        OnSend = this.OnSend,
+                    };
+                }
+                else
+                {
+                    transmission = new StubTransmission(transmission.TelemetryItems)
+                    {
+                        OnSave = this.OnSave,
+                        OnSend = this.OnSend,
+                    };
+                }
+            }
+
+            return (StubTransmission)transmission;
+        }
+
+        public int CountOfItems()
+        {
+            if (this.TelemetryItems != null)
+            {
+                return this.TelemetryItems.Count;
+            }
+            else
+            {
+                var compress = this.ContentEncoding == JsonSerializer.CompressionType;
+                string[] payloadItems = JsonSerializer
+                    .Deserialize(this.Content, compress)
+                    .Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+                return payloadItems.Length;
+            }
         }
     }
 }
--- a/Test/ServerTelemetryChannel.Test/TestFramework/Shared/StubTransmission.cs
+++ b/Test/ServerTelemetryChannel.Test/TestFramework/Shared/StubTransmission.cs
@@ -46,6 +46,11 @@
         {
             var ret = base.Split(calculateLength);
 
+            if (ret.Item2 == null)
+            {
+                return Tuple.Create((Transmission)this, (Transmission) null);
+            }
+
             return Tuple.Create((Transmission)this.Convert(ret.Item1), (Transmission)this.Convert(ret.Item2));
         }
 
--- a/Test/ServerTelemetryChannel.Test/TestFramework/Shared/StubTransmission.cs
+++ b/Test/ServerTelemetryChannel.Test/TestFramework/Shared/StubTransmission.cs
@@ -18,12 +18,12 @@
         public Func<HttpWebResponseWrapper> OnSend = () => null;
 
         public StubTransmission()
-            : base(new Uri("any://uri"), new byte[0], string.Empty, string.Empty)
+            : base(new Uri("any://uri"), new byte[0], JsonSerializer.ContentType, string.Empty)
         {
         }
 
         public StubTransmission(byte[] content)
-            : base(new Uri("any://uri"), content, string.Empty, string.Empty)
+            : base(new Uri("any://uri"), content, JsonSerializer.ContentType, string.Empty)
         {
         }
 
--- a/src/Core/Managed/Shared/Channel/Transmission.cs
+++ b/src/Core/Managed/Shared/Channel/Transmission.cs
@@ -67,7 +67,7 @@
         /// Initializes a new instance of the <see cref="Transmission"/> class.
         /// </summary>
         public Transmission(Uri address, ICollection<ITelemetry> telemetryItems, TimeSpan timeout = default(TimeSpan)) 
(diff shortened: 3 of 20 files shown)


src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionSender.cs
```diff
@@ -100,15 +164,105 @@ protected void OnTransmissionSent(TransmissionProcessedEventArgs args)
...
+                "Transmission was split by local throttling policy",
+                null,
+                System.Net.WebExceptionStatus.Success,
+                null
+            );
+            this.OnTransmissionSent(new TransmissionProcessedEventArgs(rejectedTransmission, exception, new HttpWebResponseWrapper()
+            {
+                StatusCode = ResponseStatusCodes.ResponseCodeTooManyRequests,
+                StatusDescription = "Internally Throttled",
+                RetryAfterHeader = null
```
dnduffy: What would happen if it had a retry after header of 1 second?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionSender.cs
```diff
@@ -100,15 +164,105 @@ protected void OnTransmissionSent(TransmissionProcessedEventArgs args)
...
+            if (!this.applyThrottle)
+            {
+                return numEvents;
+            }
+
+            var throttleWindowId = (long)(DateTime.UtcNow.Subtract(new DateTime(1970, 1, 1)).TotalMilliseconds / this.ThrottleWindow);
+            var isInThrottleWindow = throttleWindowId == currentThrottleWindowId;
+            if (isInThrottleWindow && currentItemsCount < this.ThrottleLimit)
+            {
+                var numAccepted = Math.Min(numEvents, this.ThrottleLimit - this.currentItemsCount);
```
dnduffy: Can this.ThrottleLimit - this.currentItemsCount never be negative (since this appears to be inside a check that currentItemsCount is less than the throttle limit)?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionSender.cs
```diff
@@ -100,15 +164,105 @@ protected void OnTransmissionSent(TransmissionProcessedEventArgs args)
...
+        private int IsTransmissionSendable(int numEvents = 1, bool peek = false)
+        {
+            if (!this.applyThrottle)
+            {
+                return numEvents;
+            }
+
+            var throttleWindowId = (long)(DateTime.UtcNow.Subtract(new DateTime(1970, 1, 1)).TotalMilliseconds / this.ThrottleWindow);
+            var isInThrottleWindow = throttleWindowId == currentThrottleWindowId;
+            if (isInThrottleWindow && currentItemsCount < this.ThrottleLimit)
```
dnduffy: nitpick: sometimes you have this. and sometimes not (I believe that StyleCop compliance requires it and one of the builds will complain).

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionSender.cs
```diff
@@ -100,15 +164,105 @@ protected void OnTransmissionSent(TransmissionProcessedEventArgs args)
...
+        /// </summary>
+        /// <returns>The number of events that are sendable</returns>
+        private int IsTransmissionSendable(int numEvents = 1, bool peek = false)
+        {
+            if (!this.applyThrottle)
+            {
+                return numEvents;
+            }
+
+            var throttleWindowId = (long)(DateTime.UtcNow.Subtract(new DateTime(1970, 1, 1)).TotalMilliseconds / this.ThrottleWindow);
```
dnduffy: Please don't use var in place of primitives (long, bool, etc.) and ideally not on anything that isn't anonymous.

Test/ServerTelemetryChannel.Test/Shared.Tests/Helpers/StubTelemetrySerializer.cs
```diff
@@ -9,7 +9,7 @@ internal class StubTelemetrySerializer : TelemetrySerializer
     {
         public Action<IEnumerable<ITelemetry>> OnSerialize = _ => { };
 
-        public override void Serialize(IEnumerable<ITelemetry> items)
+        public override void Serialize(ICollection<ITelemetry> items)
```
OsvaldoRosado: The underlying method was changed, so the this override was changed to match. The underlying method was changed to allow accessing the count of items without enumerating.

Test/ServerTelemetryChannel.Test/TestFramework/Shared/StubTransmission.cs
```diff
@@ -35,5 +41,58 @@ public override Task<HttpWebResponseWrapper> SendAsync()
...
+        private StubTransmission Convert(Transmission transmission)
+        {
+            if (transmission != null)
+            {
+                if (transmission.TelemetryItems == null)
+                {
+                    transmission = new StubTransmission(transmission.Content)
+                    {
+                        OnSave = this.OnSave,
+                        OnSend = this.OnSend,
```
OsvaldoRosado: Adding the extra properties still didn't allow for the equality check to pass. Instead the stub uses the same functionality as them real method: the exact object is returned if no split is necessary. The Ids changing on actual splits makes sense (they are new transmissions). Ids are only used for logging purposes from what I can find, so there should be no functional impact.

src/Core/Managed/Shared/Channel/Transmission.cs
```diff
@@ -57,12 +57,25 @@ public Transmission(Uri address, byte[] content, string contentType, string cont
...
             this.client = new HttpClient() { Timeout = this.Timeout };
 #endif
         }
 
-        internal Transmission(Uri address, IEnumerable<ITelemetry> telemetryItems, string contentType, string contentEncoding, TimeSpan timeout = default(TimeSpan)) 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Transmission"/> class.
+        /// </summary>
+        public Transmission(Uri address, ICollection<ITelemetry> telemetryItems, TimeSpan timeout = default(TimeSpan)) 
+            : this(address, JsonSerializer.Serialize(telemetryItems, true), "application/x-json-stream", JsonSerializer.CompressionType, timeout)
```
OsvaldoRosado: This was effectively always the case, the serializing just now happens in this constructor directly. What would it even do if you gave a different content type? We only serialize to JSON.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ErrorHandlingTransmissionPolicy.cs
```diff
@@ -33,7 +34,7 @@ private void HandleTransmissionSentEvent(object sender, TransmissionProcessedEve
             var webException = e.Exception as WebException;
             if (webException != null)
             {
-                HttpWebResponse httpWebResponse = webException.Response as HttpWebResponse;
+                HttpWebResponseWrapper httpWebResponse = e.Response;
```
OsvaldoRosado: It always was. The change is that it will now be available on all exceptions. This allows us to use e.Response with confidence rather than webException.Response (which is not the wrapper)

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryChannelEventSource.cs
```diff
@@ -451,6 +451,17 @@ public void TransmissionStorageUnexpectedRetryOfBadFile(string fileName, string
             this.WriteEvent(63, fileName, this.ApplicationName);
         }
 
+        [Event(64, Message = "Transmission locally throttled. Throttle Limit: {0}. Attempted: {1}. Accepted: {2}. ", Level = EventLevel.Warning)]
```
OsvaldoRosado: I'm not opposed to reducing the level here. It is user actionable if not caused from retrieval from storage (data rate needs to decrease)

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionSender.cs
```diff
@@ -100,15 +164,105 @@ protected void OnTransmissionSent(TransmissionProcessedEventArgs args)
                 int currentCapacity = Interlocked.Decrement(ref this.transmissionCount);
                 if (exception == null)
                 {
-                    TelemetryChannelEventSource.Log.TransmissionSentSuccessfully(transmission.Id, currentCapacity);
+                    TelemetryChannelEventSource.Log.TransmissionSentSuccessfully(acceptedTransmission.Id, currentCapacity);
+
+                    if (responseContent == null && exception is WebException)
```
OsvaldoRosado: Oops!

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionSender.cs
```diff
@@ -100,15 +164,105 @@ protected void OnTransmissionSent(TransmissionProcessedEventArgs args)
...
+                    {
+                        var response = (HttpWebResponse)((WebException)exception).Response;
+                        responseContent = new HttpWebResponseWrapper()
+                        {
+                            StatusCode = (int)response.StatusCode,
+                            StatusDescription = response.StatusDescription,
+                            RetryAfterHeader = response.Headers.Get("Retry-After")
```
OsvaldoRosado: Yes, that's correct

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionSender.cs
OsvaldoRosado: The policy that handles throttling would respect that rather than use exponential backoff. Is that better?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionSender.cs
OsvaldoRosado: It should never be negative. If it is, this method allowed more than the throttle limit to go by.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionSender.cs
OsvaldoRosado: Oops! Thought I caught them all

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TransmissionSender.cs
```diff
@@ -100,15 +164,105 @@ protected void OnTransmissionSent(TransmissionProcessedEventArgs args)
...
+                this.OnTransmissionSent(new TransmissionProcessedEventArgs(acceptedTransmission, exception, responseContent));
+            }
+        }
+
+        /// <summary>
+        /// Checks if the transmission throttling policy allows for sending another request.
+        /// If so, this method will add a request to the current throttle count (unless peeking)
+        /// </summary>
+        /// <returns>The number of events that are sendable</returns>
+        private int IsTransmissionSendable(int numEvents = 1, bool peek = false)
```
OsvaldoRosado: This was used in an incarnation that didn't support decoding raw content byte arrays when ITelemetry items weren't available on the Transmission object. I can remove this now.

Test/ServerTelemetryChannel.Test/TestFramework/Shared/StubTransmission.cs
dnduffy: True, until they leave the box the Ids are meaningless I think.

src/Core/Managed/Shared/Channel/Transmission.cs
```diff
@@ -212,6 +233,112 @@ public virtual Task<HttpWebResponseWrapper> SendAsync()
...
+                        itemsB);
+                }
+            }
+            else if (this.ContentType == JsonSerializer.ContentType)
+            {
+                // We have to decode the payload in order to split
+                bool compress = this.ContentEncoding == JsonSerializer.CompressionType;
+                string[] payloadItems = JsonSerializer
+                    .Deserialize(this.Content, compress)
+                    .Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
```
dnduffy: This deserialization code is in a couple places, is there an easy way to only have one copy?

PR #339, opened 2016-10-25
Iusafaro/nodename

iusafaro:
Move logic to populate RoleInstance with machine name from WindowsServer to Core and add population of NodeName in the same way.


src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -428,6 +428,17 @@ public void Initialize(ITelemetry telemetry)
                 var version = LazyInitializer.EnsureInitialized(ref this.sdkVersion, this.GetSdkVersion);
                 telemetry.Context.Internal.SdkVersion = version;
             }
+
+            // set RoleInstance and NodeName to the machine name if they're not initialized yet
+            if (string.IsNullOrEmpty(telemetry.Context.Internal.NodeName))
+            {
+                telemetry.Context.Internal.NodeName = PlatformSingleton.Current.GetMachineName();
```
SergeyKanzhelev: you clearly need a `try{}catch` somewhere while getting the computer name

SergeyKanzhelev: Also should we try to use roleINstance first and then fall back to the computer name? Or we do not want to trust the customer's choice or roleInstance?

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -428,6 +428,17 @@ public void Initialize(ITelemetry telemetry)
...
+            }
+
+            if (string.IsNullOrEmpty(telemetry.Context.Cloud.RoleInstance))
+            {
+                telemetry.Context.Cloud.RoleInstance = PlatformSingleton.Current.GetMachineName();
```
SergeyKanzhelev: Since you already setting roleInstance manually - you can just let our endpoint populate the nodeName from the roleInstance. No need to set two equal values

src/Core/Managed/Shared/TelemetryClient.cs
iusafaro: 1) added try/catch in PlatformImplementation
2) yes, it's intentional - i don't want to rely on customers choice for role instance as it may be done for display purposes and conflict with billing

src/Core/Managed/Shared/TelemetryClient.cs
SergeyKanzhelev: can you add this check. If both were not set - only send roleInstance

PR #352, opened 2016-11-08
Increased timeout for test that was sporadically failing on the build environment.

dnduffy:


Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TransmissionStorageTest.cs
```diff
@@ -654,7 +654,7 @@ public void SkipsTransmissionFileBeingCurrentlyLoadedByAnotherThreadAndTriesLoad
                 Assert.NotNull(dequeued);
             }
 
-            [TestMethod, Timeout(200)]
+            [TestMethod, Timeout(250)]
```
kamilszostak: Why 250 :)?

PR #357, opened 2016-11-10
Hand coded sanitization logic to match the bond schema, unit tests.

cijothomas:
Hand coded sanitization logic to match the bond schema, unit tests. This is Short-Term fix until we fully implement auto generation of sanitization logic from bond schema.

### DIFF
--- a/Test/CoreSDK.Test/Shared/Extensibility/Implementation/PropertyTest.cs
+++ b/Test/CoreSDK.Test/Shared/Extensibility/Implementation/PropertyTest.cs
@@ -9,6 +9,7 @@
     using Microsoft.VisualStudio.TestPlatform.UnitTestFramework;
 #endif
     using Assert = Xunit.Assert;
+    using DataContracts;
 
     [TestClass]
     public class PropertyTest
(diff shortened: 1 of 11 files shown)


src/Core/Managed/Shared/Extensibility/Implementation/Property.cs
```diff
@@ -120,6 +150,18 @@ public static string SanitizeAvailabilityMessage(this string value)
             return TrimAndTruncate(value, Property.MaxAvailabilityMessageLength);
         }
 
+        public static void SanitizeTelemetryContext(this TelemetryContext telemetryContext)
+        {
+            SanitizeComponentContext(telemetryContext.Component);
```
SergeyKanzhelev: typical event doesn't have all the context properties set. Also every get/set operation is a lookup and copying of the dictionary. 

Instead of this implementation - loop thru the tags collection directly and only truncate if needed.

SergeyKanzhelev:
@dnduffy can you pls also review and merge if ok

Test/CoreSDK.Test/Shared/Extensibility/Implementation/PropertyTest.cs
```diff
@@ -291,6 +293,82 @@ public void SanitizeMeasurementsReplacesNegativeInfinityWith0()
             Assert.Equal(0, original["Key"]);
         }
 
+        [TestMethod]
+        public void SanitizeTelemetryContextTest()
+        {            
+            var telemetryContext = new TelemetryContext();
```
dnduffy: nitpick: I would prefer to see the concrete type instead of var when the variable is not an anonymous type.

src/Core/Managed/Shared/Extensibility/Implementation/Property.cs
```diff
@@ -17,15 +19,42 @@ internal static class Property
...
         public const int MaxNameLength = 1024;
         public const int MaxMessageLength = 32768;
         public const int MaxUrlLength = 2048;
-        public const int MaxCommandNameLength = 2 * 1024;
+        public const int MaxDataLength = 8 * 1024;
         public const int MaxTestNameLength = 1024;
         public const int MaxRunLocationLength = 2024;
         public const int MaxAvailabilityMessageLength = 8192;
 
+        public static readonly IDictionary<string, int> TagSizeLimits = new Dictionary<string, int>()
```
dnduffy: Why do these need to be in a dictionary as opposed to constants?
Do we need to look them up at runtime by name, as done in SanitizeTelemetryContext currently, or do we have enough control to generate out the specific property references so we don't pay the hashing and lookup costs?

dnduffy: For future, when generating from the Bond schema files, I think we should be explicit and not do lookups.

src/Core/Managed/Shared/Extensibility/Implementation/Property.cs
```diff
@@ -221,6 +269,6 @@ private static string MakeKeyUnique<TValue>(string key, IDictionary<string, TVal
             }
 
             return key;
-        }
+        }
```
dnduffy: nitpick: extra spaces got added on the end of the line.

PR #359, opened 2016-11-11
Drop "aggregator" from names

vitalyf007:


olegsych:
Love it! 😍 Do you want to shorten `Metric.MetricName` to simply `Metric.Name`?

SergeyKanzhelev:
@olegsych @vitalyf007 let's also get back to use `MetricTelemetry` instead of `AggregatedMetricTelemetry`? Since there are no `Aggregate` anywhere now.

vitalyf007:
Not sure why I can't comment in line... So, I'll comment here. I'll implement Oleg's suggestion (MetricName => Name). As far as Sergey's suggestion, we can't really do it. MetricTelemetry already exists and I'd like to keep AggregatedMetricTelemetry as is for now.

SergeyKanzhelev:
@vitalyf007 any reasons to keep `AggregatedMetricTelemetry` instead of adding properties to `MetricTelemetry`?

vitalyf007:
The reason to keep MT and introduce AMT is two fold. First, back compat. Try to figure out how to add properties to existing MT such that none of the allowed usages are broken is hard. Usually it leads to situations when object is opened up to inconsistent state problems. MT represents two things and I want to deprecate it in favor of more clear object AMT.

SergeyKanzhelev:
I understand what you want to represent. But on the surface for the customer they are the same except `Value` is not `Sum` and other properties are not `nullable`. Usage of the class is exactly the same except one of constructors do not exists.

I just worry that creation of a new class is our attempt to make it clearer in our own heads. 

@olegsych - do you have an opinion here?

olegsych:
@SergeyKanzhelev Regarding keeping `MetricTelemetry` vs introducing `AggregatedMetricTelemetry` I agree with the reasons @vitalyf007 cited. It's an important step toward cleaning up the confusion around `MetricTelemetry`. However, I feel we should only do it if we can commit to cleaning this up in the HTTP API and AA schemas within a reasonably short amount of time, ideally within one year. If that's not realistic, keeping the existing `MetricTelemetry` would be a lesser evil.

SergeyKanzhelev:
@olegsych we removed the term "Aggregation" from everywhere. We just saying "the metric". So we are sending the metric value as a `MetricTelemetry` object. This is my reasoning that the new type clears the situation in our mind, but confuses the customer. We do not expose metric kind in UI or AA. Maybe export shows it. But in general for the customer - metric got reported, value (most times aggregated) can be read in UI and AA.

olegsych:
@SergeyKanzhelev Good point. So instead of introducing `AggregatedMetricTelemetry` as a new type, we could also keep `MetricTelemetry`, add `Sum` property to it and deprecate its existing `Value` property. With the `TelemetryClient.TrackMetric(string, double, IDictionary)` method already deprecated, the confusion around whether metrics are pre-aggregated or not is also removed. @vitalyf007 I think I like this better than introducing a new telemetry class.

Issue #358, opened 2016-11-11
Accuracy issue with duration for dependency operation.


cijothomas:
Commit de66830f changed the implementation of calculating duration for dependency, from using Stopwatch to DatTimeOffset.Now. This results in duration to be zero when the time difference is extremely small. (like ~1-5 milli seconds.)
(This has resulted in FUNC test failures for SQL dependency only so far as SQL access times are extremely small.)

Some relevant discussion:
http://stackoverflow.com/questions/28637/is-datetime-now-the-best-way-to-measure-a-functions-performance

cijothomas:
Some unit tests are also failing because of this.
For eg:
Failed   OnEndSendsCorrectTelemetry_1ArgumentOverride
The test uses Stopwatch() but product uses DateTIme... So this fails randomnly.

PR #360, opened 2016-11-15
Implement higher precision clock using StopWatch.

cijothomas:
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/358

Fix for issue 358 - use higher prevision clock for calculating duration for Operation.


src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
```diff
@@ -32,6 +32,12 @@ public DateTimeOffset StartTime
         /// </summary>  
         public abstract string Id { get; set; }
 
+        /// <summary>  
+        /// Gets or sets TimeSpan representing begin time of the operation. Used internally
+        /// for calculating duration.
+        /// </summary>  
+        public TimeSpan BeginTime { get; set; }
```
cijothomas: Not sure if this should be public or if this is the right place to store the time.

src/Core/Managed/Shared/Extensibility/Implementation/Clock.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    /// <summary>
+    /// A highly-accurate, precise and testable clock.
+    /// </summary>
+    internal class Clock : IClock
+    {
+        // DateTimeOffset.Now is accurate to ~16ms. We want accurate and precise timestamps,
+        // so use Stopwatch which is based on perf counters and is highly accurate and precise.
+        // Making a static instance of Stopwatch shared by all operations.        
+        private static readonly Stopwatch InitialStopWatch = Stopwatch.StartNew();
```
SergeyKanzhelev: Why is it called `Initial`?

src/Core/Managed/Shared/Extensibility/Implementation/IClock.cs
```diff
@@ -4,6 +4,6 @@
 
     internal interface IClock
     {
-        DateTimeOffset Time { get; }
+        TimeSpan Time { get; }
```
SergeyKanzhelev: was this interface used anywhere else? Why are you returning timespan instead of a number of ticks representing the point in time. First, the name of the property is misleading to what it returns. Second - you are loosing in precision as you converts to the lower-precision value twice. Third - you actually doing two conversions (arithmetic operations) instead of one that is required.

src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
SergeyKanzhelev: definitely internal. You should use `Start` and `Stop` extension methods

src/Core/Managed/Shared/Extensibility/Implementation/IClock.cs
cijothomas: This Interface is not used elsewhere.
Any advantage of using Ticks instead of storing TimeSpan? StopWatch can return Ticks or TimeSpan - i used Timespan as Duration is already in TimeSpan.
Not clear about where the two conversions are happening?

src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
cijothomas: Not clear - this field is already used in Start/Stop extension methods.

src/Core/Managed/Shared/Extensibility/Implementation/Clock.cs
cijothomas: yea Initial was not very useful name. will change to StopWatchShared?

src/Core/Managed/Shared/Extensibility/Implementation/Clock.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public static IClock Instance
+        {
+            get { return instance; }
+            protected set { instance = value ?? new Clock(); }
+        }
+
+        public TimeSpan Time
+        {
+            get { return InitialStopWatch.Elapsed; }
```
cijothomas: This is not thread-safe, so need to use alternate static method..

cijothomas: Stopwatch.GetTimestamp() is the alternate. So dont need to have a internal wrapperaround stopwatch as well.

Test/CoreSDK.Test/Shared/Extensibility/Implementation/ClockTest.cs
```diff
@@ -0,0 +1,55 @@
...
+        public void TimeReturnsHighPrecisionDateTimeOffsetValue()
+        {
+            IClock clock = Clock.Instance;
+
+            TimeSpan timestamp1 = clock.Time;
+            new ManualResetEvent(false).WaitOne(TimeSpan.FromTicks(50));
+            TimeSpan timestamp2 = clock.Time;
+
+            Assert.True(timestamp2 - timestamp1 > TimeSpan.FromTicks(5));
+            Assert.True(timestamp2 - timestamp1 < TimeSpan.FromTicks(5000));
```
dnduffy: If you only waited 50 ticks then checking for less than 5000 seems a bit high...
This test doesn't seem to be particularly useful with such a big margin of error and given that it is timing dependent I don't see how you could make it precise enough to be useful.

src/Core/Managed/Shared/Extensibility/Implementation/Clock.cs
dnduffy: I'm not really sure what these comments are/were referring to?

src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
dnduffy: +1 for internal

src/Core/Managed/Shared/Extensibility/Implementation/Clock.cs
dnduffy: I don't think it needs a fancy name, it's private so it's not "shared" in the context of the class.

src/Core/Managed/Shared/OperationTelemetryExtensions.cs
```diff
@@ -16,9 +16,12 @@ public static class OperationTelemetryExtensions
...
         public static void Start(this OperationTelemetry telemetry)
-        {
-            var startTime = DateTimeOffset.UtcNow;
-            telemetry.Timestamp = startTime;
+        {            
+            telemetry.Timestamp = DateTimeOffset.UtcNow;
+
+            // Begin time is used internally for calculating duration of operation at the end call,
+            // and hence is stored using higher precision Clock.
+            telemetry.BeginTime = Clock.Instance.Time;
```
dnduffy: If this is the only place that Clock is used do we really need a class for it or should the singleton Stopwatch instance just go someplace more direct?

src/Core/Managed/Shared/OperationTelemetryExtensions.cs
```diff
@@ -16,9 +17,13 @@ public static class OperationTelemetryExtensions
...
-        {
-            var startTime = DateTimeOffset.UtcNow;
-            telemetry.Timestamp = startTime;
+        {            
+            telemetry.Timestamp = DateTimeOffset.UtcNow;
+
+            // Begin time is used internally for calculating duration of operation at the end call,
+            // and hence is stored using higher precision Clock.
+            // Stopwatch.GetTimestamp() is used (instead of ElapsedTicks) as it is thread-safe.
+            telemetry.BeginTimeInTicks = Stopwatch.GetTimestamp();
```
dnduffy: I'm curious what thread safety you are worried about when simply reading a value?

cijothomas: Original code was using a globally shared StopWatch instance and reading Ticks using stopwatch's Elapsed property
which is not thread-safe as per MSDN doc below.
https://msdn.microsoft.com/en-us/library/system.diagnostics.stopwatch(v=vs.110).aspx

So I switched to thread-safe static method StopWatch.GetTimeStamp(), which eliminated the need for having a Globally shared instance of stopwatch in product code.

dnduffy: I doubt the property read would have been an issue but it doesn't matter because I like this simplified version that doesn't need us to keep an instance better anyway.

cijothomas:
Closing this issue as this is fixed for base sdk. For web sdk, new issue is opened: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/286

PR #369, opened 2016-11-21
Metric agg: drop "aggregated" from all names plus a bit of feedback

vitalyf007:


Test/CoreSDK.Test/Shared/DataContracts/MetricTelemetryTest.cs
```diff
@@ -205,7 +225,48 @@ public void SerializeReplacesNaNStandardDeviationOn0()
 
             Assert.Equal(0, original.StandardDeviation.Value);
         }
+
+        [TestMethod]
+        public void SerializeReplacesNaNSumOn0()
```
olegsych: Why do the names of tests for the `Sanitize` method start with _Serialize_?

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -205,10 +205,9 @@ public void TrackMetric(string name, double value, IDictionary<string, string> p
         }
 
         /// <summary>
-        /// Send a <see cref="MetricTelemetry"/> for aggregation in Metric Explorer.
+        /// Send a <see cref="MetricTelemetry"/> for representing aggregated metric data.
         /// Create a separate <see cref="MetricTelemetry"/> instance for each call to <see cref="TrackMetric(MetricTelemetry)"/>.
         /// </summary>
-        [Obsolete("This method is obsolete. Use TrackAggregatedMetric method to send pre-aggregated metric data or MetricManager class to create metrics.")]
         [EditorBrowsable(EditorBrowsableState.Never)]
```
olegsych: Remove `[EditorBrowsable(Never)]` from `TrackMetric(MetricTelemetry)`?

Test/CoreSDK.Test/Shared/DataContracts/MetricTelemetryTest.cs
vitalyf007: Makes sense. Will change (the naming is historical but you bring a good point).

src/Core/Managed/Shared/TelemetryClient.cs
vitalyf007: My bad :( Fixing.

olegsych: @vitalyf007 Thanks for removing `[EditorBrowseable]` from the `MetricTelemetry` class. I didn't catch that. Should we also remove it from the `TrackMetric(MetricTelemetry)` method now that it's no longer `[Obsolete]`?

vitalyf007: Yes.  Interesting that my Git client missed that change and only picked it up once I updated to the latest version...

PR #373, opened 2016-11-22
Source property on the request telemetry object was never getting ser…

nizarq:
…ialized.


SergeyKanzhelev:
why have you closed the same PM for `develop` branch? I think it is reasonable to take it for the 2.2 stable

PR #374, opened 2016-11-22
Serialize source property

nizarq:


nizarq:
This is the same change that has already been reviewed and merged into 2.3.0-beta1 (https://github.com/Microsoft/ApplicationInsights-dotnet/pull/373). Doing the same for develop

Issue #364, opened 2016-11-17
TelemetryClient creates InMemoryChannel in Track but never disposes it


Dmitry-Matveev:
__TelemetryClient__ has a logic to create __InMemoryChannel__ in case no other channel is configured.
The code is [here](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/de488729c28b721a758e0024ea2d4f79909ff797/src/Core/Managed/Shared/TelemetryClient.cs#L357).

__InMemoryChannel__ creates __InMemoryTransmitter__ that in turn starts a task (Runner()) on a thread pool thread. When **TelemetryClient** is no longer used and the new **TelemetryClient** is created, the old thread stays in **WaitOneNative**() waiting for the next send interval. Dispose() method of **InMemoryTransmitter**/**InMemoryChannel** would fire the event to stop waiting and will disable the Runner(), however neither seems to be currently invoked if Telemetry Client is GC'd.

The issue will reproduce if custom or default configuration object is passed into **TelemetryClient** on each creation, the issue should not reproduce if **TelemetryConfiguration.Active** is used - one channel will remain active.

The stack of the stale threads is:
System.Threading.WaitHandle.WaitOneNative
System.Threading.WaitHandle.InternalWaitOne
System.Threading.WaitHandle.WaitOne
Microsoft.ApplicationInsights.Channel.InMemoryTransmitter.Runner()

dnduffy:
Proper Dispose cleanup is hampered by the fact that the TelemetryChannel setter on TelemetryConfiguration is public.  Someone can replace the channel instance on a configuration instance without disposing of the old channel which makes for a messed up anti-pattern.
I think we need to remove some of the overly open property access in future.

dnduffy:
We have several scenarios supported in the existing code:
1) Default configuration, singleton instance (Active)
2) Create new "default" configuration, for dependency injection
3) Custom telemetry channel, set either when XML config is read or when developer does something like "TelemetryConfiguration.Active.TelemetryChannel = new SomeCustomChannel();"

TelemetryConfiguration currently has no non-default constructor and needs a setter on the TelemetryChannel property but this poses the problem about when a channel is created and who is responsible for disposing of it.

dnduffy:
PR: https://github.com/Microsoft/ApplicationInsights-dotnet/pull/406/files

olegsych:
[...]

PR #375, opened 2016-11-23
Refactor TaskTimer

macrogreg:
Hi folks,

After some discussions with Vitaly and Oleg, my goal is to improve on some of the background processing code in the SDK, specifically around the TaskTimer class.
There are several issues with that class. An instance of it is a mix between something that represents one work item and something that is used to start multiple work items. As a result both, usage and code and confusing. But the first improvement I would like to make is making it non-disposable. For that, I am making cancellation less aggressive and removing the cancellation token source, which was the reason for being disposable. I also checked all usages and removed IDisposable from all classes  where TaskTimer was the only requirement for that interface.

### DIFF
--- a/src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
+++ b/src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
@@ -17,15 +17,22 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
     /// <summary>
     /// Runs a task after a certain delay and log any error.
     /// </summary>
-    public class TaskTimer : IDisposable
+    public class TaskTimer
     {
+        private class DelayedWork
+        {
+            public DelayedWork(Func<Task> runFunc) { this.RunFunc = runFunc; this.IsCancellationRequested = false; }
+            public Func<Task> RunFunc { get; private set; }
+            public bool IsCancellationRequested { get; set; }
+        }
+
         /// <summary>
         /// Represents an infinite time span.
         /// </summary>
         public static readonly TimeSpan InfiniteTimeSpan = new TimeSpan(0, 0, 0, 0, Timeout.Infinite);
 
         private TimeSpan delay = TimeSpan.FromMinutes(1);
-        private CancellationTokenSource tokenSource;
+        private DelayedWork latestWork = null;
 
         /// <summary>
         /// Gets or sets the delay before the task starts. 
@@ -53,7 +60,10 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
         /// </summary>
         public bool IsStarted
         {
-            get { return this.tokenSource != null; }
+            get
+            {
+                return (this.latestWork != null);
+            }
         }
 
         /// <summary>
@@ -62,34 +72,50 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
         /// <param name="elapsed">The task to run.</param>
         public void Start(Func<Task> elapsed)
         {
-            var newTokenSource = new CancellationTokenSource();
+            DelayedWork delayedWork = new DelayedWork(elapsed);
+            DelayedWork previousWork = Interlocked.Exchange(ref this.latestWork, delayedWork);
 
-            TaskEx.Delay(this.Delay, newTokenSource.Token)
-                .ContinueWith(
+            if (previousWork != null)
+            {
+                previousWork.IsCancellationRequested = true;
+            }
+
 #if !NET40
-                async previousTask =>
+            Task delayTask = Task.Delay(this.Delay);
 #else
-                    previousTask =>
+            Task delayTask = TaskEx.Delay(this.Delay);
 #endif
+            delayTask.ContinueWith(
+                    (dlyTask) =>
                     {
-                        CancelAndDispose(Interlocked.CompareExchange(ref this.tokenSource, null, newTokenSource));
+                        Interlocked.CompareExchange(ref this.latestWork, null, delayedWork);
+                        if (delayedWork.IsCancellationRequested)
+                        {
+                            return;
+                        }
+
                         try
                         {
-                            Task task = elapsed();
+                            // Task may (or may not) be executed syncronously
+                            Task task = (delayedWork.RunFunc == null)
+                                                ? null
+                                                : delayedWork.RunFunc();
 
-                            // Task may be executed syncronously
-                            // It should return Task.FromResult but just in case we check for null if someone returned null
-                            if (task != null)
+                            // Just in case we check for null if someone returned null
+                            if (task == null)
                             {
-#if !NET40
-                                await task.ConfigureAwait(false);
-#else
-                                task.ContinueWith(
-                                    userTask =>
+                                return;
+                            }
+
+                            task.ContinueWith(
+                                    (userTask) =>
                                     {
                                         try
                                         {
-                                            userTask.RethrowIfFaulted();
+                                            if (task.IsFaulted)
+                                            {
+                                                throw new AggregateException(task.Exception).Flatten();
+                                            }
                                         }
                                         catch (Exception exception)
                                         {
@@ -98,9 +124,8 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
                                     },
                                     CancellationToken.None,
                                     TaskContinuationOptions.ExecuteSynchronously,
-                                    TaskScheduler.Default);
-#endif
-                            }
+                                    TaskScheduler.Default
+                                );
                         }
                         catch (Exception exception)
                         {
@@ -110,8 +135,6 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
                     CancellationToken.None,
                     TaskContinuationOptions.OnlyOnRanToCompletion | TaskContinuationOptions.ExecuteSynchronously,
                     TaskScheduler.Default);
-
-            CancelAndDispose(Interlocked.Exchange(ref this.tokenSource, newTokenSource));
         }
 
         /// <summary>
@@ -119,17 +142,13 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
         /// </summary>
         public void Cancel()
         {
-            CancelAndDispose(Interlocked.Exchange(ref this.tokenSource, null));
+            DelayedWork currentWork = this.latestWork;
+            if (currentWork != null)
+            {
+                currentWork.IsCancellationRequested = true;
+            }
         }
 
-        /// <summary>
-        /// Releases unmanaged and - optionally - managed resources.
-        /// </summary>
-        public void Dispose()
-        {
-            this.Dispose(true);
-            GC.SuppressFinalize(this);
-        }
 
         /// <summary>
         /// Log exception thrown by outer code.
@@ -148,23 +167,6 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
             }
 
             CoreEventSource.Log.LogError(exception.ToInvariantString());
-        }
-
-        private static void CancelAndDispose(CancellationTokenSource tokenSource)
-        {
-            if (tokenSource != null)
-            {
-                tokenSource.Cancel();
-                tokenSource.Dispose();
-            }
-        }
-
-        private void Dispose(bool disposing)
-        {
-            if (disposing)
-            {
-                this.Cancel();
-            }
         }
     }
 }
--- a/src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
+++ b/src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
@@ -21,8 +21,8 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
     {
         private class DelayedWork
         {
-            public DelayedWork(Func<Task> runFunc) { this.RunFunc = runFunc; this.IsCancellationRequested = false; }
-            public Func<Task> RunFunc { get; private set; }
+            public DelayedWork(Func<Task> runWhenElapsedFunc) { this.RunWhenElapsedFunc = runWhenElapsedFunc; this.IsCancellationRequested = false; }
+            public Func<Task> RunWhenElapsedFunc { get; private set; }
             public bool IsCancellationRequested { get; set; }
         }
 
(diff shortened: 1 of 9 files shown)


SergeyKanzhelev:
Can you also look at this issue https://github.com/Microsoft/ApplicationInsights-dotnet/issues/364?

macrogreg:
@SergeyKanzhelev Does #364 have any relation to this change?
If not, I'd liek to finish looking at this and then I can take a look at #364. Works?

SergeyKanzhelev:
oh, channel doesn't use task timer. I thought it is and also affected by non-disposing =) SOrry

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
```diff
@@ -17,15 +17,22 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
     /// <summary>
     /// Runs a task after a certain delay and log any error.
     /// </summary>
-    public class TaskTimer : IDisposable
+    public class TaskTimer
     {
+        private class DelayedWork
+        {
+            public DelayedWork(Func<Task> runFunc) { this.RunFunc = runFunc; this.IsCancellationRequested = false; }
+            public Func<Task> RunFunc { get; private set; }
```
olegsych: Could you name the `DelayedWork.RunFunc` property consistently with the `Start` method parameter?

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
```diff
@@ -62,34 +72,50 @@ public bool IsStarted
...
+                            // Task may (or may not) be executed syncronously
+                            Task task = (delayedWork.RunFunc == null)
+                                                ? null
+                                                : delayedWork.RunFunc();
 
-                            // Task may be executed syncronously
-                            // It should return Task.FromResult but just in case we check for null if someone returned null
-                            if (task != null)
+                            // Just in case we check for null if someone returned null
+                            if (task == null)
```
olegsych: I don't think return of `null` from the `Func<Task>` passed to the `TaskTimer.Start` method is an expected runtime behavior. We should let the `NullReferenceException` be thrown below so that it is tracked as an internal error instead of being ignored.

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
```diff
@@ -62,34 +72,50 @@ public bool IsStarted
         /// <param name="elapsed">The task to run.</param>
         public void Start(Func<Task> elapsed)
         {
-            var newTokenSource = new CancellationTokenSource();
+            DelayedWork delayedWork = new DelayedWork(elapsed);
+            DelayedWork previousWork = Interlocked.Exchange(ref this.latestWork, delayedWork);
+
+            if (previousWork != null)
+            {
+                previousWork.IsCancellationRequested = true;
```
olegsych: In an application that tracks a large volume of telemetry, `TelemetryBuffer` may `Start` and `Cancel` its `TaskTimer` a lot. By replacing the `CancellationTokenSource` with `DelayedWork.IsCancellationRequested` we may be significantly increasing the number of `Delay` tasks that will be allowed to continue running to completion instead of being cancelled. How does the new implementation affect memory allocation?

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
```diff
@@ -62,34 +72,50 @@ public bool IsStarted
...
+
+                            task.ContinueWith(
+                                    (userTask) =>
                                     {
                                         try
                                         {
-                                            userTask.RethrowIfFaulted();
+                                            if (task.IsFaulted)
+                                            {
+                                                throw new AggregateException(task.Exception).Flatten();
```
olegsych: Why do we need to wrap then flatten the original exception here when `LogException` already does it?

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
```diff
@@ -62,34 +72,50 @@ public bool IsStarted
...
+                            return;
+                        }
+
                         try
                         {
-                            Task task = elapsed();
+                            // Task may (or may not) be executed syncronously
+                            Task task = (delayedWork.RunFunc == null)
+                                                ? null
+                                                : delayedWork.RunFunc();
```
olegsych: The `elapsed` parameter (and the `DelayedWork.ReadFunc` property) should never be `null` here. The original code was [failing fast](http://martinfowler.com/ieeeSoftware/failFast.pdf) by allowing the `NullReferenceException` to be thrown and tracked by the `LogException` call inside the `catch` statement below. I think its important to preserve the original behavior.

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
```diff
@@ -110,26 +135,20 @@ public void Start(Func<Task> elapsed)
...
+            if (currentWork != null)
+            {
+                currentWork.IsCancellationRequested = true;
+            }
         }
 
-        /// <summary>
-        /// Releases unmanaged and - optionally - managed resources.
-        /// </summary>
-        public void Dispose()
```
olegsych: To keep functionality of classes that use `TaskTimer` "as is", they should now `Cancel` their timers instead of `Disposing` them. Simply removing the `Dispose` calls drops the `Cancel` logic of the previous implementation.

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
```diff
@@ -17,15 +17,22 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation
     /// <summary>
     /// Runs a task after a certain delay and log any error.
     /// </summary>
-    public class TaskTimer : IDisposable
+    public class TaskTimer
```
SergeyKanzhelev: I wonder - do we need to make this class internal and mark the public version as Obsolete for now? I do not understand why Application Insights library should provide this functionality

olegsych: Good point. `TaskTimer` is defined in the `Microsoft.ApplicationInsights` assembly because it is still used here as well as in other AI packages that depend on it. However, it doesn't need to be public anymore because we no longer publish WinRT assemblies. The class was implemented as public because WinRT didn't support internal class visibility at the time. Instead we placed it in the "Implementation" namespace and decided that semver does not apply to public types in this namespace. It should be safe to make `TaskTimer` internal straight away.

src/Core/Managed/Shared/Extensibility/MetricManager.cs
```diff
@@ -119,12 +119,6 @@ public void Flush()
         /// </summary>
         public void Dispose()
         {
-            if (this.snapshotTimer != null)
-            {
-                this.snapshotTimer.Dispose();
-                this.snapshotTimer = null;
-            }
-
             this.Flush();
```
vitalyf007: It appears to me that in case the only job of `MetricManager.Dispose()` class is to call `Flush()`, we can make `MetricManager` itself non-disposable?

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
vitalyf007: +1 to Oleg's comment. Looks real strange...

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
macrogreg: "runFunc" is the better name. The Start method parameter ("elapsed") must corresponds to a specific usage of this public class. However, Start being public, I did not want to change the parameter name. Let me call this "runWhenElapsedFunc" to underline the correspondence.

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
macrogreg: When I am thinking of respective scenarios, I imagine maybe a dozen of additional timers. This should not matter. Can you think of scenarios where the impact is bigger. If so, please clarify and I may need to rethink my approach.
@olegsych

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
macrogreg: I will preserve the original behavior for now.

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
macrogreg: FYI:
```
        public static void RethrowIfFaulted(this Task task)
        {
            if (!task.IsCompleted)
            {
                throw new ArgumentException("Task is not yet completed");
            }

            if (task.IsFaulted)
            {
                throw new AggregateException(task.Exception).Flatten();
            }
        }
```

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
macrogreg: Good point. I added Dispose and marked it Obsolete.

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
macrogreg: I generally agree, but I wanted to focus on Disposable right here since technically making it internal is still a breaking change.

src/Core/Managed/Shared/Extensibility/MetricManager.cs
macrogreg: We could. Or we could leave it to underline that Flush *must* be called. I think either decision is not a blocker for this change.

vitalyf007: I think leaving disposable on `MetricManager `is an abuse of the pattern... It was designed to free consumed resources, not to remind implementor to do something... I'd like to have `IDisposable `implementation removed from `MetricManager`. I get your point that it is not a pre-req for this change, but we must not forget... Removing it later will be a breaking change...

macrogreg: I tend to agree, but it is still a separate issue. Once this is approved, I will complete the PR and then you have make that change as a part of the general MetricsManager dev work. Would that be OK?

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
olegsych: The number of `Delay` tasks created by the `Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation.TelemetryBuffer` depends on the rate of telemetry tracking. At 500 items per second it will reach 30; at 1000 items per second it will reach 60, etc.

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
olegsych: Ok, then why do we need to re-implement `RethrowIfFaulted` then? 😄

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
olegsych: @macrogreg I see that the `TaskTimer.Dispose` method is now `[Obsolete]` but the `TelemetryBuffer`, `MetricManager` and other consumers no longer call it. It also doesn't look like they call `TaskTimer.Cancel` which is supposed to replace it. This is a significant change in the implementation of the `TaskTimer` consumers that may leave background tasks running, indefinitely, if the background task restarts the timer.

src/Core/Managed/Shared/Extensibility/MetricManager.cs
olegsych: With the risk of kicking the dead horse, I disagree. Since it's introduction for deterministic release of unmanaged resources, the disposable pattern has evolved to general purpose lifetime management in modern .NET APIs. Any class with special lifetime considerations can implement the `IDisposable` interface to its advantage. Take the `MetricManager` as an example. The removal of `IDisposable` implementation from `TaskTimer` triggered the removal of its disposal from `MetricManager`. Now the `MetricManager.Dispose` should `Cancel` the `snapshotTimer` instead of disposing it; otherwise the `SnapshotAndReschedule` will continue running indefinitely. If `MetricManager` is no longer an `IDisposable`, it'll still need to be "cancellable", "stoppable", etc. because it still needs a special lifetime management.

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
olegsych: Point taken. I didn't realize the original implementation was coded defensively.

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimer.cs
macrogreg: Thanks for the info.
I do not see any significant concern with this. Even at say, 5K req/sec we would thus get 300 tasks. Let's say this is the <1000 order of magnitude. These are 1000 objects that are already in memory anyway. These tasks will be scheduled for processing, a simple flag will be examined and then no operation will occur (the queue will be cleaned up). Notable is that all this is happening already. The difference is only in when all this occurs (at cancellation time vs when the tie interval expires). We do this now a little later, but we save all the allocations for the cancellation signals.

Normal perf runs will point out if we are missing something, but I do not see any need for any specialized testing. Agreed?

olegsych: My gut feel tells me it _should_ be OK and we can close this comment. I'm still concerned about the increased potential of memory allocations captured by the `Start` lambdas being promoted to the second generation by GC. I'd feel much more comfortable if we can rule this out with perf tests.

SergeyKanzhelev:
Please rebase against `2.4.0-beta1`. Note, I made this class internal. I think it's ok to take this change in

SergeyKanzhelev:
No response here. Closing

PR #379, opened 2016-11-30
Changed the logic for loading the iKey from an environment variable to not override config or code settings.

dnduffy:


Test/CoreSDK.Test/Shared/TelemetryClientTest.cs
```diff
@@ -494,29 +498,42 @@ public void ChannelIsInitializedInTrackWhenTelemetryConfigurationIsConstructedVi
...
-            Environment.SetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY", expectedKey); // Set via env. variable
-            Assert.DoesNotThrow(() => client.TrackTrace("Test Message"));
+            Environment.SetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY", environmentKey); // Set via the environment variable.
 
+            // Set in code method one.
+            client.Context.InstrumentationKey = expectedKey;
+            Assert.DoesNotThrow(() => client.TrackTrace(message));
+            Assert.Equal(expectedKey, sentTelemetry.Context.InstrumentationKey);
+
+            // Set in code method two.
```
SergeyKanzhelev: Please split into two unit tests. Generally we are trying to keep one test for the single scenario/check

Test/CoreSDK.Test/Shared/TelemetryClientTest.cs
```diff
@@ -494,29 +498,42 @@ public void ChannelIsInitializedInTrackWhenTelemetryConfigurationIsConstructedVi
...
+            sentTelemetry = null; // Make sure we got a new instance.
+            client.Context.InstrumentationKey = string.Empty; // Clear method one value.
+            TelemetryConfiguration.Active.InstrumentationKey = expectedKey; // Set in code method two, default config.
+            Assert.DoesNotThrow(() => client.TrackTrace(message));
             Assert.Equal(expectedKey, sentTelemetry.Context.InstrumentationKey);
 
             Environment.SetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY", null);
+            TelemetryConfiguration.Active.InstrumentationKey = oldActiveKey; // Restore state to not impact other tests.
```
SergeyKanzhelev: You called 'ClearActiveTelemetryConfiguration' in the beginning. So you still need to restore keys?

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -379,11 +379,12 @@ public void Initialize(ITelemetry telemetry)
 
             if (string.IsNullOrEmpty(instrumentationKey))
             {
-                instrumentationKey = PlatformSingleton.Current.GetEnvironmentVariable(InstrumentationKeyWebSitesEnvironmentVariable);
-                
-                if (string.IsNullOrEmpty(instrumentationKey))
+                // If no configuration override was passed to the constructor then the default configuration instance will have been used.
+                // If no iKey was present in the config then the environment variable will have been loaded into the default instance.
+                instrumentationKey = this.configuration.InstrumentationKey;
+                if (string.IsNullOrEmpty(instrumentationKey) && this.configuration != TelemetryConfiguration.Active)
```
SergeyKanzhelev: Do we need this logic at all? Do we have test for non-`Active` config and why do we need to use environment variable for it?

SergeyKanzhelev: So the decision is to use the ikey from the `ApplicaitonInsights.config` to take priority? `TelemtryConfigurationFactory` seems to kept this logic. Can you please add a test for this case or just point to it

Test/CoreSDK.Test/Shared/TelemetryClientTest.cs
dnduffy: I did that originally, it was just a lot of copied and pasted code, and the test naming got weirder but I can do it again.

src/Core/Managed/Shared/TelemetryClient.cs
dnduffy: I think we need this logic because if you specify a custom config with no iKey it won't pickup from the environment variable but this way provides a fallback when trying to send telemetry.

dnduffy: I believe that the test which I modified covers this but we can figure out how to make it clearer.

PR #389, opened 2016-12-06
Updated version information for stable release.

dnduffy:


GlobalStaticVersion.props
```diff
@@ -8,12 +8,12 @@
     <SemanticVersionMajor>2</SemanticVersionMajor>
     <SemanticVersionMinor>2</SemanticVersionMinor>
     <SemanticVersionPatch>0</SemanticVersionPatch>
-    <PreReleaseMilestone>beta6</PreReleaseMilestone>
+    <PreReleaseMilestone></PreReleaseMilestone>
```
SergeyKanzhelev: you do not need this change. Just ruin "stable" build and it will not append the postfix

GlobalStaticVersion.props
```diff
@@ -8,12 +8,12 @@
...
     <!-- 
       Date when Semantic Version was changed. 
       Update for every public release.
     -->
-    <SemanticVersionDate>2016-06-02</SemanticVersionDate>
+    <SemanticVersionDate>2016-12-05</SemanticVersionDate>
```
SergeyKanzhelev: this will crew up file versions. Please do not do it before the major or minor version change. Like 2.3 should have this date increased

GlobalStaticVersion.props
dnduffy: You do need to remove the beta6, the only thing a stable build removes is the build number.

SergeyKanzhelev: Ok. I thought otherwise

GlobalStaticVersion.props
SergeyKanzhelev: minor version stays the same. FileVersion will be something like `2.2.<number of 5 minutes intervals since SemanticVersionDate>` So resetting it now will generate file version that is less than for all `-beta` versions before.

dnduffy: Okay, so we want it to still output something like 2.2.0.850, got it.  I always like file versions with .0 on the end but I see how this would conflict with updating from a beta.

Issue #393, opened 2016-12-08
TelemetryClient.Track() invoke sanitize() irrespective of RickPayLoadEventSource is enabled or not


cijothomas:
This code:
// logs rich payload ETW event for any partners to process it
                telemetry.Sanitize();
                RichPayloadEventSource.Log.Process(telemetry);

If RickPayloadEventSource is not enabled (which is often the case), then the effort done in sanitize() is waste of cpu cycles as sanitize() is only needed in RichPayloadEventSource.

PR #394, opened 2016-12-08
Removed sanitize() from TelemetryClient.Track() 

cijothomas:
Removed sanitize() call from Track() method to avoid unnecessary perf hit if RichEventPayloadSource is not enabled. Sanitize() is invoked after checking if the event source is enabled.

Fix for this issue, 
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/393

And may be this issue as well as this could be leading to the high memory (but that will be investigated anyway):
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/227


SergeyKanzhelev:
Code makes sense for me. Would it be better to put it into 2.3.0-beta1 branch to avoid conflicts on merge later?

cijothomas:
It appears tht 2.3.0-beta1 is failing in build, so i dont want to put this in there now. (i am trying to get a perf run asap with my changes, to see if it was the cause of high memory usage)

SergeyKanzhelev:
makes sense

Issue #367, opened 2016-11-18
Mark aggregated metrics with the dedicated sdkVersion


SergeyKanzhelev:
@vitalyf007 - it will be nice to have an sdkVersion set to the special value on metrics aggregated using the new aggregators.

The only gotcha here is that for the modules using those aggregators we need to allow to set their own sdkVersion. Like exception statistics tracking module will have it's own sdkVersion.

SergeyKanzhelev:
Here is how it done for web:

Helper class:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/75373e57dcf8d646c54ee188461c373f2cc98939/Src/Web/Web.Net45/Implementation/SdkVersionUtils.cs

Use of it. In your case you can set it on telemetry item itself:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/75373e57dcf8d646c54ee188461c373f2cc98939/Src/Web/Web.Shared.Net/ExceptionTrackingTelemetryModule.cs#L59

Fallback to the default (for reference):
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Core/Managed/Shared/TelemetryClient.cs#L419

AlexBulankou:
Already done?

PR #396, opened 2016-12-08
Adding property (sdk version) to signify metric was created using metric manager

vitalyf007:
In this PR `MetricManager` sets special value into Sdk Version property of context such that we can tell if metric telemetry was created using `MetricManager`. 
The reason is that we want to know how many people use that approach. SdkVersion 'game' we play here is a standard way to measure exactly these things used by Web Sdk already, so, we're just re-using the approach.


SergeyKanzhelev:
Mentioning issue to correlate with PR: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/367

PR #403, opened 2016-12-13
Adding a tag to all metrics genrated via MetricManager

vitalyf007:
This PR was done in 2.3.0-beta1 branch andnow we're reading to merge to develop. SergeyK signed off on original PR.

### DIFF
--- a/src/Core/Managed/Shared/Extensibility/Implementation/SdkVersionUtils.cs
+++ b/src/Core/Managed/Shared/Extensibility/Implementation/SdkVersionUtils.cs
@@ -1,0 +1,34 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation
+{
+    using System;
+    using System.Globalization;
+    using System.Linq;
+    using System.Reflection;
+
+    internal class SdkVersionUtils
+    {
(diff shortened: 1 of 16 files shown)


src/Core/Managed/Shared/Extensibility/Implementation/SdkVersionUtils.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Builds a string representing file version of the assembly with added prefix
+        /// in format prefix:major.minor-revision.
+        /// </summary>
+        /// <param name="versionPrefix">Prefix to add to version.</param>
+        /// <returns>String representation of the version with prefix added.</returns>
+        internal static string GetSdkVersion(string versionPrefix)
+        {
+#if !CORE_PCL
```
dnduffy: I thought we replaced CORE_PLC with .NET Core?

vitalyf007: This is code I *moved* from TelemetryClient class out so that it can be used in two places. I did not modify the code.

PR #406, opened 2017-01-10
Improved Dispose and telemetry channel handling.

dnduffy:
This is to fix https://github.com/Microsoft/ApplicationInsights-dotnet/issues/364.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ This changelog will be used to generate documentation on [release notes page](ht
 ## Version 2.3.0-beta2
 - Added constructor overloads for TelemetryConfiguration and added creation of a default InMemoryChannel when no channel is specified for a new instance.
   TelemetryClient will no longer create an InMemoryChannel on the configuration instance if TelemetryChannel is null, instead the configuration instances will always have a channel when created.
-- Added an explicit finalizer to TelemetryConfiguration so that hopefully garbage collection will eventually clean up abondoned ITelemetryChannels which developers did not properly dispose of.
+- TelemetryConfiguration will no longer dispose of user provided ITelemetryChannel instances.  Users must properly dispose of any channel instances which they create, the configuration will only auto-dispose of default channel instances it creates when none are specified by the user.
 
 ## Version 2.3.0-beta1
 - Added metric aggregation functionality via MetricManager and Metric classes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 This changelog will be used to generate documentation on [release notes page](http://azure.microsoft.com/en-us/documentation/articles/app-insights-release-notes-dotnet/).
 
+## Version 2.3.0-beta2
+- Added constructor overloads for TelemetryConfiguration and added creation of a default InMemoryChannel when no channel is specified for a new instance.
+  TelemetryClient will no longer create an InMemoryChannel on the configuration instance if TelemetryChannel is null, instead the configuration instances will always have a channel when created.
+- Added an explicit finalizer to TelemetryConfiguration so that hopefully garbage collection will eventually clean up abondoned ITelemetryChannels which developers did not properly dispose of.
+
 ## Version 2.3.0-beta1
 - Added metric aggregation functionality via MetricManager and Metric classes.
 - Exposed a source field on RequestTelemetry. This can be used to store a representation of the component that issued the incoming http request. 
--- a/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
+++ b/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
@@ -114,7 +114,7 @@
                 throw new ArgumentNullException("item");
             }
 
-            if (isDisposed)
+            if (this.isDisposed)
             {
                 CoreEventSource.Log.InMemoryChannelSendCalledAfterBeingDisposed();
                 return;
@@ -145,7 +145,7 @@
         public void Flush(TimeSpan timeout)
         {
             this.transmitter.Flush(timeout);
-            if (isDisposed)
+            if (this.isDisposed)
             {
                 CoreEventSource.Log.InMemoryChannelFlushedAfterBeingDisposed();
             }
@@ -162,14 +162,14 @@
 
         private void Dispose(bool disposing)
         {
-            if (disposing && !isDisposed)
+            if (disposing && !this.isDisposed)
             {
                 if (this.transmitter != null)
                 {
                     this.transmitter.Dispose();
                 }
 
-                isDisposed = true;
+                this.isDisposed = true;
             }
         }
     }
--- a/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
+++ b/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
@@ -10,9 +10,14 @@
     public class InMemoryChannel : ITelemetryChannel
     {
         private readonly TelemetryBuffer buffer;
-        private readonly InMemoryTransmitter transmitter;
+        private InMemoryTransmitter transmitter;
         private bool? developerMode = false;
         private int bufferSize;
+
+        /// <summary>
+        /// Indicates if this instance has been disposed of.
+        /// </summary>
+        private bool isDisposed = false;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="InMemoryChannel" /> class.
@@ -109,6 +114,11 @@
                 throw new ArgumentNullException("item");
             }
 
+            if (isDisposed)
+            {
+                throw new ObjectDisposedException("InMemoryChannel", "The telemetry channel has been disposed.");
+            }
+
             try
             {
                 this.buffer.Enqueue(item);
@@ -133,7 +143,14 @@
         /// <param name="timeout">Timeout interval to abort sending.</param>
         public void Flush(TimeSpan timeout)
         {
-            this.transmitter.Flush(timeout);
+            if (!isDisposed)
+            {
+                this.transmitter.Flush(timeout);
+            }
+            else
+            {
+                throw new ObjectDisposedException("InMemoryChannel", "The telemetry channel has been disposed.");
+            }
         }
 
         /// <summary>
@@ -147,12 +164,15 @@
 
         private void Dispose(bool disposing)
         {
-            if (disposing)
+            if (disposing && !isDisposed)
             {
                 if (this.transmitter != null)
                 {
                     this.transmitter.Dispose();
+                    this.transmitter = null;
                 }
+
+                isDisposed = true;
             }
         }
     }
--- a/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
+++ b/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
@@ -10,7 +10,7 @@
     public class InMemoryChannel : ITelemetryChannel
     {
         private readonly TelemetryBuffer buffer;
-        private InMemoryTransmitter transmitter;
+        private readonly InMemoryTransmitter transmitter;
         private bool? developerMode = false;
         private int bufferSize;
 
--- a/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
+++ b/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
@@ -116,7 +116,8 @@
 
             if (isDisposed)
             {
-                throw new ObjectDisposedException("InMemoryChannel", "The telemetry channel has been disposed.");
+                CoreEventSource.Log.InMemoryChannelSendCalledAfterBeingDisposed();
+                return;
             }
 
             try
@@ -143,13 +144,10 @@
         /// <param name="timeout">Timeout interval to abort sending.</param>
         public void Flush(TimeSpan timeout)
         {
-            if (!isDisposed)
+            this.transmitter.Flush(timeout);
+            if (isDisposed)
             {
-                this.transmitter.Flush(timeout);
-            }
-            else
-            {
-                throw new ObjectDisposedException("InMemoryChannel", "The telemetry channel has been disposed.");
+                CoreEventSource.Log.InMemoryChannelFlushedAfterBeingDisposed();
             }
         }
 
@@ -169,7 +167,6 @@
                 if (this.transmitter != null)
                 {
                     this.transmitter.Dispose();
-                    this.transmitter = null;
                 }
 
                 isDisposed = true;
--- a/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
+++ b/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
@@ -175,12 +175,11 @@
         {
             if (disposing && !this.isDisposed)
             {
+                this.isDisposed = true;
                 if (this.transmitter != null)
                 {
                     this.transmitter.Dispose();
                 }
-
-                this.isDisposed = true;
             }
         }
     }
--- a/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
+++ b/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
@@ -163,6 +163,8 @@ namespace Microsoft.ApplicationInsights.Channel
                     // call Set to to prevent waiting for the next interval. 
                     this.startRunnerEvent.Set();
                 }
+
+                this.Flush(default(TimeSpan));
             }
         }
     }
--- a/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
+++ b/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
@@ -40,7 +40,6 @@ namespace Microsoft.ApplicationInsights.Channel
                 
         internal InMemoryTransmitter(TelemetryBuffer buffer)
         {
-            this.startRunnerEvent = new AutoResetEvent(false);
             this.buffer = buffer;
             this.buffer.OnFull = this.OnBufferFull;
 
@@ -90,13 +89,16 @@ namespace Microsoft.ApplicationInsights.Channel
         /// </summary>
         private void Runner()
         {
-            while (this.enabled)
+            using (this.startRunnerEvent = new AutoResetEvent(false))
             {
-                // Pulling all items from the buffer and sending as one transmissiton.
-                this.DequeueAndSend(timeout: default(TimeSpan)); // when default(TimeSpan) is provided, value is ignored and default timeout of 100 sec is used
+                while (this.enabled)
+                {
+                    // Pulling all items from the buffer and sending as one transmissiton.
+                    this.DequeueAndSend(timeout: default(TimeSpan)); // when default(TimeSpan) is provided, value is ignored and default timeout of 100 sec is used
 
-                // Waiting for the flush delay to elapse
-                this.startRunnerEvent.WaitOne(this.sendingInterval);
+                    // Waiting for the flush delay to elapse
+                    this.startRunnerEvent.WaitOne(this.sendingInterval);
+                }
             }
         }
 
@@ -159,9 +161,16 @@ namespace Microsoft.ApplicationInsights.Channel
 
                 if (this.startRunnerEvent != null)
                 {
-                    // Call Set to prevent waiting for the next interval in the runner, and then Dispose to prevent a race condition with dispose vs. the wait timeout.
-                    this.startRunnerEvent.Set();
-                    this.startRunnerEvent.Dispose();
+                    // Call Set to prevent waiting for the next interval in the runner.
+                    try
+                    {
+                        this.startRunnerEvent.Set();
+                    }
+                    catch (ObjectDisposedException)
+                    {
+                        // We need to try catch the Set call in case the auto-reset event wait interval occurs between setting enabled
+                        // to false and the call to Set then the auto-reset event will have already been disposed by the runner thread.
+                    }
                 }
 
                 this.Flush(default(TimeSpan));
--- a/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
+++ b/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
@@ -40,6 +40,7 @@ namespace Microsoft.ApplicationInsights.Channel
                 
         internal InMemoryTransmitter(TelemetryBuffer buffer)
         {
+            this.startRunnerEvent = new AutoResetEvent(false);
             this.buffer = buffer;
             this.buffer.OnFull = this.OnBufferFull;
 
@@ -89,16 +90,13 @@ namespace Microsoft.ApplicationInsights.Channel
         /// </summary>
         private void Runner()
         {
-            using (this.startRunnerEvent = new AutoResetEvent(false))
+            while (this.enabled)
             {
-                while (this.enabled)
-                {
-                    // Pulling all items from the buffer and sending as one transmissiton.
-                    this.DequeueAndSend(timeout: default(TimeSpan)); // when default(TimeSpan) is provided, value is ignored and default timeout of 100 sec is used
+                // Pulling all items from the buffer and sending as one transmissiton.
+                this.DequeueAndSend(timeout: default(TimeSpan)); // when default(TimeSpan) is provided, value is ignored and default timeout of 100 sec is used
 
-                    // Waiting for the flush delay to elapse
-                    this.startRunnerEvent.WaitOne(this.sendingInterval);
-                }
+                // Waiting for the flush delay to elapse
+                this.startRunnerEvent.WaitOne(this.sendingInterval);
             }
         }
 
@@ -156,13 +154,14 @@ namespace Microsoft.ApplicationInsights.Channel
         {
             if (Interlocked.Increment(ref this.disposeCount) == 1)
             {
-                // Stops the runner.
+                // Stops the runner loop.
                 this.enabled = false;
 
                 if (this.startRunnerEvent != null)
                 {
-                    // call Set to to prevent waiting for the next interval. 
+                    // Call Set to prevent waiting for the next interval in the runner, and then Dispose to prevent a race condition with dispose vs. the wait timeout.
                     this.startRunnerEvent.Set();
+                    this.startRunnerEvent.Dispose();
                 }
 
                 this.Flush(default(TimeSpan));
--- a/src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
+++ b/src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
@@ -3,12 +3,12 @@
     using System;
     using System.Collections.Generic;
     using System.Collections.ObjectModel;
+    using System.ComponentModel;
     using System.Threading;
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.DataContracts;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
     using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
-    using System.ComponentModel;
 
     /// <summary>
     /// Encapsulates the global telemetry configuration typically loaded from the ApplicationInsights.config file.
--- a/src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
+++ b/src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
@@ -55,6 +55,7 @@
         /// <param name="instrumentationKey">The instrumentation key this configuration instance will provide.</param>
         public TelemetryConfiguration(string instrumentationKey) : this(instrumentationKey, new InMemoryChannel())
         {
+            this.shouldDisposeChannel = true;
         }
 
         /// <summary>
@@ -240,7 +241,7 @@
                 if (oldChannel != null && oldChannel != value && this.shouldDisposeChannel)
                 {
                     oldChannel.Dispose();
-                    this.shouldDisposeChannel = false; // The new onw wasn't created by us so it should be managed by whoever created it.
+                    this.shouldDisposeChannel = false; // The new one wasn't created by us so it should be managed by whoever created it.
                 }
             }
         }
--- a/src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
+++ b/src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
@@ -85,16 +85,6 @@
         }
 
         /// <summary>
-        /// Finalizes an instance of the TelemetryConfiguration class.
-        /// Frees resources when a TelemetryConfiguration instance is garbage collected if it was not properly disposed of.
-        /// </summary>
-        ~TelemetryConfiguration()
-        {
-            // Cleanup managed resources if the instance was not properly disposed of.
-            this.Dispose(true);
-        }
-
-        /// <summary>
         /// Gets the active <see cref="TelemetryConfiguration"/> instance loaded from the ApplicationInsights.config file. 
         /// If the configuration file does not exist, the active configuration instance is initialized with minimum defaults 
         /// needed to send telemetry to Application Insights.
@@ -336,17 +326,16 @@
                 this.isDisposed = true;
                 Interlocked.CompareExchange(ref active, null, this);
 
-                ITelemetryChannel telemetryChannel = this.TelemetryChannel;
-                if (telemetryChannel != null)
+                if (this.telemetryChannel != null)
                 {
-                    telemetryChannel.Dispose();
-                    telemetryChannel = null;
+                    this.telemetryChannel.Dispose();
+                    this.telemetryChannel = null;
                 }
 
-                TelemetryProcessorChain processorChain = this.telemetryProcessorChain;
-                if (processorChain != null)
+                if (this.telemetryProcessorChain != null)
                 {
-                    processorChain.Dispose();
+                    // Not setting this.telemetryProcessorChain to null because calls to the property getter would reinitialize it.
+                    this.telemetryProcessorChain.Dispose();
                 }
             }
         }
--- a/src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
+++ b/src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
@@ -22,11 +22,77 @@
         private static TelemetryConfiguration active;
 
         private readonly SnapshottingList<ITelemetryInitializer> telemetryInitializers = new SnapshottingList<ITelemetryInitializer>();
+        private ITelemetryChannel telemetryChannel = null;
         private TelemetryProcessorChain telemetryProcessorChain;
         private string instrumentationKey = string.Empty;
         private bool disableTelemetry = false;
         private TelemetryProcessorChainBuilder builder;
         private SnapshottingList<IMetricProcessor> metricProcessors = new SnapshottingList<IMetricProcessor>();
+
+        /// <summary>
+        /// Indicates if this instance has been disposed of.
+        /// </summary>
+        private bool isDisposed = false;
+
+        /// <summary>
+        /// Initializes a new instance of the TelemetryConfiguration class.
+        /// </summary>
+        public TelemetryConfiguration() : this(new InMemoryChannel())
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the TelemetryConfiguration class.
+        /// </summary>
+        /// <param name="instrumentationKey">The instrumentation key this configuration instance will provide.</param>
+        public TelemetryConfiguration(string instrumentationKey) : this(instrumentationKey, new InMemoryChannel())
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the TelemetryConfiguration class.
+        /// </summary>
+        /// <param name="channel">The telemetry channel to provide with this configuration instance.</param>
+        public TelemetryConfiguration(ITelemetryChannel channel)
+        {
+            if (channel == null)
+            {
+                throw new ArgumentNullException("channel");
+            }
+
+            this.telemetryChannel = channel;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the TelemetryConfiguration class.
+        /// </summary>
+        /// <param name="instrumentationKey">The instrumentation key this configuration instance will provide.</param>
+        /// <param name="channel">The telemetry channel to provide with this configuration instance.</param>
+        public TelemetryConfiguration(string instrumentationKey, ITelemetryChannel channel)
+        {
+            if (channel == null)
+            {
+                throw new ArgumentNullException("channel");
+            }
+
+            if (instrumentationKey == null)
+            {
+                throw new ArgumentNullException("instrumentationKey");
+            }
+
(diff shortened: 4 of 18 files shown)


src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -17,14 +17,19 @@
     /// <summary>
     /// Send events, metrics and other telemetry to the Application Insights service.
     /// </summary>
-    public sealed class TelemetryClient
+    public sealed class TelemetryClient : IDisposable
```
SergeyKanzhelev: Please do not make it disposable. It was intentionally not made disposable to avoid making every customer having the reference to telemetry client make a containing class disposable. 

@olegsych and @vitalyf007 for opinion

olegsych: I don't think appropriate use of the `IDisposable` interface is a problem. But why does the `TelemetryClient` need it now? @dnduffy Could you expand the description of this pull request to help us understand [what is the main change here and why is it made](http://chris.beams.io/posts/git-commit/#why-not-how)?

dnduffy: The problem is with the fact that there is/was no control or cleanup of channels created on the configuration.  Users can define the channel at any time, replace an existing channel, or leave it up to the telemetry client to provide a default.  So when someone creates multiple new telemetry clients, and provides custom config each time, a bunch of non-cleaned up scenarios become possible.

olegsych: I don't think `TelemetryClient` should be managing the `TelemetryConfiguration.TelemetryChannel` or the `TelemetryConfiguration`. It doesn't own these objects; they are given to the `TelemetryClient` constructor by their owner. Disposing `TelemetryChannel` should be done by its proper owner; in case of #364 I believe its the user code. 

#364 is caused by a user error. The proper behavior for `TelemetryChannel(TelemetryConfiguration)` would be to throw an `ArgumentException` when `TelemetryConfiguration.TelemetryChannel` is null. This should never be the case by default because `TelemetryConfiguration.Active.TelemetryChannel` is automatically initialized by the `TelemetryConfigurationFactory`. 

This may be reopening the old debate regarding whether an instrumentation API should throw exception when user messes up or try to handle these mistakes defensively in the SDK. I think this bug, the amount of effort required to keep the code running in face of user errors and the resulting API impact are a strong argument in favor of exceptions.

dnduffy: I agree 100%.
TelemetryChannel should not be creating or managing the channel, but currently it does and we will potentially break customers if we start throwing exceptions now (not that I'm opposed to that if others are okay with it).
I will happily change it if Sergey agrees because that will simplify and remove a lot of this problem and fix #364.
I 100% want to throw exceptions instead of trying to work around user mistakes, what it does right now is just open to mistakes and/or abuse.

dnduffy:
@SergeyKanzhelev Changes made, please look again.

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
```diff
@@ -147,12 +164,14 @@ public void Dispose()
...
         private void Dispose(bool disposing)
         {
-            if (disposing)
+            if (disposing && !isDisposed)
             {
                 if (this.transmitter != null)
                 {
                     this.transmitter.Dispose();
                 }
+
```
SergeyKanzhelev: I'd also suggest `this.transmitter = null`

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -22,13 +22,85 @@ public sealed class TelemetryConfiguration : IDisposable
...
+        /// </summary>
+        public TelemetryConfiguration() : this(new InMemoryChannel())
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the TelemetryConfiguration class.
+        /// </summary>
+        /// <param name="instrumentationKey">The instrumentation key this configuration instance will provide.</param>
+        public TelemetryConfiguration(string instrumentationKey) : this(new InMemoryChannel())
```
SergeyKanzhelev: why not `this(instrumentationKey, new InMemoryChannel())`?

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -44,6 +45,12 @@ public TelemetryClient(TelemetryConfiguration configuration)
             }
 
             this.configuration = configuration;
+
+            // If someone created configuration from scratch and forgot to initialize the channel then use the default.
```
cijothomas: This comment and code doesn't agree with each other. From the comment what i understand is if channel is null, default is used, but in code we throw Ex if channel is null.

cijothomas: i see that this comment was from earlier code - please modify the comment as well.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -173,7 +245,26 @@ internal set
...
+            set
+            {
+                ITelemetryChannel oldChannel = this.telemetryChannel;
+                this.telemetryChannel = value;
+
+                // If we have a previously assigned channel which is not the same one as the "new" value
+                // passed in then we need to dispose of the old channel to keep from leaking resources.
+                if (oldChannel != null && oldChannel != value)
+                {
+                    oldChannel.Dispose();
```
SergeyKanzhelev: so you dispose it even if it was passed to us. Just checking - with ASP.NET Core DI we will register those two as singleton. Will DI call dispose on both of them?

I thought you planned to only dispose channels that was created by this configuration itself

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -44,6 +45,12 @@ public TelemetryClient(TelemetryConfiguration configuration)
...
+            if (this.configuration.TelemetryChannel == null)
+            {
+                throw new ArgumentException("The specified configuration does not have a telemetry channel.", "configuration");
```
SergeyKanzhelev: no, our policy is not to throw exceptions from API.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -22,13 +22,85 @@ public sealed class TelemetryConfiguration : IDisposable
...
+        }
+
+        /// <summary>
+        /// Finalizes an instance of the TelemetryConfiguration class.
+        /// Frees resources when a TelemetryConfiguration instance is garbage collected if it was not properly disposed of.
+        /// </summary>
+        ~TelemetryConfiguration()
+        {
+            // Cleanup managed resources if the instance was not properly disposed of.
+            this.Dispose(true);
```
olegsych: - Why does `TelemetryConfiguration` define a finalizer? It doesn't hold any unmanaged resources.
- If it's really required, the implementation should call `this.Dispose(false)` because the object is being finalized, not disposed.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
olegsych: + 1.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -22,13 +22,85 @@ public sealed class TelemetryConfiguration : IDisposable
...
+            }
+
+            this.instrumentationKey = instrumentationKey;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the TelemetryConfiguration class.
+        /// </summary>
+        /// <param name="channel">The telemetry channel to provide with this configuration instance.</param>
+        public TelemetryConfiguration(ITelemetryChannel channel)
```
olegsych: What is the purpose of the new `TelemetryConfiguration` constructors? The default constructor is used by the default initialization logic, the constructors taking the ikey communicate to users that it's required, but why do we need the 
the `TelemetryConfiguration(ITelemetryChannel)`? It doesn't create a valid object - the ikey is missing.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -240,16 +331,22 @@ public void Dispose()
...
+            if (!this.isDisposed && disposing)
             {
+                this.isDisposed = true;
                 Interlocked.CompareExchange(ref active, null, this);
 
                 ITelemetryChannel telemetryChannel = this.TelemetryChannel;
                 if (telemetryChannel != null)
                 {
                     telemetryChannel.Dispose();
+                    telemetryChannel = null;
```
olegsych: Was this intended to clear the `TelemetryChannel` property?

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -22,13 +22,85 @@ public sealed class TelemetryConfiguration : IDisposable
...
+        private ITelemetryChannel telemetryChannel = null;
         private TelemetryProcessorChain telemetryProcessorChain;
         private string instrumentationKey = string.Empty;
         private bool disableTelemetry = false;
         private TelemetryProcessorChainBuilder builder;
         private SnapshottingList<IMetricProcessor> metricProcessors = new SnapshottingList<IMetricProcessor>();
 
         /// <summary>
+        /// Indicates if this instance has been disposed of.
+        /// </summary>
```
olegsych: Are we doc-commenting private members? 😱 Doesn't the field name already say this?

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
olegsych: This is problematic, I realize `TelemetryConfiguration.Dispose` was already disposing the `TelemetryChannel`, however it doesn't own these objects - it receives them through property and constructor injection, so it shouldn't be disposing them. Combined with the `ObjectDisposedException` now thrown by the `InMemoryChannel` this is a significant change in behavior, unexpected from a well-behaved .NET API.

src/Core/Managed/Shared/TelemetryClient.cs
olegsych: Would our customers prefer an `ArgumentException` with an actionable message here or a `NullReferenceException`, deep in the call to the `TelemetryClient.Track(ITelemetry)?

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
```diff
@@ -133,7 +143,14 @@ public void Flush()
...
         public void Flush(TimeSpan timeout)
         {
-            this.transmitter.Flush(timeout);
+            if (!isDisposed)
+            {
+                this.transmitter.Flush(timeout);
+            }
+            else
+            {
+                throw new ObjectDisposedException("InMemoryChannel", "The telemetry channel has been disposed.");
```
olegsych: Because this is public API, I think we need tests for exception throwing logic.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -22,13 +22,85 @@ public sealed class TelemetryConfiguration : IDisposable
...
+        private bool isDisposed = false;
+
+        /// <summary>
+        /// Initializes a new instance of the TelemetryConfiguration class.
+        /// </summary>
+        public TelemetryConfiguration() : this(new InMemoryChannel())
```
olegsych: Do the new constructors have unit tests?

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
olegsych: Shouldn't the `InMemoryTransmitter` decide what to do when its `Flush` method is called after `Dispose`?

src/Core/Managed/Shared/TelemetryClient.cs
SergeyKanzhelev: Neither should throw

SergeyKanzhelev:
@dnduffy can you please minimize the fix by just mark channel as self-created when it is constructed by our internal code. Either from `TelemetryClient` or from Factory. Dispose channel only if self created in `get`-ter and `Dispose` method. I think this limited change will address the problem without "patching" of the design.

Bigger design change can be done with the upcoming features like multi-channel support

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: I was trying to be friendly to the garbage collector, it may not really be necessary these days but it was a common pattern to null references after disposing.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: Oops, I added that overload after this one and didn't update.

src/Core/Managed/Shared/TelemetryClient.cs
dnduffy: The only way this can throw is if someone explicitly takes steps to create a configuration with the channel set to null (i.e. they make a config instance and then set the channel to null).  No existing code for any customer should be doing that since it was unnecessary previously and makes no sense anyway and all existing code that relied on the default behavior of a channel being created for them will keep working.  Where the channel was created has changed but the fact that it is auto-created has not.  I explicitly made it keep that behavior per Sergey's request.  Throwing now does not break the API, it will not throw for any existing customer or scenario.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: I realize this is a misuse of the pattern, this was the only workaround to the problem Sergey and I could come to an agreement on that didn't potentially break someone.  For V3.0 we need to rework how a lot of this stuff works and lock it down properly.
Currently we have a problem that the existing architecture exposes where a user can create multiple TelemetryClient instances with new configuration and each of them was in turn creating a new InMemoryChannel with InMemoryTransmitter that spawned a new thread and none of those threads were getting garbage collected.  If developers fail to properly dispose of these disposable objects the garbage collector won't be able to clean up the InMemoryTransmitters and threads because of the existing references.  Hacking around the problem by adding a finalizer to the config to force cleanup was what we came up with.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: Missing an IKey is allowed...

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: One of them does directly, the others do indirectly through existing tests that now use them.  It's not explicit but the coverage exists.

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
dnduffy: We know this object is disposed and it is it's flush that was called.  I could put it in the transmitter also but I didn't.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: For DI it will give us the same instance over and over if it was registered as a singleton.
If it was not registered as a singleton then it will give a new instance each time.

src/Core/Managed/Shared/TelemetryClient.cs
dnduffy: P.S. @SergeyKanzhelev I have preserved the behavior of the scenario's we discussed, this throw will not happen for any existing customer but it does help improve for future if someone explicitly tries the following silly code:
```
var conf = new TelemetryConfigration();
conf.TelemetryChannel = null;
var client = new TelemetryClient(conf);
```

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: Personally I would doc comment everything (I haven't had time to do it retroactively) and make all variables names as descriptive as possible even if it seems redundant to have both.

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
dnduffy: Okay, I did that originally but I had to make transmitter not be readonly and changed my mind, I'll put it back.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
SergeyKanzhelev: why to have redundant docs?

dnduffy: It makes much nicer docs to read when you generate them.
And for some learning types it helps when you are new to the code and trying to figure out what things are.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: FYI: Our SDK for .NET Core registers everything as singletons.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
SergeyKanzhelev: @olegsych is right - we cannot access managed objects in finalizers. Seems that what we discussed was wrong. 

@dnduffy did you validated whether our channels will prevent themselves from the garbage collection by the reference from a timer? In any case disposing of self-created channels should address this issue I think.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: I would be happy to add an optional boolean parameter, defaulted to false, the way StreamReader provides "leaveOpen" for the Stream it wraps in newer versions of .NET.
The current behavior was already to dispose of the channel whether it was passed in or not, and this is consistent with how StreamReader closes (which disposes) the Stream it uses.
So I don't believe we should change that behavior at this moment in time.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: I agree it is an abuse, but "should not" != "can not".
We already discussed that the right way might break some customers which you are opposed to.
So we're left with either, don't fix the bug, or hack around the design limitations.

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
olegsych: I think this logic should live in the `InMemoryTransmitter`; the `InMemoryChannel` only assumes that `InMemoryTransitter.Flush` is invalid after `InMemoryTransmitter.Dispose` but it's the responsibility of the `InMemoryTransmitter` to enforce that.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
olegsych: I feel that this particular comment is noise that only makes reading the code harder. Comments quickly become obsolete and misleading. It's better for the code to be [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and [self-documenting](https://en.wikipedia.org/wiki/Self-documenting_code) first; and doc-commented only as needed. ReSharper and Ghost Doc make creating useless comments too easy and StyleCop actually encourages the bad practice! 😱 Does a doc comment saying a constructor initializes a new instance of a class have any value?

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
olegsych: We should have tests for all public APIs, including tests for exception throwing logic. Without them it will be hard to tell if a particular change breaks backward compatibility and requires a major version release as opposed to a minor release or a patch.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
olegsych: Thanks for the explanation - it helps. I don't know if it still holds true, but a while back, a missing `InstrumentationKey` was a common source of missing telemetry. As a user of this API, I'd say that indicating that the instrumentation key is optional is misleading. Unless you have strong customer evidence that says otherwise, I'd suggest dropping the `TelemetryConfiguration(ITelemetryChannel)` constructor for now.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
olegsych: If the `InMemoryChannel`  or `InMemoryTransmitter` have to be disposed, shouldn't _they_ be finalizable? How does the `TelemetryConfiguration` know that disposing _all possible_ `ITelemetryChannel` implementations is safe during finalization?

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
olegsych: This code clears the local _variable_ called `telemetryChannel`, not the `telemetryChannel` field of the `TelemetryConfiguration`.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
olegsych: I was afraid `StreamReader` would come up 😭 It's such a horrible anti-pattern. 

I'm a bit rusty on lifetime management of singletons in the SDK. 
- Could you elaborate on .NET core registering everything as singletons? Does it mean that the service container disposes them at application shutdown?
- How does `TelemetryConfiguration.Active` get disposed on other .NET platforms?

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
SergeyKanzhelev: in this case it is "can not" as Oleg said as we cannot rely on GC ordering and the managed objects may be in the process of garbage collecting while we try to call Dispose on them.

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
dnduffy: While the logic could live somewhere it would not change the end result here; this object has been disposed, the fact that it happens to call transmitter.Flush inside it's flush method is not relevant.  What is relevant is that this object has been disposed and calling anything on it once it has been disposed of should throw the ObjectDisposedException.
Even if we added the check to the transmitter, which users can't access anyway, it would still get thrown when this method is called after this object has been disposed of because this channel disposes of its private transmitter instance when the channel is disposed of.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: I agree that having no iKey seems silly, but this is an explicitly requested scenario from Visual Studio.  VS wants to have telemetry flowing in VS in debug mode to the console before customers have actually onboarded to AI "for real" and received a valid iKey.  It's existing behavior that is heavily relied on which we can't change; we have code ourselves which creates configuration instances and only sets the channel but nothing else so all I've done is add an explicit constructor to clarify the existing API usage.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: Per our discussion yesterday I'm removing this finalizer hack and it will be rightly up to the customer not to leak objects by not disposing of them.  Nothing would have broken but since we all agree it's the wrong "solution" anyway this isn't relevant.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
olegsych: Just to clarify, I'm only suggesting to remove the `TelemetryConfiguration(ITelemetryChannel)` constructor added in this pull request. Visual Studio doesn't depend on the presence of this constructor, correct? By removing this constructor we keep Visual Studio integration working and avoid confusing users further by implying that ikey is optional.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: I agree there are useless comments generated by ReSharper (among other bad things it does) but StyleCop is supposed to get people to write good comments, unfortunately it gets too picky about constructor wordings.  I also agree that code should be named clearly so it is "self documenting".  However, I really hate it when public code like the .NET framework has no comments and you can't get any documentation from MSDN on a class or method.

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
olegsych: I disagree that where program logic is implemented is not relevant because I think this design incorrectly places responsibility of one class in another. I also don't agree that object use after disposal as a clear user error - it's an API usage scenario that needs to be considered carefully. 

As an application developer, I may need to track telemetry during application shutdown in code don't understand or control, such as in a legacy system or a system where object disposal is managed by a framework I don't understand. If I'm investigating a crash or a hang at shutdown of my application, I might want to explicitly flush the telemetry channel explicitly without knowing it was already disposed. I'd be frustrated when this attempt resulted in an `ObjectDisposedException` hiding my original problem and leaving the telemetry unsent. In this scenario, I'd want a forgiving API that helps me solve my problem instead of adding insult to injury by saying I flushed at wrong time.

Could you take a look at implementation of the `InMemoryTransmitter.Flush`? I may be mistaken, but it looks safe to invoke after `InMemoryTransmitter.Dispose`. If that's the case, does the `InMemoryChannel.Flush` really need to throw the `ObjectDisposedException`?

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
olegsych: 👍

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: You are correct, Singletons, and the Active telemetry config instance, will get disposed on app shutdown.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: VS doesn't need the constructor, I was commenting on the requirement that iKey not be required.
I want the constructor so the API is clearer about the treatment of channels.
I'm open to suggestions about how to communicate iKey requirements and behavior more clearly.
If we had it to do all over again we could have a "special" iKey value for the not configured scenario so that string.Empty wouldn't be allowed but we don't really have that option in the 2.x version.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: Oops, good catch, sorry I missed this earlier.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: I've made some extra explicit parameter validate tests.  There is some test overlap so it's not pure TDD style but everything is covered at least.

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
dnduffy: I don't think we're looking at this the same way.  Architecturally, as far as InMemoryChannel is concerned, when it is disposed it should no longer be used because all of its resources should have been released and the instance should be invalid now and free for garbage collection.  The fact that it's internal implementation uses an InMemoryTransmitter instance which has a Flush method has no bearing on whether or not the InMemoryChannel instance was disposed.

But let's walk through your real world scenario:
- The user somehow gets the channel disposed, either they did it on purpose or it happened "accidentally".
- They call Flush on the channel and we'll pretend it does not check for isDisposed and it in turn calls into the transmitter to flush but the transmitter has also been disposed.
- The transmitter should throw some kind of exception but currently it may not.  At this point you expect the user to use an object which the .NET framework should have been free to reuse the memory from.  The garbage collector may not have collected it yet if the user is holding a reference to it somehow but it is also not considered a valid object state for use anymore.

This did make me find another issue, when the transmitter is disposed it should flush but instead any remaining telemetry is just quietly lost.  The existing code for the InMemoryTransmitter didn't dispose cleanly but the .NET framework design does not allow/support a user continuing to use a disposed object.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
olegsych: In that case:
1. Keep the `TelemetryConfiguration(string, ITelemetryChannel)` constructor because it communicates that both ikey and channel are required for a valid configuration.
2. Keep the `TelemetryConfiguration(string)` constructor because it communicates that ikey is required and creates a default channel.
3. Keep the `TelemetryConfiguration()` constructor because it is required for backward compatibility. Consider applying `[EditorBrowseable(Never)]` attribute to discourage users from picking it by default.
4. Remove the `TelemetryConfiguration(ITelemetryChannel)` because it incorrectly implies that the ikey is optional for any channel.

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
olegsych: `Dispose` is just a method, how it's implemented is a design choice that needs to be tailored to a specific scenario. Not all classes in the .NET framework throw `ObjectDisposedException` and the `InMemoryChannel.Flush` shouldn't either. The only reason it's required in this PR is because the `InMemoryChannel.Dispose` now clears the `InMemoryChannel.transmitter` field. 

This is a breaking change that doesn't add user value. Please undo the `InMemoryChannel` changes.

dnduffy: It sounds to me like what you are describing is that none of these classes should be disposable.
But they are already disposable.
I've fixed the issue with lost telemetry occurring because flush was not called on dispose.
The existing code behavior already stopped the runner thread when dispose was called and anything sent after that was just left in memory.  You are correct that a user could have kept using the object after disposing of it but that "scenario" is wrong in my opinion.
I object to the idea that a disposable object, which has been disposed of, should keep working after it was disposed of.  
Maybe some user will post an issue on a forum complaining that they disposed of the object and then got an exception when they tried to keep using it but at that point the whole .NET community should be expected to reply that this is expected.
I don't believe we should hide user bugs, we don't provide any value to the user by silently ignoring their mistakes and pretending their code is fine when it isn't.

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
dnduffy: I will annotate 3 as you describe, it's probably only us in unit tests that actually use 4 so I can change our unit tests to pass string.Empty instead.

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
olegsych: @dnduffy Thanks for fixing the problem in the `InMemoryTransmitter`.
@SergeyKanzhelev I think we need you to make the decision regarding changes in the `InMemoryChannel`. I still maintain that this change should not be made because it
- Is not related to the bug #364 this pull request is meant to address
- Breaks the exception behavior contract established by the previous implementation of the `InMemoryChannel`
- Can cause user data loss by throwing unnecessary `ObjectDisposedException` when `Flush` is called for an `InMemoryChannel` already disposed by an application framework without user knowledge. 
- Unnecessarily couples `InMemoryChannel` to `InMemoryTransmitter` by over-specifying assumptions of how one class should work in another

dnduffy: I've discussed with Sergey and I'll take out the throw under protest.  I maintain that a disposed object is not valid for continued use and even in the pre-existing behavior any data a user tried to add after it was disposed would be lost.  Any framework that could create such a situation would have quite a bug and we'll just be hiding it.

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
```diff
@@ -10,11 +10,16 @@
     public class InMemoryChannel : ITelemetryChannel
     {
         private readonly TelemetryBuffer buffer;
-        private readonly InMemoryTransmitter transmitter;
+        private InMemoryTransmitter transmitter;
```
olegsych: Could you undo this change and make `transmitter` readonly again?

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
```diff
@@ -147,12 +162,14 @@ public void Dispose()
 
         private void Dispose(bool disposing)
         {
-            if (disposing)
+            if (disposing && !isDisposed)
```
olegsych: Could you undo this change and remove `isDisposed` from the `if` statement?

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
dnduffy: Yes, originally Sergey agreed with nulling it out in Dispose but since that is gone I can make it readonly again.

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
dnduffy: What would be the point of that?
Calling Dispose over and over again is supposed to be safe, part of the pattern is to check if it has already been called and ignore future calls.  Since the transmitter has already been disposed of as well it accomplishes nothing to call it again either.

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -44,6 +45,11 @@ public TelemetryClient(TelemetryConfiguration configuration)
             }
 
             this.configuration = configuration;
+
+            if (this.configuration.TelemetryChannel == null)
```
SergeyKanzhelev: why do we need this check at all? It wouldn't protect from failure in `Track` if customer set it to `null` later.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 This changelog will be used to generate documentation on [release notes page](http://azure.microsoft.com/en-us/documentation/articles/app-insights-release-notes-dotnet/).
 
+## Version 2.3.0-beta2
+- Added constructor overloads for TelemetryConfiguration and added creation of a default InMemoryChannel when no channel is specified for a new instance.
+  TelemetryClient will no longer create an InMemoryChannel on the configuration instance if TelemetryChannel is null, instead the configuration instances will always have a channel when created.
+- Added an explicit finalizer to TelemetryConfiguration so that hopefully garbage collection will eventually clean up abondoned ITelemetryChannels which developers did not properly dispose of.
```
SergeyKanzhelev: this is not actual any longer

dnduffy: Good catch, I meant to update the changelog with the information about not disposing of channels that are passed in.

src/Core/Managed/Shared/TelemetryClient.cs
dnduffy: Since the client no longer creates one if it is not provided I thought it might be nicer to have a meaningful message rather than a NullReferenceException.

CHANGELOG.md
cijothomas: abondoned - typo

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -173,7 +223,27 @@ internal set
...
+            {
+                ITelemetryChannel oldChannel = this.telemetryChannel;
+                this.telemetryChannel = value;
+
+                // If we have a previously assigned channel which was created by us and is not the same one as the
+                // "new" value passed in then we need to dispose of the old channel to keep from leaking resources.
+                if (oldChannel != null && oldChannel != value && this.shouldDisposeChannel)
+                {
+                    oldChannel.Dispose();
+                    this.shouldDisposeChannel = false; // The new onw wasn't created by us so it should be managed by whoever created it.
```
cijothomas: onw - typo

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -22,13 +23,62 @@ public sealed class TelemetryConfiguration : IDisposable
...
+        {
+            this.shouldDisposeChannel = true;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the TelemetryConfiguration class.
+        /// </summary>
+        /// <param name="instrumentationKey">The instrumentation key this configuration instance will provide.</param>
+        public TelemetryConfiguration(string instrumentationKey) : this(instrumentationKey, new InMemoryChannel())
+        {
```
cijothomas: This constructor is creating the new InMemoryChannel() - so it should set shouldDIsposeChannel = true right?

dnduffy: Good catch, I need to add more unit tests since this behavior changed in the last round of changes.

cijothomas: Please do add unit test coverage, as that's our only defense in base SDK :)

src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
```diff
@@ -156,13 +154,14 @@ private void Dispose(bool disposing)
...
-                // Stops the runner.
+                // Stops the runner loop.
                 this.enabled = false;
 
                 if (this.startRunnerEvent != null)
                 {
-                    // call Set to to prevent waiting for the next interval. 
+                    // Call Set to prevent waiting for the next interval in the runner, and then Dispose to prevent a race condition with dispose vs. the wait timeout.
                     this.startRunnerEvent.Set();
+                    this.startRunnerEvent.Dispose();
```
cijothomas: this is still risky... because if this Dispose() is called before the Runner gets a chance to finish WaitOne, then WaitOne() call in Runner can get ObjectDisposedException.

cijothomas: On heavyload systems ThreadPool starvation can cause Runner to not get a chance to immediately run even if startRunnerEvent is set, so disposing startRunnerEvent here is risky.

dnduffy: Less likely but theoretically possible, I looked through the WaitHandle code and it will throw if WaitOne is called after it was disposed so I'll go back to the try catch approach in the dispose method we discussed.

PR #408, opened 2017-01-12
Add ability to load Enum-valued configuration properties

karolz-ms:
This is necessary in the context of EventSource telemetry module work, but is universally applicable

### DIFF
--- a/Test/CoreSDK.Test/Shared/Extensibility/TelemetryConfigurationFactoryTest.cs
+++ b/Test/CoreSDK.Test/Shared/Extensibility/TelemetryConfigurationFactoryTest.cs
@@ -295,6 +295,32 @@
         }
 
         [TestMethod]
+        public void LoadInstanceHandlesEnumPropertiesWithNumericValue()
+        {
+            var definition = new XElement(
+                "Definition",
+                new XElement("EnumProperty", "3"));
+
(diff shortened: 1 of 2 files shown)


Test/CoreSDK.Test/Shared/Extensibility/TelemetryConfigurationFactoryTest.cs
```diff
@@ -295,6 +295,32 @@ public void LoadInstanceInitializesGivenInstanceAndDoesNotRequireSpecifyingTypeA
...
 
         [TestMethod]
+        public void LoadInstanceHandlesEnumPropertiesWithNumericValue()
+        {
+            var definition = new XElement(
+                "Definition",
+                new XElement("EnumProperty", "3"));
+
+            var original = new StubClassWithProperties();
+            object instance = TestableTelemetryConfigurationFactory.LoadInstance(definition, typeof(StubClassWithProperties), original, null);
```
dnduffy: I personally dislike using var for anything except anonymous types and in this case with object I still don't know what the type or interface is that is returned from this factory method.

karolz-ms: @dnduffy I hear you but I just followed the style of many tests in this file

Issue #345, opened 2016-11-01
TrackTrace without flush causes an OOM exception


haniamr:
Version: "Microsoft.ApplicationInsights" 2.1.0.
Repro of the issue:
```
TelemetryConfiguration config = TelemetryConfiguration.CreateDefault();
config.InstrumentationKey = "{KEY}";

TelemetryClient client = new TelemetryClient(config);
var dictionary = new Dictionary<string, string>()
            {
                {"testprop", "testval" },
                {"testprop2", "testval2" },
                {"testprop3", "testval3" },
            };

Parallel.For(0, 10000000, (i) =>
            {
                client.TrackTrace("test " + i, SeverityLevel.Information, dictionary);
            });
```

This code will throw an out of memory exception when running in a 32-bit process, and eventually consume up to 21GB+ of RAM when running in a 64-bit process.

Here's a snapshot of the memory usage when profiling:
![image](https://cloud.githubusercontent.com/assets/5480153/19888858/3fa46884-a029-11e6-8e2c-41bca7a0f839.png)

When I call Flush(), the issue is resolved, but shouldn't the SDK flush the buffer automatically based on some criteria as mentioned in the documentation? https://azure.microsoft.com/en-us/documentation/articles/app-insights-api-custom-events-metrics/#_flushing-data

cijothomas:
Took a look at InMemoryChannel - we already have logic to flush events once it reaches buffer capacity which is 500 items by default. There is a flush every 30 secs as well.
The above repro is sending 10million events and this is more than backend can accept, and can likely cause OOM as channel cannot dispatch that many events.

dnduffy:
[...]

Issue #413, opened 2017-01-20
Channel Improvement to guarantee items are sent when buffer is full on high load systems


cijothomas:
Both InMemoryChannel and ServerTelemetryChannel has a OnBufferFull event, which attempts to clear the buffer by serializing and sending telemetry. However, as this is scheduled to run on a ThreadPool thread, there is no guarantee that the buffer will be drained. On a high load system (too many workitems queued up in TP), its possible that the buffer is not drained frequently enough, and can grow without limits.
A short term fix will be to impose a hard limit of the maximum items in backlog after which items will be simply dropped. This fix maybe sufficient for InMemoryChannel.
For ServerTelemetryChannel where guaranteed delivery is more important, we may need to create a separate thread to handle draining of buffer.

cijothomas:
The original issue:
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/345

cijothomas:
I dont think there exists any option to configure the inner working on the channel.
Did you see an error message like "The backlog of unsent items has reached maximum size of {0}. Items will be dropped until the backlog is cleared" ?

cijothomas:
MetricAggregator uses the following approach to prevent starvation when system is busy:
After its first run, Metric Aggregator blocks the thread so that it won't go back to TP. When it wakes up, aggregation is done and sleeps again in blocking way. This ensures that there'll always be 1 TP thread for Metric Aggregation to work.
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/master/src/Microsoft.ApplicationInsights/Metrics/Implementation/DefaultAggregationPeriodCycle.cs#L135

But this is not likely sufficient as Channel still relies on free TP thread - so metric maybe aggregated, but not necessarily sent from channel due to lack of TP threads. It'd be best if both creates a new Thread() in startup, and keeps it to itself dedicatedly.

We are investigating an issue currently where the metric aggregator approach of using TP thread to sleep synchronously might be causing deadlock. new Thread() approach might be the solution here as well.

PR #414, opened 2017-01-20
Channel to drop items if buffer size is growing beyond limit

cijothomas:
Modified both channels to drop items if the buffer is exceeding a very huge size(configurable, default is 1million items).This can occur if item flush logic never gets a chance to run due to threadpool overload.

Fix: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/345
ShortTermFix for :  https://github.com/Microsoft/ApplicationInsights-dotnet/issues/413

### DIFF
--- a/Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
+++ b/Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
@@ -26,6 +26,7 @@
         {
             var buffer = new TelemetryBuffer();
             Assert.Equal(500, buffer.Capacity);
+            Assert.Equal(100000, buffer.MaximumUnsentBacklogSize);
         }
 
         [TestMethod]
@@ -33,7 +34,9 @@
         {
             var buffer = new TelemetryBuffer();
             buffer.Capacity = 42;
+            buffer.MaximumUnsentBacklogSize = 9999;
             Assert.Equal(42, buffer.Capacity);
+            Assert.Equal(9999, buffer.MaximumUnsentBacklogSize);
         }
 
         [TestMethod]
@@ -41,8 +44,10 @@
         {
             var buffer = new TelemetryBuffer();
             buffer.Capacity = 0;
+            buffer.MaximumUnsentBacklogSize = 0;
 
             Assert.Equal(buffer.Capacity, 500);
+            Assert.Equal(buffer.MaximumUnsentBacklogSize, 1000000);
         }
 
         [TestMethod]
@@ -60,21 +65,25 @@
         }
 
         [TestMethod]
-        public void TelemetryBufferDropsItemWhenBufferCapacityReached()
-        {
-            IEnumerable<ITelemetry> items = null;
-            TelemetryBuffer buffer = new TelemetryBuffer { Capacity = 2 };
+        public void TelemetryBufferDoNotGrowBeyondMaxBacklogSize()
+        {            
+            TelemetryBuffer buffer = new TelemetryBuffer { Capacity = 2, MaximumUnsentBacklogSize= 5};
             buffer.OnFull = () => { //intentionaly blank to simulate situation where buffer
                                     //is not emptied.
                                   };
 
+            // Add more items to buffer than the max backlog size
             buffer.Enqueue(new EventTelemetry("Event1"));
             buffer.Enqueue(new EventTelemetry("Event2"));
             buffer.Enqueue(new EventTelemetry("Event3"));
             buffer.Enqueue(new EventTelemetry("Event4"));
+            buffer.Enqueue(new EventTelemetry("Event5"));
+            buffer.Enqueue(new EventTelemetry("Event6"));
 
+            // validate that items are not added after maxunsentbacklogsize is reached.
+            // this also validate that items can still be added after Capacity is reached as it is only a soft limit.
             int bufferItemCount = buffer.Dequeue().Count();
-            Assert.Equal(2, bufferItemCount);
+            Assert.Equal(5, bufferItemCount);
 
         }
     }
--- a/Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
+++ b/Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
@@ -65,7 +65,7 @@
         }
 
         [TestMethod]
-        public void TelemetryBufferDoNotGrowBeyondMaxBacklogSize()
+        public void TelemetryBufferDoesNotGrowBeyondMaxBacklogSize()
         {            
             TelemetryBuffer buffer = new TelemetryBuffer { Capacity = 2, MaximumUnsentBacklogSize= 1002};
             buffer.OnFull = () => { //intentionaly blank to simulate situation where buffer
--- a/Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
+++ b/Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
@@ -40,11 +40,11 @@
         }
 
         [TestMethod]
-        public void WhenNewValueIsLessThanOneSetToDefault()
+        public void WhenNewValueIsLessThanMinimumSetToDefault()
         {
             var buffer = new TelemetryBuffer();
             buffer.Capacity = 0;
-            buffer.MaximumUnsentBacklogSize = 0;
+            buffer.MaximumUnsentBacklogSize = 1000; //1001 is the minimum, setting to anything low should be overruled with Default
 
             Assert.Equal(buffer.Capacity, 500);
             Assert.Equal(buffer.MaximumUnsentBacklogSize, 1000000);
@@ -67,23 +67,23 @@
         [TestMethod]
         public void TelemetryBufferDoNotGrowBeyondMaxBacklogSize()
         {            
-            TelemetryBuffer buffer = new TelemetryBuffer { Capacity = 2, MaximumUnsentBacklogSize= 5};
+            TelemetryBuffer buffer = new TelemetryBuffer { Capacity = 2, MaximumUnsentBacklogSize= 1002};
             buffer.OnFull = () => { //intentionaly blank to simulate situation where buffer
                                     //is not emptied.
                                   };
 
             // Add more items to buffer than the max backlog size
-            buffer.Enqueue(new EventTelemetry("Event1"));
-            buffer.Enqueue(new EventTelemetry("Event2"));
-            buffer.Enqueue(new EventTelemetry("Event3"));
-            buffer.Enqueue(new EventTelemetry("Event4"));
-            buffer.Enqueue(new EventTelemetry("Event5"));
-            buffer.Enqueue(new EventTelemetry("Event6"));
+            for(int i = 0; i < 1005; i++)
+            {
+                buffer.Enqueue(new EventTelemetry("Event" + i));
+            }
+            
+            
 
             // validate that items are not added after maxunsentbacklogsize is reached.
             // this also validate that items can still be added after Capacity is reached as it is only a soft limit.
             int bufferItemCount = buffer.Dequeue().Count();
-            Assert.Equal(5, bufferItemCount);
+            Assert.Equal(1002, bufferItemCount);
 
         }
     }
--- a/Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
+++ b/Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
@@ -26,7 +26,7 @@
         {
             var buffer = new TelemetryBuffer();
             Assert.Equal(500, buffer.Capacity);
-            Assert.Equal(1000000, buffer.MaximumBacklogSize);
+            Assert.Equal(1000000, buffer.BacklogSize);
         }
 
         [TestMethod]
@@ -34,9 +34,9 @@
         {
             var buffer = new TelemetryBuffer();
             buffer.Capacity = 42;
-            buffer.MaximumBacklogSize = 9999;
+            buffer.BacklogSize = 9999;
             Assert.Equal(42, buffer.Capacity);
-            Assert.Equal(9999, buffer.MaximumBacklogSize);
+            Assert.Equal(9999, buffer.BacklogSize);
         }
 
         [TestMethod]
@@ -44,10 +44,10 @@
         {
             var buffer = new TelemetryBuffer();
             buffer.Capacity = 0;
-            buffer.MaximumBacklogSize = 1000; //1001 is the minimum, setting to anything low should be overruled with Default
+            buffer.BacklogSize = 1000; //1001 is the minimum, setting to anything low should be overruled with minimum allowed.
 
             Assert.Equal(500,buffer.Capacity);
-            Assert.Equal(1000000, buffer.MaximumBacklogSize);
+            Assert.Equal(1001, buffer.BacklogSize);
         }
 
         [TestMethod]
@@ -57,10 +57,10 @@
             buffer.Capacity = 9999; // a value greater than the minimum allowed for MaxBacklogSize            
 
             //Attempt to set MaxBacklogSize below Capacity
-            buffer.MaximumBacklogSize = buffer.Capacity - 1;
+            buffer.BacklogSize = buffer.Capacity - 1;
 
             // Validate that MaximumBacklogSize will be set to Capacity
-            Assert.Equal(buffer.Capacity, buffer.MaximumBacklogSize);
+            Assert.Equal(buffer.Capacity, buffer.BacklogSize);
         }
 
         [TestMethod]
@@ -70,10 +70,10 @@
             var buffer = new TelemetryBuffer();            
 
             //Attempt to set Capacity above MaxBacklogSize
-            buffer.Capacity = buffer.MaximumBacklogSize + 1;
+            buffer.Capacity = buffer.BacklogSize + 1;
 
             // Validate that Capacity will be set to MaxBacklogSize
-            Assert.Equal(buffer.Capacity, buffer.MaximumBacklogSize);
+            Assert.Equal(buffer.Capacity, buffer.BacklogSize);
         }
 
         [TestMethod]
@@ -93,7 +93,7 @@
         [TestMethod]
         public void TelemetryBufferDoesNotGrowBeyondMaxBacklogSize()
         {            
-            TelemetryBuffer buffer = new TelemetryBuffer { Capacity = 2, MaximumBacklogSize = 1002};
+            TelemetryBuffer buffer = new TelemetryBuffer { Capacity = 2, BacklogSize = 1002};
             buffer.OnFull = () => { //intentionaly blank to simulate situation where buffer
                                     //is not emptied.
                                   };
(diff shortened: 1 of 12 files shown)


src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
```diff
@@ -57,6 +82,17 @@ public void Enqueue(ITelemetry item)
 
             lock (this.lockObj)
             {
+                if (this.items.Count >= this.MaximumUnsentBacklogSize)
+                {
+                    if (!this.itemDroppedMessageLogged)
```
cijothomas: a unittest to verify we only log error once per overflow?

Test/ServerTelemetryChannel.Test/Shared.Tests/Helpers/StubTelemetryBuffer.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public Action<IEnumerable<ITelemetry>> OnSerialize = _ => { };
+
+        public StubTelemetryBuffer(TelemetrySerializer serializer, IApplicationLifecycle applicationLifecycle) : base(serializer, applicationLifecycle)
+        {            
+
+        }
+        public override Task FlushAsync()
+        {
+            // Intentionally blank to simulate situation where buffer is not emptied.
```
SergeyKanzhelev: make it property as for `OnSerialize`. The class intentiallally called `Stub` to be able to override methods

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
```diff
@@ -46,7 +48,30 @@ public int Capacity
                 this.capacity = value;
             }
         }
-        
+
+        /// <summary>
+        /// Gets or sets the maximum number of telemetry items that can be in the backlog to send. Items will be dropped
+        /// once this limit is hit.
+        /// </summary>        
+        public int MaximumUnsentBacklogSize
```
SergeyKanzhelev: Items already "Sent" after you pass them into `Send` method. `Unsent` in the name doesn't make much sense.

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
```diff
@@ -46,7 +48,30 @@ public int Capacity
...
+        {
+            get
+            {
+                return this.maximumUnsentBacklogSize;
+            }
+
+            set
+            {
+                if (value < 1)
```
SergeyKanzhelev: please use much bigger value as a minimum allowed number. Just to be on the safe side

cijothomas: do you have a number in mind? Does one hundred thousand sounds okay? Should it be different for inmemory vs serverchannels?

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
cijothomas: how about just MaximumBacklogSize?

Test/ServerTelemetryChannel.Test/Shared.Tests/Helpers/StubTelemetryBuffer.cs
cijothomas: Not sure i understand the comment.. FlushAsync() is overwritten in the Stub to do nothing.

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
SergeyKanzhelev: I like it

Test/ServerTelemetryChannel.Test/Shared.Tests/Helpers/StubTelemetryBuffer.cs
cijothomas: Understood the naming confusion, this was not a stub and I'll rename

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
cijothomas: settled on 1001

Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
```diff
@@ -26,23 +26,28 @@ public void DefaultValueIsAppropriateForProductionEnvironmentAndUnitTests()
         {
             var buffer = new TelemetryBuffer();
             Assert.Equal(500, buffer.Capacity);
+            Assert.Equal(1000000, buffer.MaximumUnsentBacklogSize);
```
dnduffy: One million seems kind of high to me...

Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
```diff
@@ -26,23 +26,28 @@ public void DefaultValueIsAppropriateForProductionEnvironmentAndUnitTests()
...
             Assert.Equal(42, buffer.Capacity);
+            Assert.Equal(9999, buffer.MaximumUnsentBacklogSize);
         }
 
         [TestMethod]
         public void WhenNewValueIsLessThanOneSetToDefault()
         {
             var buffer = new TelemetryBuffer();
             buffer.Capacity = 0;
+            buffer.MaximumUnsentBacklogSize = 0;
```
dnduffy: Setting it to 0 but getting 1 million is not an intuitive behavior.
Other places in code it uses 0 to disable a feature, not to default it to a high value.

Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
```diff
@@ -58,5 +63,28 @@ public void TelemetryBufferCallingOnFullActionWhenBufferCapacityReached()
             Assert.NotNull(items);
             Assert.Equal(2, items.Count());
         }
+
+        [TestMethod]
+        public void TelemetryBufferDoNotGrowBeyondMaxBacklogSize()
```
dnduffy: nitpick: "TelemetryBufferDoesNotGrowBeyondMaxBacklogSize"

Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
```diff
@@ -58,5 +63,28 @@ public void TelemetryBufferCallingOnFullActionWhenBufferCapacityReached()
...
+            // Add more items to buffer than the max backlog size
+            buffer.Enqueue(new EventTelemetry("Event1"));
+            buffer.Enqueue(new EventTelemetry("Event2"));
+            buffer.Enqueue(new EventTelemetry("Event3"));
+            buffer.Enqueue(new EventTelemetry("Event4"));
+            buffer.Enqueue(new EventTelemetry("Event5"));
+            buffer.Enqueue(new EventTelemetry("Event6"));
+
+            // validate that items are not added after maxunsentbacklogsize is reached.
+            // this also validate that items can still be added after Capacity is reached as it is only a soft limit.
```
dnduffy: nitpick: StyleCop and sentence capitalization.

Test/ServerTelemetryChannel.Test/Shared.Tests/Helpers/StubTelemetryBuffer.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Helpers
+{
+    using System;
+    using System.Collections.Generic;
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation;
+    using System.Threading.Tasks;
+
+    internal class StubTelemetryBuffer : Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation.TelemetryBuffer
```
dnduffy: Please doc comment all new code (and any existing code you feel like cleaning up while you are editing it).

dnduffy: nitpick: Shorten Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation.TelemetryBuffer to just TelemetryBuffer.

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryBufferTest.cs
```diff
@@ -112,6 +116,31 @@ public void AddsTelemetryToBufferUntilItReachesMax()
             }
 
             [TestMethod]
+            public void TelemetryBufferDoNotGrowBeyondMaxBacklogSize()
```
dnduffy: Are there actually two copies of this method/file or is the code review diff confused?

dnduffy: nitpick: "Do" -> "Does"

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
```diff
@@ -15,19 +15,21 @@ internal class TelemetryBuffer
         public Action OnFull;
 
         private const int DefaultCapacity = 500;
+        private const int DefaultMaximumUnsentBacklogSize = 1000000;
```
dnduffy: I would lower the default a lot.
Even a factor of 10x the default capacity feels high and that would be still be only 1/200th of a million.

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
```diff
@@ -46,7 +48,30 @@ public int Capacity
...
+            get
+            {
+                return this.maximumUnsentBacklogSize;
+            }
+
+            set
+            {
+                if (value < 1)
+                {
+                    this.maximumUnsentBacklogSize = DefaultMaximumUnsentBacklogSize;
```
dnduffy: I would not swallow the user's value and set something else.  Either accept their value or throw an exception.
Since this is new code I will argue that we do not need to be afraid of throwing an exception for bad configuration since no existing customers reference this we can't break them.  But giving the customer and API we don't honor and silently do something different is a bad design.

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
dnduffy: You might want to use Interlocked.Exchange instead.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryBuffer.cs
```diff
@@ -103,6 +128,17 @@ public virtual void Process(ITelemetry item)
...
             {
+                if (this.transmissionBuffer.Count >= this.MaximumUnsentBacklogSize)
+                {
+                    if (!this.itemDroppedMessageLogged)
+                    {
+                        TelemetryChannelEventSource.Log.ItemDroppedAsMaximumUnsentBacklogSizeReached(this.MaximumUnsentBacklogSize);
+                        this.itemDroppedMessageLogged = true;
+                    }
+
+                    return;
```
dnduffy: We cannot use this kind of solution in the ServerTelemetryChannel since it is trying to guarantee that no items will be dropped.  Dropping items in this channel defeats the whole purpose of its existence.

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
SergeyKanzhelev: @cijothomas just put a comment how to get to this number.

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
SergeyKanzhelev: AI SDK doesn't throw exceptions as a design principle

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryBufferTest.cs
cijothomas: TelemetryBuffer.cs is a class which is present in InMemoryChannel and in ServerTelemetryChannel - they share the name but are independent class and files. And they are tested separately as well.

Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
cijothomas: changed

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryBufferTest.cs
dnduffy: Right, I forgot about that confusion when I left the comment.

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
dnduffy: Yes but in this case the behavior is worse than the principle (and I already disagree with the principle).  I agree with not breaking existing customers, at least without increasing the major version number, but this is net new and won't break anyone so the behavior of ignoring their setting and not telling them is clearly a bug/broken/bad design.

cijothomas:
@dnduffy Since we don't new up a dedicated Thread for dispatching items from queue, there is always a possibility of items growing indefinitely if there are too many work items scheduled to run in ThreadPool. Dropping items on the floor is the only way to prevent the queue causing OOM Exception from SDK and crashing customer app.
The following improvement can make better guarantees, but with current design I think this is the best we can do. 
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/413

Test/CoreSDK.Test/Shared/Channel/TelemetryBufferTest.cs
cijothomas: At around 1KB per telemetry item, 1million means 1GB of memory. Queuing up items upto 1GB memory in the rare situation where SDK cannot dispatch items seams reasonable...

cijothomas:
@dnduffy  do you have any comments on this? ( I will do the manual testing to confirm if this happens in ServerChannel with a VM - but that should not block this PR right?)

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryBufferTest.cs
```diff
@@ -71,21 +71,26 @@ public void DefaultValueIsAppropriateForProductionEnvironmentAndUnitTests()
...
             }
 
             [TestMethod]
-            public void ThrowsArgumentOutOfRangeExceptionWhenNewValueIsLessThanOne()
+            public void ThrowsArgumentOutOfRangeExceptionWhenNewValueIsLessThanMinimum()
             {
                 var buffer = new TelemetryBuffer(new StubTelemetrySerializer(), new StubApplicationLifecycle());
                 Assert.Throws<ArgumentOutOfRangeException>(() => buffer.Capacity = 0);
+                Assert.Throws<ArgumentOutOfRangeException>(() => buffer.MaximumBacklogSize = 0);
+                Assert.Throws<ArgumentOutOfRangeException>(() => buffer.MaximumBacklogSize = 1000); // 1001 is minimum anything low would throw.
```
dnduffy: It would be good to assert that 1001 does not throw.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryBuffer.cs
```diff
@@ -103,6 +128,17 @@ public virtual void Process(ITelemetry item)
...
+                if (this.transmissionBuffer.Count >= this.MaximumBacklogSize)
+                {
+                    if (!this.itemDroppedMessageLogged)
+                    {
+                        TelemetryChannelEventSource.Log.ItemDroppedAsMaximumUnsentBacklogSizeReached(this.MaximumBacklogSize);
+                        this.itemDroppedMessageLogged = true;
+                    }
+
+                    return;
+                }
```
dnduffy: In my opinion we need a high priority bug to track that the server telemetry channel needs rewriting, at least partially, to make sure it doesn't fail to send or drop items at the buffer level.  This is not the correct place to drop items in the server telemetry channel, if we can't get items through the buffer, through the serializer, and out over the wire or saved onto disk then we have a big problem with this channel implementation.

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
```diff
@@ -46,7 +48,30 @@ public int Capacity
...
+        public int MaximumBacklogSize
+        {
+            get
+            {
+                return this.maximumBacklogSize;
+            }
+
+            set
+            {
+                if (value < 1001)
```
dnduffy: Make 1001 a constant.
e.g. MinimumUnsentBacklogSize.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryBuffer.cs
```diff
@@ -70,6 +72,29 @@ public int Capacity
...
+        public int MaximumBacklogSize
+        {
+            get
+            {
+                return this.maximumBacklogSize;
+            }
+
+            set
+            {
+                if (value < 1001)
```
dnduffy: In the server telemetry channel I believe we need 1000000 as the minimum, not 1001.

Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryBufferTest.cs
cijothomas: done.

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
```diff
@@ -15,19 +15,21 @@ internal class TelemetryBuffer
         public Action OnFull;
 
         private const int DefaultCapacity = 500;
+        private const int DefaultMaximumUnsentBacklogSize = 1000000;
```
dnduffy: For the in memory channel this should be 1001.
For the server telemetry channel it should be 1000000.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetryBuffer.cs
cijothomas: 1001 is minimum customer can set. 1 million is the default. (for both inmem and server channel)

src/Core/Managed/Shared/Channel/TelemetryBuffer.cs
```diff
@@ -43,10 +46,45 @@ public int Capacity
...
+            get
+            {
+                return this.maximumBacklogSize;
+            }
+
+            set
+            {
+                if (value < this.minimumBacklogSize)
+                {
+                    this.maximumBacklogSize = DefaultMaximumUnsentBacklogSize;
```
dnduffy: min?

PR #415, opened 2017-01-26
Change .NETStandard version of Microsoft.ApplicationInsights from 1.5 to 1.3

karolz-ms:
1.3 means it can work with .NET Framework 4.6 (1.5 indicates 4.6.2 is a requirement)
We do not use any 4.6.2-specific features so it makes it easier to consume the library if we target lower standard.

### DIFF
--- a/src/Core/NuGet/Core/Package.nuspec
+++ b/src/Core/NuGet/Core/Package.nuspec
@@ -28,7 +28,7 @@
       <group targetFramework="wp8">
         <dependency id="Microsoft.Bcl.Async" version="1.0.168" />
       </group>
-      <group targetFramework="netstandard1.5">
+      <group targetFramework="netstandard1.3">
         <dependency id="System.Collections" version="4.0.11" />
         <dependency id="System.Diagnostics.Debug" version="4.0.11" />
         <dependency id="System.Diagnostics.Tools" version="4.0.1" />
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
@pakrym will we have any troubles with `netstandard1.3`?

pakrym:
As I said before PCL you are building for (Profile7) corresponds to `netstandard1.1` (as per https://github.com/davidfowl/NetStandard/blob/master/PclToNetStandard.md ). But there should be no problem with consuming this package from `.NET Framework 4.6` because it is already cross-compiled for `net40, net45, net46`

karolz-ms:
@pakrym The motivation here is that I want to publish a library that targets .NET Standard 1.3 and depends on Microsoft.ApplicationInsights. Having Ms.AI target 1.3 makes this possible

SergeyKanzhelev:
@karolz-ms I didn't realize this. You are saying after this change 4.6 projects will pick up `netstandard1.3`? If so - we first need to actually compile it for `netstandard1.3`. Otherwise some APIs wouldn't be available for 4.6 projects (like `StartOperation`)

pakrym:
@SergeyKanzhelev you are already compiling for `Profile7` it is `netstandard1.1` even lower then `netstandard1.3`

pakrym:
@SergeyKanzhelev after this change 4.6 project would still pick up `net46` target.

pakrym:
The advantage this change gives is that third party library authors targeting `netstandard1.3` would not need to cross compile for desktop framework to reference AI.

SergeyKanzhelev:
@pakrym - the problem with Profile7 is that we removed some APIs and features since they do not exist in Profile7. It may be bad for customers if they run on full framework and after update they are loosing those APIs.

dnduffy:
@pakrym From our conversation yesterday it sounded like we should be fine either way, is that right?

PR #419, opened 2017-02-03
Fix another random failing UnitTest

cijothomas:
Increase max time out for some tests which fail occasionally when run with CodeCoverage..
(The test always pass when ran normally, but under CodeCov tool, it just needs more time)

### DIFF
--- a/Test/CoreSDK.Test/Shared/Extensibility/Implementation/TaskTimerTest.cs
+++ b/Test/CoreSDK.Test/Shared/Extensibility/Implementation/TaskTimerTest.cs
@@ -153,7 +153,7 @@
             }
 
             [TestMethod]
-            [Timeout(1000)]
+            [Timeout(1500)]
             public void HandlesAsyncExceptionThrownByTheDelegate()
             {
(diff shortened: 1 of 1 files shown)


Test/CoreSDK.Test/Shared/Extensibility/Implementation/TaskTimerTest.cs
```diff
@@ -153,7 +153,7 @@ public void CancelsPreviousActionWhenStartIsCalledMultipleTimes()
             }
 
             [TestMethod]
-            [Timeout(1000)]
+            [Timeout(1500)]
```
dnduffy: I think we're chasing a problem on the build server because there is no reason this code should take over one second to execute.  Something is wrong that is slowing everything down.

cijothomas: it depends on something being executed on a threadpool thread  - and there is no guarantee esp. if TP started with minimum number... CodeCov tool is probably causing the additional time

PR #420, opened 2017-02-03
Added try catch for SecurityException from running in partial trust.

dnduffy:


src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -373,6 +373,15 @@ public void ItemDroppedAsMaximumUnsentBacklogSizeReached(int maxBacklogSize, str
                 this.nameProvider.Name);
         }
 
+        [Event(
+            32,
```
dnduffy: This is at 32 because another PR is adding 30 and 31.

Test/CoreSDK.Test/Shared/Extensibility/Implementation/Platform/PlatformImplementationTest.cs
```diff
@@ -45,6 +46,17 @@ public void ReadConfigurationXmlIgnoresMissingApplicationInsightsConfigurationFi
...
 
+        [TestMethod]
+        public void FailureToReadEnvironmentVariablesDoesNotThrowExceptions()
+        {
+            EnvironmentPermission permission = new EnvironmentPermission(EnvironmentPermissionAccess.NoAccess, "PATH");
+            permission.PermitOnly();
+            PlatformImplementation platform = new PlatformImplementation();
+            Assert.IsNull(platform.GetEnvironmentVariable("PATH"));
+            permission = null;
+            EnvironmentPermission.RevertAll();
```
cijothomas: RevertAll() in finally block?

cijothomas: Also, validate that an error is logged into EventLog..(many tests already use TestListener to validate expected messages are logged)

dnduffy: The point of this test is not to test that event logging works.

Issue #355, opened 2016-11-09
Remove F5DiagnosticsSender


SergeyKanzhelev:
Diagnostics sender duplicates error in VS console since we trace unconfigured traces now.

``` 
ApplicationInsights configuration file loading failed. Type 'Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.QuickPulse.Quick PulseTelemetryModule, Microsoft.AI.PerfCounterCollector' was not found. Type loading was skipped. Monitoring will continue.
Application Insights Telemetry (unconfigured): {"name":"Microsoft.ApplicationInsights.Message","time":"2016-11-09T23:37:02.5079101Z","tags":{"ai.internal.sdkVersion":"dotnet:2.2.0-31572","ai.operation.id":"Dzo/PYVTX+M=","ai.location.ip":"127.0.0.1","ai.operation.syntheticSource":"SDKTelemetry","ai.operation.name":"GET /","ai.operation.parentId":"Dzo/PYVTX+M=","ai.device.roleInstance":"sergkanz-saw-vm.redmond.corp.microsoft.com"},"data":{"baseType":"MessageData","baseData":{"ver":2,"message":"AI: ApplicationInsights configuration file loading failed. Type 'Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.QuickPulse.Quick PulseTelemetryModule, Microsoft.AI.PerfCounterCollector' was not found. Type loading was skipped. Monitoring will continue."}}}
Diagnostic message: Performance counters are unavailable when the application is running under IIS Express. Use EnableIISExpressPerformanceCounters element with a value of 'true' within the Performance Collector Module element to override this behavior.
[...]

Issue #307, opened 2016-09-12
Add interface to implementations of ITelemetry that support Metrics property


TheXenocide:
I'm writing an initializer that records custom metrics and I'm having to check many different possible implementation types to access their individual `IDictionary<string, double> Metrics { get; }` property signatures. It would be nice if I could just check `if (telemetry is ISupportMetrics)` or something similar.

SergeyKanzhelev:
Sure, thanks for the feedback. 

cc: @vitalyf007 FYI

Issue #287, opened 2016-08-18
Possible multiple enumerations of an IEnumerable in InMemoryTransmitter.Send 


tokaplan:
InMemoryTransmitter.Send enumerates the telemetryItems argument multiple times in certain paths which could break in runtime if the underlying implementation does not support that.

Issue #46, opened 2015-09-28
Merge test framework projects into one


SergeyKanzhelev:
We have too many test framework projects that needs to be merged into one. Ideally we also need to eliminate the dependency on telemetry channel from test framework if possible

PR #425, opened 2017-02-07
Sergkanz/move projects2

SergeyKanzhelev:
Fix for these bugs:

https://github.com/Microsoft/ApplicationInsights-dotnet/issues/307
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/46
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/355
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/287


Test/TestFramework/Shared/StubTelemetryInitializer.cs
```diff
@@ -19,7 +20,7 @@ public StubTelemetryInitializer()
         /// <summary>
         /// Gets or sets the callback invoked by the <see cref="Initialize"/> method.
         /// </summary>
-        public TelemetryAction OnInitialize { get; set; }
+        public Action<ITelemetry> OnInitialize = item => { };
```
dnduffy: I'm not familiar with this syntax, is this saying the default getter returns item but there is an empty method for the setter?
Why not OnInitialize { get; } = item;?

Test/TestFramework/Shared/StubWebRequest.cs
```diff
@@ -4,7 +4,7 @@
     using System.IO;
     using System.Net;
     using System.Threading.Tasks;
-#if NET45 || NET46
+#if CORE_PCL || NET45 || NET46
```
dnduffy: I thought we were getting rid of PCL and moving to .NET Core?

src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
```diff
@@ -136,7 +136,14 @@ private void DequeueAndSend(TimeSpan timeout)
         /// </summary>
         private Task Send(IEnumerable<ITelemetry> telemetryItems, TimeSpan timeout)
         {
-            if (telemetryItems == null || !telemetryItems.Any())
+            byte[] data = null;
+
+            if (telemetryItems != null)
```
dnduffy: Should this also check for && telemetryItems.Any()?

SergeyKanzhelev: yes, that was the source of multiple enumerations.

Test/TestFramework/Shared/StubWebRequest.cs
SergeyKanzhelev: Yes. For now just merging test classes to minimize duplicaiton

Test/TestFramework/Shared/StubTelemetryInitializer.cs
SergeyKanzhelev: this thing defines the `Action` with the default value of an action that does nothing

dnduffy: Sorry I did not read clearly and see Action<T>.

PR #426, opened 2017-02-08
Allow more precise durations in OperationTelemetryExtensions.Stop

pharring:
We were using TimeSpan.FromMilliseconds which rounds to the nearest millisecond(!)
Now we use TimeSpan.FromTicks which gives the most precision available.

### DIFF
--- a/src/Core/Managed/Shared/OperationTelemetryExtensions.cs
+++ b/src/Core/Managed/Shared/OperationTelemetryExtensions.cs
@@ -13,6 +13,11 @@
     public static class OperationTelemetryExtensions
     {
         /// <summary>
+        /// Multiplier to convert Stopwatch ticks to TimeSpan ticks.
+        /// </summary>
+        private static readonly double StopwatchTicksToTimeSpanTicks = (double)TimeSpan.TicksPerSecond / Stopwatch.Frequency;
+
+        /// <summary>
         /// An extension to telemetry item that starts the timer for the the respective telemetry.
         /// </summary>
         /// <param name="telemetry">Telemetry item object that calls this extension method.</param>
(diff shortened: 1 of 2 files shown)


src/Core/Managed/Shared/OperationTelemetryExtensions.cs
```diff
@@ -87,8 +92,8 @@ public static void GenerateOperationId(this OperationTelemetry telemetry)
         private static void StopImpl(OperationTelemetry telemetry, long timestamp)
         {
             long stopWatchTicksDiff = timestamp - telemetry.BeginTimeInTicks;
-            double durationInMillisecs = stopWatchTicksDiff * 1000 / (double)Stopwatch.Frequency;
-            StopImpl(telemetry, TimeSpan.FromMilliseconds(durationInMillisecs));
+            double durationInTicks = stopWatchTicksDiff * StopwatchTicksToTimeSpanTicks;
```
dnduffy: nitpick: any reason to leave this as double still?

pharring: Great question. TL;DR: Yes. It's not an accident.

Notice that ```StopwatchTicksToTimeSpan``` is defined as a ratio. As it happens, on my machine (and, I think, most modern hardware), this ratio is exactly 1. However, on some hardware this ratio may be different and it's almost certainly not integral.

Now, you may note that, if you expand out the expression you get:
```durationInTicks = stopWatchTicksDiff * TimeSpan.TicksPerSecond / Stopwatch.Frequency```
in which all three terms on the right are integral (```long```) types. The danger here is with overflow. Since ```TimeSpan.TicksPerSecond``` is 10^7, that means the calculation will overflow when ```stopWatchTicksDiff``` is greater than 922,337,203,685. At the usual Stopwatch frequency, that corresponds to an elapsed time of ~25.6 hours which, while unusually long for an ASP.NET request, doesn't seem infeasible for long-lived batch processing operations.

So, to avoid the overflow, you want to divide down first and then we're back to the first point about the ratio needing to be non-integral.

Issue #428, opened 2017-02-10
Deadlock in TraceSource Initialization in Azure Web Site extension


Dmitry-Matveev:
`TelemetryConfigurationFactory` [creates an instance](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/4d348ea233f8c68a7cd87791765c1dc4713f35fb/src/Core/Managed/Net40/Extensibility/Implementation/TelemetryConfigurationFactory.cs#L84) of `InMemoryChannel` which in turn creates an instance of the `InMemoryTransmitter`.

InMemoryTransmitter [starts a new Task which logs](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/ea81a0cc8aa842d65a87aa94d6ed153dcd5ce145/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs#L52) with `CoreEventSource.Log.LogVerbose()`.

`TelemetryConfigurationFactory` continues the execution (it does not wait for the task) and may execute Event Source logging as well leading to a possible deadlock between two threads - one executing Event Source call in the newly created task and one executing event source call on the old thread.
The deadlock seems possible because Event Source is performing some initialization logic for the first ever call and there are two locks that can be acquired in reverse:

```
Thread A:
TraceInternal.TraceEvent() took TraceInternal.critSec,  
Switch.InitializeWithStatus() is waiting for for Switch.InitializedLock

Thread B:
Switch.InitializeWithStatus took Switch.InitializedLock
DiagnosticsConfiguration.Initialize() is waiting on TraceInternal.critSec
```

While the nature of the deadlock is not entirely clear at the moment (a bug in EventSource or in usage of Event Source?), the fix might be to simply initialize Event Source synchronously before we create `InMemoryChannel`.

Dumps are available for the additional investigation, please contact me to get them.

SergeyKanzhelev:
cc: @vancem are those two sequences of locks are by design?

vancem:
[...]

PR #429, opened 2017-02-10
Deadlock in EventSource Initialization

cijothomas:
TelemetryConfigurationFactory modified to ensure it writes something to CoreEventSource so as to force initialize CoreEventSource and prevent any deadlock when other code starts writing to CoreEventSource .
Fixes the issue below by protecting SDK from causing the deadlock itself. ( User code can still cause this)
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/428

### DIFF
--- a/Test/CoreSDK.Test/Shared/Extensibility/TelemetryConfigurationFactoryTest.cs
+++ b/Test/CoreSDK.Test/Shared/Extensibility/TelemetryConfigurationFactoryTest.cs
@@ -3,17 +3,22 @@
     using System;
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
+#if CORE_PCL || NET45 || NET46
+    using System.Diagnostics.Tracing;
+#endif
     using System.Linq;
     using System.Reflection;
     using System.Xml.Linq;
+#if NET40
+    using Microsoft.Diagnostics.Tracing;
+#endif
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.ApplicationInsights.Extensibility.Implementation.Platform;
     using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
     using Microsoft.ApplicationInsights.TestFramework;
-
     using Microsoft.VisualStudio.TestTools.UnitTesting;
-    using Assert = Xunit.Assert;
+    using Assert = Xunit.Assert;    
 
     [TestClass]
     public class TelemetryConfigurationFactoryTest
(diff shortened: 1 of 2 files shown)


SergeyKanzhelev:
Is it possible to unit test this?

cijothomas:
I'll add unit tests.

src/Core/Managed/Net40/Extensibility/Implementation/TelemetryConfigurationFactory.cs
```diff
@@ -44,8 +44,27 @@ protected TelemetryConfigurationFactory()
         /// </remarks>
         public static TelemetryConfigurationFactory Instance
         {
-            get { return instance ?? (instance = new TelemetryConfigurationFactory()); }
-            set { instance = value; }
+            // Both get and set should write something to CoreEventSource so as to initialize it. This is a fix for below deadlock issue
```
Dmitry-Matveev: Is there some other place in SDK code that is guaranteed to be invoked before the instance of TelemetryConfigurationFactory is get/set? That would be a good place to log with Event Source and initialize it for the first time. Or TelemetryConfigurationFactory getter/setter is indeed such a place?

cijothomas: Yes if default config is used. If user passed an instance of IChannel to TelemetryConfiguration, then the deadlock can still occur. But attempt to use the default TelemetryConfiguration will internally use TelemetryConfigurationFactory.Instance.

Test/CoreSDK.Test/Shared/Extensibility/TelemetryConfigurationFactoryTest.cs
```diff
@@ -56,6 +61,34 @@ public void InstanceIsLazilyInitializedToSimplifyResettingOfGlobalStateInTests()
...
+        public void InstanceGetterLogsMessageToCoreEventSource()
+        {
+            using (var listener = new TestEventListener())
+            {
+                listener.EnableEvents(CoreEventSource.Log, EventLevel.Verbose);
+
+                // Invoke Getter and verify that atleast one message is logged to CoreEventSource.
+                TelemetryConfigurationFactory.Instance = null;
+                TelemetryConfigurationFactory retInstance = TelemetryConfigurationFactory.Instance;                
+                Assert.True(listener.Messages.Count() > 0, "TelemetryConfigurationFactory setter did not any message to CoreEventSource");
```
dnduffy: nitpick: grammar here and below: "TelemetryConfigurationFactory setter did not send a message to CoreEventSource."

src/Core/Managed/Net40/Extensibility/Implementation/TelemetryConfigurationFactory.cs
dnduffy: 1. This is kind of a hack that isn't guaranteed to be present or work in future as code is refactored which is worrisome.
2. Code comments should not refer to bugs or issues through links, instead describe the situation in the comment so it doesn't rely on an external link that might not stay valid.

cijothomas: @SergeyKanzhelev do we have any other place to make this change? Since User can create TelemetryConfiguration in his own way, we cannot totally prevent this issue. This only prevent the deadlock if user does not create own TC, instead rely on the default one which is create via TelemetryConfigurationFactory.
(And the original issue also pointed to TelemetryConfigurationFactory as one of the parties involved in the deadlock)

SergeyKanzhelev: The original issue is in TraceSource in Azure Web Sites trace listener. So strictly speaking this change is not necessary - you can modify Azure Web Site trace listener as recommended by CLR(BCL) team.

SergeyKanzhelev:
Closing as EventSource is not a root cause of the issue. @cijothomas please reopen if this change is still needed

PR #430, opened 2017-02-11
Overload StartOperation to specify operation Id and parent Id

### LINKED ISSUE #163: Allow to start operation with pre-defined root and parent ID
`StartOperaiton` API doesn't support scenario when parent ID and root ID were received from different layer and we want to start operation with the existing parent and root ID.

Proposed change - allow `StartOperaition` to receive RequestTelemetry type as an argument

pharring:
Fixes #163

### DIFF
--- a/Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionAsyncTests.cs
+++ b/Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionAsyncTests.cs
@@ -10,6 +10,7 @@
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
     using Assert = Xunit.Assert;
+    using Extensibility.Implementation;
     using TestFramework;
 
     /// <summary>
@@ -30,13 +31,14 @@
             configuration.InstrumentationKey = Guid.NewGuid().ToString();
             configuration.TelemetryInitializers.Add(new OperationCorrelationTelemetryInitializer());
             this.telemetryClient = new TelemetryClient(configuration);
+            AsyncLocalHelpers.SaveOperationContext(null);
         }
 
         /// <summary>
         /// Ensure that context being propagated via async/await.
         /// </summary>
         [TestMethod]
-        public void ContextPropogatesThruAsyncAwait()
+        public void ContextPropagatesThruAsyncAwait()
         {
             var task = this.TestAsync();
             task.Wait();
@@ -86,7 +88,7 @@
         /// Ensure that context being propagated via Begin/End.
         /// </summary>
         [TestMethod]
-        public void ContextPropogatesThruBeginEnd()
+        public void ContextPropagatesThruBeginEnd()
         {
             var op = this.telemetryClient.StartOperation<RequestTelemetry>("request");
             var id1 = Thread.CurrentThread.ManagedThreadId;
--- a/Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionAsyncTests.cs
+++ b/Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionAsyncTests.cs
@@ -95,16 +95,13 @@
             int id2 = 0;
             this.telemetryClient.TrackTrace("trace1");
 
-            HttpWebRequest request = WebRequest.Create(new Uri("http://bing.com")) as HttpWebRequest;
-            var result = request.BeginGetResponse(
+            var result = Task.Delay(millisecondsDelay: 50).AsAsyncResult(
                 (r) =>
                     {
                         id2 = Thread.CurrentThread.ManagedThreadId;
                         this.telemetryClient.TrackTrace("trace2");
 
                         this.telemetryClient.StopOperation(op);
-
-                        (r.AsyncState as HttpWebRequest).EndGetResponse(r);
                     },
                 null);
 
--- a/Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionAsyncTests.cs
+++ b/Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionAsyncTests.cs
@@ -38,7 +38,7 @@
         /// Ensure that context being propagated via async/await.
         /// </summary>
         [TestMethod]
-        public void ContextPropagatesThruAsyncAwait()
+        public void ContextPropagatesThroughAsyncAwait()
         {
             var task = this.TestAsync();
             task.Wait();
@@ -87,8 +87,8 @@
         /// <summary>
         /// Ensure that context being propagated via Begin/End.
         /// </summary>
-        [TestMethod]
-        public void ContextPropagatesThruBeginEnd()
+        [TestMethod, Timeout(500)]
+        public void ContextPropagatesThroughBeginEnd()
         {
             var op = this.telemetryClient.StartOperation<RequestTelemetry>("request");
             var id1 = Thread.CurrentThread.ManagedThreadId;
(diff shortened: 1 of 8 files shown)


SergeyKanzhelev:
Would you mind to make another overload that takes an entire `RequestTelemetry`?

pharring:
@SergeyKanzhelev 
> Would you mind to make another overload that takes an entire RequestTelemetry?

Certainly. Done.

Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionAsyncTests.cs
```diff
@@ -86,23 +88,20 @@ public void ContextPropogatesThruAsyncAwait()
...
+        public void ContextPropagatesThruBeginEnd()
         {
             var op = this.telemetryClient.StartOperation<RequestTelemetry>("request");
             var id1 = Thread.CurrentThread.ManagedThreadId;
             int id2 = 0;
             this.telemetryClient.TrackTrace("trace1");
 
-            HttpWebRequest request = WebRequest.Create(new Uri("http://bing.com")) as HttpWebRequest;
-            var result = request.BeginGetResponse(
+            var result = Task.Delay(millisecondsDelay: 50).AsAsyncResult(
```
pharring: FYI, this test failed for me because I was working in a place with no WiFi.

src/Core/Managed/Operation.AL.Shared/TelemetryClientExtensions.cs
```diff
@@ -21,47 +21,95 @@ public static class TelemetryClientExtensions
...
+                throw new ArgumentNullException("Telemetry client cannot be null.");
+            }
+
+            if (operationTelemetry == null)
+            {
+                throw new ArgumentNullException("operationTelemetry cannot be null.");
+            }
+
+            var operation = new AsyncLocalBasedOperationHolder<T>(telemetryClient, operationTelemetry);
+            operationTelemetry.Start();
```
pharring: I wonder if these two lines (82,83) should be switched (I kept them like this to preserve existing behavior, but I don't think it matters)
1. To get the timestamp as close as possible to the top of the method.
2. To put the setting of ```operation.ParentContext``` next to the constructor.
What do you think, @SergeyKanzhelev ?

SergeyKanzhelev: I agree - ordering is not logical. If calculating what's inside - `Start` should be the very last operation. If measuring total - it should be first. Not a big deal, thanks for noticing. You can switch the order

src/Core/Managed/Operation.CC.Shared/TelemetryClientExtensions.cs
```diff
@@ -21,47 +21,95 @@ public static class TelemetryClientExtensions
...
+        /// <summary>
+        /// Start operation creates an operation object with a respective telemetry item. 
+        /// </summary>
+        /// <typeparam name="T">Type of the telemetry item.</typeparam>
+        /// <param name="telemetryClient">Telemetry client object.</param>
+        /// <param name="operationName">Name of the operation that customer is planning to propagate.</param>
+        /// <param name="operationId">Operation ID to set in the new operation.</param>
+        /// <param name="parentOperationId">Optional parent operation ID to set in the new operation.</param>
+        /// <returns>Operation item object with a new telemetry item having current start time and timestamp.</returns>
+        public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemetryClient, string operationName, string operationId, string parentOperationId = null) where T : OperationTelemetry, new()
```
SergeyKanzhelev: One of the problem with our Sampling + Correlation is that we calculate hash on `Context.User.Id` (if set), than `Context.Operaiton.Id` (if set), than toss a coin. The idea was to either include or exclude telemetry for the user.

Passing parent and root solves the problem in case `Context.User.Id` is not set. However if it was set for the request - "childrens" wouldn't get it. 

So we may need to have override with the `userId` as well. Let me chat with PMs. This sampling by `userId` not quite working with the correlation story we are implementing today...

SergeyKanzhelev:
I'll merge later today. Giving chance @dnduffy and @cijothomas to review.

src/Core/Managed/Operation.AL.Shared/TelemetryClientExtensions.cs
pharring: Done (commit 13ebde7)

src/Core/Managed/Operation.CC.Shared/TelemetryClientExtensions.cs
pharring: Presumably, if an app wants to flow ```Context.User.Id``` from parent to child, then it would need to install an initializer like ```UserTelemetryInitializer``` from the ASP.NET repo. That seems reasonable to me.

Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionAsyncTests.cs
```diff
@@ -30,13 +31,14 @@ public void TestInitialize()
...
             this.telemetryClient = new TelemetryClient(configuration);
+            AsyncLocalHelpers.SaveOperationContext(null);
         }
 
         /// <summary>
         /// Ensure that context being propagated via async/await.
         /// </summary>
         [TestMethod]
-        public void ContextPropogatesThruAsyncAwait()
+        public void ContextPropagatesThruAsyncAwait()
```
dnduffy: nitpick: There is an additional spelling mistake, "Thru" should be "Through".

Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionAsyncTests.cs
dnduffy: There is an additional Thread.Sleep(100); call in this method that seems questionable.

I also wonder if the test method itself should be attributed for timeout so that the while (!result.IsCompleted) loop doesn't ever accidentally become an infinite loop.

Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionTests.cs
```diff
@@ -184,5 +184,78 @@ public void StartDependencyTrackingStoresTheArgumentOperationNameInContext()
             var operation = this.telemetryClient.StartOperation<DependencyTelemetry>("TestOperationName");
             Assert.AreEqual("TestOperationName", AsyncLocalHelpers.GetCurrentOperationContext().RootOperationName);
         }
+
+        [TestMethod]
+        public void ContextPropagatesThroughNestedOperations()
+        {
+            using (var operation1 = this.telemetryClient.StartOperation<RequestTelemetry>("OuterRequest"))
+            {
+                using (var operation2 = this.telemetryClient.StartOperation<DependencyTelemetry>("DependentCall"))
```
dnduffy: nitpick: I have a strong personal dislike of using var instead of known types but in these usings in particular it really isn't clear what the type returned by StartOperation is.

Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionAsyncTests.cs
pharring: Fixed

Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionAsyncTests.cs
pharring: Fixed

Test/CoreSDK.Test/Operation.AL.Shared.Tests/TelemetryClientExtensionTests.cs
pharring: Fixed (although I did not replace ```var``` in tests that were already in place before this change)

PR #431, opened 2017-02-11
Introduce ETW events for Operation start/stop

pharring:
Generate new, paired ETW start/stop events for operations (requests or dependencies) so that Application Insights Profiler can extract them from profile data. This will let Application Insights Profiler work in applications that are not hosted in ASP.NET (e.g. WebJobs).


src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
```diff
@@ -10,6 +10,8 @@
     /// </summary>
     public abstract class OperationTelemetry : ITelemetry, ISupportMetrics, ISupportProperties
     {
+        internal const string TelemetryName = "Operation";
```
SergeyKanzhelev: why this needed?

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.cs
```diff
@@ -78,6 +83,40 @@ public void Process(ITelemetry item)
...
+            {
+                operationStartStopHandler(operation, EventOpcode.Start);
+            }
+        }
+
+        /// <summary>
+        /// Record an operation stop.
+        /// </summary>
+        /// <param name="operation">The operation which has just stopped.</param>
+        public void ProcessOperationStop(OperationTelemetry operation)
```
SergeyKanzhelev: so `Stop` is different from `Track`? Is order significant? Do you see scenarios when we will need only `start/stop` without `Track` for Request and Dependency?

src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
pharring: It's used for the event name in ProcessOperationStart/Stop. I don't distinguish between Request/DependencyTelemetry - they're both "Operation". I followed the pattern in ```Process(ITelemetry)```. The alternative was to hard-code the string.

src/Core/Managed/Net45/Extensibility/Implementation/RichPayloadEventSource.cs
pharring: Yes, Stop is designed to be paired with Start. It has different keywords and opcodes.
Yes, order is significant (Start has to be before Stop)
Yes, I do see scenarios where we only listen to Start/Stop - which is why I use a different keyword for them.

SergeyKanzhelev: My point was that for every request we will get `Start`, then `Stop` and then once again `Track`. Three events instead of 2. Do we need to change `TrackRequest` to report `Stop` instead?

pharring: I think I see what you're getting at: Could we pair "Start" with "Track" instead of "Start"/"Stop".
The important thing is to use the Start and Stop opcodes so that the TraceEvent library can pair them up. (It _could_ be done with Start+Track, but it requires a custom processor).
So, Stop is designed to be paired with Start; hence a different keyword.
Order is significant (Start has to be before Stop).
I do see scenarios where we listen only to "Operations" and not general telemetry keywords.

src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
pharring: I just pushed a change where we _do_ distinguish between ```RequestTelemetry``` and any other ```OperationTelemetry```.  Note that it's possible for a user to create their own operation type derived from ```OperationTelemetry``` (it's public and abstract) and pass that to ```StartOperation<T>```. I don't know how common that is.

Background: Since these operations may be nested, we need to pass the ```EventActivityOptions.Recursive``` flag to ```EventSource.Write```. We could have added that flag to _all_ operations and called it good, but there's a concern that someone will call "StartOperation" without a "StopOperation" leaving a dangling 'start'. It's a catch 22: Either we use "Recursive" everywhere and risk dangling start operations (which can basically ruin analysis of long traces with lots of overlapping operations) or _not_ use recursive, which breaks correlation of Start/Stop events when we have nesting. We compromised by saying ```RequestTelemetry``` should be used at the 'top level' (for incoming requests) and it should not be nested inside any other operation. Other operations _can_ be nested, but users must always be careful to use the ```using``` pattern to guarantee we get a ```StopOperation```.

pharring:
@SergeyKanzhelev All my testing is complete. This is ready for review/approval now.

pharring:
@SergeyKanzhelev I just had a through that these new "Operation" events should actually be fired from within ```OperationTelmetryExtensions.Start()``` and ```Stop()```. It would be more low-level and would catch 'naked' uses of "Start"/"Stop". What do you think?

pharring:
@SergeyKanzhelev @dnduffy  This is ready for merging and I'd like it to go into 2.3. Approve?

SergeyKanzhelev:
Please update CHANGELOG.md

SergeyKanzhelev:
And update it for your previous change of StartOperaiton API modification as well

pharring:
Should it be under "Version 2.3.0-beta3" ?

pharring:
Confirmed. It's not a concern for Service Profiler.

PR #432, opened 2017-02-14
Fix failing test when stopwatch frequency isn't exactly 10^7

pharring:
@karolz-ms reported a failure in the OperationTelemetryCanRecordPreciseDurations test on a physical machine where stopwatch ticks did not match timespan ticks. In that case, we need to allow for some epsilon in the test.

I also modified the implementation to use Math.Round when converting to TimeSpan ticks which gives a further, slight accuracy improvement in those cases.


karolz-ms:
Tested the fix on my hardware--works!

PR #434, opened 2017-02-16
Upgrade Core.PCL project to netstandard1.3

### LINKED ISSUE #433: Add NetStandard 1.3 version of core SDK
We would like to use ```TelemetryClientExtensions.StartOperation``` from .NET Core apps (including within the ApplicationInsights-aspnetcore repo, for eaxmple).

Right now, when we reference Microsoft.ApplicationInsights from a .Net Core app, we pick up the "netstandard1.3" version which comes from the output of the Core.PCL project.

I tried adding a reference to the Operation.AL.Shared project to Core.PCL but it failed because it couldn't find ```AsyncLocal<T>```. Interestingly, ```AsyncLocal<T>``` _is_ available in netstandard1.3, but Core.PCL can't reference it because it targets .Net 4.5 (which corresponds to netstandard1.1)

I propose that we create a new "Core.NetStandard" project and add in the necessary pieces to support ```TelemetryClientExtensions``` (including the recent change I made to fire ETW events). It would target netstandard1.3 (equivalent of Net46) and the NuGet package would pull in netstandard1.3 binaries from the output of that project instead of Core.PCL

I'd be willing to do this, but it's a substantial change, so I'm seeking guidance before I embark (e.g. perhaps you already have this planned).

pharring:
Fixes #433 

This turned out to be simpler than I had expected. I thought I'd have to create a whole new project, but in fact I could just upgrade the existing PCL project (since we're still targeting netstandard1.3)
The project still builds in VS 2015.

The main point of this is to be able to bring in the "Operation.Al.Shared" project and get TelemetryClientExtensions and RichPayloadEventSource.


src/Core/Managed/PCL/Core.PCL.csproj
```diff
@@ -13,10 +13,9 @@
     <DefaultLanguage>en-US</DefaultLanguage>
     <FileAlignment>512</FileAlignment>
     <ProjectTypeGuids>{786C830F-07A1-408B-BD7F-6EE04809D6DB};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
-    <TargetFrameworkProfile>Profile7</TargetFrameworkProfile>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
-    <NuGetPackageImportStamp>
-    </NuGetPackageImportStamp>
+    <TargetFrameworkProfile>
+    </TargetFrameworkProfile>
+    <TargetFrameworkVersion>v5.0</TargetFrameworkVersion>
```
SergeyKanzhelev: what this target framework indicates? Is it `netstandard1.3`?

src/Core/Managed/PCL/Core.PCL.csproj
```diff
@@ -25,20 +24,65 @@
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' " />
   <ItemGroup>
-    <None Include="packages.config" />
+    <None Include="project.json" />
```
SergeyKanzhelev: will it use csproj or project.json to build? Will signing work?

SergeyKanzhelev:
Can you please explain how it is done? @karolz-ms mentioned that we have to use the new csproj format that only supported by VS 2017.

SergeyKanzhelev:
Would it be OK if I'll create a separate branch for this change and we will try to run signed build first?

pharring:
In VS 2015, right click on "Core.PCL" and select "Properties". On the first page, "Library", in the "Targeting" section, there's a hyperlink that says "Target Standard Library Profile". That will cause VS to unload and reload the project (converting packages.config to project.json).

(I'll let you into a secret, though: I did this first on a brand new PCL project, got it working and then, once I'd figured it out, I hand-patched all the changes back to the original Core.PCL.csproj file)

Yes, please do as much verification as you need on this. There's a good chance there's a scenario I don't know about that will be broken.

src/Core/Managed/PCL/Core.PCL.csproj
pharring: I believe "v5.0" means ".Net Standard". The actual netstandard version is set in project.json

src/Core/Managed/PCL/Core.PCL.csproj
pharring: It uses .csproj to build and project.json to specify the references (NuGet packages)

SergeyKanzhelev:
Can you please resubmit this PR to `pclToNetStandardConversion`. I'll start a singed build for it

src/Core/Managed/PCL/project.json
```diff
@@ -0,0 +1,10 @@
+﻿{
+  "supports": {},
+  "dependencies": {
+    "Microsoft.NETCore.Portable.Compatibility": "1.0.1",
```
SergeyKanzhelev: Do we need to mention all these explicitly? https://github.com/Microsoft/ApplicationInsights-dotnet/blob/530b6f53a19d3290cfd0735b9813c723aafc226f/src/Core/NuGet/Core/Package.nuspec#L32-L50

``` xml
        <dependency id="System.Collections" version="4.0.11" />
        <dependency id="System.Diagnostics.Debug" version="4.0.11" />
        <dependency id="System.Diagnostics.Tools" version="4.0.1" />
        <dependency id="System.Diagnostics.Tracing" version="4.1.0" />
        <dependency id="System.Globalization" version="4.0.11" />
        <dependency id="System.IO" version="4.1.0" />
        <dependency id="System.IO.Compression" version="4.1.0" />
        <dependency id="System.Linq" version="4.1.0" />
        <dependency id="System.Net.Http" version="4.1.0" />
        <dependency id="System.Reflection" version="4.1.0" />
        <dependency id="System.Reflection.Extensions" version="4.0.1" />
        <dependency id="System.Runtime" version="4.1.0" />
        <dependency id="System.Runtime.Extensions" version="4.1.0" />
        <dependency id="System.Text.Encoding" version="4.0.11" />
        <dependency id="System.Text.Encoding.Extensions" version="4.0.11" />
        <dependency id="System.Text.RegularExpressions" version="4.1.0" />
        <dependency id="System.Threading" version="4.0.11" />
        <dependency id="System.Threading.Tasks" version="4.0.11" />
        <dependency id="System.Xml.XDocument" version="4.0.11" />
```

SergeyKanzhelev:
if this change will work - we may need to merge this project and Net46 as they uses the same APIs.

src/Core/Managed/PCL/project.json
pharring: Good point. I suspect they can be removed now. I'll try it.

SergeyKanzhelev:
Merging to a temporary branch for build validation

src/Core/Managed/PCL/project.json
pharring: Yup. These seem to be safe to delete. It works fine without it. On the consuming side, the dependencies are automatically picked up.
I'll wait for the results of your verification before making this change.

PR #435, opened 2017-02-16
Fix for AclsAreAppliedToLocalAppData and AclsAreAppliedToTemp

karolz-ms:
AclsAreAppliedToLocalAppData and AclsAreAppliedToTemp are failing if VS is running elevated


SergeyKanzhelev:
Thanks!

PR #436, opened 2017-02-16
mark InMemoryChannel as internal sdk operation. This will allow excep…

SergeyKanzhelev:
…tions statistics to pick mark those exceptions as internal

This PR creates a foundation to address the feedback for exception statistics: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/212:

>One of the biggest feedbacks so far for exception statistics is that internal SDK exceptions should be marked somehow or filtered. 

If approach looks OK - I'll mark server telemetry channel and configuration reading logic as well.

### DIFF
--- a/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
+++ b/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
@@ -6,12 +6,10 @@ namespace Microsoft.ApplicationInsights.Channel
 {
     using System;
     using System.Collections.Generic;
-    using System.Diagnostics;
     using System.Globalization;
-    using System.Linq;
-    using System.Net;
     using System.Threading;
     using System.Threading.Tasks;
+    using Extensibility;
     using Extensibility.Implementation;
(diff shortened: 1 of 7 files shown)


src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs
```diff
@@ -80,7 +78,19 @@ public void Dispose()
         /// </summary>
         internal void Flush(TimeSpan timeout)
         {
-            this.DequeueAndSend(timeout);
+#if !CORE_PCL
+            SdkInternalOperationsMonitor.Enter();
+            try
+            {
+#endif
+                this.DequeueAndSend(timeout);
```
cijothomas: Can you instead do Enter/Exit in the body of DequeueAndSend() so Flush() and Runner() need not separately be marked with Enter/Exit.

SergeyKanzhelev: I want to do it once around the "while" instead of doing it at every cycle of the loop

PR #437, opened 2017-02-16
Sergkanz/task timer deprecation

SergeyKanzhelev:
There is no place for `TaskTimer` class in Application Insights API. It was implemented before.

>`TaskTimer` is defined in the `Microsoft.ApplicationInsights` assembly because it is still used here as well as in other AI packages that depend on it. However, it doesn't need to be public anymore because we no longer publish WinRT assemblies. The class was implemented as public because WinRT didn't support internal class visibility at the time. Instead we placed it in the "Implementation" namespace and decided that semver does not apply to public types in this namespace. It should be safe to make `TaskTimer` internal straight away.


cijothomas:
@macrogreg 
You may need to close this PR and make changes to the updated code instead. https://github.com/Microsoft/ApplicationInsights-dotnet/pull/375

PR #438, opened 2017-02-16
Pcl to net standard conversion

SergeyKanzhelev:
CC: @dnduffy @karolz-ms @cijothomas @pharring

This seems to be working fine. https://www.myget.org/F/applicationinsights/api/v2 Version `2.3.0-beta2-build27851`

@karolz-ms you can do the same in logging repo. Can you?


karolz-ms:
@SergeyKanzhelev Most likely yes, I can, but project.json is/will soon be obsolete. I'd rather wait a couple of weeks till dev15 ships and use the new tooling.

SergeyKanzhelev:
@karolz-ms project.json looks very tiny. So not much investments here... Using this approach will allow to support .NET Core `ILogger` and `EventSource` by //build

karolz-ms:
@SergeyKanzhelev OK, sounds good. Will submit a PR for `EventSource` in the next couple of days

pharring:
As Sergey noted in #434 and I verified, we can also remove the netstandard1.3 dependencies from Package.nuspec. That could be done here or in a follow-up PR,

SergeyKanzhelev:
We'll wait for VS 2017 RTM to use new `.csproj` approach

Issue #439, opened 2017-02-16
[Metrics] Feedback on MetricManager API


AlexBulankou:
Will the following code result into pre-aggregated  ``test`` metric sent:
```
                new MetricManager(new TelemetryClient()).CreateMetric("test").Track(1);
                new MetricManager(new TelemetryClient()).CreateMetric("test").Track(2);
```
or 2 separate MetricTelemetry items? As a customer I would like API to suggest the optimal approach to me and from this perspective having to initialize new MetricManager object and having to Create and keep around my instance of Metric appears too much work. I already learnt how to do this:
```
new TelemetryClient().TrackMetric("test", 1);
```
Additional confusion is that ``Metric`` is different from ``MetricTelemetry`` that we all know. 
Can we add ``TrackAggregated`` without having to create MetricManager and keep the reference to metric.
```
new TelemetryClient().TrackMetricAggregated("test", 1) 
```

*Suggestion:* Add ``TrackAggregatedMetric`` API on TelemetryClient for more intuitive use and to not require to create instance of MetricManager and Metric.

SergeyKanzhelev:
cc: @vitalyf007 @olegsych @macrogreg 

We discussed multiple implementations of this API. this is a good feedback and I'd appreciate if you'll provide a specific example when you had such confusion rather than opinion.

AlexBulankou:
Yesterday I upgraded my app to 2.3 Beta2 (from 2.2 stable) and, as a customer would, saw warnings saying that `TrackMetric(name, value)` is now obsolete. I understand that metrics are now aggregated on the client, however my first thought was, why do I have to do anything different, please go ahead and aggregate my metric on the client, not the server. Ah, I see,  I have to pick dimensions now. Well, can you pick them for me? Sure, I realize, this may be not so straightforward. So I can pick dimensions and even start using different API, but why am I asked to change my program structure, create `MetricManager` and create and retain Metric objects. I think we could benefit from API that would be make upgrade easier:

- My first preference would be: no changes to `TrackMetric` but now there's client-side aggregation. Pick important dimensions for me.
- My second preference: modify `TrackMetric(name, ...)` and `TrackMetric(MetricTelemetry, ...)` APIs so they support aggregated option
- My third preference: introduce `TrackMetricAggregated` that supports same usage model as `TrackMetric`

SergeyKanzhelev:
[...]

PR #441, opened 2017-02-23
Move Metrics related code to a dedicated namespace / directory.

macrogreg:


src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
```diff
@@ -22,6 +23,8 @@ public sealed class SamplingTelemetryProcessor : ITelemetryProcessor
         private const string RequestTelemetryName = "Request";
         private const string TraceTelemetryName = "Trace";
 
+        private const string SamplingRateMetricName = "Sampling Rate (Preview)";
```
SergeyKanzhelev: Why do we need this metric? It is not in design spec.

Also here is the patter we recommend - chain multiple sampling processors to set various sampling algorithms for different types: https://github.com/SergeyKanzhelev/TR24DEVCT304/blob/master/ApplicationInsightsDataROI/demo2.cs#L43-L65

When you have a chain like this this approach will send contradicting metrics.

dnduffy: If you make the metric name unique based on the sampling telemetry processor configuration such as IncludedTypes vs. ExcludedTypes then it would be safe since there would be multiple metrics, one for each sampling processor.

dnduffy: As Vitaly commented in the meeting, dimensions would be better.

macrogreg: I will make the change and update the PR.

macrogreg: PR updated.

src/Core/Managed/Shared/Extensibility/Metrics/Metric.cs
```diff
@@ -1,13 +1,12 @@
-﻿namespace Microsoft.ApplicationInsights.Extensibility
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Metrics
```
dnduffy: I am worried that this namespace change will break customers using the beta SDK with the old namespace (I know I just used this functionality in AIMC and it will break again when updating to the next SDK release).
How important is it to rename the namespace?
And are you willing to change all of the internal applications with references to the old namespace and upgrade them when the SDK releases?

macrogreg: Well, but isn't it the purpose of a Beta to be able to iterate breaking changes for the new Beta features?

dnduffy: Beta's are to get customer feedback and make sure we don't have major bugs in new functionality.
If we have to make a breaking change, in new functionality only, to fix something then we can.
I don't know that tweaking a namespace is that important but I am open to hear why it should be renamed vs. left as is.

src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
```diff
@@ -139,72 +157,145 @@ public string IncludedTypes
...
+            double samplingPercentage = this.SamplingPercentage;
+
+            // If sampling rate is 100% we log the sapling rate as a metric and do nothing else:
+            if (samplingPercentage >= 100.0 - 1.0E-12)
             {
-                var samplingSupportingTelemetry = item as ISupportSampling;
+                
+                TrackSamplingRate(100.0);
+                this.Next.Process(item);
+                return;
```
dnduffy: nitpick: They all call this.Next.Process(item) followed by return, can this be shortened up with else statements?

src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
```diff
@@ -139,72 +157,145 @@ public string IncludedTypes
...
+                if (metricManager == null)
+                {
+                    return;
+                }
+
+                // There is an edge case where there may be several sampling processors in the pipeline.
+                // To account for that, the aggregated metric documents will be marked with sufficinet info to differentiate the sampling rates.
+                // In general, if the user is only interested in whether the sampling rate is 100% or not 100% across, these properties my be ignored.
+                samplingMetric = metricManager.CreateMetric(SamplingRateMetricName,
+                                                            new Dictionary<string, string>()
```
dnduffy: nitpick: un-indent to readable level or pre-build the dictionary for StyleCop multi-line compliance.

src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
```diff
@@ -139,72 +157,145 @@ public string IncludedTypes
...
+                                                            {
+                                                                [UniqueProcessorIdMetricPropertyName] = this.uniqueProcessorId,
+                                                                [IncludedTypesMetricPropertyName] = this.IncludedTypes ?? "null",
+                                                                [ExcludedTypesMetricPropertyName] = this.ExcludedTypes ?? "null",
```
dnduffy: Question: Is it better to always set the dimensions to "null" vs. leave them out when there is no value?

src/Core/Managed/Shared/Extensibility/Implementation/Metrics/SimpleMetricStatisticsAggregator.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Metrics
```
vitalyf007: Isn't this a breaking change? If so, you can't do it in minor release...

src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
vitalyf007: I suggest you do metric for each *type* of telemetry being sampled. If it isn't sampled, no data is "tracked". That way your time series won't be "jumpy" meaning the set of dimensions will not shift on you all of a sudden as you (or somebody else) updates included/excluded types.

src/Core/Managed/Shared/Extensibility/Metrics/Metric.cs
macrogreg: This is new functionality. was planning to update the Web SDK to adjust to this change, but not other tooling, which should be updated next time it is touched.
There are three reasons for this change:
a) internal. since this is a current area of investment, separating metrics specific code helps organizing.
b) user facing. since there are a whole bunch of public APIs that only make sense in the context of metrics, this will help keep intellisense clutter-free for customers who do not need these APIs and do not import ...Metrics.
c) Some nuances form this discussion: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/439

src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
macrogreg: In this case I believe null is better. If makes an explicit statement "no excluded types" rather than the "I do not know anything about excluded types" statement implied by an absent property.

src/Core/Managed/Shared/Extensibility/Implementation/Metrics/SimpleMetricStatisticsAggregator.cs
macrogreg: aren't we within beta versions?

src/TelemetryChannels/ServerTelemetryChannel/Shared/SamplingTelemetryProcessor.cs
macrogreg: Changing incl/excl types or having multiple samplers will occur, but it will be a rare occurrence. I would prefer to optimize for the typical case, which is what I am doing here. In the vast majority of cases we will have only one data series. If we did this by type, we would multiply the cost for this by the number of telemetry types for everybody and only gain anything for edge cases. Unless you have serious concerns I would prefer to leave the current approach.

vitalyf007: Ok. Makes sense.

src/Core/Managed/Shared/Extensibility/Implementation/Metrics/SimpleMetricStatisticsAggregator.cs
vitalyf007: I just want to make sure that we are...

macrogreg:
I am abandoning this in favor of a new PR of this branch into 2.4.0-beta1:
https://github.com/Microsoft/ApplicationInsights-dotnet/pull/446

Issue #440, opened 2017-02-22
[Exceptions Statistics] Mark configuration reading and server telemetyr channel as "internal" execution


SergeyKanzhelev:
Mark SDK work as internal operations using `SdkInternalOperationsMonitor`

PR #443, opened 2017-02-24
mark server telemetyr channel with internal SDK operation

SergeyKanzhelev:
Step toward https://github.com/Microsoft/ApplicationInsights-dotnet/issues/440

### DIFF
--- a/src/Core/Managed/Shared/OperationTelemetryExtensions.cs
+++ b/src/Core/Managed/Shared/OperationTelemetryExtensions.cs
@@ -40,7 +40,9 @@
             // Stopwatch.GetTimestamp() is used (instead of ElapsedTicks) as it is thread-safe.
             telemetry.BeginTimeInTicks = timestamp;
 
+#if !CORE_PCL
             RichPayloadEventSource.Log.ProcessOperationStart(telemetry);
+#endif
         }
 
         /// <summary>
@@ -111,7 +113,9 @@
                 telemetry.Timestamp = DateTimeOffset.UtcNow;
             }
 
+#if !CORE_PCL
             RichPayloadEventSource.Log.ProcessOperationStop(telemetry);
+#endif
         }
     }
 }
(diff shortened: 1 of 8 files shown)


src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/TelemetrySerializer.cs
```diff
@@ -8,8 +8,9 @@
 
     internal class TelemetrySerializer
     {
+        internal readonly Transmitter Transmitter;
```
dnduffy: nitpick: Why the capitalization change?  This isn't a property or a constant, does StyleCop want readonly variables to be capitalized also?

SergeyKanzhelev: Internal should be capitalized according to StyleCop

dnduffy: I think StyleCop is confused in this case. :(

SergeyKanzhelev: I'll be happy to follow any other ruleset. Keeping it compliant to the current ruleset for now.

src/Core/Managed/Shared/OperationTelemetryExtensions.cs
```diff
@@ -40,7 +40,9 @@ public static void Start(this OperationTelemetry telemetry, long timestamp)
             // Stopwatch.GetTimestamp() is used (instead of ElapsedTicks) as it is thread-safe.
             telemetry.BeginTimeInTicks = timestamp;
 
+#if !CORE_PCL
```
pharring: I just saw your IM. Thanks for fixing this. I didn't realize it had been broken for a while now. Yes, eventually we want RichPayloadEventSource in Core.PCL.

PR #444, opened 2017-02-27
Introduce Metric Extractors and use them to Auto-Collect some default metrics.

macrogreg:


src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public override void Process(ITelemetry item)
+        {
+            DependencyTelemetry dependencyCall = item as DependencyTelemetry;
+            if (dependencyCall == null)
+            {
+                return;
+            }
+
+            bool isFailed = (dependencyCall.Success != null) && (dependencyCall.Success == false);
```
tokaplan: You mean dependencyCall.Success != true?

src/Core/Managed/Shared/Extensibility/Metrics/RequestMetricExtractor.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            RequestTelemetry request = item as RequestTelemetry;
+            if (request == null)
+            {
+                return;
+            }
+
+            _requestsMetric.Track(1.0);
+
+            bool isFailed = (request.Success != null) && (request.Success == false);
```
tokaplan: request.Success == false is probably enough :)

src/Core/Managed/Shared/Extensibility/Metrics/MetricExtractorTelemetryProcessorBase.cs
```diff
@@ -0,0 +1,159 @@
...
+            }
+
+            extractionPipelineInfo = extractionPipelineInfo + thisExtractorInfo;
+            item.Context.Properties[MetricTerms.Extraction.PipelineInfo.PropertyKey] = extractionPipelineInfo;
+        }
+
+        private void InvokeNextProcessor(ITelemetry item)
+        {
+            ITelemetryProcessor next = _nextProcessorInPipeline;
+            if (next != null)
```
tokaplan: You mean next?.Process(item) ?

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
macrogreg: .Success is a nullable value and the call is not faulted unless it is explicitly faulted.

src/Core/Managed/Shared/Extensibility/Metrics/RequestMetricExtractor.cs
macrogreg: .Success is a nullable value and the call is not faulted unless it is explicitly faulted.

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.ApplicationInsights.DataContracts;
+using System.Collections.Generic;
+
+namespace Microsoft.ApplicationInsights.Extensibility.Metrics
+{
+    public class DependencyMetricExtractor : MetricExtractorTelemetryProcessorBase
+    {
+        private const string Version = "1.0";
+
+        private Metric _dependencyCallsDurationMetric;
```
vitalyf007: I think we were using this.var notation for referencing member variables of the class before (not sure those - that could have changed). Better to stick with the same pattern everywhere.

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.DataContracts;
+using System.Collections.Generic;
+
+namespace Microsoft.ApplicationInsights.Extensibility.Metrics
+{
+    public class DependencyMetricExtractor : MetricExtractorTelemetryProcessorBase
```
vitalyf007: The other way to design this would be to have single metric processor called "MetricExtractorTelemetryProcessor" and have a list of metric extractors within that. Each approach has certain pros and cons. I'd like you to consider those and make a decision (whatever that would be). If you already considered and made a choice - I'd be interested to seewhy this coded approach won.

src/Core/Managed/Shared/Extensibility/Metrics/MetricExtractorTelemetryProcessorBase.cs
```diff
@@ -0,0 +1,164 @@
...
+using System.Globalization;
+
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+
+namespace Microsoft.ApplicationInsights.Extensibility.Metrics
+{
+    public abstract class MetricExtractorTelemetryProcessorBase : ITelemetryProcessor, ITelemetryModule, IDisposable
+    {
+        private MetricManager _metricManager = null;
```
vitalyf007: Adding to my previous comment. The number of processors you create would define the number of managers you got (and number of timers within them. Not a big deal, but that is one of the "cons" of the approach you have coded up...

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
dnduffy: Correct, the convention is this.var, no underscores.
StyleCop is your friend (even if you don't like what it tells you to do ;) ).

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        private const string Version = "1.0";
+
+        private Metric _dependencyCallsDurationMetric;
+        private Metric _dependencyCallsFailuresMetric;
+
+        public DependencyMetricExtractor(ITelemetryProcessor nextProcessorInPipeline)
```
dnduffy: The class and all public items need doc comments (and I'll be quite happy if you also comment the private stuff as well).

src/Core/Managed/Shared/Extensibility/Metrics/MetricExtractorTelemetryProcessorBase.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+
+            try
+            {
+                Process(item);
+                AddExtractorInfo(item);
+            }
+            catch(Exception ex)
+            {
+                // Is this te right way to log erroes here?
```
dnduffy: nitpick: "te" = "the"
The answer is that you should usually make a specific message in the CoreEventSource rather than calling Log.LogError.

src/Core/Managed/Shared/Extensibility/Metrics/MetricExtractorTelemetryProcessorBase.cs
dnduffy: That newer short form is more compact to write but it's the same compiled.

src/Core/Managed/Shared/Extensibility/Metrics/MetricTerms.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    internal static class MetricTerms
+    {
+        private const string MetricPropertiesNamePrefix = "Microsoft.ApplicationInsights.Metrics";
+        public static class Extraction
+        {
+            //Microsoft.ApplicationInsights.Metrics.MetricIsAutocollected
+            public static class ConsideredByProcessors
+            {
+                public static class Moniker
```
dnduffy: Why do you have 4 nested classes pretending to be a namespace for a single const string?

src/Core/Managed/Shared/Extensibility/Metrics/MetricTerms.cs
```diff
@@ -0,0 +1,48 @@
...
+                public const string Key = MetricPropertiesNamePrefix + ".MetricIsAutocollected";
+                public const string Value = "True";
+
+            }
+
+            public static class MetricNames
+            {
+                private const string PreviewLabel = " (Preview)";
+
+                public static class Requests
```
dnduffy: nitpick: C# has namespaces, and long variable names, these nested classes are not a style I care for just to get dots.

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
dnduffy: Which way is easier to configure for users?
Which way is more consistent with the existing design philosophy in the SDK?

src/Core/Managed/Shared/Extensibility/Metrics/MetricExtractorTelemetryProcessorBase.cs
macrogreg: I prefer the new form in assignments like x = foo?.ComputeValue(), but not in statements like do this when x isn't null. I think, both is generally fine, though. :)

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
macrogreg: I had no StyleCop issues, but happy to change it if that's the style elsewhere. Will update the PR accordingly. (Although I personally really do not like the this.var style, code-base consistency definitely trumps.)

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
macrogreg: @vitalyf007 @dnduffy I have actually considered both approached form the start. I picked this because I wanted to follow guidance by @SergeyKanzhelev to make things "plugable as if I was a customer". So a customer should be able to create a metric extractor and to plug it in using a familiar paradigm. This means that we have two approaches:
- The one I used.
- A similar, but from the ExtractorBase I would derive a single "Auto-Collected Metric Extractor" that would do all *our* extraction and leave the users the possibility to create more extractors.
(- on contrary, a processor that contains all extractors would require the user to configure things differently from everything else)
Please let me know if you prefer me to go into more detail about who I chose the first of the two.

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
macrogreg: Yes, I was just waiting for general agreement before adding those + unit tests.

src/Core/Managed/Shared/Extensibility/Metrics/MetricTerms.cs
macrogreg: Readability

src/Core/Managed/Shared/Extensibility/Metrics/MetricTerms.cs
macrogreg: This was, actually, just to get better readability and intellisense. If you feel very passionate about this, I can change it, but myself - I really like the added readability with the dots. And I am sure that other people with dyslexia do too. :)

src/Core/Managed/Shared/Extensibility/Metrics/MetricExtractorTelemetryProcessorBase.cs
macrogreg: Ok, so to continue my response above - how do I chose between the two options.

* Making thing configurable publicly "just for ourselves" is not a good approach. It creates a lot of work and hardening requirements. So the extractors should be either not configurable at all, or - if they appear in ai.config - then customers should be able to write their own. As mentioned, based on Sergey's recommendation, I am going with the appear in ai.config route.

So, not I want to make it super easy to write an extractor. So I put common logic in a common base. But a key engineering principle is to avoid any unnecessary abstractions. So to test whether my base is factored correctly i should design to have at least two specializations. So instead of having just one specialization for "all auto-collected stuff", i am going with 2 specializations for "all request related stuff" and "all dependency related stuff". And additional benefit of this approach is that if in the future we add a package to collect some kind of SpecialTelemetry we will be able to have that package add SpecialTelemetryMetricExtractor using a consistent method.

src/Core/Managed/Shared/Extensibility/Metrics/MetricTerms.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            public static class DependencyCall
+            {
+                public static class PropertyName
+                {
+                    public const string Success = "Success";
+                    public const string TypeName = "Type";
+                }
+
+                public static class TypeName
```
dnduffy: What I was getting at with my earlier comment was that you can make a namespace that gives you the same dotted naming to type in and get intellisense for without having to resort to these strange nested public classes.
I'm not objecting to dots, I like dots too, I'm objecting to reinventing namespaces in a way that makes this code less readable.

macrogreg: Hi @dnduffy ,

I have actually seen the current approach in the past and have used it myself.
I am not sure what you mean, so let m see if I understand you correctly by making an example.
Are you looking for this?:
---
namespace Microsoft.ApplicationInsights.Extensibility.Metrics.MetricTerms
{
    internal static class Shared
    {
        public const string MetricPropertiesNamePrefix = "Microsoft.ApplicationInsights.Metrics";
    }

    namespace Extraction.ProcessedByExtractors
    {
        internal static class Moniker
        {
            public const string Key = Shared.MetricPropertiesNamePrefix + ".Extraction.ProcessedByExtractors";
        }
    }

    namespace Autocollection
    {
        internal static class Moniker
        {
            public const string Key = Shared.MetricPropertiesNamePrefix + ".MetricIsAutocollected";
            public const string Value = "True";
        }

        namespace MetricNames
        {
            internal static class Request
            {
                public const string Duration = "Server response time";
            }

            internal static class DependencyCall
            {
                public const string Duration = "Dependency duration";
            }
        }

        namespace Request
        {
            internal static class PropertyName
            {
                public const string Success = "Success";
            }
        }

        namespace DependencyCall
        {
            internal static class PropertyName
            {
                public const string Success = "Success";
                public const string TypeName = "Type";
            }

            internal static class TypeName
            {
                public const string Other = "Other";
            }
        }
    }
}
---
If yes, then that approach has a number of problems:
- Namespaces are not a proper encapsulation. Default tools will tend to "use" then ratehr than typing out everything from MetricTerms... This will result in short identifiers like Request.Duration in code, and so - defeat the purpose.
- It is not clear that even if the user was to type out a longer ID, then MetricTerms is the part to start while the outer components should be "used".
- Intellisense will begin showing empty nonsensical namespaces to people who reference the assembly.
- The convention of correspondence between folder structure and namespaces gets broken.

All in all, if I understood your proposal correctly, then I do not think it is a better alternative. I am not sure why you have concerns with the current approach. It plays very well together with scoping in general and with how intellisense works.

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
```diff
@@ -0,0 +1,169 @@
...
+        private const string Version = "1.0";
+
+        private IDictionary<string, Metric> successDurationMetricsByType = null;
+        private IDictionary<string, Metric> failureDurationMetricsByType = null;
+
+        private Metric successDurationMetricDefault = null;
+        private Metric failureDurationMetricDefault = null;
+
+        private string[] dependencyTypes = null;
+        private string dependenccyTypesString = String.Empty;
```
SergeyKanzhelev: why do you need to keep both? For consistency it's better to only have one source of truce

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
```diff
@@ -0,0 +1,169 @@
...
+                        new Dictionary<string, string>() {
+                            [MetricTerms.Autocollection.DependencyCall.PropertyName.Success] = Boolean.FalseString,
+                        });
+            }
+            else
+            {
+                var successDurationsByType = new Dictionary<string, Metric>();
+                var failureDurationsByType = new Dictionary<string, Metric>();
+
+                for (int i = 0; i < depTypes.Length; i++)
```
SergeyKanzhelev: do you pre-populate for efficiency? What about not configured types?

src/Core/Managed/Shared/Extensibility/Metrics/DependencyMetricExtractor.cs
```diff
@@ -0,0 +1,169 @@
...
+            if (durationMetricsByType == null || dependencyType == null)
+            {
+                metric = (isFailed ? this.failureDurationMetricDefault : this.successDurationMetricDefault);
+            }
+            else
+            {
+                bool isKnownType = durationMetricsByType.TryGetValue(dependencyType, out metric);
+                if (! isKnownType)
+                {
+                    metric = (isFailed ? this.failureDurationMetricDefault : this.successDurationMetricDefault);
```
SergeyKanzhelev: this will not going to fly. Pre-configuring dependency types is not a good approach - we do not want this extra configuration problem for our customers

src/Core/Managed/Shared/Extensibility/Metrics/MetricExtractorTelemetryProcessorBase.cs
SergeyKanzhelev: @macrogreg we have a versioned set of expected preaggs that should be documented. It is not a free-form list of aggs.

src/Core/Managed/Shared/Extensibility/Metrics/RequestMetricExtractor.cs
```diff
@@ -0,0 +1,61 @@
...
+                                                                            [MetricTerms.Autocollection.Request.PropertyName.Success] = Boolean.TrueString,
+                                                                        });
+
+            this.responseFailureTimeMetric = MetricManager.CreateMetric(MetricTerms.Autocollection.MetricNames.Request.Duration,
+                                                                        new Dictionary<string, string>() {
+                                                                            [MetricTerms.Autocollection.Request.PropertyName.Success] = Boolean.FalseString,
+                                                                        });
+        }
+
+        public override void ExtractMetrics(ITelemetry fromItem, out bool isItemProcessed)
```
SergeyKanzhelev: please check with VitalyG whether we need to filter out requests marked as `Synthetic` for those metrics.

PR #446, opened 2017-03-07
Introduce Metric Extractors for Auto-Collected Metrics.

macrogreg:
This is instead of the previous PR:
https://github.com/Microsoft/ApplicationInsights-dotnet/pull/441


SergeyKanzhelev:
1. Why do we need a separate namespace? I proposed to move it to `Experimantal` namespace so we can change API if we need in 2.4 without slowly deprecating it.

2. I do not like an idea of a metric every minute just to answer the question whether sampling is on or off. Use a query like [this](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-sampling#how-do-i-know-whether-sampling-is-in-operation) instead

Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
```diff
@@ -374,34 +385,90 @@ private static void TelemetryTypeDoesNotSupportSampling(Func<TelemetryProcessorC
...
             var channelBuilder = new TelemetryProcessorChainBuilder(tc);            
             channelBuilder.UseSampling(samplingPercentage, excludedTypes, includedTypes);
             channelBuilder.Use(next => new StubTelemetryProcessor(next) { OnProcess = t => sentTelemetry.Add(t) });
             
             channelBuilder.Build();
 
-            return tc.TelemetryProcessorChain;
+            TelemetryProcessorChain processors = tc.TelemetryProcessorChain;
+
+            foreach (ITelemetryProcessor processor in processors.TelemetryProcessors)
```
SergeyKanzhelev: Thanks!

macrogreg:
@SergeyKanzhelev  I am not sure I fully understand. We we discussed on the phone the query you suggest will be hitting a storage backend that is different from our core metrics back-end. Also, it will not work at all if a sampler is mis-configured in a way that all telemetry of a certain type is sampled out.

SergeyKanzhelev:
Have you read/addressed comments I left here: https://github.com/Microsoft/ApplicationInsights-dotnet/pull/444?

GlobalStaticVersion.props
```diff
@@ -5,9 +5,9 @@
       Semantic Version. See http://semver.org for full details. 
       Update for every public release. 
     -->
-    <SemanticVersionMajor>2</SemanticVersionMajor>
-    <SemanticVersionMinor>3</SemanticVersionMinor>
-    <SemanticVersionPatch>0</SemanticVersionPatch>
+    <SemanticVersionMajor>5</SemanticVersionMajor>
+    <SemanticVersionMinor>0</SemanticVersionMinor>
+    <SemanticVersionPatch>6</SemanticVersionPatch>
     <PreReleaseMilestone>beta3</PreReleaseMilestone>
```
dnduffy: Should be 2.4.0-beta1.

Test/CoreSDK.Test/Shared/Extensibility/MetricManagerTest.cs
```diff
@@ -150,8 +150,8 @@ public void AggregatedMetricTelemetryIntervalDurationPropertyIsPositiveInteger()
             Assert.Equal(1, aggregatedMetric.Count);
             Assert.Equal(1, aggregatedMetric.Properties.Count);
 
-            Assert.True(aggregatedMetric.Properties.ContainsKey("IntervalDurationMs"));
-            Assert.True(long.Parse(aggregatedMetric.Properties["IntervalDurationMs"]) > 0);
+            Assert.True(aggregatedMetric.Properties.ContainsKey("Microsoft.ApplicationInsights.Metrics.Aggregation.IntervalMs"));
+            Assert.True(long.Parse(aggregatedMetric.Properties["Microsoft.ApplicationInsights.Metrics.Aggregation.IntervalMs"]) > 0);
```
dnduffy: Are these long strings going to need to show in a UI somewhere?

Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
```diff
@@ -374,34 +385,73 @@ private static void TelemetryTypeDoesNotSupportSampling(Func<TelemetryProcessorC
...
+                                                          int samplingPercentage = 10)
         {
             const int ItemsToGenerate = 100;
-            const int SamplingPercentage = 10;
             var sentTelemetry = new List<ITelemetry>();
-            var telemetryProcessorChainWithSampling = CreateTelemetryProcessorChainWithSampling(sentTelemetry, SamplingPercentage, excludedTypes, includedTypes);
+            TelemetryProcessorChain telemetryProcessorChainWithSampling = CreateTelemetryProcessorChainWithSampling(sentTelemetry,
+                                                                                                                    samplingPercentage,
+                                                                                                                    excludedTypes,
+                                                                                                                    includedTypes);
```
dnduffy: nitpick: what is with this formatting?

Test/CoreSDK.Test/Shared/Extensibility/MetricManagerTest.cs
macrogreg: No.

Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
macrogreg: I tried to stick to existing formatting everywhere and StyleCop runs clean. I could have missed something, so please let me know if you would like me to adjust something specific. On the first glance, this looks fine to me.

dnduffy: It's just that now the parameters are on multiple lines and indented far to the right side so they make this diff look crazy.
The default StyleCop rules used to complain if the parameters were not either all on one line or all on their own lines which isn't the case here since one of them is still with the method call but personally I would just format it like before as all one line or put the parameters together on the second line with only a single indent, whatever you prefer.

src/Core/Managed/Shared/Extensibility/Implementation/Metrics/MetricTerms.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    internal static class MetricTerms
+    {
+        private const string MetricPropertiesNamePrefix = "Microsoft.ApplicationInsights.Metrics";
+
+        public static class Aggregation
+        {
+            public static class Interval
+            {
+                public static class Moniker
```
dnduffy: nitpick: I still think you should not try to reinvent namespaces like this.

macrogreg:
I addressed this. ☺

G))

src/Core/Managed/Shared/Extensibility/Implementation/Metrics/DependencyMetricsExtractor.cs
```diff
@@ -0,0 +1,361 @@
...
+                                                            MetricTerms.Autocollection.MetricNames.DependencyCall.Duration,
+                                                            new Dictionary<string, string>()
+                                                            {
+                                                                [MetricTerms.Autocollection.DependencyCall.PropertyNames.Success] = Boolean.FalseString, // FAILURE metric
+                                                                [MetricTerms.Autocollection.DependencyCall.PropertyNames.TypeName] = depType,
+                                                            }));
+                                            }
+                                        });
+                            }
+                            catch (InvalidOperationException)
```
Dmitry-Matveev: It looks like we are using exceptions to control execution flow, is there some other pattern we can use here to achieve the same result without the unnecessary exception?

src/Core/Managed/Shared/Extensibility/Implementation/Metrics/MetricTerms.cs
```diff
@@ -0,0 +1,77 @@
...
+                    public const string ExtractorInfoTemplate = "(Name:{0}, Ver:{1})";      // $"(Name:{ExtractorName}, Ver:{ExtractorVersion})"
+                }
+            }
+        }
+
+        public static class Autocollection
+        {
+            public static class Moniker
+            {
+                public const string Key = MetricPropertiesNamePrefix + ".MetricIsAutocollected";
```
Dmitry-Matveev: If this propagates all the way to AI Endpoint, we will count this huge string towards the bill. We might want to reduce the footprint if it indeed does.

macrogreg: It seems to me that given all the other goo AI counts, this is a drop in the ocean. Name-qualifying auto-added properties is a good way to avoid collisions with user properties. In future versions we will need to add new metric documents to support things like percentiles etc. Than we can move this into the internal context..

src/Core/Managed/Shared/Extensibility/Implementation/Metrics/MetricTerms.cs
macrogreg: I d not want to use namespaces here as discussed on the other thread. If you guys object to this stricture naming I can use just constants with long names instead if you like. :) However, it seems to me like a minor point.

src/Core/Managed/Shared/Extensibility/Implementation/Metrics/DependencyMetricsExtractor.cs
macrogreg: I thought about this for a while. The problem is that CuncurrentDictionary.GetOrAdd does not allow the delegate to say "I changed my min, do not add anything". Null values are also allowed. For example, the following code:

            var cd = new ConcurrentDictionary<string, string>();

            cd.GetOrAdd("key1", (k) => k + "-value");
            cd.GetOrAdd("key2", (k) => null);

            Console.WriteLine("Entries in cd: " + cd.Count);

            try
            {
                cd.GetOrAdd("key3", (k) => throw new Exception("Not adding an element"));
            }
            catch (Exception ex)
            {
                Console.WriteLine("Observed error: " + ex);
            }

results in this output:

Entries in cd: 2
Observed error: System.Exception: Not adding an element
   at MiscTests01.Program.<>c.<Main>b__2_2(String k) in C:\00\Code\Sandbox\MiscTests01\MiscTests01\Program.cs:line 47
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at MiscTests01.Program.Main(String[] args) in C:\00\Code\Sandbox\MiscTests01\MiscTests01\Program.cs:line 47

So I cannot see a better way, but if you have a specific suggestion, I'd love to know. :)
In the meantime, let me expand on the exception message in case someone stumbles over the first chance... :)

src/Core/Managed/Shared/Extensibility/Implementation/Metrics/MetricTerms.cs
Dmitry-Matveev: In the situation with drops there is always the last drop. Currently the billing size of the metric is about 600 bytes. The string of 50 characters increases the metric size by 8.33%, rather a large drop.

macrogreg: @Dmitry-Matveev I have calculated the billing size for metrics recently and it was around 2KBytes. 50 chars in UTF8 is 5 bytes, which is on the order of 2%, not 8%. Also, most of the long names appear only in auto-collected metrics, which are quite few.

src/Core/Managed/Shared/Extensibility/Implementation/Metrics/MetricTerms.cs
dnduffy: I personally find it confusing when reading code that looks like a namespace and then isn't because it's harder to locate where things come from when you don't have a full IDE with intellisense running (and I run into that situation regularly).  I admit that it is a nitpick and personal preference but in this case I have a strong preference for constants with long names.

Test/CoreSDK.Test/Shared/Extensibility/MetricManagerTest.cs
dnduffy: That's good, but to Dmitry's point below, are they going to increase the cost of transmission and billing to the customer?

PR #463, opened 2017-03-14
Exclude events from sampling

SergeyKanzhelev:
Do not sample events out by default


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 This changelog will be used to generate documentation on [release notes page](http://azure.microsoft.com/en-us/documentation/articles/app-insights-release-notes-dotnet/).
 
+## Version 2.4.0-beta1
+- Event telemetry is set to be sampled separately from all other telemetry types. It potentially can double the bill. The reason for this change is that Events are mostly used for usage analysis and should not be subject to sampling on high load of requests and dependencies. Edit `ApplicationInsights.config` file to revert to the previous behavior.
```
AlexBulankou: Why not apply the same approach to traces then - sample them separately from requests and dependencies? Traces are the same category, i.e. telemetry that is user-generated as opposed to system-generated.

SergeyKanzhelev: I think there is a big difference between Events and Traces. You enable Traces in bulk by using one of trace adaptors. Events are always tracked by explicit call to `TrackEvent`. 

Also I'd expect to see traces for correlated requests. So sampling together makes sense. When events represent a usage event and may not be interesting outside of the request context.

PR #472, opened 2017-03-15
Move operation to shared

### LINKED ISSUE #454: Move the Operation.XX core projects into Core\Shared
Following on from #451 we should move the core from the two flavors of Operation.XX.Shared projects into Core\Shared with and eliminate those shared projects.

pharring:
Fixes #454 

The trick to this change was agreeing on a name for the 'neutral' version of ```OperationContextForCallContext``` and ```OperationContextForAsyncLocal```. I really wanted to call it ```OperationContext``` but, of course, that's already in use. In the end I settled on "OperationContextForCallContext" since it describes what it does even though it might be misleading about the implementation. Similarly for ```CallContextHelpers``` - it describes what it does, but it might be implemented using ```AsyncLocal<T>```. Once I'd settled on that, fortunately, all the other code differences melted away, modulo one or two conditional compilation directives.


SergeyKanzhelev:
Now it has conflicts... can you pls resolve

pharring:
@SergeyKanzhelev conflicts resolved.

SergeyKanzhelev:
I was thinking to improve code by storing `RequestTelemetry` object itself in `AsyncLocal`. This way it will be less allocations and it will allow to update telemetry item and context on the thread the same time.

I like the merge you've done though to simplify logic. I think we can still will be able to do what I planned anyway - let me know if you think otherwise.

SergeyKanzhelev:
@lmolkova who will be doing some correlation work...

pharring:
Yes, storing the ```RequestTelemetry```` (actually, the ```OperationTelemetry``` base class) as the context object seems reasonable. It does subtly change the observable implementation, though: Users would then be able to modify the parent request context after we have stored it. However, I think that might actually be less surprising than the current implementation which snapshots only some ```OperationContext``` properties.
Oh, it also might not work for .Net 4.0 and .Net 4.5 which require that the object stored in the CallContext is serializable :-(

SergeyKanzhelev:
Yes, it will not work in 4.0... this stopped me from implementing it immediately.

I was thinking of three options for 4.0:
- Check whether `AsyncLocal` is available using reflection (case when compiled for 4.0 and running on 4.5) and switch to `AsyncLocal`
- Implement a lookup table
- do nothing, just document the limitation

Maybe @lmolkova can help with making this change

lmolkova:
The OperationContext similar to Activity API.
My intention is to remove OperationContextForCallContext on net46+ and substitute it with Activity. 

However:
1. We discussed with @SergeyKanzhelev before that bringing System.Diagnostic.DiagnosticSource dependency to Base SDK may be a problem. **I'd like to reconsider this option**. Otherwise we would have duplicated context on 4.6 and competing/redundant telemetry initialization in Web and Base.
2. There is no Activity for net40 and net45, so current implementation may remain here.

SergeyKanzhelev:
@lmolkova we should preserve the scenario when a single dll is sufficient to report telemetry. Having said that I realized that we can take a reference anyway. We just need to ensure that code will function (gratefully degrade) when diagnostics source dll cannot be found.

Do you think it is doable?

PR #475, opened 2017-03-15
Add missing dependency on NETStandard.Library for .netstandard target

pharring:
Also remove unnecessary dependency on EventSource.Register

(Based on a review from @eerhardt)


SergeyKanzhelev:
we still need EventRegister, do we? This package validates that EventSource is authored correctly. Why removing it?

pharring:
We need only the Redist piece of EventSource. We do not need EventRegister. It's a tool that is used to create the EventSource class definitions. It applies only to the .Net 4.0 project (it's included in newer frameworks without requiring a separate package). Core.csproj builds fine without it.

Also, it's not in the .nuspec, so customers were not already getting a dependency.

SergeyKanzhelev:
@pharring - EventRegister helps detecting the issues when I do "Write" with the different ID than defined in method attributes. Are you saying it is now integrated into the framework? See this discussion for instance: https://github.com/Microsoft/ApplicationInsights-SDK-Labs/issues/99

pharring:
Ah, I think you're right. EventRegister _also_ works for System.Diagnostics, so it's not limited to net40.
Looks like we can add it back for all frameworks, but it should _not_ be in the nuspec since it's just used during the build. Stand by, I'll update this PR.

pharring:
@SergeyKanzhelev fixed.

PR #478, opened 2017-03-18
Add CorrelationContext and InternalsVisibleTo Web and DependencyColle…

lmolkova:
This change 
* adds internal CorrelationContext to TelemetryContext
* InternalVisibleTo attribute to expose internals to AI.Web and DependencyCollector

This is a part of dependency correlation feature being implemented in AI.Web and DependencyCollector: https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/368

### DIFF
--- a/src/Core/Managed/Shared/DataContracts/TelemetryContext.cs
+++ b/src/Core/Managed/Shared/DataContracts/TelemetryContext.cs
@@ -14,6 +14,7 @@
     public sealed class TelemetryContext
     {
         private readonly IDictionary<string, string> properties;
+        private readonly IDictionary<string, string> correlationContext;
         private readonly IDictionary<string, string> tags;
 
         private string instrumentationKey;
@@ -40,6 +41,7 @@
             Debug.Assert(properties != null, "properties");
             this.properties = properties;
             this.tags = new ConcurrentDictionary<string, string>();
+            this.correlationContext = new ConcurrentDictionary<string, string>();
         }
 
         /// <summary>
@@ -119,6 +121,15 @@
         public IDictionary<string, string> Properties
         {
             get { return this.properties; }
+        }
+
+        /// <summary>
+        /// Gets a Correlation-Context for the operation.
+        /// <see href="https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md"/> 
+        /// </summary>
+        internal IDictionary<string, string> CorrelationContext
+        {
+            get { return this.correlationContext; }
         }
 
         internal InternalContext Internal
--- a/src/Core/Managed/Shared/DataContracts/TelemetryContext.cs
+++ b/src/Core/Managed/Shared/DataContracts/TelemetryContext.cs
@@ -1,5 +1,6 @@
 ﻿namespace Microsoft.ApplicationInsights.DataContracts
 {
+    using System;
     using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Diagnostics;
@@ -14,7 +15,7 @@
     public sealed class TelemetryContext
     {
         private readonly IDictionary<string, string> properties;
-        private readonly IDictionary<string, string> correlationContext;
+        private readonly Lazy<IDictionary<string, string>> correlationContext;
         private readonly IDictionary<string, string> tags;
 
         private string instrumentationKey;
@@ -41,7 +42,7 @@
             Debug.Assert(properties != null, "properties");
             this.properties = properties;
             this.tags = new ConcurrentDictionary<string, string>();
-            this.correlationContext = new ConcurrentDictionary<string, string>();
+            this.correlationContext = new Lazy<IDictionary<string, string>>(() => new Dictionary<string, string>());
         }
 
         /// <summary>
@@ -129,7 +130,7 @@
         /// </summary>
         internal IDictionary<string, string> CorrelationContext
         {
-            get { return this.correlationContext; }
+            get { return this.correlationContext.Value; }
         }
 
         internal InternalContext Internal
--- a/src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
+++ b/src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
@@ -81,6 +81,14 @@
         internal long BeginTimeInTicks { get; set; }
 
         /// <summary>
+        /// Sets random operation Id.
+        /// </summary>
+        public void GenerateOperationId()
+        {
+            this.Id = Convert.ToBase64String(BitConverter.GetBytes(WeakConcurrentRandom.Instance.Next()));
+        }
+
+        /// <summary>
         /// Sanitizes the properties based on constraints.
         /// </summary>
         void ITelemetry.Sanitize()
--- a/src/Core/Managed/Shared/Extensibility/OperationCorrelationTelemetryInitializer.cs
+++ b/src/Core/Managed/Shared/Extensibility/OperationCorrelationTelemetryInitializer.cs
@@ -1,6 +1,7 @@
 ﻿namespace Microsoft.ApplicationInsights.Extensibility
 {
     using Implementation;
+    using Microsoft.ApplicationInsights;
     using Microsoft.ApplicationInsights.Channel;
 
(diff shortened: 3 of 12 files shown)


src/Core/Managed/Shared/DataContracts/TelemetryContext.cs
```diff
@@ -121,6 +123,15 @@ public LocationContext Location
             get { return this.properties; }
         }
 
+        /// <summary>
+        /// Gets a Correlation-Context for the operation.
+        /// <see href="https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md"/> 
+        /// </summary>
+        internal IDictionary<string, string> CorrelationContext
```
lmolkova: let's not expose it now.
It could be confusing for customers and we don't really expect them to use it. Moreover, we encourage them to avoid it.
We can always expose it later, it we have a real need for it.

src/Core/Managed/Shared/Properties/AssemblyInfo.cs
```diff
@@ -9,11 +9,20 @@
...
+[assembly: InternalsVisibleTo("Microsoft.AI.DependencyCollector" + AssemblyInfo.AiWebPublicKey)]
+
 [assembly: InternalsVisibleTo("Microsoft.ApplicationInsights.TelemetryChannel.Net40.Tests" + AssemblyInfo.PublicKey)]
 [assembly: InternalsVisibleTo("Microsoft.ApplicationInsights.TelemetryChannel.Net45.Tests" + AssemblyInfo.PublicKey)]
 
 internal static class AssemblyInfo
 {
     // Public key; assemblies are delay signed or OSS signed
     public const string PublicKey = ", PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9";
+#if PUBLIC_RELEASE
```
lmolkova: please advise how to change build tools/scripts to use it

src/Core/Managed/Shared/Extensibility/OperationCorrelationTelemetryInitializer.cs
```diff
@@ -46,6 +46,29 @@ public void Initialize(ITelemetry telemetryItem)
...
+                foreach (var item in parentContext.CorrelationContext)
+                {
+                    if (!telemetryItem.Context.CorrelationContext.ContainsKey(item.Key))
+                    {
+                        telemetryItem.Context.CorrelationContext.Add(item);
+                    }
+                }
+            }
+
+            // TODO: change backend so it tracts correlaton context as properties, so we don't need to copy them
```
pharring: "tracks correlation" (two typos)

src/Core/Managed/Shared/Extensibility/OperationCorrelationTelemetryInitializer.cs
```diff
@@ -46,6 +46,29 @@ public void Initialize(ITelemetry telemetryItem)
...
+            // we don't expect any Correlaiton-Context at all, or in the worst case expect a few item,
```
pharring: "Correlation" (typo)

src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
```diff
@@ -81,6 +81,14 @@ public DateTimeOffset StartTime
         internal long BeginTimeInTicks { get; set; }
 
         /// <summary>
+        /// Sets random operation Id.
+        /// </summary>
+        public void GenerateOperationId()
```
pharring: Should this be public? Internal, perhaps.
Should it throw if ```Id``` is already set?

src/Core/Managed/Shared/TelemetryClientExtensions.cs
```diff
@@ -32,8 +35,9 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
         /// <param name="operationName">Name of the operation that customer is planning to propagate.</param>
         /// <param name="operationId">Operation ID to set in the new operation.</param>
         /// <param name="parentOperationId">Optional parent operation ID to set in the new operation.</param>
+        /// <param name="correlationContext">CorrelationContext that is added to telemetry properties and propagated with outgoing HTTP calls.</param>
```
pharring: Nit: prefix with "Optional"

src/Core/Managed/Shared/TelemetryClientExtensions.cs
```diff
@@ -57,9 +61,22 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
                 operationTelemetry.Context.Operation.ParentId = parentOperationId;
             }
 
+            if (correlationContext != null)
+            {
+                foreach (var item in correlationContext)
```
pharring: We're merging dictionaries here. Perhaps that could be extracted into a helper method. Consider an extension method on ```IDictionary<string, string>``` so you can write:
```C#
operationTelemetry.Context.CorrelationContext.Merge(correlationContext);

File: src/Core/Managed/Shared/DataContracts/TelemetryContext.cs [hunk 8]
```diff
@@ -40,6 +41,7 @@ internal TelemetryContext(IDictionary<string, string> properties)
             Debug.Assert(properties != null, "properties");
             this.properties = properties;
             this.tags = new ConcurrentDictionary<string, string>();
+            this.correlationContext = new ConcurrentDictionary<string, string>();
```
pharring: ```ConcurrentDictionary<K,V>``` is pretty heavyweight. To be creating one on every telemetry item could be a performance problem. (I notice we already have one for ```tags```).
I have a few pieces of advice:
1. If the correlationContext is often empty, then use a lazy initializer to create the dictionary on first use. Take care that just _reading_ the empty context doesn't create a new dictionary.
2. Consider an ```ImmutableDictionary<K,V>``` (from System.Collections.Immutable).
3. If you must use a ```ConcurrentDictionary``` call the constructor that allows you to set the maximum concurrency, and set it to 1. (You also have to set the expected capacity, so I would set it to zero). That way you get thread-safety without the bulk of the overhead.

pharring: Actually, I really like the idea of ```ImmutableDictionary<K,V>``` because they're zero-cost to copy which will help in ```StartOperation```
However, I know @SergeyKanzhelev has previously expressed a desire not to include additional dependencies.

src/Core/Managed/Shared/DataContracts/TelemetryContext.cs
pharring: I agree.

src/Core/Managed/Shared/TelemetryClientExtensions.cs
lmolkova: I was thinking about it and want to remove this parameter at all: there is no way get correlation-context (as TelemetryContext.CorrelationContext is internal) and should be no way to  set it in user code.
If there is a Correlation-context from upstream service, we just propagate it further. If  AI SDK needs to pass something for internal needs - it's always there as internal.

src/Core/Managed/Shared/Extensibility/Implementation/OperationTelemetry.cs
lmolkova: There was a public extension method in OperationTelemetryExtensions and this is a small refactoring that substitutes it with class method. DependencyTelemetry and RequestTelemetry initialized Id with `Convert.ToBase64String(BitConverter.GetBytes(WeakConcurrentRandom.Instance.Next()))` and now call `GetOperationId`

Ican make it internal since there is still extension method (in case someone called OperationTelemetryExtensions.GenerateOperationId(operation)).

Throwing exception would change behavior the extension method has now.

PR #481, opened 2017-03-23
Remove MetricManager APIs from 2.3 public stable release.

macrogreg:


src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -195,7 +195,7 @@ public void TrackTrace(TraceTelemetry telemetry)
         /// <param name="name">Metric name.</param>
         /// <param name="value">Metric value.</param>
         /// <param name="properties">Named string values you can use to classify and filter metrics.</param>
-        [Obsolete("This method is obsolete. Use TrackMetric(metricTelemetry) method to send pre-aggregated metric data or MetricManager class to create metrics.")]
+        [Obsolete("This method is obsolete. Use TrackMetric(metricTelemetry) method to send pre-aggregated metric data.")]
```
SergeyKanzhelev: I don't think we should deprecate this method if we removed the `MetricManager`.

macrogreg:
@SergeyKanzhelev "Remove the Obsolete from TrackMetric" is completed as requested.

PR #483, opened 2017-03-25
Expose logic for determining Request Success from ResponseCode for internal consumption and use it in the Request Metrics Extractor.

macrogreg:


Test/CoreSDK.Test/Shared/Extensibility/MetricManagerTest.cs
```diff
@@ -150,8 +150,8 @@ public void AggregatedMetricTelemetryIntervalDurationPropertyIsPositiveInteger()
             Assert.Equal(1, aggregatedMetric.Count);
             Assert.Equal(1, aggregatedMetric.Properties.Count);
 
-            Assert.True(aggregatedMetric.Properties.ContainsKey("Microsoft.ApplicationInsights.Metrics.Aggregation.IntervalMs"));
-            Assert.True(long.Parse(aggregatedMetric.Properties["Microsoft.ApplicationInsights.Metrics.Aggregation.IntervalMs"]) > 0);
+            Assert.True(aggregatedMetric.Properties.ContainsKey("Microsoft.AppInsights.Metrics.Aggregation.IntervalMs"));
```
SergeyKanzhelev: you should refrain from using `AppInsights`. `ApplicationInsights` is the correct term

src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
```diff
@@ -223,19 +223,31 @@ void ITelemetry.Sanitize()
...
-                {
-                    this.Success = true;
-                }
+                this.Success = this.GetSuccessFromResponseCode();
             }
 
             this.Context.SanitizeTelemetryContext();
         }
+
+        internal bool GetSuccessFromResponseCode()
```
SergeyKanzhelev: I don't think this logic belong here. If customer override the response code - customer should also override the `Success` value. 

I'd say - remove this logic completely with the mention in CHANGELOG.md

Test/CoreSDK.Test/Shared/Extensibility/MetricManagerTest.cs
macrogreg: @SergeyKanzhelev As mentioned by email, this code was not included in the previous merge by mistake. The change being tested here was previously approved by the team and merged. If we want to go back on this, we can, but let us not block this PR on that.

src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
macrogreg: @SergeyKanzhelev , refactoring this is not in scope for this particular change. We can do this separately, but right now I need to do a minimal, non-impacting change to the SDK that will unblock my progress. This PR is exactly that. All I did was to move a few lines into a separate method so that I can reuse them.

Test/CoreSDK.Test/Shared/Extensibility/MetricManagerTest.cs
SergeyKanzhelev: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/484

src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs
SergeyKanzhelev: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/485

PR #486, opened 2017-03-29
Support initialization from Activity and use Activity in StartOperation

lmolkova:
This is a part of dependency correlation feature being implemented in .NET. ASP.NET (Core) and AppInsights (Classic and Core)

OperationCorrelaitonTelemetryInitializer initializes telemetry from the Activity.
Notes:
- StartOperation behavior does not change, however it forces OperationTelemetry.Id generation. Id becomes hierarchical
- if pure AppInsights.dll is used and DiagnosticSource.dll is not available, StartOperation/ falls back to CallContext/AsyncLocal

CorrelationHelper class is introduced to make correlation possible for AppInsights.Web on 4.0 where there is no Activity.

Some background about the feature:
- [Activity User Guide](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/ActivityUserGuide.md)
- [Http protocol](https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md)
- [HttpClient instrumentation on .NET Core ](https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs)


Test/CoreSDK.Test/Shared/TelemetryClientExtensionTests.cs
```diff
@@ -77,14 +89,23 @@ public void StartDependencyTrackingCreatesADependencyTelemetryItemWithTimeStamp(
...
         public void StartDependencyTrackingAddsOperationContextStoreToCallContext()
         {
             Assert.IsNull(CallContextHelpers.GetCurrentOperationContext());
             var operation = this.telemetryClient.StartOperation<DependencyTelemetry>(operationName: null);
             Assert.IsNotNull(CallContextHelpers.GetCurrentOperationContext());
         }
-
+#else
+        [TestMethod]
+        public void StartDependencyTrackingAddsOperationContextStoreToCallContext()
```
pharring: Rename method? e.g. ```StartDependencyTrackingAddsOperationToCurrentActivity```

Test/CoreSDK.Test/Shared/TelemetryClientExtensionTests.cs
```diff
@@ -131,6 +168,30 @@ public void StartDependencyTrackingHandlesMultipleContextStoresInCallContext()
             this.telemetryClient.StopOperation(operation);
             Assert.IsNull(CallContextHelpers.GetCurrentOperationContext());
         }
+#else
+
+        [TestMethod]
+        public void StartDependencyTrackingHandlesMultipleContextStoresInCallContext()
```
pharring: Rename

src/Core/Managed/Shared/ActivityProxy.cs
```diff
@@ -0,0 +1,21 @@
...
+    internal static class ActivityProxy
+    {
+        public static bool TryRun(Func<bool> method)
+        {
+            Debug.Assert(method != null, "Method must not be null");
+            try
+            {
+                return method.Invoke();
+            }
+            catch (System.IO.FileNotFoundException)
```
pharring: This needs an explanatory comment. What is ```TryRun``` used for and why is FileNotFoundException handled specially?

src/Core/Managed/Shared/Extensibility/Implementation/OperationHolder.cs
```diff
@@ -74,17 +78,38 @@ protected virtual void Dispose(bool disposing)
                     if (!this.isDisposed)
                     {
                         var operationTelemetry = this.Telemetry;
+                        bool isActivityEnabled = false;
+#if !NET40
+                        isActivityEnabled = ActivityProxy.TryRun(() =>
```
pharring: Is this doing a runtime availability check for the DiagnosticsSource assembly?
If so, consider making ```isActivityEnabled``` a static field so it's initialized just once and not each time this is called.

src/Core/Managed/Shared/Extensibility/OperationCorrelationTelemetryInitializer.cs
```diff
@@ -22,27 +28,86 @@ public void Initialize(ITelemetry telemetryItem)
...
+            isActivityEnabled = ActivityProxy.TryRun(() =>
             {
-                var parentContext = CallContextHelpers.GetCurrentOperationContext();
-                if (parentContext != null)
+                if (Activity.Current == null)
+                {
+                    return false;
+                }
+
+                var currentActivity = Activity.Current;
```
pharring: Move this up above line 35 and use it for the null check. That way you access the ```.Current``` property just once. It'll be cheaper (since ```Activity.Current``` isn't simply a static field).

src/Core/Managed/Shared/Extensibility/Implementation/OperationHolder.cs
```diff
@@ -74,17 +78,38 @@ protected virtual void Dispose(bool disposing)
...
+                        {
+                            var currentActivity = Activity.Current;
+                            if (currentActivity == null || operationTelemetry.Id != currentActivity.ParentId ||
+                                operationTelemetry.Context.Operation.Name != Activity.Current.Tags.FirstOrDefault(t => t.Key == "OperationName").Value)
```
pharring: The expression: ```.Tags.FirstOrDefault(t => t.Key == "OperationName").Value``` appears 4 times in this review. Can it be refactored into a helper method? (Extension method on ```Activity```, perhaps?)

pharring: Also, replace the string ```"OperationName"``` everywhere with a named constant.

src/Core/Managed/Shared/TelemetryClientExtensions.cs
```diff
@@ -105,14 +107,60 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
                 operationTelemetry.Context.Operation.Name = operationTelemetry.Name;
             }
 
+            bool isActivityEnabled = false;
+#if !NET40
+            isActivityEnabled = ActivityProxy.TryRun(() =>
+            {
+                bool operationNameIsSet = false;
+                var activity = new Activity("Internal");
```
pharring: Is there any significance to the string ```"Internal"```? Maybe it should be more descriptive. e.g. "Application Insights Operation Context". Is there value in keeping it short?

src/Core/Managed/Shared/TelemetryClientExtensions.cs
```diff
@@ -105,14 +107,60 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
+                if (!string.IsNullOrEmpty(operationTelemetry.Context.Operation.Name))
```
pharring: Consider extracting the expression ```operationTelemetry.Context.Operation``` into a local variable to avoid repetition.

src/Core/Managed/Shared/TelemetryClientExtensions.cs
```diff
@@ -105,14 +107,60 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
+                    operationNameIsSet = true;
+                }
+
+                if (Activity.Current == null)
+                {
+                    if (!string.IsNullOrEmpty(operationTelemetry.Context.Operation.Id))
+                    {
+                        activity.SetParentId(operationTelemetry.Context.Operation.Id);
+                    }
+                                else if (!string.IsNullOrEmpty(operationTelemetry.Context.Operation.ParentId))
```
pharring: Nit: Strange spacing here.

src/Core/Managed/Shared/TelemetryClientExtensions.cs
```diff
@@ -105,14 +107,60 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
-            var operationContext = new OperationContextForCallContext();
-            operationContext.ParentOperationId = operationTelemetry.Id;
-            operationContext.RootOperationId = operationTelemetry.Context.Operation.Id;
-            operationContext.RootOperationName = operationTelemetry.Context.Operation.Name;
-            CallContextHelpers.SaveOperationContext(operationContext);
+            if (!isActivityEnabled)
+            {
+                // Update the call context to store certain fields that can be used for subsequent operations.
+                var operationContext = new OperationContextForCallContext();
+                operationContext.ParentOperationId = operationTelemetry.Id;
```
pharring: Nit: Could use initializer syntax here. i.e.
```C#
var operationContext = new OperationContextForCallContext
{
    ParentOperationId = operationTelemetry.Id,
    RootOperationId = operationTelemetry.Context.Operation.Id,
    // etc.
}

File: src/Core/NuGet/Package.nuspec [hunk 11]
```diff
@@ -22,14 +22,18 @@
...
       </group>
       <group targetFramework="net45">
+        <dependency id="System.Diagnostics.DiagnosticSource" version="4.4.0-beta-25124-07" />
       </group>
       <group targetFramework="net46">
+        <dependency id="System.Diagnostics.DiagnosticSource" version="4.4.0-beta-25124-07" />
       </group>
       <group targetFramework="wp8">
         <dependency id="Microsoft.Bcl.Async" version="1.0.168" />
+        <dependency id="System.Diagnostics.DiagnosticSource" version="4.4.0-beta-25124-07" />
```
pharring: Not needed. We no longer build the wp8 flavor. It's just a binary copy of an older stable version.

Test/CoreSDK.Test/Net45/Core.Net45.Tests.csproj
```diff
@@ -30,6 +30,9 @@
     </Reference>
     <Reference Include="System" />
     <Reference Include="Microsoft.VisualStudio.QualityTools.UnitTestFramework" />
+    <Reference Include="System.Diagnostics.DiagnosticSource, Version=4.0.2.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
+      <HintPath>..\..\..\..\packages\System.Diagnostics.DiagnosticSource.4.4.0-beta-25124-07\lib\net45\System.Diagnostics.DiagnosticSource.dll</HintPath>
```
lmolkova: It seems those versions are not related. E.g. if you install stable DiagnosticSource you'll see 
```
    <Reference Include="System.Diagnostics.DiagnosticSource, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
      <HintPath>..\packages\System.Diagnostics.DiagnosticSource.4.3.0\lib\net46\System.Diagnostics.DiagnosticSource.dll</HintPath>
      <Private>True</Private>
    </Reference>
```

The same happens with any other .NET package.

lmolkova:
There is an open question on how to test AppInsights.dll standalone (without DiagnosticSource and Activity) on .NET4.5/4.6.

So far I tested it manually, but since there is a requirement that it could be used as a standalone dll, it should be automatically tested. It is fragile so any future changes may easily break it.

Are there any functional tests/manual release testing for Core SDK?

NuGet.config
```diff
@@ -18,5 +18,6 @@
   <packageSources>
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
     <add key="applicationinsights" value="https://www.myget.org/F/applicationinsights/api/v3/index.json" />
+    <add key="dotnet" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" protocolVersion="3" />
```
SergeyKanzhelev: will it be available on nugget before the 2.4 release? Would you please create an issue to update the reference and remove this source

SergeyKanzhelev:
@lmolkova we do not have many tests. Do you have suggestion on how we can test it?

NuGet.config
lmolkova: Yes, .NET team agreed to release DiagSource when we will need it. I assume the deadline is around mid of April?

I created #487 to track it

lmolkova:
@SergeyKanzhelev 
We can have a separate app/test lib that references AppInsights.dll only and run it in the CI pipeline.

I've created #488 and will take care of it after 2.4.0 code freeze, since it's not urgent

SergeyKanzhelev:
do you need a signed build of Base to try in dotnet-server repository? I can kick off one

src/Core/Managed/Shared/ActivityExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+    internal static class ActivityExtensions
+    {
+        private const string OperationNameTag = "OperationName";
+        private static bool isInitialized = false;
+        private static bool isEnabled = false;
+
+        /// <summary>
+        /// Checks that Activity API could be called (DiagnosticSource DLL is loaded).
+        /// </summary>
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
```
pharring: ```NoInlining``` makes sense, but ```NoOptimization``` shouldn't be necessary. In fact, you probably want a fast, inlinable check for isEnabled. Refactor this into the fast and slow parts like this:
i.e.
```C#
// Fast, inlinable
public static bool IsActivityEnabled() => isInitialized ? isEnabled : SlowPath();

// Slow
[MethodImpl(MethodImplOptions.NoInlining)]
public static bool SlowPath()
{
    // etc.
}

File: src/Core/Managed/Shared/ActivityExtensions.cs [hunk 15]
```diff
@@ -0,0 +1,55 @@
...
+            }
+
+            try
+            {
+                var activity = new Activity("IsEnabled");
+                isEnabled = true;
+                isInitialized = true;
+                return true;
+            }
+            catch (System.IO.FileNotFoundException)
```
pharring: It seems a bit indirect to rely on ```FileNotFoundException``` to catch this. I'm curious how you don't end up with a ```TypeLoadException``` when this method is JIT-ted (but the DiagnosticsSource assembly is not present). Is that, perhaps, why you added ```NoOptimization```?

A more direct approach is to try and instantiate the ```Activity``` type either directly via ```Activator.CreateInstance``` using the fully-qualified name or indirectly by calling a method that uses it and catching the ```TypeLoadException```.

src/Core/Managed/Shared/ActivityExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        private const string OperationNameTag = "OperationName";
+        private static bool isInitialized = false;
+        private static bool isEnabled = false;
+
+        /// <summary>
+        /// Checks that Activity API could be called (DiagnosticSource DLL is loaded).
+        /// </summary>
+        [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)] 
+        public static bool IsActivityEnabled()
```
pharring: Suggest ```IsActivityAvailable```  ?

src/Core/Managed/Shared/Extensibility/Implementation/OperationHolder.cs
```diff
@@ -74,17 +77,36 @@ protected virtual void Dispose(bool disposing)
                     if (!this.isDisposed)
                     {
                         var operationTelemetry = this.Telemetry;
+                        bool isActivityEnabled = false;
```
pharring: ```isActivityAvailable``` ?

src/Core/Managed/Shared/Extensibility/Implementation/OperationHolder.cs
```diff
@@ -74,17 +77,36 @@ protected virtual void Dispose(bool disposing)
...
+#if !NET40
+                        if (isActivityEnabled = ActivityExtensions.IsActivityEnabled())
+                        {
+                            var currentActivity = Activity.Current;
+                            if (currentActivity == null || operationTelemetry.Id != currentActivity.ParentId ||
+                                operationTelemetry.Context.Operation.Name != Activity.Current.GetOperationName())
```
pharring: ```Activity.Current.GetOperationName()``` -> ```currentActivity.GetOperationName()```

src/Core/Managed/Shared/TelemetryClientExtensions.cs
```diff
@@ -93,26 +94,78 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
-            if (string.IsNullOrEmpty(operationTelemetry.Context.Operation.Name))
+            if (string.IsNullOrEmpty(telemetryContext.Name))
             {
-                operationTelemetry.Context.Operation.Name = operationTelemetry.Name;
+                telemetryContext.Name = operationTelemetry.Name;
             }
 
+            bool isActivityEnabled = false;
+#if !NET40
+            if (ActivityExtensions.IsActivityEnabled())
```
pharring: You need to assign ```isActivityEnabled``` within the if statement here (like you did elsewhere).

src/Core/Managed/Shared/TelemetryClientExtensions.cs
```diff
@@ -93,26 +94,78 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
+            {
+                bool operationNameIsSet = false;
+
+                var childActivity = new Activity("Microsoft.ApplicationInsights.OperationContext");
```
pharring: Should this string be a ```const``` somewhere and should we check for it within ```StopOperation``` ?

src/Core/Managed/Shared/TelemetryClientExtensions.cs
```diff
@@ -93,26 +94,78 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
+                    if (!string.IsNullOrEmpty(telemetryContext.Id))
+                    {
+                        childActivity.SetParentId(telemetryContext.Id);
+                    }
+                    else if (!string.IsNullOrEmpty(telemetryContext.ParentId))
+                    {
+                        childActivity.SetParentId(telemetryContext.ParentId);
+                    }
+
+                    operationTelemetry.Id = childActivity.ParentId;
```
pharring: This doesn't look right. If the Id has already been set (e.g. by the caller or via a TelemetryInitializer), then we're overwriting it here. What is the purpose?

src/Core/Managed/Shared/TelemetryClientExtensions.cs
```diff
@@ -93,26 +94,78 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
+                    if (!operationNameIsSet)
+                    {
+                        var parentOperationName = parentActivity.GetOperationName();
+                        if (!string.IsNullOrEmpty(parentOperationName))
+                        {
+                            childActivity.SetOperationName(parentOperationName);
+                        }
+                    }
+
+                    operationTelemetry.Id = parentActivity.Id;
```
pharring: Again, this doesn't look right.

pharring:
@lmolkova Please see my latest comments (sorry for not getting to it before it was merged).

SergeyKanzhelev:
@lmolkova please also update `CHANGELOG.md` @pharring sorry for merging fast. Trying to unblock the work in `-server` repo.

src/Core/Managed/Shared/ActivityExtensions.cs
lmolkova: TypeLoadException will be thrown if the assembly was loaded, but it does not have the Activity type in it. In our case, we are sure Activity is there (because we insist on DiagnosticSource 4.4.0), but we assume that someone will try to run AppInsights.dll without the DiagnosticSource.dll. In this case assembly will not be found and it will cause FileNotFoundException.

I like the approach with Activator.CreateInstance and I assume it will not be optimized away.

lmolkova: In fact, BindingFlags that are needed for Activator.CreateInstance are not available in netstandard 1.3. So, I'll call Assembly.Load

src/Core/Managed/Shared/TelemetryClientExtensions.cs
lmolkova: It makes sense to make it const, but in the StopOperation we check for activity.Id which is globally unique, so there is no reason to check for Activity name. 

I also want to stop checking (AI) operation name: it's redundant and could cause telemetry being lost if operation name changes

lmolkova: Operation Name in ApplicationInsights is quite unfortunately not the same as Activity.OperationName,

The reason is that AppInsights for ASP.NET classic uses incoming request routes for the Operation.Name. And the routes are calculated somewhere late in the pipeline, after Activity for the request is started.
Activity Name could only be provided in ctor and could never change. 
Also Activity does not support null or empty names.

So AI will use special Tag  "OperationName" for it's own  operation name.

src/Core/Managed/Shared/TelemetryClientExtensions.cs
lmolkova: OperationTelemetry.Id is used as a Request-Id (or x-ms-request-id) header.
With Activity/new Http protocol support, it becomes hierarchical.

So, when the incoming request is received, it takes Request-Id and makes it a ParentId of the RequestTelemetry: e.g. 
```
requestTelemetry.Opertion.Context.ParentId = "|guid."
requestTelemetry.Id = "|guid.abc123_"
```

When the dependency call happens, new Activity is created for it (through StartOperation) or directly by Activity.Start(). So the dependency gets an Id:
```
dependencyTelemetry.Opertion.Context.ParentId = "|guid.abc123_"
dependencyTelemetry.Id = "|guid.abc123_1."
```
In case of HTTP, dependencyTelemetry.Id is injected into outgoing request `Request-Id` header.

AppInsights gets root node (`guid`) from each telemetry and sets `Telemetry.Operation.Context.Id` from it.

So what happens here: 
1. Assuming there is no custom Telemetry.Id generation 
if someone has called StartOperation without any parent Activity, operationTelemetry.Id is default one, generated by AppInsgihts in telemetry ctor. We need to align it with the root Operation.Id once we know it (after initialization).
If someone has called StartOperation when there is a parent activity: that activity represents context for this operation, we always make sure Telemetry.Id is correct.

2. Custom Telemetry.Id
I highly doubt there is any reason to override telemetry.Id. @SergeyKanzhelev are you aware of any customer doing this?
 
The idea behind Activity to provide **default**  and **automagical** way to correlate telemetry.
Overriding this behavior breaks the whole correlation feature and we should protect ourselves and customers from it. 
If someone really wants to break default correlation behavior, they can still do it, however I even prefer it to be hard and hacky.

lmolkova: I agree that it is a breaking change.

We could consider generating default telemetry Id from the Activity.Current **before** initialization in OperatonTelemetry ctor, however Activity.Current may be null (because Activity is not started yet).

PR #489, opened 2017-03-30
Allow prerelease DiagnosticSource package

lmolkova:
Fix dependency version for DiagnosticSource: there is no stable 4.4.0 version yet

### DIFF
--- a/src/Core/NuGet/Package.nuspec
+++ b/src/Core/NuGet/Package.nuspec
@@ -22,17 +22,17 @@
         <dependency id="Microsoft.Diagnostics.Tracing.EventSource.Redist" version="1.1.28"/>
       </group>
       <group targetFramework="net45">
-        <dependency id="System.Diagnostics.DiagnosticSource" version="4.4.0" />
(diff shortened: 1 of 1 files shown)


Dmitry-Matveev:
You may want to reference specific version of DiagnosticSource (pre-release) to make sure that changes in DiagnosticSource on its way to release won't fail SDK builds in case or the broken backcompat. Otherwise builds will always pick up the latest available pre-release and it may not be compatible or may break some scenarios unexpectedly.

lmolkova:
I'm monitoring any changes in DiagnosticSource and it's also quite important to know that some changes there break compatibility with AppInsights, sooner we know about it the better 👍

lmolkova:
@SergeyKanzhelev @Dmitry-Matveev 
Hey, could you please merge it? Without it it's not possible to install fresh AI nuget, because it depends on something that does not exist yet.

me and @karolz-ms are blocked on it

SergeyKanzhelev:
I'll kick off the signed build in a sec

lmolkova:
@SergeyKanzhelev  Thanks!

PR #495, opened 2017-03-30
Activity support Core SDK continued

lmolkova:
This change addresses comment for #486 posted after the merge


pharring:
👍

src/Core/Managed/Shared/ActivityExtensions.cs
```diff
@@ -50,6 +38,31 @@ internal static void SetOperationName(this Activity activity, string operationNa
             Debug.Assert(activity != null, "Activity must not be null");
             activity.AddTag(OperationNameTag, operationName);
         }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static bool Initialize()
+        {
+            try
+            {
+                Assembly.Load(new AssemblyName("System.Diagnostics.DiagnosticSource, Version=4.0.2.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51"));
```
SergeyKanzhelev: can you please have a unit test that will check the version in the string matches the version of actual dependency. So we will keep it in sync and have a verification that will fail if we forgot to update this string

Test/CoreSDK.Test/Shared/ActivityExtensionsTests.cs
```diff
@@ -0,0 +1,22 @@
...
+    [TestClass]
+    public class ActivityExtensionsTests
+    {
+        [TestMethod]
+        public void CanLoadDiagnosticSourceAssembly()
+        {
+            // verifies that currently referenced DiagnosticSource version, culture and public token matches one checked in ActivityExtension.Initialize
+            // if fails, fix the ActivityExtension as well
+            Assert.DoesNotThrow(() => 
+                Assembly.Load(new AssemblyName("System.Diagnostics.DiagnosticSource, Version=4.0.2.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51")));
```
SergeyKanzhelev: can you please share this string as const. So fixing it here will fix product code as well

lmolkova: ActivityExtensions are internal, there is no reason except for the test code to have public const with this string

src/Core/Managed/Shared/ActivityExtensions.cs
```diff
@@ -12,30 +14,25 @@ internal static class ActivityExtensions
         private static bool isEnabled = false;
 
         /// <summary>
-        /// Checks that Activity API could be called (DiagnosticSource DLL is loaded).
```
lmolkova: unfortunately previous version did not work. I assume runtime tries to resolve Activity and crashed when entering StartOperation method.
I'm working on the standalone test now

src/Core/NuGet/Package.nuspec
```diff
@@ -22,17 +22,17 @@
         <dependency id="Microsoft.Diagnostics.Tracing.EventSource.Redist" version="1.1.28"/>
       </group>
       <group targetFramework="net45">
-        <dependency id="System.Diagnostics.DiagnosticSource" version="4.4.0" />
+        <dependency id="System.Diagnostics.DiagnosticSource" version="4.4.0-beta-25130-03" />
```
lmolkova: :( was wrong again.
This way Nuget tried to reference the lowest possible version and ended up with the version that did not have Activity...

Test/CoreSDK.Test/Shared/ActivityExtensionsTests.cs
SergeyKanzhelev: Test has internalsVisibleTrue to the product code. So you can have internal const string

Issue #480, opened 2017-03-23
SDK stops sending telemetry and does not start again.


dnduffy:
There is logic in the ServerTelemetryChannel to set the sender and buffer capacities to 0 in the event of a server error response (408 timeout, 500 internal error, and 503 service unavailable) and then a timer is supposed to re-enable sending after some back-off time period.
We have had at least one confirmed instance where sending does not get re-enabled and stays stuck at 0.  We speculate we have had a couple other instances as well but we don't have the logs to prove those.
Eventually the disk buffer fills and all telemetry after that is discarded but the overhead of the logic attempting to send and save the unsent data still exists.

Dmitry-Matveev:
Let's investigate if it's a regression in 2.3 and whether we should take it as part of the stable release. We've got dump for the repro.

iusafaro:
Investigation showed that it can happen if SDK received two different error responses (PartialSuccess and InternalServerError for instance) at the same time, then it goes to the situation with sender and buffer capacities equal to 0 and never came back. Let's see if there're any other cases leading to the same situation.

PR #501, opened 2017-03-31
Update TaskTimeInternal to not cancel previous operation

iusafaro:
SDK stops sending telemetry and does not start again. #480

### DIFF
--- a/Test/CoreSDK.Test/Shared/Extensibility/Implementation/TaskTimerInternalTest.cs
+++ b/Test/CoreSDK.Test/Shared/Extensibility/Implementation/TaskTimerInternalTest.cs
@@ -135,28 +135,28 @@
             [TestMethod]
             public void StartIsCalledMultipleTimes()
             {
-                var timer = new TaskTimerInternal { Delay = TimeSpan.FromMilliseconds(1) };
+                var timer = new TaskTimerInternal { Delay = TimeSpan.FromMilliseconds(25) };
 
-                int invokationCount = 0;
+                int invocationCount = 0;
                 var firstActionInvoked = new ManualResetEventSlim();
                 var secondActionInvoked = new ManualResetEventSlim();
                 timer.Start(() => Task.Factory.StartNew(
                     () =>
                     {                       
-                        Interlocked.Increment(ref invokationCount);
+                        Interlocked.Increment(ref invocationCount);
                         firstActionInvoked.Set();
                     }));
                 timer.Start(
                     () => Task.Factory.StartNew(
                         () =>
                             {
-                                Interlocked.Increment(ref invokationCount);
+                                Interlocked.Increment(ref invocationCount);
                                 secondActionInvoked.Set();
                             }));
 
                 Assert.True(firstActionInvoked.Wait(50));
                 Assert.True(secondActionInvoked.Wait(50));
-                Assert.Equal(2, invokationCount);
+                Assert.Equal(2, invocationCount);
             }
 
             [TestMethod]
@@ -214,29 +214,29 @@
             [TestMethod]
             public void CancelAbortsOnlyLastTask()
             {
-                var timer = new TaskTimerInternal { Delay = TimeSpan.FromMilliseconds(1) };
+                var timer = new TaskTimerInternal { Delay = TimeSpan.FromMilliseconds(25) };
 
-                int invokationCount = 0;
+                int invocationCount = 0;
                 var firstActionInvoked = new ManualResetEventSlim();
                 var secondActionInvoked = new ManualResetEventSlim();
                 timer.Start(() => Task.Factory.StartNew(
                     () =>
                     {
-                        Interlocked.Increment(ref invokationCount);
+                        Interlocked.Increment(ref invocationCount);
                         firstActionInvoked.Set();
                     }));
                 timer.Start(
                     () => Task.Factory.StartNew(
                         () =>
                         {
-                            Interlocked.Increment(ref invokationCount);
+                            Interlocked.Increment(ref invocationCount);
                             secondActionInvoked.Set();
                         }));
 
                 timer.Cancel();
 
                 Assert.True(firstActionInvoked.Wait(50));
-                Assert.Equal(1, invokationCount);
+                Assert.Equal(1, invocationCount);
             }
         }
     }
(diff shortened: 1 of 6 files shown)


Test/CoreSDK.Test/Shared/Extensibility/Implementation/TaskTimerInternalTest.cs
```diff
@@ -203,6 +210,34 @@ public void AbortsPreviousAction()
...
+                        () =>
+                        {
+                            Interlocked.Increment(ref invokationCount);
+                            secondActionInvoked.Set();
+                        }));
+
+                timer.Cancel();
+
+                Assert.True(firstActionInvoked.Wait(50));
+                Assert.Equal(1, invokationCount);
```
SergeyKanzhelev: invocation, not invokation

SergeyKanzhelev: Why do you need both - increment and event for this check?

Test/CoreSDK.Test/Shared/Extensibility/Implementation/TaskTimerInternalTest.cs
```diff
@@ -203,6 +210,34 @@ public void AbortsPreviousAction()
...
+                    }));
+                timer.Start(
+                    () => Task.Factory.StartNew(
+                        () =>
+                        {
+                            Interlocked.Increment(ref invokationCount);
+                            secondActionInvoked.Set();
+                        }));
+
+                timer.Cancel();
```
SergeyKanzhelev: 1 ms may not be sufficient between the `Start` and `Cancel`. You are waiting 50 milliseconds so maybe change timer to 25 milliseconds?

Test/CoreSDK.Test/Shared/Extensibility/Implementation/TaskTimerInternalTest.cs
```diff
@@ -203,6 +210,34 @@ public void AbortsPreviousAction()
...
+
+                Assert.True(firstActionInvoked.Wait(50));
```
SergeyKanzhelev: Does it accept `TimeSpan`? It's better to use timespan so it's clear how long this wait it without reading docs

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimerInternal.cs
```diff
@@ -82,7 +82,7 @@ public void Start(Func<Task> elapsed)
                             if (task != null)
                             {
 #if !NET40
-                                await task.ConfigureAwait(false);
+                                await task.ConfigureAwait(false);
```
SergeyKanzhelev: can you pls get rid of those trailing spaces?

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimerInternal.cs
```diff
@@ -150,6 +150,14 @@ private static void LogException(Exception exception)
...
 
+        private static void DisposeToken(CancellationTokenSource tokenSource)
+        {
+            if (tokenSource != null)
+            {
+                tokenSource.Dispose();
+            }
+        }
+
         private static void CancelAndDispose(CancellationTokenSource tokenSource)
```
SergeyKanzhelev: it's not in this PR - are there any calls to this method left?

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimerInternal.cs
```diff
@@ -72,7 +72,7 @@ public void Start(Func<Task> elapsed)
                     previousTask =>
 #endif
                     {
-                        CancelAndDispose(Interlocked.CompareExchange(ref this.tokenSource, null, newTokenSource));
+                        DisposeToken(Interlocked.CompareExchange(ref this.tokenSource, null, newTokenSource));
```
SergeyKanzhelev: Can you please update the comment for this method in `<summary>` task clarifying this behavior

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimerInternal.cs
iusafaro: still used in Cancel()

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimerInternal.cs
iusafaro: done

Test/CoreSDK.Test/Shared/Extensibility/Implementation/TaskTimerInternalTest.cs
```diff
@@ -133,23 +133,30 @@ public void InvokesActionAfterDelay()
             }
 
             [TestMethod]
-            public void CancelsPreviousActionWhenStartIsCalledMultipleTimes()
+            public void StartIsCalledMultipleTimes()
```
dnduffy: nitpick: Should be slightly more descriptive new name. e.g. EachCallStartsSeparateTimerTaskWhenStartIsCalledMultipleTimes.

Test/CoreSDK.Test/Shared/Extensibility/Implementation/TaskTimerInternalTest.cs
```diff
@@ -203,6 +210,34 @@ public void AbortsPreviousAction()
...
+                    }));
+                timer.Start(
+                    () => Task.Factory.StartNew(
+                        () =>
+                        {
+                            Interlocked.Increment(ref invocationCount);
+                            secondActionInvoked.Set();
+                        }));
+
+                timer.Cancel();
```
dnduffy: Cancel should actually take the token to cancel instead of just cancelling the most recent.

src/Core/Managed/Shared/Extensibility/Implementation/TaskTimerInternal.cs
```diff
@@ -111,7 +111,7 @@ public void Start(Func<Task> elapsed)
                     TaskContinuationOptions.OnlyOnRanToCompletion | TaskContinuationOptions.ExecuteSynchronously,
                     TaskScheduler.Default);
 
-            CancelAndDispose(Interlocked.Exchange(ref this.tokenSource, newTokenSource));
+            DisposeToken(Interlocked.Exchange(ref this.tokenSource, newTokenSource));
```
Dmitry-Matveev: It looks like we only dispose token in this code but never use it to cancel or manage task lifetime. Should we simplify the approach and use only one token across or no token at all if we do not rely on it?

PR #503, opened 2017-04-03
Sergkanz/fw4

SergeyKanzhelev:
Distinguish `fw4` telemetry

### DIFF
--- a/src/Core/Managed/Shared/Extensibility/Implementation/SdkVersionUtils.cs
+++ b/src/Core/Managed/Shared/Extensibility/Implementation/SdkVersionUtils.cs
@@ -19,7 +19,7 @@
                     .OfType<AssemblyFileVersionAttribute>()
                     .First()
                     .Version;
-            
+
 #else
             string versionStr = typeof(TelemetryClient).GetTypeInfo().Assembly.GetCustomAttributes<AssemblyFileVersionAttribute>()
                     .First()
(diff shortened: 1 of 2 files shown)


src/Core/Managed/Shared/Extensibility/Implementation/SdkVersionUtils.cs
```diff
@@ -19,15 +20,20 @@ internal static string GetSdkVersion(string versionPrefix)
...
 #endif
 
             Version version = new Version(versionStr);
-            return versionPrefix + version.ToString(3) + "-" + version.Revision;
+            string postfix = version.Revision.ToString(CultureInfo.InvariantCulture);
+#if NET40
+            postfix += "-fw4";
+#endif
+
+            return versionPrefix + version.ToString(3) + "-" + postfix;
```
Dmitry-Matveev: It might so happen that AI Endpoint relies on "-" to split SDK version into prefix and postfix, with this change, it will get "Version-Revision" as version (due to extra "-" from fw4) and might not apply necessary logic that depends on SDK version.

SergeyKanzhelev: Validated. We are using semver package to do comparison.

Code:

``` js
const semver = require('semver')

var versionStr = '1.2.0-1323-fw4'

if (semver.valid(versionStr))
    console.log("Yes, it's valid!")

console.log("prerelease is " + semver(versionStr).prerelease)

if (semver.gt(versionStr, '1.1.0'))
    console.log("Yes, it is greater!")
```

Output:

``` 
Yes, it's valid!
prerelease is 1323-fw4
Yes, it is greater!
```

PR #505, opened 2017-04-03
Fix StartOperation: do not start child activity 

lmolkova:
StartOperation is the same as Start Activity: every OperationTelemetry must have a corresponding Activity.
Imagine following scenario:

In the previous implementation, new Activity was created to represent children of the operation, however it would mean that every child telemetry in scope of parent would have just one activity representing them.
In case of multiple dependency calls within the same request, all of them must have different activities and different activity.Id (telemetry.id).

So this change fixes it:
- Activity is started to represent current operation being started in the StartOperation, not the child one
- OperationCorrelationTelemtryInitializer sets ParentId with Activity.Current.Id

This way every dependency call started with StartOperation will have unique Telemetry.Id

### DIFF
--- a/src/Core/Managed/Shared/ActivityExtensions.cs
+++ b/src/Core/Managed/Shared/ActivityExtensions.cs
@@ -11,7 +11,7 @@ namespace Microsoft.ApplicationInsights
     {
         private const string OperationNameTag = "OperationName";
         private static bool isInitialized = false;
-        private static bool isEnabled = false;
+        private static bool isAvailable = false;
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Core/Managed/Shared/ActivityExtensions.cs
```diff
@@ -11,7 +11,7 @@ internal static class ActivityExtensions
     {
         private const string OperationNameTag = "OperationName";
```
lmolkova: @pharring I'll try to address your concerns regarding Action in  a different PR

PR #512, opened 2017-04-04
Keithh 2.4.0 beta1

kstevenham:


src/Core/Managed/Shared/DataContracts/ExceptionTelemetry.cs
```diff
@@ -173,7 +189,43 @@ internal IList<ExceptionDetails> Exceptions
...
+        {
+            List<StackFrame> orderedStackTrace = new List<StackFrame>();
+
+            if (this.Exceptions != null && this.Exceptions.Count > 0)
+            {
+                if (frames != null && frames.Length > 0)
+                {
+                    int stackLength = 0;
+
+                    this.Exceptions[0].parsedStack = new List<StackFrame>();
```
SergeyKanzhelev: why do you override and not append? Appending would be logical here

src/Core/Managed/Shared/DataContracts/ExceptionTelemetry.cs
```diff
@@ -173,7 +189,43 @@ internal IList<ExceptionDetails> Exceptions
...
+                    this.Exceptions[0].hasFullStack = true;
+
+                    for (int level = 0; level < frames.Length; level++)
```
SergeyKanzhelev: do you need to do it synchronously? Does the exceptions stack trace parsed synchronously as well?

src/Core/Managed/Shared/DataContracts/ExceptionTelemetry.cs
```diff
@@ -173,7 +189,43 @@ internal IList<ExceptionDetails> Exceptions
...
+                    {
+                        StackFrame sf = ExceptionConverter.GetStackFrame(frames[level], level);
+
+                        stackLength += ExceptionConverter.GetStackFrameLength(sf);
+
+                        if (stackLength > ExceptionConverter.MaxParsedStackLength)
```
SergeyKanzhelev: should this be part of `Sanitize`? Why do we need this check here?

PR #513, opened 2017-04-05
SDK stops sending telemetry and does not start again. #480

iusafaro:
Revert TaskTimeInternal changes and make each policy to have its own instance of the timer

### DIFF
--- a/src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/BackoffLogicManager.cs
+++ b/src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/BackoffLogicManager.cs
@@ -19,7 +19,7 @@
         
         private bool exponentialBackoffReported = false;
         private int consecutiveErrors;
-        private DateTimeOffset nextMinTimeToUpdateConsecutiveErrors = DateTimeOffset.MinValue;        
+        private DateTimeOffset nextMinTimeToUpdateConsecutiveErrors = DateTimeOffset.MinValue;
 
         public BackoffLogicManager()
         {
--- a/src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/BackoffLogicManager.cs
+++ b/src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/BackoffLogicManager.cs
@@ -8,7 +8,7 @@
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
     using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation;
 
-    internal class BackoffLogicManager : IDisposable
+    internal class BackoffLogicManager
     {
         private const int SlotDelayInSeconds = 10;
         private const int MaxDelayInSeconds = 3600;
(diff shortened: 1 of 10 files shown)


src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/BackoffLogicManager.cs
```diff
@@ -19,22 +16,23 @@ internal class BackoffLogicManager : IDisposable
 
         private readonly object lockConsecutiveErrors = new object();
         private readonly TimeSpan minIntervalToUpdateConsecutiveErrors;
-
-        private TaskTimerInternal pauseTimer = new TaskTimerInternal { Delay = TimeSpan.FromSeconds(SlotDelayInSeconds) };
+        
         private bool exponentialBackoffReported = false;
         private int consecutiveErrors;
-        private DateTimeOffset nextMinTimeToUpdateConsecutiveErrors = DateTimeOffset.MinValue;
+        private DateTimeOffset nextMinTimeToUpdateConsecutiveErrors = DateTimeOffset.MinValue;
```
dnduffy: nitpick: spaces on end of line

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ErrorHandlingTransmissionPolicy.cs
```diff
@@ -12,9 +12,10 @@
     using TaskEx = System.Threading.Tasks.Task;
 #endif
 
-    internal class ErrorHandlingTransmissionPolicy : TransmissionPolicy
+    internal class ErrorHandlingTransmissionPolicy : TransmissionPolicy, IDisposable
     {
         private BackoffLogicManager backoffLogicManager;
+        private TaskTimerInternal pauseTimer = new TaskTimerInternal { Delay = TimeSpan.FromSeconds(BackoffLogicManager.SlotDelayInSeconds) };
```
dnduffy: nitpick: the delay value here is throw away because it has to be replaced with the latest value from the backoff logic manager when handling a transmission error and actually starting the timer.

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ErrorHandlingTransmissionPolicy.cs
```diff
@@ -59,8 +66,8 @@ private void HandleTransmissionSentEvent(object sender, TransmissionProcessedEve
                             this.backoffLogicManager.ReportBackoffEnabled((int)httpWebResponse.StatusCode);
                             this.Transmitter.Enqueue(e.Transmission);
 
-                            this.backoffLogicManager.ScheduleRestore(
-                               httpWebResponse.RetryAfterHeader,
+                            this.pauseTimer.Delay = this.backoffLogicManager.GetBackOffTimeInterval(httpWebResponse.RetryAfterHeader);
+                            this.pauseTimer.Start(
```
dnduffy: What if it has already been started somehow?

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ThrottlingTransmissionPolicy.cs
```diff
@@ -54,13 +61,13 @@ private void HandleTransmissionSentEvent(object sender, TransmissionProcessedEve
...
+                    this.pauseTimer.Start(
                         () =>
-                            {
-                                this.ResetPolicy();
-                                return TaskEx.FromResult<object>(null);
-                            });
+                                    {
+                                        this.ResetPolicy();
+                                        return TaskEx.FromResult<object>(null);
+                                    });
```
dnduffy: nitpick: formatting indents

src/TelemetryChannels/ServerTelemetryChannel/Shared/Implementation/ErrorHandlingTransmissionPolicy.cs
iusafaro: then previous is cancelled and after next delay interval capacities will be set back to null - that's exactly what expected

PR #518, opened 2017-04-05
Metrics: Address more bugs that are blocking for 2.4.-Beta1 release.

macrogreg:


src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -399,6 +399,26 @@ public void FailedToLoadEnvironmentVariables(string ex, string appDomainName = "
             this.WriteEvent(32, ex, this.nameProvider.Name);
         }
 
+        // Verbosity is Error - so it is always sent to portal; Keyword is Diagnostics so throttling is not applied.
+        [Event(33,
+            Message = "Metrics Extractor should be used before Sampling Processors or any other Telemetry Processors that might filter out Telemetry Items. Otherwise, extracted metrics may be incorrect.",
+            Level = EventLevel.Error,
+            Keywords = Keywords.Diagnostics | Keywords.UserActionable)]
+        public void MetricExtractorAfterSampling_Error(string appDomainName = "Incorrect")
```
SergeyKanzhelev: We do not use underscore in other names here. Let's keep it this way

src/Core/Managed/Shared/Extensibility/AutocollectedMetricsExtractor.cs
```diff
@@ -252,6 +263,29 @@ private void ExtractMetrics(ITelemetry fromItem)
             }
         }
 
+        private bool EnsureItemNotSampled(ITelemetry item)
+        {
+            ISupportSampling potentiallySampledItem = item as ISupportSampling;
+
+            if (potentiallySampledItem != null && potentiallySampledItem.SamplingPercentage.HasValue)
```
SergeyKanzhelev: I'd recommend to check whether it's not 100-0.001. Just in case somebody wants to ensure the item will be send and set the sampling rate explicitly. Like:

``` csharp
var veryImportantEvent = new EventTelemetry("app is down") { SamplingPercentage = 100; }
client.TrackEvent(veryImportantEvent);
```

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -399,6 +399,26 @@ public void FailedToLoadEnvironmentVariables(string ex, string appDomainName = "
             this.WriteEvent(32, ex, this.nameProvider.Name);
         }
 
+        // Verbosity is Error - so it is always sent to portal; Keyword is Diagnostics so throttling is not applied.
+        [Event(33,
+            Message = "Metrics Extractor should be used before Sampling Processors or any other Telemetry Processors that might filter out Telemetry Items. Otherwise, extracted metrics may be incorrect.",
```
SergeyKanzhelev: can you state a problem first. "Metric Extractor detected an element with SamplingPercentage set to {0}. Metrics Extractor should be used before Sampli..."

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -399,6 +399,26 @@ public void FailedToLoadEnvironmentVariables(string ex, string appDomainName = "
...
+            Level = EventLevel.Error,
+            Keywords = Keywords.Diagnostics | Keywords.UserActionable)]
+        public void MetricExtractorAfterSampling_Error(string appDomainName = "Incorrect")
+        {
+            this.WriteEvent(33, this.nameProvider.Name);
+        }
+
+        // Verbosity is Verbose - so it is always sent to portal; Keyword is Diagnostics so throttling is not applied.
+        [Event(34,
+            Message = "Metrics Extractor should be used before Sampling Processors or any other Telemetry Processors that might filter out Telemetry Items. Otherwise, extracted metrics may be incorrect.",
```
SergeyKanzhelev: text here may be smaller. Just "MetricExctractor will ignore a telemetry item of type {0} with the SamplingPercentage {1}"

SergeyKanzhelev: This will not be send to portal, only analyzed by support personnel

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
macrogreg: Agreed. Addressing.

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
macrogreg: Ok, will do.

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
macrogreg: > MetricExctractor will ignore a telemetry item of type {0} with the SamplingPercentage {1}

This incorrectly describes the issue. Yes, we will ignore it, but it indicates a bigger problem that the current message describes. SO I will leave the current message.

> This will not be send to portal, only analyzed by support personnel

I have copied the comment from elsewhere so it must be wrong there too. Will remove the incorrect comment.

src/Core/Managed/Shared/Extensibility/AutocollectedMetricsExtractor.cs
macrogreg: Ok. However, is your use case what we recommend? This makes the condition complex. For the above use case we should instead expose a special boolean like "SkipSampling" or similar.

Also, the sampling processor instead uses (100.0 - 1.0E-12) as the threshold.

macrogreg: @SergeyKanzhelev

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -399,6 +399,26 @@ public void FailedToLoadEnvironmentVariables(string ex, string appDomainName = "
             this.WriteEvent(32, ex, this.nameProvider.Name);
         }
 
+        // Verbosity is Error - so it is always sent to portal; Keyword is Diagnostics so throttling is not applied.
+        [Event(33,
+            Message = "A Metric Extractor detected a telemetry item with SamplingPercentage < 100. Metrics Extractors should be used before Sampling Processors or any other Telemetry Processors that might filter out Telemetry Items. Otherwise, extracted metrics may be incorrect.",
```
SergeyKanzhelev: I meant to pass an actual sampling percentage and item type we found instead of ` SamplingPercentage < 100`. Consider updating, but not a big deal.

PR #521, opened 2017-04-06
no special logic for .NET Core exceptions when serializing

SergeyKanzhelev:


pharring:
👍

src/Core/Managed/Core.csproj
```diff
@@ -37,6 +37,10 @@
   <ItemGroup Condition="'$(TargetFramework)' != 'net40'">
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.0-beta-25124-07" />
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.3'">
+    <PackageReference Include="System.Diagnostics.StackTrace" Version="4.3.0" />
```
SergeyKanzhelev: @pharring is it the right way to add dependencies? Why we do not have a list of other packages we depend on?

PR #523, opened 2017-04-10
get changes from develop to 2.4.0beta1

cijothomas:


pharring:
I've been merging develop into 2.4.0-beta1 quite regularly. Are you sure you want all these additional changes? In particular, I believe the changes around the Metric APIs were supposed to be for 2.3 only and not ported to 2.4

pharring:
Recommend 👎 
You'll have to double-check, but I think all the required commits from develop are already in 2.4.0-beta1.

cijothomas:
I will abandon this, and pick any required bug fixes (if needed)

PR #533, opened 2017-04-12
Fix the issue that ExceptionDetails class doesn't decorate the EventData attribute correctly

yantang-msft:
As titled. Without this fix, it may throw exception when track exception
@pharring @SergeyKanzhelev @lmolkova

### DIFF
--- a/src/Core/Managed/Shared/Extensibility/Implementation/External/ExceptionDetails_types.cs
+++ b/src/Core/Managed/Shared/Extensibility/Implementation/External/ExceptionDetails_types.cs
@@ -26,48 +26,48 @@
 
 namespace Microsoft.ApplicationInsights.Extensibility.Implementation.External
 {
-    using System.Collections.Concurrent;
+    using System;
     using System.Collections.Generic;
 
-    
-    
+
+
     [System.CodeDom.Compiler.GeneratedCode("gbc", "0.4.1.0")]
     internal partial class ExceptionDetails
     {
-        
-        
+
+
         public int id { get; set; }
 
-        
-        
+
+
         public int outerId { get; set; }
 
-        
-        
-        
+
+
+
         public string typeName { get; set; }
 
-        
-        
-        
+
+
+
         public string message { get; set; }
 
-        
-        
+
+
         public bool hasFullStack { get; set; }
 
-        
-        
-        
+
+
+
         public string stack { get; set; }
 
-        
-        
+
+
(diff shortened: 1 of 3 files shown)


src/Core/Managed/Shared/Extensibility/Implementation/External/ExceptionDetails_types.cs
```diff
@@ -77,5 +77,30 @@ protected ExceptionDetails(string fullName, string name)
             stack = "";
             parsedStack = new List<StackFrame>();
         }
+
+        /// <summary>
+        /// Creates a new instance of ExceptionDetails from a System.Exception and a parent ExceptionDetails.
+        /// </summary>
+        internal static ExceptionDetails CreateWithoutStackInfo(Exception exception, ExceptionDetails parentExceptionDetails)
```
SergeyKanzhelev: this is generated file. Please do not modify manually

SergeyKanzhelev: It's patial so just move implementation into the different file

dnduffy: Yes, any changes here will be lost the next time we regenerate.

yantang-msft: Moved the implementation to a third partial class to keep the pattern consistent.

src/Core/Managed/Shared/Extensibility/Implementation/External/ExceptionDetails_types.cs
```diff
@@ -29,45 +29,45 @@ namespace Microsoft.ApplicationInsights.Extensibility.Implementation.External
     using System.Collections.Concurrent;
     using System.Collections.Generic;
 
-    
-    
+
```
pharring: I would undo all changes in this file. Even though there are only whitespace changes, it's still different from what was auto-generated by the tool and could cause problems.

yantang-msft: Reverted. VS auto formatted it for me when I copied the source code. :)

pharring:
👍

SergeyKanzhelev:
Should it be in `2.4.0-beta1` or can wait?

+ @cijothomas

Issue #453, opened 2017-03-13
Create a Core.NetStandard.Tests project


pharring:
Right now we test .Net 4.0, 4.5 and 4.6. We should also test .Net Standard 1.3

This is a follow on from #451

PR #536, opened 2017-04-13
Allow .NET Core apps to use applicationinsights.config

pharring:
This allows .Net core applications referencing the base SDK to use applicationinsights.config via ```TelemetryConfiguration.CreateDefault()```

Created a common base class with a couple of abstract methods where NetStandard1.3 differs from desktop .Net FX.

### DIFF
--- a/src/Core/Managed/netstandard1.3/Extensibility/Implementation/Platform/PlatformImplementation.cs
+++ b/src/Core/Managed/netstandard1.3/Extensibility/Implementation/Platform/PlatformImplementation.cs
@@ -1,53 +1,21 @@
 ﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Platform
 {
     using System;
-    using System.Collections.Generic;
-    using Microsoft.ApplicationInsights.Extensibility.Implementation.External;
 
-    internal class PlatformImplementation : IPlatform
+    /// <summary>
+    /// The .Net standard 1.3 implementation of the <see cref="IPlatform"/> interface.
+    /// </summary>
+    internal sealed class PlatformImplementation : PlatformImplementationBase
     {
-        private IDebugOutput debugOutput = null;
-        
-        public IDictionary<string, object> GetApplicationSettings()
-        {
-            return null;
-        }
-
-        public string ReadConfigurationXml()
-        {
-            return null;
-        }
-
-        public ExceptionDetails GetExceptionDetails(Exception exception, ExceptionDetails parentExceptionDetails)
-        {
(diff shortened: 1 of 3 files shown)


pharring:
@karolz-ms Can you pull this down and verify that it works for your .net core test app?

Separately, this change reminds me that we need to implement .Net Core unit tests for the base SDK: #453

src/Core/Managed/netstandard1.3/Extensibility/Implementation/Platform/PlatformImplementation.cs
```diff
@@ -1,51 +1,28 @@
...
 
         /// <summary>
         /// Returns the machine name.
         /// </summary>
-        /// <returns>The machine name.</returns>
-        public string GetMachineName()
+        /// <returns>Null always (Environment.MachineName is not available on .Net Standard 1.3).</returns>
+        protected override string GetHostName()
         {
             return null;
```
SergeyKanzhelev: can we try to get it from environment variable?

pharring: Yes, in Windows we can use COMPUTERNAME as a reasonable approximation. There was some discussion over that in the coreclr repo: https://github.com/dotnet/corefx/issues/4306 
As a result, I think .NetStandard 2.0 has it directly on the Environment object, just like desktop FX.

pharring:
This is for (all) .Net core applications, not just ASP.NET Core. I don't know your plans for DI in the aspnetcore SDK. I'd be interested in learning off-line. Regardless, I would think you'd still want to support ```TelemetryConfiguration.CreateDefault()```. That's really what this change is about.

pharring:
Ping @SergeyKanzhelev I made the change you suggested.
Also FYI @dnduffy since Sergey mentioned you're thinking of another pattern for ASP.NET Core - although, I'll repeat that _this_ change is for any .Net Core app and doesn't stop ASP.NET Core doing its own thing.

pharring:
Ping @SergeyKanzhelev @dnduffy please respond. I made the change that was suggested.

SergeyKanzhelev:
Sorry for delay. I'm not sure what side effects it can bring in some corner cases of leftover `AI.config` so I'm waiting for the branch to be opened for `2.5-beta1` contributions. Let me know if there are reasons to push it for `2.4` stable

SergeyKanzhelev:
Closing PR for now. Will re-open if we will decide it's needed in this issue: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/505

PR #537, opened 2017-04-14
contributing instructions

SergeyKanzhelev:
Based on @Dmitry-Matveev feedback on the number of the last minute PRs lacking the context we can use the following pr template. Please suggest better wording or other ideas on how to streamline the process.

(see details about [pr templates](https://github.com/blog/2111-issue-and-pull-request-templates))

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,8 @@
+Fix # .
+
+
+For significant contribution make sure you completed the list:
+
+- [ ] Design discussion issue #
+- [ ] Changes in public surface reviewed
+- [ ] CHANGELOG.md updated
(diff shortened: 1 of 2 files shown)


SergeyKanzhelev:
@Dmitry-Matveev signoff?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,8 @@
+Fix # .
+
+
+For significant contribution make sure you completed the list:
```
dnduffy: nitpick: "For significant contributions please make sure you have completed the following items:"

Dmitry-Matveev:
My general feedback is that it's a nice check list but it comes too late in the process. 100 files PR submitter will find out about "Design review" in check list only when all those 100 files are ready and are in PR, however the design might not be something that team would've approved.

We need a common approach regarding contributions across our repos. Contributing.MD might be the right place but it grows big with time and people tend to skim through or skip it towards the end, so this idea of Design review should be front and center there and almost the first thing to see. It also requires some formalization of the "Design Discussion Issue" format, so that everyone follows the similar approach.

PR #543, opened 2017-04-21
Cherry pick nuget package specs fix from develop into master

cijothomas:
This is needed for 2.4.0-beta1


Microsoft.ApplicationInsights.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26228.9
+VisualStudioVersion = 15.0.26228.4
```
dnduffy: This downgrade shouldn't be relevant or necessary.

PR #557, opened 2017-05-17
Small metrics-related fixes.

macrogreg:
* Metrics: When we made MetricManager private, we forgot the Metric class. Catching up on that.
* Metrics: Add a temporary workaround for a suspected but unconfirmed issue with Extractor performance on telemetry when no metrics are extracted.


SergeyKanzhelev:
Thanks!

PR #571, opened 2017-06-21
Added ps scripts to enable/disable strong name validation in build machines

cijothomas:


disablestrongnamevalidation.ps1
```diff
@@ -0,0 +1,5 @@
+Set-Location HKLM:
+$registryPath = "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\.NETFramework"
+$Name = "AllowStrongNameBypass"
```
SergeyKanzhelev: do we need it for a specific assembly or entire machine? I just afraid it may create a security vulnerability on the machine.

cijothomas:
@SergeyKanzhelev i modified the script to disable SN for just the required assembly. But this required the usage of sn.exe directly, and its location may change in future requiring work to modify build definitions.
Anyway as long as its unblocking build/test infra, i'll merge.

PR #572, opened 2017-06-23
Fixed parameter passing bug in event source logging call.

dnduffy:


dnduffy:
I am wondering if it could in any way contribute to IcM 40548436 which I was investigating when I noticed it was the only call that tried to send two parameters to a trace that doesn't take two.
I wrote a couple unit tests to check the behavior but they didn't fail so it just seems the extra parameter was ignored.

Issue #559, opened 2017-05-20
EventFlow multi-channel support


SergeyKanzhelev:
With the [EventFlow](https://github.com/Azure/diagnostics-eventflow) integration we enable SDK to send data to more backends than just Application Insights. Typical scenario would be send sampled and pre-aggregated data to Application Insights, ALL data to blobs or ElasticSearch and set of critical telemetry to EventHub for real-time processing.

To enable this capability we need to make a number of modifications to the SDK. The biggest change is related to configuration and initialization—to allow multiple channels, we need to somehow indicate that there can be multiple chains of TelemetryProcessors and make it easy to plug the new chain. We also want to make sure that the standard feature of Application Insights like VS experience, live stream, self-diagnostics and error reporting are not affected by this change.

We discussed with @karolz-ms there are two options - create a broadcast telemetry processor that can have multiple "children" chains or have a new concept in SDK, We agreed that the best approach will be to extend `ApplicationInsights.config` with the set of "additional channels" and keep the standard channel in place. This will ensure there is a single place to hook a new telemetry subscriber that easy to implement in external NuGet. 

The specific name of this collection is TBD. The idea is that SDK will clone the telemetry item before sending it to the channel so modification of telemetry item in a single channel will not affect another one. Every channel will have it's own set of telemetry processors

CC: @karolz-ms @Dmitry-Matveev @cijothomas @AlexBulankou

karolz-ms:
@SergeyKanzhelev My assumption is that these "additional channels" could have their own telemetry processors and metric processors. Can you confirm?

What about telemetry initializers? I am leaning towards not including channel-specific telemetry initializers in the first iteration of this feature

SergeyKanzhelev:
I think we got rid of MetricProcessors in latest stable. So no need to include those initially. Do you think there should be a separate concept of channel or every "additional channel" is just a `ITelemetryProcessor` that starts the chain?

karolz-ms:
Hmm, the develop branch still has MetricProcessors on TelemetryConfiguration (internal property). Should I base this change off of different branch?

Since the primary chain is fully explorable via TelemetryConfiguration I thought it would be desirable to do the same for "additional channels". That is, I expected they will have a separate configuration-like class with processors and a telemetry channel.

karolz-ms:
Also, do we want to have a capability for telemetry/metric processors that apply to all channels? Or do the telemetry processors that are defined at the top configuration level only applicable to the primary channel?

[...]

PR #584, opened 2017-07-06
Add multi-channel capability (part 1: runtime)

karolz-ms:
This is related to issue #559 

This change does not include support for the new "sink" concept in configuration, but I thought it is big and complete enough (see tests) to be a good starting point for discussion and suggestions. 

I also made some unrelated changes to tests. On my development laptop the tests that have been updated were quite unreliable initially because timeouts were too short.


Test/TestFramework/Shared/DeterministicTaskScheduler.cs
```diff
@@ -0,0 +1,60 @@
+﻿namespace Microsoft.ApplicationInsights.TestFramework
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// TaskScheduker for executing tasks on the same thread that calls RunTasksUntilIdle() or RunPendingTasks().
```
pharring: Typo.

src/Core/Managed/Shared/Channel/IDeepCloneable.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Microsoft.ApplicationInsights.Channel
```
pharring: Should this interface be in the `Microsoft.ApplicationInsights` namespace? I realize that it's associated with the work you're doing in channels/sinks, but it seems more general purpose.

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
```diff
@@ -46,6 +46,19 @@ public AvailabilityTelemetry(string name, DateTimeOffset timeStamp, TimeSpan dur
...
+        /// Initializes a new instance of the <see cref="AvailabilityTelemetry"/> class by cloning an existing instance.
+        /// </summary>
+        /// <param name="source">Source instance of <see cref="AvailabilityTelemetry"/> to clone from.</param>
+        private AvailabilityTelemetry(AvailabilityTelemetry source)
+        {
+            this.Data = source.Data.DeepClone();
+            this.context = source.context.DeepClone(this.Data.properties);
+            this.Sequence = source.Sequence;
+            this.Timestamp = source.Timestamp;
+            this.Data.id = Convert.ToBase64String(BitConverter.GetBytes(WeakConcurrentRandom.Instance.Next()));
```
pharring: Surprising that we generate a new ID. I guess this is required for your scenario, but it does taint the concept of 'cloning'.

src/Core/Managed/Shared/Extensibility/TelemetrySink.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        /// <summary>
+        /// Gets or sets an instance of the <see cref="TelemetryProcessorChainBuilder"/> that this sink is using.
+        /// </summary>
+        public TelemetryProcessorChainBuilder TelemetryProcessorChainBuilder
+        {
+            get
+            {
+                LazyInitializer.EnsureInitialized(ref this.telemetryProcessorChainBuilder, () => new TelemetryProcessorChainBuilder(this.telemetryConfiguration, this));
```
pharring: Note that `EnsureInitialized` returns the value too. So this can be a one-liner:
`return LazyInitializer.EnsureInitialized(ref ...)`

src/Core/Managed/Shared/Channel/IDeepCloneable.cs
SergeyKanzhelev: We'd like to keep the root namespace empty so the only thing there is `TelemetryClient`. Avoiding confusions on how to use Application Insights. Maybe place alongside other telemetry items in `Microsoft.ApplicationInsights.DataContracts`?

src/Core/Managed/Shared/Channel/IDeepCloneable.cs
```diff
@@ -0,0 +1,16 @@
...
+    /// </summary>
+    /// <typeparam name="T">The type of the object that results from the cloning operation.</typeparam>
+    public interface IDeepCloneable<out T>
+    {
+        /// <summary>
+        /// Clones the object deeply, so that the original object and its clones share no state 
+        /// and can be modified independently.
+        /// </summary>
+        /// <returns>The cloned object.</returns>
+        T DeepClone();
```
SergeyKanzhelev: Would it make more sense just expand the `ITelemetry` interface to include this method?

SergeyKanzhelev: Not sure what else may be deep cloneable in Application Insights

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -214,31 +206,40 @@ internal set
...
-                return this.telemetryChannel;
+                // We do not ensure not disposed here because TelemetryChannel is accessed during configuration disposal.
+                return this.telemetrySinks[DefaultSinkIndex].TelemetryChannel;
             }
 
             set
             {
-                ITelemetryChannel oldChannel = this.telemetryChannel;
-                this.telemetryChannel = value;
+                this.EnsureNotDisposed();
```
SergeyKanzhelev: we typically wouldn't throw in situations like this to not crash customer code. Just return nothing.

src/Core/Managed/Shared/Extensibility/TelemetrySink.cs
```diff
@@ -0,0 +1,149 @@
...
+        private bool shouldDisposeChannel;
+        private TelemetryProcessorChain telemetryProcessorChain;
+        private TelemetryProcessorChainBuilder telemetryProcessorChainBuilder;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TelemetrySink"/> class.
+        /// </summary>
+        /// <param name="telemetryConfiguration">Telemetry configuration to use for the new <see cref="TelemetrySink"/> instance.</param>
+        /// <param name="telemetryChannel">Telemetry channel to use for the new <see cref="TelemetrySink"/> instance.</param>
+        public TelemetrySink(TelemetryConfiguration telemetryConfiguration, ITelemetryChannel telemetryChannel = null)
```
SergeyKanzhelev: If one implements `ITelemetryProcessor` it typically takes the next processor as a constructor argument

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
```diff
@@ -313,6 +314,43 @@ public void Dispose()
         }
 
         /// <summary>
+        /// Adds a new telemetry sink to the configuration.
+        /// </summary>
+        /// <param name="sink">Sink to add.</param>
+        public void AddSink(TelemetrySink sink)
```
SergeyKanzhelev: why do you need a method for this? Shouldn't it be added via configuration the same way it works for telemetry initializers?

src/Core/Managed/Shared/Extensibility/Implementation/BroadcastProcessor.cs
```diff
@@ -0,0 +1,148 @@
...
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
+
+    /// <summary>
+    /// An <see cref="ITelemetryProcessor"/> that asynchronously sends the data to multiple child telemetry processors.
+    /// </summary>
+    internal class BroadcastProcessor : ITelemetryProcessor, IDisposable
```
SergeyKanzhelev: with the telemetry channel today channel `Send` method makes everything async. So we move the complexity of it outside of core. 

Would it be better to do the same in this case? Just call sinks synchronously and ask every sync developer make Send method very fast?

Test/CoreSDK.Test/Shared/DataContracts/PerformanceCounterTelemetryTest.cs
```diff
@@ -2,8 +2,9 @@
 {
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using System;
```
karolz-ms: This was compiled with Release. I assume System namespaces should precede AI namespaces? Not sure why Stylecop did not flag this... Will fix

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
karolz-ms: I was under assumption that any pipeline that can be created via configuration should also be buildable programmatically.

I did not want to expose the TelemetrySinks collection directly because of the invariant that the first sink is the "default" sink, always present.

src/Core/Managed/Shared/Extensibility/TelemetrySink.cs
karolz-ms: I mostly made TelemetrySink an ITelemetryProcessor out of convenience (can treat all the elements in the pipeline uniformly).

If you feel strongly about the requirement that ITelemetryProcessor has to have the abovementioned constructor, I'd rather remove ITelemetryProcessor designation from TelemetrySink. The constructor taking "next" processor just does not make sense for TelemetrySink.

src/Core/Managed/Shared/Extensibility/Implementation/BroadcastProcessor.cs
karolz-ms: I see your point, but this complexity buys you nice isolation between sinks/channels--see TelemetrySink tests FailingChannelDoesNotAffectOtherChannels and SlowChannelDoesNotAffectOtherChannels. Also it implicitly makes all sinks work in parallel, which we probably want regardless whether we call them synchronously or not. I felt these advantages outweigh the concerns about additional complexity.

src/Core/Managed/Shared/Channel/IDeepCloneable.cs
karolz-ms: Sergey, please confirm that 3rd party implementing a custom telemetry class is a non-scenario for AI SDK. If that is the case, adding deep cloning capability directly to ITelemetry sounds like a good idea

pharring:
I spoke to @karolz-ms offline, but just so I get my thoughts recorded: This has the potential to create a performance problem due to the deep cloning of every telemetry item. This proposed implementation takes care to clone only when there's more than one channel, but if multi-channel is going to become common - particularly at scale - then we should be mindful of the performance impact.

We discussed a few possible design patterns to reduce the allocations:
1. Immutable objects.
2. Copy-on-write
3. "Safe" telemetry processors.

Immutable objects would be most efficient, but unfortunately it would be a widespread breaking change for telemetry items. So, it's ruled out.

Copy-on-write would still allocate, but the allocations would be fairly cheap and they could be satisfied from a pool. It breaks reference identity, but so does deep cloning, so I assume that's allowed. The implementation would add complexity to every property setter and dictionary.

Safe (i.e. non-mutating) telemetry processors means that we can identify the set of non-mutating telemetry processors and avoid a deep clone if we find a chain consisting of only safe ones. "Safe" could be opt-in or opt-out. It could be discovered by reflection (attributes) or a marker interface.

Test/TestFramework/Shared/DeterministicTaskScheduler.cs
karolz-ms: Will fix, thx

src/Core/Managed/Shared/Channel/IDeepCloneable.cs
karolz-ms: I think it makes more sense as a separate interface. There are internal data contract classes that benefit from having it. 

I am going to move it to DataContracts namespace. 

We could consider making the interface internal, let me know if you want me to pursue that.

src/Core/Managed/Shared/Channel/IDeepCloneable.cs
karolz-ms: Will put the interface inside `Microsoft.ApplicationInsights.DataContracts`

src/Core/Managed/Shared/DataContracts/AvailabilityTelemetry.cs
karolz-ms: After thinking about it a bit I do not think that generating new ID makes much sense. Will fix.

src/Core/Managed/Shared/Extensibility/TelemetrySink.cs
karolz-ms: good point--will fix

Test/CoreSDK.Test/Shared/DataContracts/PerformanceCounterTelemetryTest.cs
karolz-ms: OK, I think I know why this was not flagged--StyleCop is not running for test code. As I said, will fix

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
karolz-ms: OK

karolz-ms:
I pushed a commit that should address most minor issues. The issues that still need to be agreed on

1. Whether we have `AddSink()` and `RemoveSink()` on `TelemetrySink`
2. Whether the telemetry is sent asynchronously to sinks (and what exactly is the alternative if it isn't)
3. Performance impact of the cloning, described by Paul above

Regarding the last issue, it is worth noting that copy-on-write is an implementation detail that can be transparently introduced in future AI SDK versions. OTOH non-mutating telemetry processors and channels require some public interface or attribute, which could still be introduced later, but is not as transparent, i.e. to benefit from it, the processor/channel need to declare itself as non-mutating.

Test/CoreSDK.Test/Net40/Core.Net40.Tests.csproj
```diff
@@ -24,6 +24,9 @@
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' " />
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' " />
   <ItemGroup>
+    <Reference Include="KellermanSoftware.Compare-NET-Objects, Version=3.8.0.0, Culture=neutral, PublicKeyToken=d970ace04cc85217, processorArchitecture=MSIL">
```
cijothomas: @Dmitry-Matveev - do we need any additional approvals for using new libraries?

src/Core/Managed/Shared/Channel/IDeepCloneable.cs
SergeyKanzhelev: I want to make sure that every telemetry is cloanable. And Application Insights SDK will never have anything else defined cloanable. With such a small public surface I think it is a safe assumption. And what if somebody will take a dependency on this interface? Application Insights SDK is not a general purpose library with common classes and this sort of dependency will force us to keep the interface around

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
SergeyKanzhelev: I'd than have a default sink and sink collection. Exposing collection is the way SDK configures things via config file

src/Core/Managed/Shared/Extensibility/TelemetrySink.cs
SergeyKanzhelev: Agree, let's remove. Do you think `ITelemetryChannel`'s `Send` method is suitable here?

src/Core/Managed/Shared/Extensibility/Implementation/BroadcastProcessor.cs
SergeyKanzhelev: My concern this brings complexity and make the behavior unclear. With this implementation you cannot implement the sink that will slow down `Track` so sending will keep up with the load if you need to implement back-pressure design for guarantee delivery instead of dropping items.

Code is good and can be used as a base class for event flow outputs.

SergeyKanzhelev:
@karolz-ms @pharring at first I was hesitant of idea to clone by default. However with the copy on write approach it shouldn't be such a big concern and it will not change the logic. (In most cases the only thing changed in item is a sampling percentage. So there will almost never be an allocation). And it also can be implemented later as we see performance issues. Thank you!

Test/CoreSDK.Test/Net40/Core.Net40.Tests.csproj
karolz-ms: @cijothomas @Dmitry-Matveev According to NuGet https://github.com/GregFinzer/Compare-Net-Objects/wiki/Licensing this component is licensed under MS public license (Ms-PL) so I do not think any extra approval is necessary... but maybe we need to "register the use" or something.

src/Core/Managed/Shared/Channel/IDeepCloneable.cs
karolz-ms: Roger. Will remove IDeepCloneable and add DeepClone() method to ITelemetry

src/Core/Managed/Shared/Extensibility/TelemetryConfiguration.cs
karolz-ms: OK, will try that

src/Core/Managed/Shared/Extensibility/TelemetrySink.cs
karolz-ms: OK will make sure TelemetrySink is not an ITelemetryProcessor.

It is not a ITelemetryChannel either (since it can have its own set of telemetry processors), so I am thinking it will have its own `Process` method.

src/Core/Managed/Shared/Extensibility/Implementation/BroadcastProcessor.cs
karolz-ms: Well, technically, you could still implement thread-affinity back-pressure using the current design. The logic in `AsyncCall.Post` today is "either enqueue, or drop". `Post` is synchronous, so the logic could be changed to "enqueue or wait" without changing the caller code.

My biggest concern though is I really do not have a good alternative in mind for the current implementation of the BroadcastProcessor. Assuming I have a `sinks` collection and a new `telemetry` object these are the alternatives I have considered so far

1. Simple iteration
```csharp
foreach(var sink in sinks) {
  try {
    sink.Process(telemetry); 
  } catch (Exception e) {
    // Process error
  }
}
```
Problem: sinks invoked one at a time, making the pipeline slower than it could be

2. Parallel iteration
```csharp
Parallel.For(0, sinks.Count, (sinkId) => {  
  try {
    sink.Process(telemetry); 
  } catch (Exception e) {
    // Process error
  }
});
```
Problem: I am really worried about the cost of setting up the Parallel.For machinery for every single piece of telemetry, esp. since the number of sinks will usually be small.

3. The current design, which is essentially a queue-per-sink, individually populated

@SergeyKanzhelev can you think of anything else?

SergeyKanzhelev:
``` csharp
foreach(var sink in sinks) {
  try {
    sink.Process(telemetry); 
  } catch (Exception e) {
    // Process error
  }
}
```
>Problem: sinks invoked one at a time, making the pipeline slower than it could be

If every sink implementation just do what broadcast thing does - put item in queue of sort and return the control back, performance will be the same, will it? This makes code much easier to understand and follow. And opens so many possibilities. Like if you have a sampling processor that samples very aggressively it may be more efficient to sample before scheduling task. You will save a lot of overhead of scheduler by queuing 1/100th of all telemetry

karolz-ms: OK, so I think you are saying we should execute sink processors synchronously and rely on the fact that both processors and `channel.Send` are implemented efficiently (with channel having its own internal queueing). With these assumptions the synchronous "simple iteration" variant will have acceptable performance.

I will put a comment to that effect int he `BroadcastProcessor` code and make it synchronous. I will also probably remove the AsyncCall implementation from the PR since it won't be used by the rest of the code anymore. It is part of the GitHub history now and we can resurrect it should it become necessary for some future channel implementation.

karolz-ms:
Hi @SergeyKanzhelev with the latest commit I believe I have addressed all the feedback. Please take another look-thx!

Test/CoreSDK.Test/Net40/Core.Net40.Tests.csproj
SergeyKanzhelev: I registered the use of component.

src/Core/Managed/Shared/Extensibility/Implementation/TelemetryConfigurationFactory.cs
```diff
@@ -83,15 +83,20 @@ public virtual void Initialize(TelemetryConfiguration configuration, TelemetryMo
                     configuration.InstrumentationKey = environmentIKey;
                 }
 
-                // Creating the default channel if no channel configuration supplied
-                configuration.TelemetryChannel = configuration.TelemetryChannel ?? new InMemoryChannel();
```
SergeyKanzhelev: does this change modifies existing behavior?

dnduffy: Yes it does, I expect some things to break if there is no default channel created somewhere...

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -419,6 +419,14 @@ public void MetricExtractorAfterSamplingVerbose(string appDomainName = "Incorrec
             this.WriteEvent(34, this.nameProvider.Name);
         }
 
+        [Event(35,
+            Message = "At least one telemetry channel cannot consume incoming telemetry fast enough. Some telemetry was dropped to prevent out of memory condition.",
```
dnduffy: nitpick: "...prevent an out of memory condition."

Can we also include the number of items that were dropped and which channel it was that dropped them in the trace message?

src/Core/Managed/Shared/Extensibility/Implementation/TelemetryConfigurationFactory.cs
karolz-ms: David is correct. This was simply moved from `TelemetryConfigurationFactory`. Tests seem to confirm that the behavior has not been changed

src/Core/Managed/Shared/Extensibility/Implementation/Tracing/CoreEventSource.cs
karolz-ms: Good suggestions, but... this event became unnecessary after the last round of changes. I just missed it when doing the refactoring.

I have pushed a commit that removes the new event entirely.

karolz-ms:
@SergeyKanzhelev so, any chance to review/approve this change soon? I have another one (configuration support) coming shortly...

karolz-ms:
Super, thanks @SergeyKanzhelev !! I will submit a PR that adds config. support later today

PR #586, opened 2017-07-09
Remove the use of concurrent diuctionary for context properties on hot path

### LINKED ISSUE #585: Minimize the number of concurent dictionary lookups
Performance tests for 2.4 release shown quite a few concurrent dictionary lookups on hot path. Not a root cause of performance overhead, but worth fixing.

SergeyKanzhelev:
Fix #585.

This PR changes the behavior of `Sanitize` and `Track` methods. Before - all context properties were sanitized after the call to `Track`. Now they would not be. This is not a big deal as we explicitly tell not to re-use items in documentation. Also with multi-sink support feature this behavior should be changed anyway.

- [X] Design discussion issue #585
- [X] Changes in public surface reviewed
- [X] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 This changelog will be used to generate documentation on [release notes page](http://azure.microsoft.com/en-us/documentation/articles/app-insights-release-notes-dotnet/).
 
+## Version 2.5.0-beta1
+- Method `Sanitize` on classes implementing `ITelemetry` does not modify the `TelemetryContext` fields any longer. Serailized event json and ETW event will still have context tags sanitized.
+
 ## Version 2.4.0
 - Updated version of DiagnosticSource to 4.4.0 stable
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 This changelog will be used to generate documentation on [release notes page](http://azure.microsoft.com/en-us/documentation/articles/app-insights-release-notes-dotnet/).
 
 ## Version 2.5.0-beta1
-- Method `Sanitize` on classes implementing `ITelemetry` does not modify the `TelemetryContext` fields any longer. Serailized event json and ETW event will still have context tags sanitized.
+- Method `Sanitize` on classes implementing `ITelemetry` no longer modifies the `TelemetryContext` fields. Serialized event json and ETW event will still have context tags sanitized.
 
 ## Version 2.4.0
 - Updated version of DiagnosticSource to 4.4.0 stable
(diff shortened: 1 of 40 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 This changelog will be used to generate documentation on [release notes page](http://azure.microsoft.com/en-us/documentation/articles/app-insights-release-notes-dotnet/).
 
+## Version 2.5.0-beta1
+- Method `Sanitize` on classes implementing `ITelemetry` does not modify the `TelemetryContext` fields any longer. Serailized event json and ETW event will still have context tags sanitized.
```
pharring: Suggest:
Method `Sanitize` on classes implementing `ITelemetry` no longer modifies the `TelemetryContext` fields.

Also: Typo: Serailized -> Serialized

Test/CoreSDK.Test/Shared/DataContracts/TelemetryContextTest.cs
```diff
@@ -26,10 +26,11 @@ public void TelemetryContextIsSealedToSupportCompilationAsWinmd()
         }
 
         [TestMethod]
+        [Ignore]
```
pharring: Why `Ignore`? Add a comment or, if you never intend to bring it back, delete the test.

src/Core/Managed/Shared/Extensibility/Implementation/External/Tags.cs
```diff
@@ -147,38 +64,29 @@ internal static void SetDateTimeOffsetValueOrRemove(this IDictionary<string, str
...
-            }
-        }
-
         internal static void SetTagValueOrRemove<T>(this IDictionary<string, string> tags, string tagKey, T tagValue)
         {
             SetTagValueOrRemove(tags, tagKey, Convert.ToString(tagValue, CultureInfo.InvariantCulture));
         }
 
-        internal static void InitializeTagValue<T>(this IDictionary<string, string> tags, string tagKey, T tagValue)
+        internal static string CopyTagValue(string sourceValue, string targetValue)
```
pharring: Consider this form:
```C#
internal static void CopyTagValue(string srouceValue, ref string target)
{
    if (!string.IsNullOrEmpty(sourceValue) || string.IsNullOrEmpty(target))
    {
        target = sourceValue;
    }
}
```

Then callers do this:
```Tags.CopyTagValue(other.FooTag, ref this.FooTag)```

Since the assignment is done within the body of CopyTagValue, that reduces the code that callers have to write (and eliminates one potential source of bugs where they assign to the wrong field) and it eliminates the write if the value is not to be updated (small perf win).

PR #589, opened 2017-07-24
Add multi-channel capability (part 2: configuration)

karolz-ms:
This is related to [issue 559](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/559) and completes the implementation of multi-channel capability.

### DIFF
--- a/src/Core/Managed/Shared/Extensibility/Implementation/TelemetryConfigurationFactory.cs
+++ b/src/Core/Managed/Shared/Extensibility/Implementation/TelemetryConfigurationFactory.cs
@@ -19,6 +19,7 @@
     {
         private const string AddElementName = "Add";
         private const string TypeAttributeName = "Type";
+        private const string NameAttributeName = "Name";
         private const string InstrumentationKeyWebSitesEnvironmentVariable = "APPINSIGHTS_INSTRUMENTATIONKEY";
         private static readonly MethodInfo LoadInstancesDefinition = typeof(TelemetryConfigurationFactory).GetRuntimeMethods().First(m => m.Name == "LoadInstances");
         private static readonly XNamespace XmlNamespace = "http://schemas.microsoft.com/ApplicationInsights/2013/Settings";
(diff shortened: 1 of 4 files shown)


src/Core/Managed/Shared/Extensibility/Implementation/TelemetryConfigurationFactory.cs
```diff
@@ -202,14 +204,48 @@ protected static void BuildTelemetryProcessorChain(XElement definition, Telemetr
...
-                        return (ITelemetryProcessor)LoadInstance(addElement, typeof(ITelemetryProcessor), telemetryConfiguration, constructorArgs, null);
+                        return (ITelemetryProcessor)LoadInstance(addElement, typeof(ITelemetryProcessor), null, constructorArgs, null);
                     });
                 }
             }
 
             builder.Build();
         }
 
+        protected static void LoadTelemetrySinks(XElement definition, TelemetryConfiguration telemetryConfiguration)
```
SergeyKanzhelev: I wonder why we need a special function to read it. Is it to avoid `<Add` under `<Sinks`?

karolz-ms: No, I am not trying to avoid `<Add>` element under `<Sinks>`, in fact, using `<Add>` is the way to create sinks. There is a couple of reasons why special handling for TelemetrySink is necessary

1. Just like for TelemetryProcessors, the sink has to be created from a constructor that takes parameters (telemetry configuration in the sink case)

2. As you can see, there is special logic involving default sink (always present, and designated by the name "default"). Default sink is never created.

3. Finally, if the sink is referred to multiple times in the configuration (by the same name), we apply all the configuration to the single, named instance. This part is debatable, but I had to choose some behavior in case I am facing repeated sink configuration with the same name and I think this is a reasonable choice.

Take a look at the tests and it will make things clearer.

SergeyKanzhelev: Can you please copy this comment verbatim as a summary of the method for the future reference

SergeyKanzhelev: I'm OK with this explanaition

karolz-ms: Comment added

PR #595, opened 2017-07-31
Fix typo in Transmitter log about moving data from buffer to sender

xtt129:
In the ApplyPolicies method of Transimitter, the log information should be 'MovedFromBufferToSender' instead of 'MovedFromSenderToBuffer'.


SergeyKanzhelev:
Thank you!

Issue #458, opened 2017-03-13
Upgrade ServerTelemetryChannel projects to SDK projects


pharring:
#451 shows the way for removing shared projects and using the new SDK project types.
We could do the same thing for the projects under ServerTelemetryChannel and its NuGet project.

PR #598, opened 2017-08-03
Adaptive sampling for .NET core and couple clean up tasks

SergeyKanzhelev:
Fix:
- Change OSS signing to the supported way to do it https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/639
- Need to enable adaptive sampling for pure .NET Core apps https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/458
- Fixing https://github.com/Microsoft/ApplicationInsights-dotnet/issues/458
- Also make StyleCop an error

- [X] Design discussion issue **no discussions needed**
- [X] Changes in public surface reviewed - **expose existing types for netstandard1.3 target**
- [X] CHANGELOG.md updated - **added**


src/Core/NuGet/Core.NuGet.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildProjectDirectory), 'Common.props'))\Common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net45</TargetFrameworks>
+    <TargetFrameworks>netstandard1.3</TargetFrameworks>
```
Dmitry-Matveev: Is it supposed to be only net1.3? Will it install over 4.0/4.5/4.6?
I see that other projects you were changing are now targeting several net versions at the same time.

SergeyKanzhelev: it is just the "last" target so the packing of the NuGet will happen after Base SDK were compiled for all projects.

PR #599, opened 2017-08-07
Fix random build breaks

SergeyKanzhelev:
Fix build breaks related to tests synchronization by introducing context aware event source listener.

Also fix for NuGet package location

### DIFF
--- a/Test/TestFramework/Shared/TestEventListener.cs
+++ b/Test/TestFramework/Shared/TestEventListener.cs
@@ -1,5 +1,6 @@
 ﻿namespace Microsoft.ApplicationInsights.TestFramework
 {
+    using Microsoft.ApplicationInsights.Extensibility;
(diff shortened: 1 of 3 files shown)


Test/TestFramework/Shared/TestEventListener.cs
```diff
@@ -10,24 +11,105 @@
...
+
+            public static void Enter()
+            {
+                try
+                {
+                    CallContext.LogicalSetData(InternalOperationsMonitorSlotName, syncObj);
+                }
+                catch (Exception)
+                {
+                    // CallContext may fail in partially trusted environment
```
Dmitry-Matveev: Does it mean that some tests may pass in this case? No exception will happen and it looks like no traces will be collected (including error traces) because enter() failed for test trace listener itself.

PR #604, opened 2017-08-10
Make instrumentation key optional

karolz-ms:
Addresses [issue 344](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/344) and completes work on [issue 559](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/559)

### DIFF
--- a/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
+++ b/src/Core/Managed/Shared/Channel/InMemoryChannel.cs
@@ -133,6 +133,16 @@
                 return;
             }
 
+            if (string.IsNullOrEmpty(item.Context.InstrumentationKey))
+            {
+                if (CoreEventSource.Log.IsVerboseEnabled)
+                {
+                    CoreEventSource.Log.ItemRejectedNoInstrumentationKey(item.ToString());
+                }
+
+                return;
+            }
+
             try
             {
                 this.buffer.Enqueue(item);
--- a/src/Core/Managed/Shared/TelemetryClient.cs
+++ b/src/Core/Managed/Shared/TelemetryClient.cs
@@ -411,11 +411,6 @@
             {
                 this.Initialize(telemetry);
 
-                if (string.IsNullOrEmpty(telemetry.Context.InstrumentationKey))
-                {
-                    TelemetryDebugWriter.WriteTelemetry(telemetry);
-                }
-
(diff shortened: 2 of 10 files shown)


SergeyKanzhelev:
Please update `CHANGELOG.md`

SergeyKanzhelev:
@tokaplan will QuickPulse Module be OK with this change or you need to add the null check for instrumentation key in your module as well?

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
```diff
@@ -133,6 +133,16 @@ public void Send(ITelemetry item)
                 return;
             }
 
+            if (string.IsNullOrEmpty(item.Context.InstrumentationKey))
+            {
+                if (CoreEventSource.Log.IsVerboseEnabled)
+                {
+                    CoreEventSource.Log.ItemRejectedNoInstrumentationKey(item.ToString());
```
SergeyKanzhelev: what `ToString` does to ITelemetry? Maybe output item type instead?

src/Core/Managed/Shared/TelemetryClient.cs
```diff
@@ -414,7 +414,6 @@ public void Track(ITelemetry telemetry)
                 if (string.IsNullOrEmpty(telemetry.Context.InstrumentationKey))
                 {
                     TelemetryDebugWriter.WriteTelemetry(telemetry);
```
SergeyKanzhelev: I'm not sure this statement will be needed any longer. It may start output the same item twice in debug window.

karolz-ms: OK, will remove

SergeyKanzhelev: Can you please double check that debug output will still have this telemetry item written

karolz-ms: I have verified that the item still shows up in the Debug output

src/Core/Managed/Shared/Channel/InMemoryChannel.cs
karolz-ms: item.ToString() will output item type (or at least it does for TraceTelemetry). Do you think this is good enough?

PR #606, opened 2017-08-11
no more core

SergeyKanzhelev:
No more mentions of Core as it was very confusing


src/Microsoft.ApplicationInsights/NuGet/Package.nuspec
```diff
@@ -40,32 +40,32 @@
   </metadata>
   <files>
     <!-- Binaries for .NET Standard 1.3 -->
-    <file src="$configuration$\Core\Managed\netstandard1.3\Microsoft.ApplicationInsights.dll" target="lib\netstandard1.3" />
-    <file src="$configuration$\Core\Managed\netstandard1.3\Microsoft.ApplicationInsights.pdb" target="lib\netstandard1.3" />
-    <file src="$configuration$\Core\Managed\netstandard1.3\Microsoft.ApplicationInsights.xml" target="lib\netstandard1.3" />
+    <file src="$configuration$\Microsoft.ApplicationInsights\Managed\netstandard1.3\Microsoft.ApplicationInsights.dll" target="lib\netstandard1.3" />
```
Dmitry-Matveev: I remember we had an issue with the long paths in the nuget packages location (reference file paths were too lengthy within the application being on-boarded in VS), and here we add ~25 extra characters. However, if this nuget will mostly be installed with Microsoft.ApplicationInsights.Web - then we should be OK.

SergeyKanzhelev: This is compile-time only path. Resulting NuGet will not change a bit

PR #608, opened 2017-08-14
Get rid of xunit

SergeyKanzhelev:
We only used xunit for fancy asserts and not for running actual tests. Major problem with xunit is that latest version of it targets 452+ when SDK targets 40+. So we used weird combination of older versions of xunit including some pre-release versions. To the extend that some APIs were not quite supported.

With this commit I'm removing all xunit references from the project


Test/TestFramework/Shared/AssertExtensions.cs
```diff
@@ -0,0 +1,252 @@
...
+#endif
+
+
+    internal class AssertEx
+    {
+        public static void AreEqual<T>(IEnumerable<T> expected, IEnumerable<T> actual)
+        {
+            var a = actual.GetEnumerator();
+
+            foreach(var t in expected)
```
Dmitry-Matveev: This seems to rely on equal ordering of the collections. I see that this assert is mostly used for Arrays above, so should not be an issue right now, but someone might want to use it to compare non-ordered collections later and this limitation is not implied in the method name.

PR #612, opened 2017-08-17
Fix #610 and few more things

### LINKED ISSUE #610: Default sink get stuck with the standard channel even after re-placing it from code
See https://github.com/Microsoft/ApplicationInsights-aspnetcore/pull/517 for working theory of what's broken

SergeyKanzhelev:
Fix #610.

Also:
- convert telemetry channel tests 4.5 to the new project type
- enable pdb generation for better unit test expirience
- do not package xdt files as contentFiles so they would not appear as content in new projects


SergeyKanzhelev:
Sorry for mixing more stuff in the bug fix

Issue #654, opened 2017-11-20
ServerTelemetryChannel for .Net Core Apps


cijothomas:
ServerTelemetryChannel was enabled for .NET Core Applications starting 2.5.0-beta1 (https://github.com/Microsoft/ApplicationInsights-dotnet/releases/tag/v2.5.0-beta1)
This enabled .NET Core apps to use full functionalities provided by ServerTelemetryChannel including persistent storage, retries, adaptive sampling etc. 
However, for non-windows users, this release caused an issue. (https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/551)

ServerTelemetryChannel relied on windows specific mechanisms to determine (and secure) local storage folder where telemetry items are stored when it cannot be sent to Application Insights due to temporary network issues. A workaround is currently made (Insert PR when available).
With the above fix in place, starting with beta2 following are the options for Non-Windows users to fully use persistent storage capabilities.
1. Create a local storage folder manually, and specify this in ServerTelemetryChannel.StorageFolder. Please make sure that the process running your application has write access to this folder, but also make sure this folder is protected from other users to avoid telemetry being read by unintended users.
2. Alternatively, if you don't care about whether the folder where ServerTelemetryChannel stores telemetry items is restricted to current/admin user, then set AllowUnsecureLocalStorage to true on ServerTelemetryChannel. This allows ServerTelemetryChannel to create a sub-directory in /tmp, and store telemetry items there.

If neither of above is done, telemetry will still be collected and sent, but in case of network issues, items will get dropped.

Issue #416, opened 2017-01-30
TelemetryClient.Initialize should not crash in partial trust scenario


SergeyKanzhelev:
Reported here: https://github.com/Microsoft/ApplicationInsights-Home/issues/87

`TelemetryClient` and other components should not crash when app is running in partial trust

`System.Security.SecurityException
 at System.Security.CodeAccessSecurityEngine.Check(Object demand, StackCrawlMark&amp; stackMark, Boolean isPermSet)
 at System.Security.CodeAccessPermission.Demand()
 at System.Environment.GetEnvironmentVariables()
 at Microsoft.ApplicationInsights.Extensibility.Implementation.Platform.PlatformImplementation..ctor()
 at Microsoft.ApplicationInsights.Extensibility.Implementation.Platform.PlatformSingleton.get_Current()
 at Microsoft.ApplicationInsights.TelemetryClient.Initialize(ITelemetry telemetry)
 at Microsoft.ApplicationInsights.TelemetryClient.Track(ITelemetry telemetry)
 at Microsoft.ApplicationInsights.TelemetryClient.TrackTrace(TraceTelemetry telemetry)
 at Microsoft.ApplicationInsights.TelemetryClient.TrackTrace(String message)`

dnduffy:
PR https://github.com/Microsoft/ApplicationInsights-dotnet/pull/420

dberryalliance:
@dnduffy The constructor for `TransmissionProcessor` calls into the `PlatformSingleton` to asses debugging status.  This also breaks in PartialTrust mode.

fedejousset:
I'm still getting an error trying to use TelemetryClient in partial trust mode:

System.Security.SecurityException: Request for the permission of type 'System.Security.Permissions.FileIOPermission, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'

> at System.Security.CodeAccessSecurityEngine.Check(Object demand, StackCrawlMark&amp; stackMark, Boolean isPermSet)
> at System.Security.CodeAccessPermission.Demand()
> at System.AppDomainSetup.VerifyDir(String dir, Boolean normalize)
> at Microsoft.ApplicationInsights.Extensibility.Implementation.Platform.PlatformImplementation.ReadConfigurationXml()
[...]

PR #655, opened 2017-11-20
Non windows persistent storage short-term fix

cijothomas:
Fix issue https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/551
Short description of the fix:
For non-windows users ServerTelemetryChannel will not automatically create a directory and store telemetry items when network issues occur. This is because, currently .net core does not have an api to set file permissions, so this temp directory cannot be secured. Users can, however, set a flag AllowUnsecureLocalStorage to true, on ServerTelemetryChannel. to override this behavior. Alternatively, users can create a directory themselves, and pass it to ServerTelemetryChannel.StorageFolder. (this functionality existed originally)

Checklist:
- [x] I ran all unit tests locally


cijothomas:
Some notes:
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/654

src/ServerTelemetryChannel/Managed/Shared/Implementation/TelemetryChannelEventSource.cs
```diff
@@ -369,7 +369,7 @@ public void TransmissionSendingFailedWarning(string transmissionId, string excep
         [Event(
             55, 
             Keywords = Keywords.UserActionable,
-            Message = "Access to the local storage was denied (User: {1}). If you want Application Insights SDK to store telemetry locally on disk in case of transient network issues please give the process access either to %LOCALAPPDATA% or to %TEMP% folder. After you gave access to the folder you need to restart the process. Currently monitoring will continue but if telemetry cannot be sent it will be dropped. Error message: {0}.", 
+            Message = "Local storage access has resulted in an error (User: {1}). If you want Application Insights SDK to store telemetry locally on disk in case of transient network issues please give the process access to %LOCALAPPDATA% or %TEMP% folder. After you gave access to the folder you need to restart the process. Currently monitoring will continue but if telemetry cannot be sent it will be dropped. Error message: {0}.",
```
cijothomas: might need to link to an actual documentation instead of this long but in-complete (due to lack of instruction for linux/docker) text.

cijothomas:
based on offline dicussion with dmitry,
(This will be modified to discontinue use of the confusing flag - AllowUnsecureLocalStorage’ , and allow local persistence folder in Non-Windows ONLY if user explicitly supply a StorageFolder.)

src/ServerTelemetryChannel/Implementation/ApplicationFolderProvider.cs
```diff
@@ -11,12 +11,15 @@
...
     {
+        internal Func<DirectoryInfo, bool> ApplySecurityToDirectory;
+
         private readonly IDictionary environment;
         private readonly string customFolderName;
-        private readonly WindowsIdentity currentIdentity = WindowsIdentity.GetCurrent();
+        private IIdentityProvider identityProvider;
 
         public ApplicationFolderProvider(string folderName = null)
             : this(Environment.GetEnvironmentVariables(), folderName)
```
SergeyKanzhelev: you may not have permissions to environment variables and this code will crash. Something like #416 will happen

src/ServerTelemetryChannel/Implementation/ApplicationFolderProvider.cs
```diff
@@ -11,12 +11,15 @@
...
+    using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Shared.Implementation;    
 
     internal class ApplicationFolderProvider : IApplicationFolderProvider
     {
+        internal Func<DirectoryInfo, bool> ApplySecurityToDirectory;
+
         private readonly IDictionary environment;
         private readonly string customFolderName;
-        private readonly WindowsIdentity currentIdentity = WindowsIdentity.GetCurrent();
+        private IIdentityProvider identityProvider;
```
SergeyKanzhelev: it can still be readonly. It's initialized in constructor

cijothomas: done

src/ServerTelemetryChannel/Implementation/ApplicationFolderProvider.cs
cijothomas: Yes. Will track separately as I don't want to make additional changes with this fix.
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/657

src/ServerTelemetryChannel/Implementation/ApplicationFolderProvider.cs
```diff
@@ -128,6 +154,10 @@ private IPlatformFolder CreateAndValidateApplicationFolder(string rootPath, bool
                     if (createSubFolder)
                     {
                         telemetryDirectory = this.CreateTelemetrySubdirectory(telemetryDirectory);
+                        if (!this.ApplySecurityToDirectory(telemetryDirectory))
+                        {                            
+                            throw new SecurityException("Unable to apply security restrictions to the storage directory.");
```
Dmitry-Matveev: It looks like this exception/message is going to be thrown every time this channel is created in non-Windows environment because `this.ApplySecurityToDirectory()` will return `false` there. Comment in the `catch` below states that "// The caller does not have code access permission to create the directory" is the reason for this `SecurityException`. However, the actual reason is that we simply do not support that platform.

src/ServerTelemetryChannel/Implementation/TelemetryChannelEventSource.cs
```diff
@@ -481,6 +482,12 @@ public void ItemRejectedNoInstrumentationKey(string item, string appDomainName =
             this.WriteEvent(67, item ?? string.Empty, this.ApplicationName);
         }
 
+        [Event(68, Message = "Failed to set access permissions on storage directory {0}. Error : {1}.", Level = EventLevel.Warning)]
```
Dmitry-Matveev: Should it direct user to the creation of the custom folder for the telemetry storage in case of non-Windows platforms?

cijothomas: This is a warning message and won't be automatically picked up y SDK diagnostics. There is an error message which will inform the user to create folder themselves.

PR #663, opened 2017-12-01
AdaptiveSampling causes starvation of Telemetry over time

d3r3kk:
My proposed fix for Microsoft/ApplicationInsights-dotnet-server#756.

- [x] Design discussion issue Microsoft/ApplicationInsights-dotnet-server#756

The problem: `AdaptiveSamplingTelemetryProcessor` in the default configuration within the 
ApplicationInsights.config file will cause starvation of telemetry over time.

The root cause: `AdaptiveSamplingTelemetryProcessor` makes use of `SamplingTelemetryProcessor`
chained directly to `SamplingPercentageEstimatorTelemetryProcessor`. This causes all telemtry,
sampled or not, to pass into the estimator and become counted as part of the calculation of the
moving average, and therefore part of the adapted sampling average applied to the chain. Some
telemetry passing through the AdaptiveSamplingTelemetryProcessor should never be counted in that
calculation (for instance, if the telemetry is of a type that should never be sampled at all, or
if the telemetry item has already been sampled by a prior ITelemetryProcessor).

The proposed fix: Alter the `SamplingTelemetryProcessor` so that it has two `Next` properties
to send `ITelemetry` items to, one for sampled items and one for unsampled items. 

- Make an internal constructor that populates the new `UnsampledNext` property, and update the 
  public one to ensure that the `Next` and `UnsampledNext` properties are both pointing at a valid
 `ITelemetryProcessor` later in the pipeline.
- Update the `AdaptiveSamplingTelemetryProcessor` constructor to use the internal constructor of
  `SamplingTelemetryProcessor`, giving it the true next `ITelemetryProcessor` in the pipeline as
  the unsampledNext processor, and the `SamplingPercentageEstimatorTelemetryProcessor` as the
  sampledNext processor.
- Update the `SamplingTelemetryProcessor.Process` method to call the `UnsampledNext` processor 
  whenever there should be no sampling done on the current item at all.

## Root Cause (with pictures!) ##

The way that the `SamplingTelemetryProcessor` works is that it either accepts or rejects telemetry. This is entirely desirable behavior and users can employ the `SamplingTelemetryProcessor` if they wish to do so, when their situation does not require any adaptive corrections of data rates over time. Telemetry items are sampled out based on the percent of data desired, and any skipped item that it encounters is allowed to continue through the pipeline.

samplingdataflow

And the `AdaptiveSamplingTelemetryProcessor` leverages the `SamplingTelemetryProcessor`, but injects a new processor after it in the pipeline, which calculates a new sample percent as items are handed to it, which feeds back to into the `SamplingTelemetryProcessor` over time to achieve a constant amount of data flowing through the pipeline, and that will respond to changes in the quantity of telemetry being sent.

adaptivesamplingdataflow-problem

Unfortunately, not all telemetry items being sent through the `SamplingTelemetryProcessor` are the same. Some are evaluated and dropped, some are evaluated and allowed to progress through the pipeline, and some are simply unevaluated and allow to progress through the pipeline. It's the last class of telemetry item that is a problem here - unevaluated items should not be counted against the current sample percentage.

adaptivesamplingdataflow-solutionenhanced

## Data Tests ##

Using the above hypothesis I made use of a testing program (thanks @zakimaksyutov!) that I extended to collect data and organize it into a spreadsheet-friendly format. I have a lot of data from the tests, but here's some summary charts (full data attached below).

The test program runs for N minutes (I ran it for 30) and emits 100 RequestTelemetry items per second, and 10 EventTelemetry items per second throughout the entire test period. A custom ITelemetryProcessor is injected into the end of the pipeline to count the number of Requests and Events that are actually emitted, making it through the `AdaptiveSamplingTelemetryProcessor`.

*Note that the data in these summary charts include data after the initial 2-minute sampling period where the data is left unsampled at 100%*

Requests and Events emitted over 30 minutes, prior to fix (using SDK 2.5.0-beta1):

beforefix-dataafterninitialsampleperiod

Requests and Events emitted over 30 minutes, after applying my fix (SDK 2.5.0-beta2-local_build):

fixappliedandupdated-dataafterninitialsampleperiod

~~From the data above it appears that we've corrected the initial starvation problem. However, there appears to be another issue where the data flow increases after about 1000 seconds (16 minutes) and the rate is higher than I might expect. I am currently looking at that problem as a separate issue.~~ (Fixed)

The raw data with added charts & summarized tables:
[Work-Comparison-AdaptiveSamplingStarve.xlsx](https://github.com/Microsoft/ApplicationInsights-dotnet/files/1522943/Work-Comparison-AdaptiveSamplingStarve.xlsx)

*EDIT*: Added template checkboxes for PRs
*EDIT*: After further fixes, updated resulting data emitted chart, updated XLS data analysis file, and the solution flow diagram

### DIFF
--- a/src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
+++ b/src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
@@ -2,7 +2,6 @@
 {
     using System;
     using System.Collections.Generic;
-    using System.Globalization;
     using System.Threading;
     
     using Microsoft.ApplicationInsights.Channel;
@@ -45,6 +44,11 @@
 
             this.SamplingPercentage = 100.0;
             this.Next = next;
+            if (this.UnsampledNext == null)
+            {
+                this.UnsampledNext = next;
+            }
+
             this.excludedTypesHashSet = new HashSet<Type>();
             this.includedTypesHashSet = new HashSet<Type>();
             this.allowedTypes = new Dictionary<string, Type>(6, StringComparer.OrdinalIgnoreCase)
@@ -56,6 +60,16 @@
                 { RequestTelemetryName, typeof(RequestTelemetry) },
                 { TraceTelemetryName, typeof(TraceTelemetry) },
             };
+        }
+
+        internal SamplingTelemetryProcessor(ITelemetryProcessor unsampledNext, ITelemetryProcessor sampledNext) : this(sampledNext)
+        {
+            if (unsampledNext == null)
+            {
+                throw new ArgumentNullException("unsampledNext");
+            }
+
+            this.UnsampledNext = unsampledNext;
         }
 
         /// <summary>
@@ -140,6 +154,19 @@
         private ITelemetryProcessor Next { get; set; }
 
         /// <summary>
+        /// Gets or sets the next TelemetryProcessor to call in the chain if the ITelemetry item passed in is /not/ sampled.
+        /// 
+        /// <remarks>
+        /// Sampling telemetry processor is used standalone (can be added to the Application Insights config file
+        /// directly) as well as in a combination within the processing chain sandwiched between the 
+        /// AdaptiveSamplingTelemetryProcessor and the SamplingPercentageEstimatorTelemetryProcessor. We cannot count
+        /// unsampled items in this class's processor routine towards adaptive sampling samples, so we need to bypass
+        /// those unsampled items passed the estimator to be accurate. See issue (blah) for more details.
+        /// </remarks>
+        /// </summary>
+        private ITelemetryProcessor UnsampledNext { get; set; }
+
+        /// <summary>
         /// Process a collected telemetry item.
         /// </summary>
         /// <param name="item">A collected Telemetry item.</param>
@@ -160,7 +187,7 @@
             var samplingSupportingTelemetry = item as ISupportSampling;
             if (samplingSupportingTelemetry == null)
             {
-                this.Next.Process(item);
+                this.UnsampledNext.Process(item);
                 return;
             }
 
@@ -172,7 +199,7 @@
                     TelemetryChannelEventSource.Log.SamplingSkippedByType(item.ToString());
                 }
 
-                this.Next.Process(item);
+                this.UnsampledNext.Process(item);
                 return;
             }
 
@@ -180,7 +207,7 @@
             bool itemAlreadySampled = samplingSupportingTelemetry.SamplingPercentage.HasValue;
             if (itemAlreadySampled)
             {
-                this.Next.Process(item);
+                this.UnsampledNext.Process(item);
                 return;
             }
 
--- a/src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
+++ b/src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
@@ -43,11 +43,8 @@
             }
 
             this.SamplingPercentage = 100.0;
-            this.Next = next;
-            if (this.UnsampledNext == null)
-            {
-                this.UnsampledNext = next;
-            }
+            this.SampledNext = next;
+            this.UnsampledNext = next;
 
             this.excludedTypesHashSet = new HashSet<Type>();
             this.includedTypesHashSet = new HashSet<Type>();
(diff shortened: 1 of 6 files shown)


src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -45,6 +44,11 @@ public SamplingTelemetryProcessor(ITelemetryProcessor next)
 
             this.SamplingPercentage = 100.0;
             this.Next = next;
+            if (this.UnsampledNext == null)
```
zakimaksyutov: I think it might be always true

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -58,6 +62,16 @@ public SamplingTelemetryProcessor(ITelemetryProcessor next)
             };
         }
 
+        internal SamplingTelemetryProcessor(ITelemetryProcessor unsampledNext, ITelemetryProcessor sampledNext) : this(sampledNext)
```
zakimaksyutov: [not blocking] If we need it ourselves - does it mean that we should expose it for customers as well?

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -139,6 +153,19 @@ public string IncludedTypes
...
 
+        /// <summary>
+        /// Gets or sets the next TelemetryProcessor to call in the chain if the ITelemetry item passed in is /not/ sampled.
+        /// 
+        /// <remarks>
+        /// Sampling telemetry processor is used standalone (can be added to the Application Insights config file
+        /// directly) as well as in a combination within the processing chain sandwiched between the 
+        /// AdaptiveSamplingTelemetryProcessor and the SamplingPercentageEstimatorTelemetryProcessor. We cannot count
+        /// unsampled items in this class's processor routine towards adaptive sampling samples, so we need to bypass
+        /// those unsampled items passed the estimator to be accurate. See issue (blah) for more details.
```
zakimaksyutov: ?

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -139,6 +153,19 @@ public string IncludedTypes
         /// </summary>
         private ITelemetryProcessor Next { get; set; }
```
zakimaksyutov: Potentially rename it to SampledNext?

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -160,7 +187,7 @@ public void Process(ITelemetry item)
             var samplingSupportingTelemetry = item as ISupportSampling;
```
zakimaksyutov: I think the condition which checks samplingPercentage >= 100 should honor Next/UnsampledNext.

zakimaksyutov: Without it different stream will be affecting and causing oscillation...

zakimaksyutov: And we can see this is happening in "after fix" picture (love pictures! =))

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
d3r3kk: Yes, I've removed the if statement and am just setting Next = UnsampledNext in the public constructor.

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
d3r3kk: Not certain anyone else uses it in the way we have with `AdaptiveSamplingTelemetryProcessor` (see below for a flow chart).

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
d3r3kk: Removing the <remark> part of this comment altogether. This PR and the issue will serve as the documentation.

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
d3r3kk: Sure, done.

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
d3r3kk: I made the change in a way that keeps the original intent of the `SamplingTelemetryProcessor` having an early exit for items when the sampling rate is at 100%, but forces evaluations to occur when it's being used as part of the `AdaptiveSamplingTelemetryProcessor` chain.

Same test program and time range as the original 2 charts, but notice there is no lag in events immediately following the initial sampling:

fixappliedandupdated-dataafterninitialsampleperiod

zakimaksyutov: Looks great!

d3r3kk:
Thanks @Dmitry-Matveev!

dhaval24:
@d3r3kk This is a great observation. I will make sure to keep this in mind when implementing this in Java (in near few days).

Issue #644, opened 2017-11-01
Introduce new TrackWork API


lmolkova:
We now have `StartOperation` and `StopOperation` `TelemetryClient` extension methods to track custom operations.

While it's pretty useful and flexible, typical scenario requires customer to write something like 

```C#
    var operation = telemetryClient.StartOperation<RequestTelemetry>("name");

    try
    {
        await DoWorkAsync();
        operation.Telemetry.Success = true;
    }
    catch (Exception e)
    {
        telemetryClient.TrackException(e);
        operation.Telemetry.Success = false;
        throw;
    }
    finally
    {
        telemetryClient.StopOperation(operation);
    }
```

This approach potentially has an issue with execution context: it may not be preserved between Start/Stop, so the operation may not be tracked at all.
All above code except `DoWorkAsync` has nothing to do with user code.

# Proposal
Introduce new TelemetryClient extensions methods:
- `async Task TrackWorkAsync( Func<Task> work, Activity activity=null)`
- `async Task TrackWorkAsync( Func<Task> work, OperationTelemetry telemetry)`

- other necessary overloads

These methods would do the same as above example does.

# Real world scenario
I'm working on the queues instrumentation and majority of queue clients (Azure Storage Queue, EventHub, certain ServiceBus usage scenarios) have API to receive a message(s) from the queue but do not include API to process a message.

I. e. if dependency tracing is implemented, AppInsights then would track 'receive' call to the queue, but message processing should be instrumented manually. 

Queues will provide method to extract Activity from the messages, so customers are expected to write code
``` C#
var message = queue.Receive();
var activity = message.ExtractActivity();

telemetryClient.TrackWork( () => DoWorkAsync(), activity );
```

@SergeyKanzhelev @Dmitry-Matveev  @TomMilos thoughts?

lmolkova:
@SergeyKanzhelev @Dmitry-Matveev @TomMilos ping

zakimaksyutov:
@lmolkova, is TrackWork a syntax sugar for above example? Should it take operation name as parameter as well then? Or does it deduce it somehow? If yes then can you please elaborate on how it does it?

SergeyKanzhelev:
[...]

Issue #676, opened 2017-12-06
Document new StartOperation(Activity) API


lmolkova:
See #644 

Documentation to be updated:
- https://docs.microsoft.com/en-us/azure/application-insights/app-insights-api-custom-events-metrics - Operation Context section 
- https://docs.microsoft.com/en-us/azure/application-insights/application-insights-custom-operations-tracking

Documentation should explain sync/async aspects of usage.

PR #677, opened 2017-12-06
Implement new StartOperation(activity) overload

lmolkova:
Fix Issue #644 .
Introduces `telemetryClient.StartOperation(Activity)` overload.

It's expected to be used to track custom operations such as Queue messages processing
```C#
var message = queue.Receive();
var activity = message.ExtractActivity();

using (telemetryClient.StartOperation(activity))
{
  //process
}
```

Documentation update task - https://github.com/Microsoft/ApplicationInsights-dotnet/issues/676


src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -176,5 +176,63 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
 
             operation.Dispose();
         }
+
+        /// <summary>
+        /// Creates an operation object with a respective telemetry item.
```
SergeyKanzhelev: Please expand - explain whether Activity expected to be already started, which properties should already be initialized, etc.

Test/Microsoft.ApplicationInsights.Test/Standalone/AppInsightsStandaloneTests.cs
```diff
@@ -87,7 +86,7 @@ static void Main(string[] args)
                         var tc = new TelemetryClient();
                         using (var requestOperation = tc.StartOperation<RequestTelemetry>(""request"", args[0]))
                         {
-                            using (var dependencyOperation = tc.StartOperation<DependencyTelemetry>(""dependency""))
+                            using (var dependencyOperation = tc.StartOperation<DependencyTelemetry>(""dependency"", args[0]))
```
SergeyKanzhelev: can you please test both overloads.

Actually - one of the most important things for me is that `TrackTrace` will not crash due to `OperationCorrelationTelemetryInitializer` initialization failure without DS present

lmolkova: The thing is, that StartOperation<DependencyTelemetry>("dependency") does not compile any longer without DS being referenced.

For .NET project, as AppInsights SDK is installed/updated, DiagnosticSource automatically by nuget.

The only problem is non .NET (non-nuget) scripts that could still use AI. As I mention here: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/644#issuecomment-349754193
it's extremely unlikely that such scripts use StartOperation at all, as it's a generic method and could be called from proweshell only with reflection.

As for the StartOperation overloads: all of them except Activity-one are internally calling the same StartOperation(T telemetry), so they all use Activity underneath (when available).

I added OperationCorrelationTelemetryInitializer into the configuration and call to TrackTrace.

lmolkova:
@cijothomas 

Could you please help me understand build failure? It failed in Codecov.ps1 with errors like 
```2017-12-07T19:31:03.0296713Z An System.IO.DirectoryNotFoundException occured: Could not find a part of the path 'E:\A\_work\7\s\src\Core\Managed\shared\OperationTelemetryExtensions.cs'. 
2017-12-07T19:31:03.0296713Z An System.IO.DirectoryNotFoundException occured: Could not find a part of the path 'E:\A\_work\7\s\src\Core\Managed\shared\TelemetryClient.cs'. 
```
I'm almost sure it has nothing to do with my changes.

@Dmitry-Matveev @MS-TimothyMothra @cijothomas 
please review

cijothomas:
@lmolkova yes the unit tests actually passed on regualr run, but failed when ran with CodCov turned on. I have kicked off a re-run to see if this was transient issue.
(Surely this does not prevent the PR from merging)

lmolkova:
@Dmitry-Matveev @MS-TimothyMothra 
please review

lmolkova:
@cijothomas could you please merge it?

Issue #666, opened 2017-12-04
Add default SDK payload properties to the heartbeat feature


d3r3kk:
As stated in the original issue (#628) we would be providing specific default fields to the heartbeat feature that customers can turn off as desired. This issue is to track the addition of these default properties and allow for discussion of specific details as necessary.

MS-TimothyMothra:
Hi @d3r3kk,
I see that there is a related PR that's already been merged.

Is there any more work to do for this issue, or is it safe to close this? :)

d3r3kk:
The work related to this issue is complete for this repo (work continues in -dotnet-server repo, but other issues are tracking that).

PR #636, opened 2017-10-13
Initial PR for HealthHeartbeat

d3r3kk:
This is the PR for adding the HealthHeartbeat functionality to the AppInsights SDK. In this revision, I've added the HealthHeartbeat as a child to the DiagnosticsTelemetryModule. I believe this may be better suited in its own module though, outside of the ApplicationInsights base SDK - looking for feedback here.

Overall ideas, comments on structure, questions regarding functionality, or pointing out missed opportunities are all welcome. I look forward to any reviews folks can offer.

- [x] Changes in public surface reviewed
- [x] Design discussion issue #628  
- [x] CHANGELOG.md updated

PR #681, opened 2017-12-08
Default heartbeat payload complete

d3r3kk:
Fix Issue #666 

This is for adding the remainder of the proposed default payload fields for the heartbeat feature.

- [x] Design discussion issue #666 & #636


d3r3kk:
Note that the size of the payload can now be discussed, and we can sort out the best default to use for heartbeat delay.

Here is the JSON payload created by the SDK on my Linux VM in Azure (minified):
`{"name":"Microsoft.ApplicationInsights.Metric","time":"2017-12-07T18:05:06.5941919Z","seq":"0","tags":{"ai.operation.syntheticSource":"HeartbeatState","ai.internal.sdkVersion":"dotnet:2.5.0-45161"},"data":{"baseType":"MetricData","baseData":{"ver":2,"metrics":[{"name":"HeartbeatState","kind":"Aggregation","value":0,"count":1}],"properties":{"osType":"Linux","baseSdkTargetFramework":"netstandard1.3","osVersion":"Linux 4.13.0-17-generic #20-Ubuntu SMP Mon Nov 6 10:04:08 UTC 2017","runtimeFramework":".NET Core 4.6.0.0","name":"dek-linux-2","sku":"17.10","vmSize":"Standard_E8s_v3","vmId":"3596acb8-43ef-4761-adfd-f90db3dc9dad","location":"centralus","offer":"UbuntuServer","platformUpdateDomain":"0","platformFaultDomain":"0","publisher":"Canonical","version":"17.10.201711220"}}}}`

This is 785 bytes long, minified. 

Let's take this as the minimum default size (it's not) and make the maximum 785 x 2=1570 bytes. Using those as the range, here's the storage cost of using different heartbeat delay values for a service running 24/7 in a month of 30 days:

Payload Size (JSON, bytes) | Heartbeat Rate (min) | % Daily Usage @ 1Gb/Month |
------------------------------|------------------------|----------------------------------|
785 | 0.5 | 6.32 |
785 | 5.0 | 0.63 |
785 | 10.0 | 0.32 |
785 | 15.0 | 0.21
1570 | 0.5 | 12.63 |
1570 | 5.0 | 1.26 |
1570 | 10.0 | 0.63 |
1570 | 15.0 | 0.42 |

Given this information, I think to leave the default at 15 minutes is fine, but it certainly could be increased as far as every 5 minutes and still have a fairly negligible effect. Thoughts?

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -79,16 +219,59 @@ private static string GetBaseSdkTargetFramework()
...
+        }
+
+        /// <summary>
+        /// Runtime information for the underlying OS, should include Linux information here as well.
+        /// </summary>
+        /// <returns>String representing the OS, and any version/patch information reported by that OS</returns>
+        private static string GetRuntimeOsType()
+        {
+#if NET45 || NET46
+            return Environment.OSVersion.VersionString;
```
cijothomas: From the method name GetRuntimeOsType() - i expect it to return "windows", 'Linux", "osx" only.

But this seem to be getting version.
we need to double check whether these gives right and useful information.
For eg:, Environment.OSVersion.VersionString gives me "Microsoft Windows NT 6.2.9200.0" and System.Runtime.InteropServices.RuntimeInformation.OSDescription gives me "Microsoft Windows 10.0.16299 " on the same app/machine.

src/Microsoft.ApplicationInsights/Microsoft.ApplicationInsights.csproj
```diff
@@ -75,5 +79,11 @@
     </AdditionalFiles>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net46'">
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json">
```
cijothomas: Is Microsoft.Extensions.Configuration.Json needed only for netstand 1.3 and net46? what about net45? is it missing by accident?

cijothomas: if its required for all targets, then simply remove Conditions

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -3,19 +3,58 @@
...
-            "baseSdkTargetFramework"
+            "baseSdkTargetFramework",
+            "osVersion"
+        };
+
+        public static readonly string[] DefaultOptionalFields =
+         {
+            "processId",
+            // the  following are Azure Instance Metadata fields
+            "osType",
```
cijothomas: for non-AzureVM also we need osType.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -42,6 +100,84 @@ public static void PopulateDefaultPayload(IEnumerable<string> disabledFields, IH
...
+                {
+                    var allFields = await GetAzureInstanceMetadataFields(baseImdsUrl, imdsApiVersion, imdsTextFormat)
+                                    .ConfigureAwait(false);
+
+                    var enabledImdsFields = enabledFields.Intersect(allFields);
+                    foreach (string field in enabledImdsFields)
+                    {
+                        heartbeatManager.AddHeartbeatProperty(
+                            propertyName: field,
+                            propertyValue: await GetAzureInstanceMetadaValue(baseImdsUrl, field, imdsApiVersion, imdsTextFormat)
```
cijothomas: if this AzureMetaDataService has a temp. issue, we need to retry. (500 code) It might be better to handle exceptions/etc inside this helper method itself rather than an overall try catch.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -3,19 +3,58 @@
...
-        {
+        public static readonly string[] DefaultFields =
+         {
             "runtimeFramework",
-            "baseSdkTargetFramework"
+            "baseSdkTargetFramework",
+            "osVersion"
+        };
+
+        public static readonly string[] DefaultOptionalFields =
```
cijothomas: We need unittests. Maybe make baseImdsUrl as overridable from UnitTest to mock AzureMetaData, and make sure this class does not crash/handle all responses from the metadata service.
You can use in-process server to mock azure metadata. An example is here:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/NetCore.Tests/DependencyTrackingTelemetryModuleTestNetCore.cs#L190

d3r3kk: Agreed, I can spend some further time against making some proper unit tests. Thanks for the link to an existing example Cijo!

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
d3r3kk: Yes, I have that covered in the osVersion field. Further discussion of that below.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -3,19 +3,58 @@
...
         };
 
-        public static void PopulateDefaultPayload(IEnumerable<string> disabledFields, IHeartbeatProvider provider)
+        public static readonly string[] AllDefaultFields = DefaultFields.Union(DefaultOptionalFields).ToArray();
+
+        /// <summary>
+        /// Azure Instance Metadata Service exists on a single non-routable IP on machines configured
+        /// by the Azure Resource Manager. See <a href="https://go.microsoft.com/fwlink/?linkid=864683">to learn more.</a>
+        /// </summary>
+        private static string baseImdsUrl = $"http://169.254.169.254/metadata/instance/compute";
```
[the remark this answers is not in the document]
d3r3kk: Yes, that is a definite requirement here (and appears to be one of the reasons my build is failing perhaps?). Thanks again for the link to an example! I think I simply have to ensure my calls are wrapped by the internal SDK call enter- and exit- here.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -25,13 +64,32 @@ public static void PopulateDefaultPayload(IEnumerable<string> disabledFields, IH
                     switch (fieldName)
                     {
                         case "runtimeFramework":
-                            provider.AddHeartbeatProperty(fieldName, GetRuntimeFrameworkVer(), true);
+                            provider.AddHeartbeatProperty(fieldName, true, HeartbeatDefaultPayload.GetRuntimeFrameworkVer(), true);
```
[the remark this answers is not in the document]
d3r3kk: I will ensure there is a facility to capture any exceptions here, they will be logged but non-fatal.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
d3r3kk: Ok. Should we do a retry on this though since the Azure instance service should be stable and available while we run our app. If it isn't, it could be that it would never be stable and would be a waste of our resources to continue to try... I was going for a fire-once-and-forget pattern here, failures or no.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -42,6 +100,84 @@ public static void PopulateDefaultPayload(IEnumerable<string> disabledFields, IH
...
+                            propertyName: field,
+                            propertyValue: await GetAzureInstanceMetadaValue(baseImdsUrl, field, imdsApiVersion, imdsTextFormat)
+                                .ConfigureAwait(false),
+                            isHealthy: true,
+                            allowDefaultFields: true);
+                    }
+                }
+                catch (AggregateException)
+                {
+                    CoreEventSource.Log.CannotObtainAzureInstanceMetadata();
```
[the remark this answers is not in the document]
d3r3kk: I can add some of the detail from the aggregate exception perhaps, but really the expectation would be that if it fails we are not on an Azure VM (or not an Azure VM that was procured through ARM at the least).

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -79,16 +219,59 @@ private static string GetBaseSdkTargetFramework()
...
             Assembly assembly = typeof(Object).GetTypeInfo().Assembly;
 
             AssemblyFileVersionAttribute objectAssemblyFileVer =
                         assembly.GetCustomAttributes(typeof(AssemblyFileVersionAttribute))
                                 .Cast<AssemblyFileVersionAttribute>()
                                 .FirstOrDefault();
-
             return objectAssemblyFileVer != null ? objectAssemblyFileVer.Version : "undefined";
+#elif NETSTANDARD1_3
+            return System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription;
```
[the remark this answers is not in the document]
d3r3kk: Let's chat about this further offline, I've searched through .NET samples online and in the dotnet cli itself and I cannot see how we get a definitive "This is .NET Core 2.0" out of the system.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
d3r3kk: Yes, I believe it has to do with the deprecated nature of the Environment.OSVersion object [see here](https://docs.microsoft.com/en-us/dotnet/api/system.environment.osversion?view=netframework-4.7#System_Environment_OSVersion), and in [this document it backs it up by saying](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724832(v=vs.85).aspx):

> ...Applications not manifested for Windows 8.1 or Windows 10 will return the Windows 8 OS version value (6.2). To manifest your applications for Windows 8.1 or Windows 10, refer to Targeting your application for Windows.

Therefore, I think the right way forward is to make use of the System.Runtime.InteropServices.RuntimeInforamtion.OSDescription. Unfortunately, that doesn't exist until .NET Framework 4.7.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -79,16 +219,59 @@ private static string GetBaseSdkTargetFramework()
...
+        /// </summary>
+        /// <returns>String representing the OS, and any version/patch information reported by that OS</returns>
+        private static string GetRuntimeOsType()
+        {
+#if NET45 || NET46
+            return Environment.OSVersion.VersionString;
+#elif NETSTANDARD1_3
+            return System.Runtime.InteropServices.RuntimeInformation.OSDescription;
+#else
+#error Unrecognized framework
```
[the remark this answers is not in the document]
d3r3kk: The error is indicating the framework is unidentified and will happen at build time, it is not an indication that the runtime is unrecognized which would happen only during execution.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -79,16 +219,59 @@ private static string GetBaseSdkTargetFramework()
...
+        }
+
+        /// <summary>
+        /// Runtime information for the process ID currently running and consuming the SDK
+        /// </summary>
+        /// <returns>string ID of the current process' AppDomain</returns>
+        private static string GetProcessId()
+        {
+#if NET45 || NET46
+            return System.AppDomain.CurrentDomain.FriendlyName;
```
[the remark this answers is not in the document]
d3r3kk: Another good topic for you and I to chat about a bit further. I had a heck of a time finding a way to get the name of the executable (perhaps that's not what we want here anyways?) across all the .NET versions and frameworks available to Application Insights SDK. I'm suuuuper open to suggestions here.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -79,16 +219,59 @@ private static string GetBaseSdkTargetFramework()
...
+#elif NETSTANDARD1_3
+            return null;
```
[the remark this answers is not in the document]
d3r3kk: Because I cannot find something suitable for this value in the .NET Standard < 1.5! Help!

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -79,16 +219,59 @@ private static string GetBaseSdkTargetFramework()
...
+#else
+#error Unrecognized framework
```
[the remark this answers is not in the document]
d3r3kk: Again, the error here is at compile time indicating the lack of a framework to build with.

src/Microsoft.ApplicationInsights/Microsoft.ApplicationInsights.csproj
d3r3kk: I believe this was me wandering off on a tangent earlier on in my change. I will clean it up (we don't need this reference at all).

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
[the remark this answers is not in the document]
SergeyKanzhelev: Call `SdkInternalOperation.Start/Stop` around this call to address this feedback

SergeyKanzhelev: I haven't looked at it yet. Just want to make sure you know that the same IP is used in AWS. So you'd need to expect that something will be returned to you that may be incompatible with what you'd expect from Azure

src/Microsoft.ApplicationInsights/Microsoft.ApplicationInsights.csproj
```diff
@@ -45,7 +45,8 @@
   <ItemGroup>
     <PackageReference Include="Desktop.Analyzers" Version="1.1.0">
       <PrivateAssets>All</PrivateAssets>
-    </PackageReference>    
+    </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="1.1.2" />
```
SergeyKanzhelev: Please no external references. Anything that is not a framework (or DiagnosticsSource which is an exception) not allowed. If you want this reference very badly - I'd suggest moving this logic into `WindowsServer` package that already have Azure-specific telemetry initializers. And it will be logical to have this there as well.

I'd say in general it's a good idea to put it into `WindowsServer` package.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -42,6 +100,84 @@ public static void PopulateDefaultPayload(IEnumerable<string> disabledFields, IH
...
+        /// Gets the value of a specific field from the imds link asynchronously, and returns it.
+        /// </summary>
+        /// <param name="baseUrl">Url of the Azure Instance Metadata service</param>
+        /// <param name="fieldName">Specific imds field to retrieve a value for</param>
+        /// <param name="apiVersion">rest Api version to append to the constructed Uri</param>
+        /// <param name="textFormatArg">query-string argument to request text data be returned</param>
+        /// <returns>an array of field names available, or null</returns>
+        private static async Task<string> GetAzureInstanceMetadaValue(string baseUrl, string fieldName, string apiVersion, string textFormatArg)
+        { 
+            string computFieldUrl = $"{baseUrl}/{fieldName}?{textFormatArg}&{apiVersion}";
```
SergeyKanzhelev: I thought you can get a single JSON that has both - names and values. Do you really need to run an extra http request for every name?

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -42,6 +100,84 @@ public static void PopulateDefaultPayload(IEnumerable<string> disabledFields, IH
             }
         }
 
+        /// <summary>
+        /// Gathers operational data from an Azure VM, if the IMDS endpoint is present. If not, and the SDK is
+        /// not running on an Azure VM with the Azure Instance Metadata Service running, the corresponding
+        /// fields will not get set into the heartbeat payload.
+        /// </summary>
+        private static async Task AddAzureVmDetail(HeartbeatProvider heartbeatManager, IEnumerable<string> enabledFields)
```
SergeyKanzhelev: There should be a configuration switch to disable any attempts to run this discovery

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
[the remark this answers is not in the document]
d3r3kk: They are try...catch'ed - if you look outside my switch statement. Since it's not fatal for a heartbeat property to fail, why wouldn't we just catch the exception and move on? 
I think there is an argument here to have specific try...catch blocks per-property maybe, what do you think? My assumption when I first wrote this is that each property was (a) going to be fairly trivial to get, and (b) was going to be non-fatal should something go wrong.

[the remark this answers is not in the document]
cijothomas: test that if there is an issue collecting one of the item, it should not affect the rest of the items. It looks to me like any issue with any item will cause the rest of the item to also be not collected.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
d3r3kk: Wow, interesting! Both [Amazon AWS VMs](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html), as well as [Google cloud VMs](https://cloud.google.com/compute/docs/storing-retrieving-metadata), have it there as well it would appear. Will add a check to ensure an Azure field is present.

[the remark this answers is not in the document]
cijothomas: @SergeyKanzhelev ```SdkInternalOperation.Start/Stop``` is only to prevent FirstChanceExceptionmodule from reporting ex right? I was referring to the DependencyTelemetry item collected by DepCollector module. The change would be in web repository in the file linked above.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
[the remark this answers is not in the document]
cijothomas: As discussed offline, this looks correct. Gh did not show the try..catch {}. Sorry for confusion!.
Btw - if there is no unit test covering ErrorOnOnePropertyDoesNotAffectOtherProperty, then please add one.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
d3r3kk: Looking into this further this morning I see that the web requests all require special setup and URI resource paths to get any data out of the service. I doubt there is any risk of getting incompatible data here aside from someone mocking the service up on their own. I don't think there is much value protecting ourselves here (aside from looking only for 200 responses/allowing for retries).

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
d3r3kk: Agreed, done.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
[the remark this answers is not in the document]
d3r3kk: As discussed offline, let's raise another issue against this specific problem and provide a solution outside the SDK that might help us out - if we want to get super detailed on what version of the framework we're running in I doubt that code belongs within the SDK itself (perhaps a telemetry initializer module might be better here).

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
d3r3kk: Instead of going to any of this trouble, I've simplified this to simply attempt to return OS type, and completely remove any notion of OS version (this is a complicated beast in .NET/Windows it seems).

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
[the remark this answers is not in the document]
d3r3kk: As discussed, this will be a 'processSessionId' that we can use to identify when a process has been restarted.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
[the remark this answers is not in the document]
d3r3kk: Sorted - this is now a unique session ID that we create internally (guid).

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -2,46 +2,251 @@
...
+                            requestResult = content.ReadToEnd();
+                        }
+                    }
+                }
+
+#else
+#error Unknown framework
+#endif
+            }
+            catch (AggregateException ex)
```
zakimaksyutov: What if we get different exception?

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -2,46 +2,251 @@
...
+
+                var allFields = await GetAzureInstanceMetadataFields(baseImdsUrl, imdsApiVersion, imdsTextFormat)
+                                .ConfigureAwait(false);
+
+                var enabledImdsFields = enabledFields.Intersect(allFields);
+                foreach (string field in enabledImdsFields)
+                {
+                    heartbeatManager.AddHeartbeatProperty(
+                        propertyName: field,
+                        propertyValue: await GetAzureInstanceMetadaValue(baseImdsUrl, field, imdsApiVersion, imdsTextFormat)
```
zakimaksyutov: Can you check whether some fields can get changed without restarting a VM?

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -2,46 +2,251 @@
...
+        /// fields will not get set into the heartbeat payload.
+        /// </summary>
+        private static async Task AddAzureVmDetail(HeartbeatProvider heartbeatManager, IEnumerable<string> enabledFields)
+        {
+            if (HeartbeatDefaultPayload.EnableAzureInstanceMetadata && !HeartbeatDefaultPayload.isAzureMetadataCheckCompleted)
+            {
+                // only ever do this once when the SDK gets initialized
+                HeartbeatDefaultPayload.isAzureMetadataCheckCompleted = true;
+
+                var allFields = await GetAzureInstanceMetadataFields(baseImdsUrl, imdsApiVersion, imdsTextFormat)
```
zakimaksyutov: Can we determine that we're running on Azure infra before making this call? Concern - we will be doing this call for On-Premise VMs and other APM tools will be flagging this call.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -2,46 +2,251 @@
...
+        /// <summary>
+        /// Gets or sets a value indicating whether the collection of instance metadata from Azure VMs is enabled or not.
+        /// </summary>
+        public static bool EnableAzureInstanceMetadata { get => enableAzureInstanceMetadataInHeartbeat; set => enableAzureInstanceMetadataInHeartbeat = value; }
+
+        public static void PopulateDefaultPayload(IEnumerable<string> disabledFields, HeartbeatProvider provider)
+        {
+            var enabledProperties = HeartbeatDefaultPayload.RemoveDisabledDefaultFields(disabledFields);
+
+            Task.Factory.StartNew(async () => await HeartbeatDefaultPayload.AddAzureVmDetail(provider, enabledProperties).ConfigureAwait(false));
```
zakimaksyutov: Let's add comment why we're doing it in async fashion

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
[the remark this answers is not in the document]
SergeyKanzhelev: It is already implemented https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/DependencyCollector/Shared/Implementation/ApplicationInsightsUrlFilter.cs#L59

SergeyKanzhelev:
what's the status of this PR? Time to review again? Please ping when you need review

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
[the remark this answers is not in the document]
d3r3kk: I am adding the exception detail here now, and the inner exception detail as well if there is any (that will cover 99% of cases that we will capture expectedly here)

---
In reply to: [155880742](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/681#discussion_r155880742) [](ancestors = 155880742)

src/Microsoft.ApplicationInsights/Microsoft.ApplicationInsights.csproj
d3r3kk: Agreed - this was a slip-up and was never intended to be part of my changes (was added during some preliminary experimentation only). Removed #Closed

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
d3r3kk: I am not convinced it's worthwhile to write a JSON parser here, and I doubt we want to take a dependency on something like Newtonsoft either. I chose this route as it removed the necessity for both options altogether. While I believe the JSON would likely be trivial to parse, I cannot guarantee the absence of edge cases that would cause us (and our customers, especially in regions where special characters are common in text strings) grief.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
d3r3kk: Changed to capture all exceptions and log them. This call can fail without any major consequence. #Closed

d3r3kk:
@SergeyKanzhelev & @cijothomas 

Finished the (hopefully) last round of edits for this PR. 
- Enhanced the architecture of how we obtain the 'default' payload properties to make them testable
- Added & corrected unit tests to improve coverage, added coverage to the Azure default properties
- NOTE: No _functional_ tests are included in this PR, and would be out of scope here. If we feel we need functional tests (ie. tests for actually obtaining values from Azure Metadata Service) I would suggest we raise an issue on the dotnet-server repo and perform them there once this code is integrated.

SergeyKanzhelev:
@d3r3kk - I'd again suggest to move this logic to `WindowsServer` package. Please see justification earlier. Core package provides infrastructure for HeartBeat and very basic metadata. All the platform specific metadata should be obtained using corresponding packages.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModule.cs
```diff
@@ -80,6 +80,17 @@ public TimeSpan HeartbeatInterval
             set => this.HeartbeatProvider.HeartbeatInterval = value;
         }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether or not to query for and include Azure Instance Metadata in the
+        /// heartbeat properties. For more information on the Azure Instance Metadata service in Azure see
+        /// <a href="https://go.microsoft.com/fwlink/?linkid=864683">this document to learn more.</a>
+        /// </summary>
+        public bool EnableInstanceMetadata
```
cijothomas: EnableAzureInstanceMetadata as name to make it obvious.

minor: the summary comment - remove the 'to learn more' at end as its redundant.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/AzureMetadataRequestor.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            SdkInternalOperationsMonitor.Enter();
+            try
+            {
+#if NETSTANDARD1_3
+
+                using (var getFieldValueClient = new HttpClient())
+                {
+                    getFieldValueClient.DefaultRequestHeaders.Add("Metadata", "True");
+                    requestResult = await getFieldValueClient.GetStringAsync(metadataRequestUrl).ConfigureAwait(false);
```
cijothomas: dont we need to ensure http response code is OK before reading response? (we had a bug in X-Component lookup similar to this, where we didn't check for OK before reading respone)

d3r3kk: I believe in this case we can disregard this, as the entire module that reads IMS data is done in a Task that is completely ignored in terms of return value. If the response is not OK, and there is some problem raised here (which would throw an exception perhaps) then the Task terminates and the data for IMS is not injected into the heartbeat payload. At most, we would want to get a log line here I suspect.

d3r3kk:
Rebased and ready for review again.

- Removed default heartbeat payload provider for Azure Instance Metadata (moving to dotnet-server)
- Re-architected how we provide 'default' payload items
- Removed/altered tests to cope with the change

d3r3kk:
(Closing & reopening the PR to spark of builds - transient test failures are getting in my way)

d3r3kk:
This PR coupled with [PR 795](https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/795) satisfies the original requirements in #636.

d3r3kk:
Ready for review @SergeyKanzhelev and @cijothomas

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseHeartbeatProperties.cs
```diff
@@ -0,0 +1,174 @@
...
+
+            return Task.FromResult(hasSetValues);
+        }
+
+        /// <summary>
+        /// This will return the current running .NET framework version, based on the version of the assembly that owns
+        /// the 'Object' type. The version number returned can be used to infer other things such as .NET Core / Standard.
+        /// </summary>
+        /// <returns>a string representing the version of the current .NET framework</returns>
+        private string GetRuntimeFrameworkVer()
```
cijothomas: I forgot what was the conclusion of our prior discussions on this topic, apologies if this is repeat!.
in my asp.net core app, targetting netcoreapp2.0, System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription returns the following
_".NET Core 4.6.26020.03"_

If there a way to return NETCORE2.0 ?

Also imp: 
Please provide example values for diff cases as a comment here.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseHeartbeatProperties.cs
```diff
@@ -0,0 +1,174 @@
...
+#error Unrecognized framework
+            return "undefined";
+#endif
+        }
+
+        /// <summary>
+        /// Runtime information for the underlying OS, should include Linux information here as well.
+        /// </summary>
+        /// <returns>String representing the OS or 'unknown'</returns>
+        private string GetRuntimeOsType()
```
cijothomas: For net45, Environment.OSVersion.Platform.ToString() returns Win32NT
For netstand1_3, values would be WINDOWS, LINUX etc,
Any way to get consistent values in net45 and netstandard targets.

Can you provide example values in the comments here as well?

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/IHeartbeatProvider.cs
```diff
@@ -4,10 +4,22 @@
     using System.Collections.Generic;
     using Microsoft.ApplicationInsights.Extensibility;
 
-    internal interface IHeartbeatProvider : IHeartbeatPropertyManager, IDisposable
+    internal interface IHeartbeatProvider : IDisposable
```
cijothomas: Bit confused with so many interfaces.. IHeartbeatProvider has Add/Set property methods.. But its also duplicated in IHeartbeatPropertyManager. I think we should not have this duplication unless there is a strong reason to it.

cijothomas: Please add comments about all the interfaces - so it'll become clearer the purpose of each of them.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseHeartbeatProperties.cs
d3r3kk: Let's take this field out altogether until we find a concrete method that we can map from runtime version to NETCORE / NETSTANDARD / NETFRAMEWORK versions.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseHeartbeatProperties.cs
d3r3kk: I'll have a look and see if I can tighten this up a bit.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/IHeartbeatProvider.cs
d3r3kk: Will make some comments.

d3r3kk: Done. Also, I was able to remove some of the redundant duplication.

d3r3kk:
@cijothomas I completed the PR suggestions you and I walked through yesterday. Let me know if we need any further fixes/updates. Would love for the base to be in for 2.6-beta1

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
```diff
@@ -3,92 +3,45 @@
...
-            if (disabledFields == null || disabledFields.Count() <= 0)
+            foreach (var payloadProvider in DefaultPayloadProviders)
             {
-                enabledProperties = DefaultFields.ToList();
-            }
-            else
-            {
-                enabledProperties = new List<string>();
-                foreach (string fieldName in DefaultFields)
+                if (disabledProviders != null && disabledProviders.Contains(payloadProvider.Name, StringComparer.OrdinalIgnoreCase))
```
cijothomas: I don't see unit tests testing this feature - disabledProviders is something newly added right? Need to unit test that it is respected by HeartbeatDefaultPayload.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/IHeartbeatDefaultPayloadProvider.cs
```diff
@@ -0,0 +1,32 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    internal interface IHeartbeatDefaultPayloadProvider
```
cijothomas: Please comment explaining what this interface is.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseHeartbeatProperties.cs
```diff
@@ -0,0 +1,181 @@
...
+                try
+                {
+                    switch (fieldName)
+                    {
+                        // case "runtimeFramework":
+                        //    provider.AddHeartbeatProperty(fieldName, true, this.GetRuntimeFrameworkVer(), true);
+                        //    hasSetValues = true;
+                        //    break;
+                        case "baseSdkTargetFramework":
+                            provider.AddHeartbeatProperty(fieldName, true, this.GetBaseSdkTargetFramework(), true);
```
cijothomas: I think the AddHeartbeatProperty with 4 args is the only one used from IHeartbeatPropertyManager. It still contains AddHeartbeatProperty with 3 args.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/IHeartbeatProvider.cs
```diff
@@ -1,13 +1,20 @@
...
+    /// </summary>
     internal interface IHeartbeatProvider : IHeartbeatPropertyManager, IDisposable
     {
         string InstrumentationKey { get; set; }
 
         void Initialize(TelemetryConfiguration configuration);
+
+        bool AddHeartbeatProperty(string propertyName, bool overrideDefaultField, string propertyValue, bool isHealthy);
+
+        bool SetHeartbeatProperty(string propertyName, bool overrideDefaultField, string propertyValue = null, bool? isHealthy = null);
```
cijothomas: mentioned somewhere else too - if this is the only one used, why we have 3 arg version of this same method

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseHeartbeatProperties.cs
```diff
@@ -0,0 +1,181 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Reflection;
+    using System.Text;
+    using System.Threading.Tasks;
+#if NETSTANDARD1_3
+    using System.Runtime.InteropServices;
+#endif
+
+    internal class BaseHeartbeatProperties : IHeartbeatDefaultPayloadProvider
```
cijothomas: name of class not really indicative of the interface it implements.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseHeartbeatProperties.cs
```diff
@@ -0,0 +1,181 @@
...
+        // but until we understand how to make the mapping from the runtime version to the standard versions
+        // this information is not entirely useful. Leaving it commented out here for now until such a mapping
+        // can be made in an appropriate manner, or until we can provide such a mapping.
+        //
+        // <summary>
+        // This will return the current running .NET framework version, based on the version of the assembly that owns
+        // the 'Object' type. The version number returned can be used to infer other things such as .NET Core / Standard.
+        // </summary>
+        // <returns>a string representing the version of the current .NET framework</returns>
+        // private string GetRuntimeFrameworkVer()
```
cijothomas: I just realized that this code is already shipping in 2.5.0 stable. We should try to find this mapping ourselves, and post it in Github or some place for quick reference. @Dmitry-Matveev also to comment on best way to support already shipped (2.5) sdk.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModule.cs
```diff
@@ -80,13 +80,21 @@ public TimeSpan HeartbeatInterval
             set => this.HeartbeatProvider.HeartbeatInterval = value;
         }
 
+        /// <summary>
+        /// Gets a list of default heartbeat property providers that are disabled and will not contribute to the
+        /// default heartbeat properties.
+        /// </summary>
+        public IList<string> ExcludedHeartbeatPropertyProviders
```
cijothomas: Since we already provide ExcludedHeartbeatProperties for customers to exclude any field they dont want. What additional purpose does ExcludedHeartbeatPropertyProviders serve? Can you give an example of HeartbeatPropertyProviders ?

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseHeartbeatProperties.cs
Dmitry-Matveev: Having a reference list of the known combinations of base .NET version and runtime version (in the way we retrieve it in 2.5) should make this feature more useful for our customers as well as for internal investigations. We may want to create such a list for 2.5.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModule.cs
d3r3kk: Yes, this is a departure from what I originally created and there is a bit of background on why I made this choice:

Before, when the `static HeartbeatDefaultPayload` class was  created, all the default properties where immediately available (or not) and there was no problems testing the associated code flows. When I added the Azure IMS default properties to the base SDK, things changed in that I had to spawn a thread and gather properties asynchronously (to cope with http requests/timeouts). This made it very difficult to test the `HeartbeatDefaultPayload`/`HeartbeatProvider.Initialize` flow as waiting for the async call to return was not exposed to anything but two methods within the `HeartbeatDefaultPayload` class.

Once I re-architected how we gather ‘default’ heartbeat properties in an async-friendly way, *in a testable* way, I saw that it was going to be a lot of work for customers to exclude every single field from any particular provider. I decided then to allow per-field exclusions *plus* per-module exclusions. Of course, this isn’t entirely necessary at the  moment since I moved the Azure IMS module over to -dotnet-server, but I figured the extra architecture doesn’t detract from anything and it allows us to easily add new default heartbeat modules in the future.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseHeartbeatProperties.cs
d3r3kk: Changed!

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseHeartbeatProperties.cs
d3r3kk: Internally, yes, this is the only one used (as in the base SDK we allow the overriding of default properties). Externally, we only allow adding properties that do not collide with the ‘default’ property names. See my PR in -dotnet-server for an example of the use of the `AddHeartbeatProperty` with 3 args.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseHeartbeatProperties.cs
d3r3kk: I feel that this discussion needs to be sorted out separately from this PR (it's related, but tangential to the scope of this discussion somewhat). I've raised a new issue (#699) where we can follow up and deal with this in a specific place. 

I've re-enabled the code and left it as it was in 2.5.0-stable for now.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatDefaultPayload.cs
d3r3kk: You are correct, I’d skipped making these tests after creating the feature! I’ve created a few extra tests.
Thanks for catching this Cijo!

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatProvider.cs
```diff
@@ -6,14 +6,15 @@
...
+    using System.Threading.Tasks;
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.DataContracts;
     using Microsoft.ApplicationInsights.Extensibility;
 
     /// <summary>
     /// Implementation of heartbeat functionality.
     /// </summary>
-    internal class HeartbeatProvider : IDisposable, IHeartbeatProvider
+    internal class HeartbeatProvider : IDisposable, IHeartbeatProvider, IHeartbeatPropertyManager
```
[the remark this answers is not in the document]
d3r3kk: As I’d had them separate before, this class needed to implement both. I’ve corrected this oversight now.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/IHeartbeatDefaultPayloadProvider.cs
d3r3kk: Done

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/IHeartbeatProvider.cs
```diff
@@ -1,13 +1,20 @@
...
+    /// Internal interface for providing heartbeat information into the SDK pipeline. Useful
+    /// for mocking for unit test purposes.
+    /// </summary>
     internal interface IHeartbeatProvider : IHeartbeatPropertyManager, IDisposable
     {
         string InstrumentationKey { get; set; }
 
         void Initialize(TelemetryConfiguration configuration);
+
+        bool AddHeartbeatProperty(string propertyName, bool overrideDefaultField, string propertyValue, bool isHealthy);
```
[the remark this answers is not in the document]
d3r3kk: The difference is in the accessibility of each interface. Internally, I've exposed a `bool overrideDefaultField` here. This is the internal interface, and as such allows the internal caller to override any default field. Whereas the public Add/Set defined in the `IHeartbeatPropertyManager` does not allow this override functionality.

We want to make it obvious our 'default' heartbeat properties shoud not be overwritten/altered by other people to ensure the information we add there is left intact for the UX to make use of.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/IHeartbeatProvider.cs
d3r3kk: Internal/external access. Please see the above discussion.

d3r3kk:
Thanks @cijothomas and @Dmitry-Matveev for the feedback. I've addressed all concerns, please have a look and let me know if you want any other changes.

Dmitry-Matveev:
Looks OK, @cijothomas , any other comments?

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/Tracing/HeartbeatTests.cs
```diff
@@ -173,16 +257,20 @@ public void HeartbeatPayloadContainsUserSpecifiedData()
         [TestMethod]
         public void HeartbeatPayloadContainsOnlyAllowedDefaultPayloadFields()
         {
+            var assemblyDefFields = new BaseDefaultHeartbeatPropertyProvider();
```
cijothomas: What is this test testing? No asserts found .

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModule.cs
cijothomas: ok. Please give an example name of a provider in the comments.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseDefaultHeartbeatPropertyProvider.cs
```diff
@@ -0,0 +1,176 @@
...
+    using System.Runtime.InteropServices;
+#endif
+
+    internal class BaseDefaultHeartbeatPropertyProvider : IHeartbeatDefaultPayloadProvider
+    {
+        internal readonly List<string> DefaultFields = new List<string>()
+        {
+             "runtimeFramework",
+            "baseSdkTargetFramework",
+            "osType",
```
cijothomas: In the PR in web sdk repo, there is a field called 'osType'. Did you meant it to be same where Azure one overrides this when available?

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/IHeartbeatProvider.cs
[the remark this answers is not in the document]
cijothomas: ok

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/Tracing/HeartbeatTests.cs
d3r3kk: There is an assert on line 285. Testing that explicitly removed or 'disabled' properties are not in the heartbeat.

src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/BaseDefaultHeartbeatPropertyProvider.cs
d3r3kk: A interesting! I had missed this. There is two, one here and one there. The defined behavior then is that this one will be present, and the one from the Web SDK repo will be ignored silently. 

The name 'osType' is something I can control here, but I cannot control it in the Web SDK. Therefore I would have to change this name to something else for the Web SDK to provide (essentially the same) information. 

What I will do is I will revisit this issue when I get to the Web SDK PR, and either choose to rename the osType name there with a specific override, or come to some other solution. Since this is essentially redundant information though, I bet we just leave it as is and ignore the Azure Instance Metadata value.

PR #689, opened 2018-01-04
script to copy dlls for fxCop

MS-TimothyMothra:
script will copy DLLs from bin into folders split by framework for fxcop to scan

### DIFF
--- a/SDL/FxCopSetup.ps1
+++ b/SDL/FxCopSetup.ps1
@@ -21,7 +21,7 @@ Write-Host "`nCreate FxCop Directory...";
 if (Test-Path $fxCopDirectory) { Remove-Item $fxCopDirectory -Recurse; }
 
 # copy all
-Write-Host "`nCopy all files...";
+Write-Host "`nCopy all files (excluding 'Test' directories)...";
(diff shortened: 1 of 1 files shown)


SDL/FxCopSetup.ps1
```diff
@@ -0,0 +1,63 @@
...
+)
+
+# these are dlls that end up in the bin, but do not belong to us and don't need to be scanned.
+$excludedFiles = @();
+
+Write-Host "`nPARAMETERS:";
+Write-Host "`tbuildDirectory:" $buildDirectory;
+Write-Host "`tfxCopDirectory:" $fxCopDirectory;
+
+$frameworks = @("net45", "net46", "netstandard1.3");
```
cijothomas: would it make sense to move the list of target frameworks to a common props file and have it refered from there? then the main .csproj can also refer from there..
Its easy to forget to update these ps scripts when we update list of frameworks.

MS-TimothyMothra: I didn't see your comment until after I merged.
I didn't consider that because I was testing all these scripts locally where i could inspect each directory. 
But yes, i think that would be a good idea.

PR #691, opened 2018-01-11
sdl script revisions

MS-TimothyMothra:
testing excluding certain files from FxCop


MS-TimothyMothra:
@d3r3kk, thanks. I'm still testing...
We're currently scanning files that don't belong to us, but are required for the scan to finish.
I'm testing a few other changes to resolve this.

Issue #708, opened 2018-02-19
Lost many context fields in 2.5.0


stefangordon:
I am having trouble tracking down the expected behavior, but in the latest version various important context fields such as Context.Component.Version are no longer being transmitted to the service and show as "Undefined" in my data.   Upon examining the telemetry object, I see these tags are showing up under the "Sanitized" property.   Is this expected behavior and how do I resolve?  Thanks!

## Repro Steps
```
            telemetry = new TelemetryClient();

            // Telemetry setup
            telemetry.InstrumentationKey = "<key>";
            telemetry.Context.User.Id = Environment.UserName;
            telemetry.Context.Session.Id = SessionId;
            telemetry.Context.Device.OperatingSystem = Environment.OSVersion.ToString();
            telemetry.Context.Device.Id = MachineId;
            telemetry.Context.Component.Version = System.Reflection.Assembly.GetExecutingAssembly().GetName().Version.ToString() 
```

##  Actual Behavior
Version shows up in Azure portal as `[undefined]`

## Expected Behavior
Version should show up correctly....

## Version Info 
SDK Version : 2.5.0
.NET Version : 4.5.2	 						
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) : c#/nuget
OS : Windows 10

stefangordon:
Note, reverting to 2.4.0 resolved my issue.

older:
Looks like the same problem as #706

MS-TimothyMothra:
@lmolkova Can you please review this?
This seems to be related to PR #677

pharring:
We forgot to call `telemetryClient.Initialize()`

MS-TimothyMothra:
Hi @pharring, can you please clarify where?

pharring:
I'm sorry, I deleted my comment because I think it was unrelated and distracted from the issue at hand. 
But, for the record: I was reviewing the aforementioned PR #677 and noted that we didn't call telemetryClient.Initialize after constructing the OperationTelemetry here https://github.com/Microsoft/ApplicationInsights-dotnet/pull/677/files#diff-498e5cc99a0c604df36160a534acf301R236

lmolkova:
@Paul Harrington<mailto:Paul.Harrington@microsoft.com> , It’s not required as telemetry is initialized in the  Track call anyway.  In other StartOperation methods, I presume, it’s needed immediately for some reasons.

[...]

Issue #706, opened 2018-02-19
Dependency correlation is broken after upgrade to .NET SDK 2.5


older:
## Repro Steps
1. Create azure function using example from [Custom telemetry in C# functions](https://docs.microsoft.com/en-us/azure/azure-functions/functions-monitoring#custom-telemetry-in-c-functions) using ApplicationInsights 2.5 NuGet package.
2. Run azure function and check what is logged in application insights
3. Downgrade ApplicationInsights NuGet package to version 2.4
3. Run azure function and check what is logged in application insights

##  Actual Behavior

This is what I get with ApplicationInsights 2.5:
![appinsights25](https://user-images.githubusercontent.com/1029802/36370655-c99f1640-155f-11e8-9389-8310b5e30033.png)

## Expected Behavior

Screenshot using ApplicationInsights 2.4:
![appinsights24](https://user-images.githubusercontent.com/1029802/36370668-dcdd3d04-155f-11e8-975a-091a1522ea7e.png)

## Version Info 
SDK Version : 2.5
.NET Version : 4.6.1	 						
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) : Visual Studio
OS : Windows
Hosting Info (IIS/Azure WebApps/ etc) : Azure

MS-TimothyMothra:
@lmolkova Can you please review this?
This seems to be related to PR #677

lmolkova:
@older 
could you please verify if http dependency calls are collected? From the screenshots it's not clear whether the calls are not tracked or correlation between telemetry is broken.
You may filter only dependency calls in the search blade on the portal, or query `depednencies`  in analytics. [More on dependency tracking](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-asp-net-dependencies)

@MS-TimothyMothra 
#677 should not be a root cause of this issue as it introduces a new API, that is not used (yet) anywhere, but does not alter any old APIs.

older:
Calls are tracked but they are not correlated. Azure functions runtime collects requests, but doesn't automatically collect dependencies so it has to be done manually according to the document I linked. Also application version property is lost (maybe others as well, but I haven't checked).

Anyway, this is fixed with simple NuGet downgrade from 2.5 back to 2.4 so this is either intended breaking change which went undocumented or it is a bug in 2.5.

lmolkova:
@older 
[...]

Issue #692, opened 2018-01-11
Logic bug within Initialize(..,..) in TelemetryContext : null-conditional operator comparison is always null


DanSmith:
Hi,

I have set several default properties on a TelemetryClient's Context and they seem to not be copied over to custom events and exceptions.

I think the issue started when the Initialize method in TelemetryContext was updated here in this commit:

https://github.com/Microsoft/ApplicationInsights-dotnet/commit/e0208cd3bc08ab1baf6ac5ff075513886c942ba4#diff-3d363b608e70a044e45f704453bce8de

If the private context fields of the new TelemetryContext are null, the ? operator will always fail to call the copy from the source. 

This logic was previously reversed prior to the mentioned commit where the call would copy the source if the source wasn't null, which use to work as expected.

## Repro Steps
1. Set client.Context.Component.Version = "some version"
2. Create a new EventTelemetry and call TrackEvent on the client

##  Actual Behavior
Component.Version, ai.application.ver, isn't set, serialized, or reported to the endpoint.

## Expected Behavior
ai.application.ver is included in the tags in the json serialization

## Version Info 
SDK Version : dotnet:2.5.0-55494
.NET Version : 4.5

MS-TimothyMothra:
Today we released Microsoft.ApplicationInsights (v2.5.1) and Microsoft.ApplicationInsights.Web (v2.5.1) with this fix.

PR #710, opened 2018-02-21
fix for Context properties not being copied to telemetry items.

MS-TimothyMothra:
ensure that Context Properties are initialized.
add test to verify that Context Properties are copied to TelemetryItems.

Fix Issue #692.
Fix Issue #706.
Fix Issue #708.

### DIFF
--- a/Test/Microsoft.ApplicationInsights.Test/Shared/DataContracts/TelemetryContextInitializeTest.cs
+++ b/Test/Microsoft.ApplicationInsights.Test/Shared/DataContracts/TelemetryContextInitializeTest.cs
@@ -1,0 +1,152 @@
+﻿namespace Microsoft.ApplicationInsights.DataContracts
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    /// <summary>
+    /// If a user sets the context on a TelemetryClient, it is expected that those values will propagate to individual TelemetryItems.
+    /// </summary>
+    [TestClass]
+    public class TelemetryContextInitializeTest
+    {
+        private TelemetryBuffer TelemetryBuffer;
+        private ITelemetryChannel TelemetryChannel;
+        private TelemetryClient TelemetryClient;
+
+        #region Telemetry Context Properties
+        private const string TestInstrumentationKey = "00000000-0000-0000-0000-000000000000";
+
+        private const string TestComponentVersion = "TestComponentVersion";
+
+        private const string TestDeviceType = "TestDeviceType";
+        private const string TestDeviceId = "TestDeviceId";
+        private const string TestDeviceOperatingSystem = "TestDeviceOperatingSystem";
+        private const string TestDeviceOemName = "TestDeviceOemName";
+        private const string TestDeviceModel = "TestDeviceModel";
+
+        private const string TestCloudRoleName = "TestCloudRoleName";
+        private const string TestCloudRoleInstance = "TestCloudRoleInstance";
+
+        private const string TestSessionId = "TestSessionId";
+        private const bool TestSessionIsFirst = true;
+
+        private const string TestUserId = "TestUserId";
+        private const string TestUserAccountId = "TestUserAccountId";
+        private const string TestUserUserAgent = "TestUserUserAgent";
+        private const string TestUserAuthenticatedUserId = "TestUserAuthenticatedUserId";
+
+        private const string TestOperationId = "TestOperationId";
+        private const string TestOperationParentId = "TestOperationParentId";
+        private const string TestOperationCorrelationVector = "TestOperationCorrelationVector";
+        private const string TestOperationSyntheticSource = "TestOperationSyntheticSource";
+        private const string TestOperationName = "TestOperationName";
+
+        private const string TestLocationIp = "TestLocationIp";
+
+        private const string TestInternalSdkVersion = "TestInternalSdkVersion";
+        private const string TestInternalAgentVersion = "TestInternalAgentVersion";
+        private const string TestInternalNodeName = "TestInternalNodeName";
+        #endregion
+
+        [TestInitialize]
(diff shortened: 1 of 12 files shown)


src/Microsoft.ApplicationInsights/DataContracts/TelemetryContext.cs
```diff
@@ -132,13 +132,13 @@ public LocationContext Location
...
             {
                 var result = new Dictionary<string, string>();
-                this.component?.UpdateTags(result);
-                this.device?.UpdateTags(result);
-                this.cloud?.UpdateTags(result);
-                this.session?.UpdateTags(result);
-                this.user?.UpdateTags(result);
-                this.operation?.UpdateTags(result);
-                this.location?.UpdateTags(result);
+                this.Component?.UpdateTags(result);
```
Dmitry-Matveev: Is there a test like you did for `CopyFrom()` that we can do for `UpdateTags()`?

MS-TimothyMothra:
Should also write a test scenario where Context items on the TelemetryItem are overridden.

Test/Microsoft.ApplicationInsights.Test/Shared/DataContracts/TelemetryContextInitializeTest.cs
```diff
@@ -0,0 +1,152 @@
...
+    public class TelemetryContextInitializeTest
+    {
+        private TelemetryBuffer TelemetryBuffer;
+        private ITelemetryChannel TelemetryChannel;
+        private TelemetryClient TelemetryClient;
+
+        #region Telemetry Context Properties
+        private const string TestInstrumentationKey = "00000000-0000-0000-0000-000000000000";
+
+        private const string TestComponentVersion = "TestComponentVersion";
```
pharring: Nit: `nameof(TestComponentVersion);` (and all the other const strings)

src/Microsoft.ApplicationInsights/DataContracts/TelemetryContext.cs
```diff
@@ -132,13 +132,13 @@ public LocationContext Location
...
             {
                 var result = new Dictionary<string, string>();
-                this.component?.UpdateTags(result);
-                this.device?.UpdateTags(result);
-                this.cloud?.UpdateTags(result);
-                this.session?.UpdateTags(result);
-                this.user?.UpdateTags(result);
-                this.operation?.UpdateTags(result);
-                this.location?.UpdateTags(result);
+                this.Component.UpdateTags(result);
```
pharring: Changing this method defeats the purpose of making the backing fields lazy. I don't think this change is necessary.

src/Microsoft.ApplicationInsights/DataContracts/TelemetryContext.cs
```diff
@@ -156,13 +156,13 @@ internal void Initialize(TelemetryContext source, string instrumentationKey)
...
             Property.Initialize(ref this.instrumentationKey, instrumentationKey);
 
-            this.component?.CopyFrom(source);
-            this.device?.CopyFrom(source);
-            this.cloud?.CopyFrom(source);
-            this.session?.CopyFrom(source);
-            this.user?.CopyFrom(source);
-            this.operation?.CopyFrom(source);
-            this.location?.CopyFrom(source);
+            this.Component.CopyFrom(source);
```
pharring: This may be the simplest fix, but it's defeating the purpose of making the backing fields lazy.
Consider:
`if (source.component != null) this.Component.CopyFrom(source);`
etc.

tomasr: Isn't the entire check backwards? I mean..... whether the target is null or not seems (to me) irrelevant. Wouldn't it be better to check if the *source* has a property set that requires copying? And if so, the target being null is clearly not an issue since you're gonna have to create it in the first place. Am I misunderstanding the whole thing?

src/Microsoft.ApplicationInsights/DataContracts/TelemetryContext.cs
MS-TimothyMothra: at a casual glance, it looked like an identical bug that hadn't been discovered yet. After investigating I agree. This method is used as an export, and the null properties should remain null. I will revert this change.

@pharring what do you think about `this.Internal.UpdateTags(result)`? Should that remain as is?

pharring: @MS-TimothyMothra yes. Internal is different because it is not backed by a lazy-initialized field. It should remain as-is.

src/Microsoft.ApplicationInsights/DataContracts/TelemetryContext.cs
MS-TimothyMothra: @tomasr you're right.
There was some "magic" happening here to LazyInitialize fields as needed.
I've refactored all of this to make it more readable.

MS-TimothyMothra:
@pharring friendly ping for a re-review :)

src/Microsoft.ApplicationInsights/DataContracts/TelemetryContext.cs
```diff
@@ -152,18 +152,23 @@ internal TelemetryContext DeepClone(IDictionary<string, string> properties)
...
-            this.location?.CopyFrom(source);
-            this.Internal.CopyFrom(source);
+            source.component?.CopyTo(this.Component);
+            source.device?.CopyTo(this.Device);
+            source.cloud?.CopyTo(this.Cloud);
+            source.session?.CopyTo(this.Session);
+            source.user?.CopyTo(this.User);
+            source.operation?.CopyTo(this.Operation);
+            source.location?.CopyTo(this.Location);
+            source.Internal?.CopyTo(this.Internal);
```
pharring: Nit: As we touched on before, no need to use `?.` for `Internal` because it's not backed by a lazy field.

MS-TimothyMothra: good catch, thanks!

PR #719, opened 2018-03-01
Fix of the correct parentId correlation token chaining

mfpgod:
Fix Issue # .
When operation context has id & parentId set, then new telemetry id is initialized with the wrong id. Operation context id is used as a base but not the operation parentId.
Example:
type                         id                                                         parentOperationId
request:                   |RoxIgAwOW8Q.                                  RoxIgAwOW8Q.
dependency            |RoxIgAwOW8Q.1.                               |RoxIgAwOW8Q.
inner request           |RoxIgAwOW8Q.45                             |RoxIgAwOW8Q.1.
Instead of '|RoxIgAwOW8Q.45' above there should be '|RoxIgAwOW8Q.1.45'. The issue is reproducible in cross process communication when Activity chanin is lost.

- [* ] I ran [Unit Tests](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/.github/CONTRIBUTING.md) locally.

### DIFF
--- a/src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
+++ b/src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
@@ -125,7 +125,7 @@
                 if (parentActivity == null)
                 {
                     // telemetryContext.Id is always set: if it was null, it is set to opTelemetry.Id and opTelemetry.Id is never null
-                    operationActivity.SetParentId(telemetryContext.Id);
+                    operationActivity.SetParentId(telemetryContext.ParentId ?? telemetryContext.Id);
                 }
 
                 operationActivity.Start();


lmolkova:
@mfpgod 

Could you please describe the actual issue you are experiencing?
The single requirement we have for correlation is
parent.Context.Operation.Id = child.Context.Operation.Id
parent.Id = child.Context.Operation.ParentId
[more on this](https://docs.microsoft.com/en-us/azure/application-insights/application-insights-correlation)

the actual format of Ids does not matter.
Could you please also provide the exact telemtry with all fields showing where correlation is lost?

It seems in your case you are using StartOperation method and it does not work for you. Could you please provide a repro?

Note that the code you changes is intentional. 
Somebody could have called StartOperaton with 
telemetry.Id = "a"
telemetry.Context.Operation.Id = "b"
telemetry.Context.Operation.ParentId = "c"

Now (without your change) child of such operation would be correct

child.Id = "b.1"
child.Context.Operation.Id = "b"
child.Context.Operation.ParentId = "a"

With your change it would be wrong
child.Id = "c.1" !!!
child.Context.Operation.Id = "c" !!!
child.Context.Operation.ParentId = "a"

I'm also surprised tests passed meaning we don;t have validation for this case.

So, please first explain the issue you have and we can see how to fix it.

mfpgod:
Thanks for the fast response. 
Agree, that your case with telemetry.Context.Operation.Id = "b" and telemetry.Context.Operation.ParentId = "c" wont work. At the same time the case assumes that parentId was generated in a not quite correct way - without using the operationId, which, I agree, is possible.

At the same time if you look to the Hierarchical Request-Id model here (table with the example in the end) https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/HierarchicalRequestId.md, you will see that ideally parentId should be used to generate new telemetry id.


I will try to come up with a better fix that could cope with both cases.

lmolkova:
@mfpgod 
I still don't understand what is the problem you are trying to solve. Please let's first describe the problem. Without understanding the impact, I'm afraid we won't accept the fix.

As an author of the spec you are referring to, I can assure you it's a valid violation (assuming that incoming request-id is not validated and there is no guarantee it conforms to this spec).
I do not like the idea of request-id validation in AppInsights SDK as it's a pure overhead and there is no gain from it: we can ensure proper correlation without it.

As AppInsights had non-hierarchical schema before, we had to make sure different versions of the AppInsights SDK could interop in terms of correlation headers and format.

So, please describe your issue, and we'll fix it ;)

mfpgod:
Thanks @lmolkova for the detailed response!
I am working with the Service Fabric and we have business operations that span to multiple microservices. To be able to easily understand what service called what and when, I want the follwoing telemetry id generation:
|RoxIgAwOW8Q. - root request
|RoxIgAwOW8Q.* - root request inner operation traces, exceptions etc.
|RoxIgAwOW8Q.1 - dependency call
|RoxIgAwOW8Q.1.2_ - inner request call
|RoxIgAwOW8Q.1.2_* - inner request call operation traces, exceptions etc.
|RoxIgAwOW8Q.1.2_3. - inner dependency call operation traces, exceptions etc.
etc.
Note: 1, 2, 3 in the ids above are just for the reference - in reality it could be anything. So I need that every inner request/dependency id is generated not from the operation id, but from the parent operation id.

lmolkova:
@mfpgod 
Thanks!

You can use new ApplicationInsights TraceViewer to achieve it without parsing request-id. It's a new portal UI 
image

https://docs.microsoft.com/en-us/azure/application-insights/app-insights-transaction-diagnostics 

This is something we guarantee to work, while the format underneath is likely to change in future and we cannot even guarantee it will be hierarchical.

Please go through the article I mentioned. Perhaps, you can even try it - it's in preview now - just make sure you've enabled it.

Let us know if it would address the issue.

PR #721, opened 2018-03-01
Fix of the correct parentId correlation token chaining

mfpgod:
When operation context has id & parentId set, then new telemetry id is initialized with the wrong id. Operation context id is used as a base but not the operation parentId.
Example:
type id parentOperationId
request: |RoxIgAwOW8Q. RoxIgAwOW8Q.
dependency |RoxIgAwOW8Q.1. |RoxIgAwOW8Q.
inner request |RoxIgAwOW8Q.45 |RoxIgAwOW8Q.1.
Instead of '|RoxIgAwOW8Q.45' above there should be '|RoxIgAwOW8Q.1.45'. The issue is reproducible in cross process communication when Activity chanin is lost.

- [* ] I ran [Unit Tests](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/.github/CONTRIBUTING.md) locally.

Relates to https://github.com/Microsoft/ApplicationInsights-dotnet/pull/719

### DIFF
--- a/src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
+++ b/src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
@@ -125,7 +125,15 @@
                 if (parentActivity == null)
                 {
                     // telemetryContext.Id is always set: if it was null, it is set to opTelemetry.Id and opTelemetry.Id is never null
-                    operationActivity.SetParentId(telemetryContext.ParentId ?? telemetryContext.Id);
+                    if (telemetryContext.ParentId != null && GetRootId(telemetryContext.ParentId) == telemetryContext.Id)
+                    {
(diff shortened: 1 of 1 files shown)


lmolkova:
@mfpgod
Repeating my comment from the initial PR:
I still don't understand what is the problem you are trying to solve. Please let's first describe the problem. Without understanding the impact, I'm afraid we won't accept the fix.

As an author of the spec you are referring to, I can assure you it's a reasonable violation (assuming that incoming request-id is not validated and there is no guarantee it conforms to this spec).
I do not like the idea of request-id validation in AppInsights SDK as it's a pure overhead and there is no gain from it: we can ensure proper correlation without it.

As AppInsights had non-hierarchical schema before, we had to make sure different versions of the AppInsights SDK could interop in terms of correlation headers and format.

So, please describe your issue, and we'll fix it ;)

mfpgod:
Explained in the initial pull request :)

PR #726, opened 2018-03-08
Consolidated tests to Shared.Tests project.

MS-TimothyMothra:
merged duplicate tests into Shared project.
Disabled the failing "FlushCanBeAborted" test.


MS-TimothyMothra:
Not ready to merge this because "FlushCanBeAborted" test has started failing as a result of my changes.
It's not clear why, and needs to be investigated.

Issue #549, opened 2017-05-02
DependencyMetricsExtractor.ExtractMetrics threw exceptions


northtyphoon:
We observed the behavior in some ETL traces we collected. I attached one sample call stack.

"ExceptionMessage Cannot execute ExtractMetrics because this metrics extractor has not been initialized (no metrics manager)."

```
 + Thread (9800) CPU=72ms (.NET ThreadPool)
  + ntdll!?
   + kernel32!?
    + clr!Thread::intermediateThreadProc
     + clr!ThreadpoolMgr::ExecuteWorkRequest
      + clr!UnManagedPerAppDomainTPCount::DispatchWorkItem
       + picohelper!?
        + webengine4!?
         + ?!?
          + clr!UM2MDoADCallBack
           + clr!??Thread::DoADCallBack
            + clr!UM2MThunk_Wrapper
             + ?!?
              + system.web!PipelineRuntime.ProcessRequestNotification
               + system.web!PipelineRuntime.ProcessRequestNotificationHelper
                + system.web!dynamicClass.IL_STUB_PInvoke(int,value class System.Web.RequestNotificationStatus&)
                 + webengine4!?
                  + iiscore!?
                   + webengine4!?
                    + ?!?
                     + system.web!PipelineRuntime.ProcessRequestNotification
                      + system.web!PipelineRuntime.ProcessRequestNotificationHelper
                       + system.web!HttpRuntime.ProcessRequestNotificationPrivate
                        + system.web!HttpApplication.BeginProcessRequestNotification
                         + system.web!System.Web.HttpApplication+PipelineStepManager.ResumeSteps(class System.Exception)
                          + system.web!HttpApplication.ExecuteStep
                           + system.web!System.Web.HttpApplication+CallHandlerExecutionStep.System.Web.HttpApplication.IExecutionStep.Execute()
                            + system.web.mvc!IHttpAsyncHandler.EndProcessRequest
                             + system.web.mvc!MvcHandler.EndProcessRequest
                              + system.web.mvc!System.Web.Mvc.Async.AsyncResultWrapper+<>c__DisplayClass4.<MakeVoidDelegate>b__3(class System.IAsyncResult)
                               + system.web.mvc!System.Web.Mvc.MvcHandler+<>c__DisplayClass8.<BeginProcessRequest>b__3(class System.IAsyncResult)
                                + system.web.mvc!IAsyncController.EndExecute
                                 + system.web.mvc!Controller.EndExecute
[...]

PR #730, opened 2018-03-08
549 fix auto init modules

MS-TimothyMothra:
Fix Issue #549 .
When `TelemetryProcessorChainBuilder` rebuilds Processors, 
Processors that implement `ITelemetryModule` may not get initialized via this workflow.

### DIFF
--- a/Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/TelemetryProcessorChainBuilderTest.cs
+++ b/Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/TelemetryProcessorChainBuilderTest.cs
@@ -103,17 +103,11 @@
 
         private class MockProcessorModule : ITelemetryProcessor, ITelemetryModule
         {
-            public bool ModuleInitialized = false;
+            public bool ModuleInitialized { get; private set; } = false;
 
-            public void Initialize(TelemetryConfiguration configuration)
-            {
-                this.ModuleInitialized = true;
-            }
+            public void Initialize(TelemetryConfiguration configuration) => this.ModuleInitialized = true;
 
-            public void Process(ITelemetry item)
-            {
-                throw new NotImplementedException();
-            }
+            public void Process(ITelemetry item) => throw new NotImplementedException();
         }
     }
 }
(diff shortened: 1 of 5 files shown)


Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/TelemetryProcessorChainBuilderTest.cs
```diff
@@ -72,22 +71,43 @@ public void BuildUsesTelemetryProcesorFactoryOnEachCall()
...
+            var tc1 = new TelemetryConfiguration();
+            var builder1 = new TelemetryProcessorChainBuilder(tc1);
+            builder1.Use((next) => new MockProcessorModule());
+            builder1.Build();
+
+            Assert.AreEqual(2, tc1.TelemetryProcessors.Count); // Transmission is added by default
+            Assert.IsTrue(((MockProcessorModule)tc1.TelemetryProcessors[0]).ModuleInitialized, "Module was not initialized.");
+        }
+
+        private class MockProcessorModule : ITelemetryProcessor, ITelemetryModule
```
d3r3kk: I know this is common practice inside test classes/files, but I still prefer to see mocks in their own sub-folder of the test. I'll leave it to you to decide whether or not this is important, as I don't think it's commonplace in our codebase as yet.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryProcessorChainBuilder.cs
```diff
@@ -104,8 +94,15 @@ public void Build()
                 }
 
                 telemetryProcessorsList.Add(linkedTelemetryProcessor);
-            }
 
+                // If a Processor also implements ITelemtryModule, We should Initialize that Module
+                if (linkedTelemetryProcessor is ITelemetryModule telemetryModule)
+                {
+                    telemetryModule.Initialize(this.configuration);
```
d3r3kk: We protect against re-initialization across thread boundaries in some of the `ITelemetryModule`s I've seen and worked on.  However, it is entirely possible that a 3rd party may not have done so. Would it be prudent to put this inside a try...catch block? (Please safely ignore this comment if there is an outer try catch I cannot see).

MS-TimothyMothra: That's a good point. We have no control over other implementations, and our own [TelemetryConfigurationFactory calls Initialize in a try/catch](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs#L372-L379). I'll do the same here.

Issue #643, opened 2017-10-31
Azure Metrics SDK, version Alpha 0.1: request for feedback.


macrogreg:
Here I introduce the pre-release version of the Microsoft Azure Metrics SDK. I give a brief overview of the feature set and then provide a number of usage examples as a tutorial for the API surface. The goal of this publication is to collect feedback on the API surface and on the implementation strategy.

The goal of the Azure Metrics SDK is to provide an API for metric pre-aggregation and metric ingestion into Azure. In the current pre-release version we target the Application Insights ingestion endpoint. However, the SDK is designed with minimal dependencies on the Application Insights Core SDK. It is structured to be naturally used to submit preaggregated metrics to other endpoints. In the mid- to long-term, a subset of the features may be moved into more platform-level .Net APIs.

Metrics, once emitted by the source application, are always a conceptually aggregated quantity. Specific metric measurements should never be sent or stored in the original, non-aggregated form. Once the APIs discussed here reach a stable public release, the Application Insights TelemetryClient.TrackMetric(..) API should be deprecated. Advanced developers wishing to implement their own pre-aggregation logic should use TelemetryClient.Track(ITelemetry) API to send any telemetry that results from their custom aggregation. Developers who want to send specific single-point measurements should use Events.

Currently the SDK offers the following features:

•	Different kinds of aggregators:
	-- Measurements
	-- Counters
	-- Custom aggregators

•	Zero-dimensional and multi-dimensional metrics:
	-- Local metric series capping
	-- Local dimension cardinality capping
	-- Support for dynamic discovery of dimension values.

•	Data introspection: applications can reason about their own telemetry.

•	Support for centrally configured aggregation.

•	APIs structure is layered:
	-- Low level APIs are optimized for performance and memory consumption. Aggregation is lock-free and GC-friendly.
	-- Convenience APIs are discoverable and provide a smooth learning curve.

•	Support for QuickPulse/LiveMetrics.

•	Support for custom aggregation cycles, data filters, value filters.

•	Support for unit and integration of metrics-emitting code.

Currently the Azure Metrics SDK is available as source code in the ApplicationInsights-SDK-Labs repo on GitHub in the branch metrics/master. Direct link:
[https://github.com/Microsoft/ApplicationInsights-SDK-Labs/tree/metrics/master](https://github.com/Microsoft/ApplicationInsights-SDK-Labs/tree/metrics/master)

[...]

PR #735, opened 2018-03-19
Feature/PreAggMetrics.Master

macrogreg:
Hi folks,

This is the Merge PR for Metrics Pre-Aggregation. We have worked on the architecture and APIs previously.
E.g.: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/643

The amount of code so substantial that making it fit the folder structure and other conventions may take a couple of iterations. I want to start early and obtain early feedback. I'll be reaching out to people directly for specific feedback questions.


macrogreg:
@SergeyKanzhelev ,  @Dmitry-Matveev ,  @zakimaksyutov 
I just sent you an email about this, but Sergey asked to also put a comment here and mention you all:
I have refactored the code to follow the repo rules. It builds cleanly. Basically, everything is ready for review. There are additional details in the email.
Thank you!

SergeyKanzhelev:
Why you committed src/Metrics.Extensions/Materials/StdDev Test Tool.xlsx ?

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics/AccumulatorAggregator.cs
```diff
@@ -0,0 +1,127 @@
...
+{
+    internal sealed class AccumulatorAggregator : MetricSeriesAggregatorBase<double>
+    {
+        private static readonly Func<MetricValuesBufferBase<double>> MetricValuesBufferFactory = () => new MetricValuesBuffer_Double(capacity: 500);
+
+        private readonly object _updateLock = new object();
+
+        private readonly bool _restrictToUInt32Values;
+
+        private double _min;
```
SergeyKanzhelev: we don't do underscore for private fields. Has stylecop complained? If no - we need a new naming rule here

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics/GaugeAggregator.cs
```diff
@@ -0,0 +1,156 @@
...
+            lock (_updateLock)
+            {
+                _data.HasValues = true;
+                _data.Max = (bufferData.Max > _data.Max) ? bufferData.Max : _data.Max;
+                _data.Min = (bufferData.Min < _data.Min) ? bufferData.Min : _data.Min;
+                _data.Last = bufferData.Last;
+            }
+        }
+
+#pragma warning disable SA1401 // Fields must be private
```
SergeyKanzhelev: why wouldn't you create properties instead of disabling the warning? Or make it a `struct`?

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics/NaiveDistinctCountMetricSeriesAggregator.cs
```diff
@@ -0,0 +1,137 @@
...
+
+                    string stringValue = metricValue.ToString();
+
+                    if (!_caseSensitive)
+                    {
+                        stringValue = stringValue.ToLowerInvariant();
+                    }
+
+                    _uniqueValues.Add(stringValue);
+                    Interlocked.Increment(ref _totalValuesCount);
```
SergeyKanzhelev: interlocked inside the lock? Maybe it's valid, just look wierd

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
```diff
@@ -0,0 +1,200 @@
...
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+
+using Microsoft.ApplicationInsights.Metrics.Extensibility;
+using Microsoft.ApplicationInsights.Metrics.Extensions;
+
+namespace Microsoft.ApplicationInsights.Metrics
+{
+    /// <summary>
+    ///
```
SergeyKanzhelev: I'm missing these description. This code is far from being self-descriptive.

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
```diff
@@ -0,0 +1,200 @@
...
+    /// 
+    /// </summary>
+    public class MetricSeriesConfigurationForAccumulator : IMetricSeriesConfiguration
+    {
+        private readonly bool _restrictToUInt32Values;
+        private readonly int _hashCode;
+
+        static MetricSeriesConfigurationForAccumulator()
+        {
+            MetricAggregateToTelemetryPipelineConverters.Registry.Add(
```
SergeyKanzhelev: this is super long line. Why the next line starts with to many "tabs" instead of a single "tab" indent

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+            return _hashCode;
+        }
+
+        /// <summary>
+        /// Groups constants used by metric aggregates produced by aggregators that are configured by metric configurations represented through
+        /// instances of <see cref="MetricSeriesConfigurationForAccumulator"/>. This class cannot be instantiated. To access the constants, use the 
+        /// extension method <c>MetricConfigurations.Common.Accumulator().Constants()</c> or <see cref="MetricSeriesConfigurationForAccumulator.Constants"/>.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
SergeyKanzhelev: why not browsable?

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
```diff
@@ -0,0 +1,200 @@
...
+
+        /// <summary />
+        /// <returns></returns>
+        public override int GetHashCode()
+        {
+            return _hashCode;
+        }
+
+        /// <summary>
+        /// Groups constants used by metric aggregates produced by aggregators that are configured by metric configurations represented through
```
SergeyKanzhelev: Do you need it to be public? Why not simple strings? I don't understand how one suppose to use this class

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
```diff
@@ -0,0 +1,200 @@
...
+                /// </summary>
+                public string Max { get { return Constants.AggregateKindDataKeys.Max; } }
+            }
+        }
+
+        /// <summary>
+        /// Defines constants used my metric aggregates produced by aggregators that are configured by metric configurations represented
+        /// through instances of <see cref="MetricSeriesConfigurationForAccumulator"/>.
+        /// </summary>
+        public static class Constants
```
SergeyKanzhelev: Why this is browsable than?

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+            /// <summary>
+            /// The kind moniker for aggregates produced by aggregators that are configured by metric configurations represented
+            /// through instances of <see cref="MetricSeriesConfigurationForAccumulator"/>.
+            /// </summary>
+            public const string AggregateKindMoniker = "Microsoft.Azure.Accumulator";
```
SergeyKanzhelev: how customers will be using this one?

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForGauge.cs
```diff
@@ -0,0 +1,213 @@
+﻿using System;
```
SergeyKanzhelev: I'm pretty sure stylecop wants `using` be inside `namespace`

src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl.csproj
```diff
@@ -0,0 +1,114 @@
...
+  <Import Project="..\..\..\..\NuGet.Packages\MSTest.TestAdapter.1.1.11\build\net45\MSTest.TestAdapter.props" Condition="Exists('..\..\..\..\NuGet.Packages\MSTest.TestAdapter.1.1.11\build\net45\MSTest.TestAdapter.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{0272FE8B-9B25-43B0-A26B-3C71154DEBBA}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl</RootNamespace>
+    <AssemblyName>Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl</AssemblyName>
+    <TargetFrameworkVersion>v4.6.2</TargetFrameworkVersion>
```
SergeyKanzhelev: project targeted 4.5. Why tests are on 4.6.2?

src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl.csproj
```diff
@@ -0,0 +1,114 @@
...
+    <ProjectTypeGuids>{3AC096D0-A1C2-E12C-1390-A8335801FDAB};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
+    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">15.0</VisualStudioVersion>
+    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+    <ReferencePath>$(ProgramFiles)\Common Files\microsoft shared\VSTT\$(VisualStudioVersion)\UITestExtensionPackages</ReferencePath>
+    <IsCodedUITest>False</IsCodedUITest>
+    <TestProjectType>UnitTest</TestProjectType>
+    <NuGetPackageImportStamp>
+    </NuGetPackageImportStamp>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
```
SergeyKanzhelev: are you sure you need these sections? Is there defaults in global proprs?

src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl.csproj
```diff
@@ -0,0 +1,114 @@
...
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>bin\Debug\</OutputPath>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <DocumentationFile>..\..\..\..\bin\Debug\Metrics\Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl\Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl.xml</DocumentationFile>
```
SergeyKanzhelev: do you need this for test project?

src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl.csproj
```diff
@@ -0,0 +1,114 @@
...
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="Microsoft.ApplicationInsights, Version=2.3.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\..\..\..\NuGet.Packages\Microsoft.ApplicationInsights.2.3.0\lib\net46\Microsoft.ApplicationInsights.dll</HintPath>
+    </Reference>
+    <Reference Include="Microsoft.VisualStudio.TestPlatform.TestFramework, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\..\..\..\NuGet.Packages\MSTest.TestFramework.1.1.11\lib\net45\Microsoft.VisualStudio.TestPlatform.TestFramework.dll</HintPath>
+    </Reference>
+    <Reference Include="Microsoft.VisualStudio.TestPlatform.TestFramework.Extensions, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\..\..\..\NuGet.Packages\MSTest.TestFramework.1.1.11\lib\net45\Microsoft.VisualStudio.TestPlatform.TestFramework.Extensions.dll</HintPath>
```
SergeyKanzhelev: projects are using `1.2.0`

src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl.csproj
```diff
@@ -0,0 +1,114 @@
...
+    <Compile Include="Microsoft.ApplicationInsights.Metrics\ExtensionsUtilTests.cs" />
+    <Compile Include="Microsoft.ApplicationInsights.Metrics\UtilTests.cs" />
+    <Compile Include="Microsoft.ApplicationInsights.Metrics\MetricAggregationManagerTests.cs" />
+    <Compile Include="Microsoft.ApplicationInsights.Metrics\NaiveDistinctCountMetricSeriesAggregatorTests.cs" />
+    <Compile Include="Microsoft.ApplicationInsights.Metrics\MeasurementAggregator_restrictToUInt32ValuesFalse_Tests.cs" />
+    <Compile Include="Microsoft.ApplicationInsights.Metrics\MeasurementAggregator_restrictToUInt32ValuesTrue_Tests.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
```
SergeyKanzhelev: Use `PackageReference` instead

SergeyKanzhelev:
Nother naming issue: test files should not use underscore in names like: `Foo_Tests.cs`

src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test/Microsoft.ApplicationInsights.Metrics.TestUtil/StubApplicationInsightsTelemetryChannel.cs
```diff
@@ -0,0 +1,77 @@
...
+
+using Microsoft.ApplicationInsights.Channel;
+
+
+namespace Microsoft.ApplicationInsights.Metrics.TestUtil
+{
+    /// <summary>
+    /// A stub of <see cref="ITelemetryChannel"/>.
+    /// </summary>
+    public sealed class StubTelemetryChannel : ITelemetryChannel
```
SergeyKanzhelev: I'm pretty sure we have at least one stub channel in repo already

SergeyKanzhelev:
Also fix the build

macrogreg:
Seems impossible to respond to a comment within the context of that comment if the comment is not tight to a code line. :( So doing it here..

> [ @SergeyKanzhelev ] Why you committed src/Metrics.Extensions/Materials/StdDev Test Tool.xlsx ?

This is an essential tool when calculating the correct values for testing aggregators. Thanks for pointing that out. As I move the tests into the right directory, I will place this file next to them.

macrogreg:
> [ @SergeyKanzhelev ] Nother naming issue: test files should not use underscore in names like: Foo_Tests.cs

Can you please clarify what you mean> I looked and only found files with underscores under
src\Microsoft.ApplicationInsights\Extensibility\Implementation\External\
which is not part of this PR.

macrogreg:
> [ @SergeyKanzhelev ] I reviewed the form of code, not semantics. I'm really missing PublicAPI changes update and documentation for public methods to make sense of this code.

I will add the PublicAPI changes update as the next step. As discussed offline, I wanted to get this moving while i was still working on it. In terms of docs - a part is already completed, but there will not be time to describe every single API this week. Especially, it seems that we should focus on docs after we have closed on all API shape discussions.

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/MetricManagerTest.cs
```diff
@@ -20,10 +20,10 @@ public void CanCreateMetricHavingNoDimensions()
             var sentTelemetry = new List<ITelemetry>();
 
             var client = this.InitializeTelemetryClient(sentTelemetry);
-            using (MetricManager manager = new MetricManager(client))
+            using (MetricManagerV1 manager = new MetricManagerV1(client))
```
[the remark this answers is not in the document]
macrogreg: You are right, it would have been better to do this as separate PRs.Unfortunately I discovered the need for this renaming in the process of integrating and they ended up embedded. I can redo those changes in a separate branch for a separate PR, but it will take a little time. Given that they concern a separate set of files I hope that it does not affect the PR logic too much. Can we please skip this improvement?

SergeyKanzhelev:
@macrogreg it is nearly impossible to review public API without comments and even a list of how API looks like. Especially in 120 files PR that GitHub keep "optimizing" by hiding files and making them non-searcheable.

SergeyKanzhelev:
For the renaming - few clicks on refactor/rename and than merge back to this PR from develop doesn't sounds like time-consuming tasks. But it may save more time for reviewers that you've spent on implementing it.

Test/Microsoft.ApplicationInsights.Test/Shared/TelemetryClientTest.cs
```diff
@@ -140,6 +140,7 @@ public void TrackMetricSendsSpecifiedAggregatedMetricTelemetry()
             var sentTelemetry = new List<ITelemetry>();
             var client = this.InitializeTelemetryClient(sentTelemetry);
 
+#pragma warning disable CS0618 // Type or member is obsolete
```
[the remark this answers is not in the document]
macrogreg: The logic behind deprecating TrackMetric(..):
The recommended new API for metrics is now .GetMetric(..).TrackValue(..). The old TrackMetric(..) API is discouraged because it
a) Distracts from discovering the new API. (Being used to the TrackXxx(..) pattern users will discover it in intellisense and stop looking for other ways to send a metric.)
b) Promotes incorrect coding behavior. (Being used to the TrackXxx pattern users will just construct MetricTelemetry items for each data point and utilize the old API without ever reasoning about the need for pre-aggregation.)

The old TrackMetric(..) API can be used in 2 cases:
A) "Einstein" users who have implemented their own pre-aggregation. These few very advanced users can be expected to realize that TrackMetric(MetricTelemetry item) is just a thin wrapper around Track(ITelemetty item) and that the latter can be used to send the results of their custom pre-aggregation.
B) Users who want to send a single (non-pre-aggregated) measurement. However, such data is not a correct use of AppInsights' data model. If the customer really requires to send and store a single numerical observation, a custom event is the correct data item to use. If the customer is actually interested in a metric, than pre-aggregation is more appropriate. In fact, metrics (on contrary to events) should *never* leave the collection box without pre-aggregation.

src/Metrics.Extensions/Microsoft.ApplicationInsights.Metrics.Extensions.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
macrogreg: I think this is a misunderstanding. As discussed on Skype, most of the metrics work is in the Microsoft.ApplicationInsights folder and ships with the Base SDK. A few extensions are optional. they do not ship with the SDK. Customers may optionally install that separate NuGet separately and we do not plan to offer the same level of support on those features as we do for the base SDK. Those are all in the Metrics.Extensions folder. I put it in this repo to take advantage of the infrastructure, but we can move it into a separate repo if we want to. Either way, it builds and ships separately. OK?

src/Metrics.Extensions/CommonBuild.Global.props
```diff
@@ -0,0 +1,126 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
macrogreg: Please see my response on the Metrics.Extensions/Microsoft.ApplicationInsights.Metrics.Extensions.csproj file.

src/Metrics.Extensions/Microsoft.ApplicationInsights.Metrics.Extensions.props
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
macrogreg: Please see my response on the Metrics.Extensions/Microsoft.ApplicationInsights.Metrics.Extensions.csproj file.

src/Metrics.Extensions/Microsoft.ApplicationInsights.Metrics.Extensions.ruleset
```diff
@@ -0,0 +1,223 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
macrogreg: Please see my response on the Metrics.Extensions/Microsoft.ApplicationInsights.Metrics.Extensions.csproj file.

src/Metrics.Extensions/Microsoft.ApplicationInsights.Metrics.Extensions.sln
```diff
@@ -0,0 +1,22 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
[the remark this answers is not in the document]
macrogreg: Please see my response on the Metrics.Extensions/Microsoft.ApplicationInsights.Metrics.Extensions.csproj file.

src/Metrics.Extensions/src/Internal/InternalsVisibleTo.cs
```diff
@@ -0,0 +1,4 @@
+﻿using System.Runtime.CompilerServices;
+
+//[assembly: InternalsVisibleTo("Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl.dll")]
+[assembly: InternalsVisibleTo("Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl")]
```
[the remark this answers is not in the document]
macrogreg: Good point. Originally this came from the need to limit complete paths to 255 chars. But I do not like this either. I will work to change this as I move all tests into the right place.

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics.Extensions/Util.cs
```diff
@@ -0,0 +1,206 @@
...
+
+        private const string FallbackParemeterName = "specified parameter";
+
+        /// <summary>
+        /// Paramater check for Null with a little more informative exception.
+        /// </summary>
+        /// <param name="value">Value to be checked.</param>
+        /// <param name="name">Name of the parameter being checked.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void ValidateNotNull(object value, string name)
```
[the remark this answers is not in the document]
macrogreg: What do you mean by 
> moder C#
By what specific code would you like me to replace invocations of this method?

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics.Extensions/Util.cs
```diff
@@ -0,0 +1,206 @@
...
+{
+    internal static class Util
+    {
+        public const string NullString = "null";
+        private const double MicroOne = 0.000001;
+
+        private const string FallbackParemeterName = "specified parameter";
+
+        /// <summary>
+        /// Paramater check for Null with a little more informative exception.
```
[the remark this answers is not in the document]
macrogreg: Good point. Fixed in the next iteration.

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics.Extensions/Util.cs
```diff
@@ -0,0 +1,206 @@
...
+            }
+        }
+
+        /// <summary>
+        /// String paramater check with a little more informative exception.
+        /// </summary>
+        /// <param name="value">Value to be checked.</param>
+        /// <param name="name">Name of the parameter being checked.</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void ValidateNotNullOrEmpty(string value, string name)
```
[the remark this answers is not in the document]
macrogreg: Good point on comments. Fixed in the next iteration.

If you do not mind, I would prefer not to churn the entire project for this. I see this as a short and descriptive way to do informative string validation. We can use it in the rest of the project as well, but we should do this as a later PR.

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics.Extensions/Util.cs
```diff
@@ -0,0 +1,206 @@
...
+                }
+            }
+        }
+
+        public static int CombineHashCodes(int hash1)
+        {
+            int hash = 17;
+            unchecked
+            {
+                hash = hash * 23 + hash1;
```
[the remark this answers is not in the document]
macrogreg: defensively putting the compete calculation in the unchecked block.

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics.Extensions/Util.cs
```diff
@@ -0,0 +1,206 @@
...
+
+        public static int CombineHashCodes(int hash1, int hash2)
+        {
+            int hash = 17;
+            unchecked
+            {
+                hash = hash * 23 + hash1;
+                hash = hash * 23 + hash2;
+            }
+            return hash;
```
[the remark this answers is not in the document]
macrogreg: if it complaints, i will add the new line.

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics/AccumulatorAggregateToApplicationInsightsPipelineConverter.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using Microsoft.ApplicationInsights.DataContracts;
+
+namespace Microsoft.ApplicationInsights.Metrics.Extensibility
+{
+    /// <summary />
```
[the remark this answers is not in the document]
macrogreg: See previous comment on this separate folder. It's a different project. I have not attempted to enforce the Core's Style Cop rules there.

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics/AccumulatorAggregator.cs
```diff
@@ -0,0 +1,127 @@
+﻿using System;
+using System.Runtime.CompilerServices;
+using Microsoft.ApplicationInsights.Metrics.Extensibility;
+
+using Util = Microsoft.ApplicationInsights.Metrics.Extensions.Util;
```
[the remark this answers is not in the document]
macrogreg: This is a separate project so I cannot merge Utils. But let me see if I can rename the namespace. Either way - again - we are not sipping this as discussed.

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics/GaugeAggregator.cs
macrogreg: Just an encapsulation private to this aggregator. Minimizing code. struct would make it less efficient for passing around.

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
macrogreg: We are still in the extensions assembly

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
macrogreg: Reduced indent in the next iteration.

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
macrogreg: The class does not need to pollute intellisense in the metrics namespace. the extensions methods its defines are still browsable after the dot of the this parameter.

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
macrogreg: This is to display all the right things after every dot in intellisense when working thought a chain like  MetricConfigurations.Common.Accumulator().Constants().Xxx

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
macrogreg: There are examples in
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/feature/PreAggMetrics.Master/src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test/Examples.cs

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
macrogreg: There are examples in
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/feature/PreAggMetrics.Master/src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test/Examples.cs

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForGauge.cs
macrogreg: still separate project

src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl.csproj
macrogreg: Good point. Need to fix while moving tests.

src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl.csproj
macrogreg: will try to remove

src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl.csproj
macrogreg: nope, good point.

src/MetricsTests/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl/Microsoft.ApplicationInsights.Metrics.Core.Test.Intrnl.csproj
macrogreg: will adjust

macrogreg:
@SergeyKanzhelev  I have pushed a change that re-enables PublicApiAnalyzer and lists the APIs in the right files.

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/MetricManagerTest.cs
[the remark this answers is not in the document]
SergeyKanzhelev: Have you tried opening and reviewing this PR? It hangs chrome on my machine =)

It literally takes a few minutes to merge as there will be only handful conflicts. I just tried. And will save a lot of time and nerves to reviewers. So please consider doing it

Test/Microsoft.ApplicationInsights.Test/Shared/TelemetryClientTest.cs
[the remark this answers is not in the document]
SergeyKanzhelev: Discouraging doesn't mean you need to deprecate it. You can make it non editor browseable. I was OK obsoleting it when we discussed introducing new TrackAggregaterMetric API. Since we still have a VALID scenario when one will want to use API - we should not mark it as obsolete.

src/Metrics.Extensions/Microsoft.ApplicationInsights.Metrics.Extensions.csproj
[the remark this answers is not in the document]
SergeyKanzhelev: Not OK. Either code follows all practices and builds in a single build or should be moved out of repository

src/Metrics.Extensions/Microsoft.ApplicationInsights.Metrics.Extensions.props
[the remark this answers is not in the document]
SergeyKanzhelev: Please see my reply there

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics.Extensions/Util.cs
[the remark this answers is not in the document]
SergeyKanzhelev: `value ?? throw new ArgumentNullException(paramName: nameof(value);` Or any of examples in this repo like https://github.com/Microsoft/ApplicationInsights-dotnet/blob/7321d2bf7b9c4733fb43b6c186e381863931eb1a/src/ServerTelemetryChannel/Implementation/PlatformFile.cs#L12-L15

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics/AccumulatorAggregateToApplicationInsightsPipelineConverter.cs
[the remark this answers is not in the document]
SergeyKanzhelev: then move the project out please

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics/GaugeAggregator.cs
SergeyKanzhelev: So why not properties? Compiler will optimize everything out

src/Metrics.Extensions/src/Public/Microsoft.ApplicationInsights.Metrics/MetricSeriesConfigurationForAccumulator.cs
SergeyKanzhelev: but why?

src/Metrics.Extensions/src/Internal/Microsoft.ApplicationInsights.Metrics.Extensions/Util.cs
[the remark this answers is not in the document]
macrogreg: I really prefer the ValidateNotNull helper method for its expressiveness. I would prefer not to expand the scope of this change to having the entire SDK use it. I see no harm in having a specific feature area using a helper method while others do not.

src/Metrics.Extensions/Microsoft.ApplicationInsights.Metrics.Extensions.csproj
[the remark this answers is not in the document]
macrogreg: I ll move it out to short-cut this particular discussion line.

Test/Microsoft.ApplicationInsights.Test/Shared/TelemetryClientTest.cs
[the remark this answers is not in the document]
macrogreg: Typical usages of TrackMetric are the "wrong" ones. It is hardly likely that most users wrote their own pre-aggregators and much more likely that they are sending non-preaggregated values. Deprecating the method will highlight the problem to those customers and tell them how to fix it. it will also prevent them from making this mistake in the future. The few Einstein Pre-Aggregation authors are advanced users who should not be too inconvenienced by switching to Track. Especially, because the fact that they fall into this category indicates that they will understand why we are doing this.

macrogreg:
Hi @Dmitry-Matveev  & @SergeyKanzhelev ,

I have critically evaluated the new API surface.
A one of the features offered by the library is to allow customers to plug in their own aggregator into the system and have it magically take effect.
For example I plan to ship a separate optional nuget (the aforementioned Metrics.Extensions) package that uses this feature to provide Accumulation and Distinct Count aggregators.

However, if we hold off on releasing this feature in the first beta, we can significantly reduce the size of the public API surface.

We can manage the size of the public change by doing this. However, I would like to re-enable the respective APIs in a future beta.
If you are OK with this plan, I will make change the corresponding APIs from public to internal for this beta release.

Are you OK with this plan?

macrogreg:
I have pushed a change that reduces the public surface significantly, as discussed in my previous comments. Please note that there is still a significant number of lines in the PublicAPI.Unshipped.txt files, but a lot of them are overloads of the same methods.

Heads-up: @SergeyKanzhelev   @Dmitry-Matveev

macrogreg:
Thank you, @Dmitry-Matveev , for the review.

> Please coordinate with Sergey on TrackMetric() API. There are still concerns about it being obsolete with this PR.

@SergeyKanzhelev : You have asked some questions to which I all replied. Here is a link to that conversation for your convenience:
https://github.com/Microsoft/ApplicationInsights-dotnet/pull/735/files/1f151e21907f4408a8eb783fc0c1a05553fff20e#diff-a4553125b36a87ff02a1256a752265fc

Do you have any follow-up questions? 

Thanks!

Test/Microsoft.ApplicationInsights.Test/Shared/TelemetryClientTest.cs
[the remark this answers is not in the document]
macrogreg: I have not heard back in 3 days on this particular issue, so I am assuming that my explanation makes sense.
Notably, a beta release is the right place to deprecate this API and to see what customer feedback we get from it.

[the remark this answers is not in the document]
OsvaldoRosado: As a non-maintainer of this SDK who doesn't have a ton of context for why this API is the way it is, I'm curious at why these changes are incompatible with `TrackMetric`

@macrogreg You noted `TrackMetric` should be marked obsolete because it:
> Promotes incorrect coding behavior. (Being used to the TrackXxx pattern users will just construct MetricTelemetry items for each data point and utilize the old API without ever reasoning about the need for pre-aggregation.)

Isn't this entirely our fault? `TrackMetric` does whatever we say it does... If it should have been aggregating the whole time why don't we fix the disconnect from user expectations and actual behavior by making `TrackMetric` work the _right_ way? Why is it important that users of our `TelemetryClient` API need to reason separately over metrics/aggregation and use different API methods that are inconsistent from standard `TelemetryClient` methods for tracking these when we can transparently aggregate it behind the scenes?

Is the reason for all of this so users can specify the type of metric / means of aggregation? Can't that be an optional parameter to `TrackMetric` and it can handle all the necessary plumbing with the aggregator internally?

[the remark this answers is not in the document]
macrogreg: Hi @OsvaldoRosado ,

As this topic has been discussed at great length and detail so many times, let me please give you a brief overview and then we can chat offline for additional details:

- Established convention for TrackXxx methods is that the thing being tracked will be sent to ingestion. However, in this case the item being tracked will not be sent. Now when the method is called, nor later. It will merely contribute to an aggregation that will be sent at some future time. So having this method aggregate would give it an unexpected behavior. We would be immediately responding to tons of customer requests of the kind "I tracked the metric value X, why do you not show it to me that way?".

- Changing the behavior of an existing API is a highly discouraged practice.

- One of the requirements is that thousands of metric measurements may be observed per second. We need APIs that allow contributing values to aggregates without object creation and lookups. This requires "handles" to metric series.

- Local dimension capping is required to protect against memory problems due to unthoughtful telemetry collection.

- Different kinds of aggregation should be supported.

- QuickPulse integration must be possible.

- Multiple dimensions must be possible.

We have evaluated multiple options. This includes your suggestion, separate, strongly typed metric classes, the present solution, where metrics are configured in one place and tracked all over the code, as well as other options. While there is no ideal solution to a complex problem, the feature team believed that this was the most optimal balance.

I would be happy to chat more. :)

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
```diff
@@ -33,5 +33,237 @@ Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.IHeartbeatPro
 Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.IHeartbeatPropertyManager.IsHeartbeatEnabled.get -> bool
 Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.IHeartbeatPropertyManager.IsHeartbeatEnabled.set -> void
 Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.IHeartbeatPropertyManager.SetHeartbeatProperty(string propertyName, string propertyValue = null, bool? isHealthy = null) -> bool
-Microsoft.ApplicationInsights.DataContracts.PageViewTelemetry.Id.get -> string
-Microsoft.ApplicationInsights.DataContracts.PageViewTelemetry.Id.set -> void
\ No newline at end of file
+Microsoft.ApplicationInsights.Metric
```
SergeyKanzhelev: historically we keep namespace `Microsoft.ApplicationInsights` clean. So developer has only option in intelliSense after typing `Microsoft.ApplicaitonInsights` to create a `TelemetryClient`. I remember we looked at implementation and only thing in root namespace was change in `TelemetryClient`. Is my recollection incorrect?

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
```diff
@@ -33,5 +33,237 @@ Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.IHeartbeatPro
...
+Microsoft.ApplicationInsights.Metric.GetDimensionValues(int dimensionNumber) -> System.Collections.Generic.IReadOnlyCollection<string>
+Microsoft.ApplicationInsights.Metric.Identifier.get -> Microsoft.ApplicationInsights.Metrics.MetricIdentifier
+Microsoft.ApplicationInsights.Metric.SeriesCount.get -> int
+Microsoft.ApplicationInsights.Metric.TrackValue(double metricValue) -> void
+Microsoft.ApplicationInsights.Metric.TrackValue(object metricValue) -> void
+Microsoft.ApplicationInsights.Metric.TryGetDataSeries(out Microsoft.ApplicationInsights.Metrics.MetricSeries series) -> bool
+Microsoft.ApplicationInsights.Metric.TryGetDataSeries(out Microsoft.ApplicationInsights.Metrics.MetricSeries series, bool createIfNotExists, params string[] dimensionValues) -> bool
+Microsoft.ApplicationInsights.Metric.TryGetDataSeries(out Microsoft.ApplicationInsights.Metrics.MetricSeries series, string dimension1Value) -> bool
+Microsoft.ApplicationInsights.Metric.TryGetDataSeries(out Microsoft.ApplicationInsights.Metrics.MetricSeries series, string dimension1Value, string dimension2Value) -> bool
```
SergeyKanzhelev: why do we need `Try` alternative? Why not limit it to unconditional method? What customer will do when it returned false?

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
```diff
@@ -33,5 +33,237 @@ Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.IHeartbeatPro
...
+Microsoft.ApplicationInsights.Metrics.MetricManager.CreateNewSeries(string metricNamespace, string metricId, Microsoft.ApplicationInsights.Metrics.IMetricSeriesConfiguration config) -> Microsoft.ApplicationInsights.Metrics.MetricSeries
+Microsoft.ApplicationInsights.Metrics.MetricManager.CreateNewSeries(string metricNamespace, string metricId, System.Collections.Generic.IEnumerable<System.Collections.Generic.KeyValuePair<string, string>> dimensionNamesAndValues, Microsoft.ApplicationInsights.Metrics.IMetricSeriesConfiguration config) -> Microsoft.ApplicationInsights.Metrics.MetricSeries
+Microsoft.ApplicationInsights.Metrics.MetricManager.Flush() -> void
+Microsoft.ApplicationInsights.Metrics.MetricManager.MetricManager(Microsoft.ApplicationInsights.Metrics.Extensibility.IMetricTelemetryPipeline telemetryPipeline) -> void
+Microsoft.ApplicationInsights.Metrics.MetricManager.Metrics.get -> Microsoft.ApplicationInsights.Metrics.MetricsCollection
+Microsoft.ApplicationInsights.Metrics.MetricSeries
```
SergeyKanzhelev: do we need to expose it? I haven't looked at it yet, but curious. I understand there is a `Metric` that you get via `GetMetric`, aggregator that you can configure for customization. What else do we need?

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
```diff
@@ -33,5 +33,237 @@ Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.IHeartbeatPro
...
+Microsoft.ApplicationInsights.TelemetryClient.GetMetric(string metricId, string dimension1Name, string dimension2Name, string dimension3Name, string dimension4Name, Microsoft.ApplicationInsights.Metrics.MetricConfiguration metricConfiguration) -> Microsoft.ApplicationInsights.Metric
+Microsoft.ApplicationInsights.TelemetryClient.GetMetric(string metricId, string dimension1Name, string dimension2Name, string dimension3Name, string dimension4Name, Microsoft.ApplicationInsights.Metrics.MetricConfiguration metricConfiguration, Microsoft.ApplicationInsights.MetricAggregationScope aggregationScope) -> Microsoft.ApplicationInsights.Metric
+const Microsoft.ApplicationInsights.MetricDimensionNames.TelemetryContext.Cloud.RoleInstance = "TelemetryContext.Cloud.RoleInstance" -> string
+const Microsoft.ApplicationInsights.MetricDimensionNames.TelemetryContext.Cloud.RoleName = "TelemetryContext.Cloud.RoleName" -> string
+const Microsoft.ApplicationInsights.MetricDimensionNames.TelemetryContext.Component.Version = "TelemetryContext.Component.Version" -> string
+const Microsoft.ApplicationInsights.MetricDimensionNames.TelemetryContext.Device.Id = "TelemetryContext.Device.Id" -> string
```
SergeyKanzhelev: why do you need all these constants exposed publicly? Especially I concerned for the deprecated properties like device ID and screen resolution

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
```diff
@@ -33,5 +33,237 @@ Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.IHeartbeatPro
...
+static Microsoft.ApplicationInsights.Metrics.Extensibility.MetricSeriesExtensions.GetConfiguration(this Microsoft.ApplicationInsights.Metrics.MetricSeries metricSeries) -> Microsoft.ApplicationInsights.Metrics.IMetricSeriesConfiguration
+static Microsoft.ApplicationInsights.Metrics.Extensibility.TelemetryClientExtensions.GetMetricManager(this Microsoft.ApplicationInsights.TelemetryClient telemetryClient, Microsoft.ApplicationInsights.MetricAggregationScope aggregationScope) -> Microsoft.ApplicationInsights.Metrics.MetricManager
+static Microsoft.ApplicationInsights.Metrics.MetricIdentifier.DefaultMetricNamespace.get -> string
+static Microsoft.ApplicationInsights.Metrics.MetricIdentifier.DefaultMetricNamespace.set -> void
+static Microsoft.ApplicationInsights.Metrics.TelemetryConfigurationExtensions.GetMetricManager(this Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration telemetryPipeline) -> Microsoft.ApplicationInsights.Metrics.MetricManager
+static readonly Microsoft.ApplicationInsights.MetricConfigurations.Common -> Microsoft.ApplicationInsights.MetricConfigurations
```
SergeyKanzhelev: what is it?

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
```diff
@@ -33,5 +33,237 @@ Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.IHeartbeatPro
...
+const Microsoft.ApplicationInsights.Metrics.MetricSeriesConfigurationForMeasurement.Constants.AggregateKindDataKeys.Count = "Count" -> string
+const Microsoft.ApplicationInsights.Metrics.MetricSeriesConfigurationForMeasurement.Constants.AggregateKindDataKeys.Max = "Max" -> string
+const Microsoft.ApplicationInsights.Metrics.MetricSeriesConfigurationForMeasurement.Constants.AggregateKindDataKeys.Min = "Min" -> string
+const Microsoft.ApplicationInsights.Metrics.MetricSeriesConfigurationForMeasurement.Constants.AggregateKindDataKeys.StdDev = "StdDev" -> string
+const Microsoft.ApplicationInsights.Metrics.MetricSeriesConfigurationForMeasurement.Constants.AggregateKindDataKeys.Sum = "Sum" -> string
+const Microsoft.ApplicationInsights.Metrics.MetricSeriesConfigurationForMeasurement.Constants.AggregateKindMoniker = "Microsoft.Azure.Measurement" -> string
+override Microsoft.ApplicationInsights.Metrics.MetricConfiguration.Equals(object obj) -> bool
+override Microsoft.ApplicationInsights.Metrics.MetricConfiguration.GetHashCode() -> int
+override Microsoft.ApplicationInsights.Metrics.MetricIdentifier.Equals(object otherObj) -> bool
```
SergeyKanzhelev: Judging by name `MetricIdentifier` sounds like too low level of a detail

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
```diff
@@ -33,5 +33,237 @@ Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.IHeartbeatPro
...
+static Microsoft.ApplicationInsights.Metrics.Extensibility.TelemetryClientExtensions.GetMetricManager(this Microsoft.ApplicationInsights.TelemetryClient telemetryClient, Microsoft.ApplicationInsights.MetricAggregationScope aggregationScope) -> Microsoft.ApplicationInsights.Metrics.MetricManager
+static Microsoft.ApplicationInsights.Metrics.MetricIdentifier.DefaultMetricNamespace.get -> string
+static Microsoft.ApplicationInsights.Metrics.MetricIdentifier.DefaultMetricNamespace.set -> void
+static Microsoft.ApplicationInsights.Metrics.TelemetryConfigurationExtensions.GetMetricManager(this Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration telemetryPipeline) -> Microsoft.ApplicationInsights.Metrics.MetricManager
+static readonly Microsoft.ApplicationInsights.MetricConfigurations.Common -> Microsoft.ApplicationInsights.MetricConfigurations
+virtual Microsoft.ApplicationInsights.Metrics.MetricConfiguration.Equals(Microsoft.ApplicationInsights.Metrics.MetricConfiguration other) -> bool
```
SergeyKanzhelev: `MetricConfiguration` namespace looks empty - only two methods?

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
macrogreg: Of course I may not have the right historical perspective.
However, this is not quite correct:
 >  I remember we looked at implementation and only thing in root namespace was change in TelemetryClient.

The public API service is identical to what we reviewed at the end of 2017 (barring the types that we made internal this week as a follow up to @Dmitry-Matveev 's feedback). At the time, I made a set of examples available, which are now also in the Core repo: 

https://github.com/Microsoft/ApplicationInsights-dotnet/blob/feature/PreAggMetrics.Master/Test/Microsoft.ApplicationInsights.Test/Shared/Metrics/MetricsExamples.cs

In that set of examples, you can find use case that get progressively more advanced. The goal was to enable the use cases from example 1 to be discovered via Intellisense without the need for additional namespace imports. In order to achieve that, the base namespace includes all types that may appear as parameters to the metric related TelemetryClient methods.

Please let me know if this answers your question.
If you require adjustments to these APIs, please request precisely the changes you would like to see so that I can make them.
Thank you.

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
macrogreg: Thank you for this question.
The Try-pattern is employed here in-line with the Framework Design Guidelines, because the method for tracking a value may fail even if the user specifies valid parameters under non-trivial conditions:

A key goal of the metric APIs is to ensure that a service cannot fail due to mistakes in tracing and telemetry collection (nothing is absolute, so we strive for "unlikely to fail"). This includes preventing memory starvation due to unintentional combinatorial explosion in the number if time series.
For that a `MetricConfiguration` has options for `SeriesCountLimit` and `ValuesPerDimensionLimit`. If these limits are reached, but a `TryTrackValue(..)` call would result in the creation of a new time series, the call will return `false` and the value will not be tracked.

There are different approaches for handling dimension capping, so rather than prescribing the customer a specific approach, we make it easy to detect this case and to react to the circumstance. Please let me know if you would like me to discuss some of these approaches.

Notably, we do not offer exception-throwing `TrackValue(..)` equivalents to all `TryTrackValue(..)` methods. This is because appropriate usage of such `TrackValue(..)` methods would require them to always be used in a try-catch block. This is not appropriate for tracing and telemetry collection APIs, which should be easy to incorporate in existing code and must not break existing code.

Please let me know if this answers your question.
If you require adjustments to these APIs, please request *precisely* the changes you would like to see so that I can make them.
Thank you.

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
macrogreg: I apologize, but I am not sure what you mean by "it" in your question "do we need to expose it?". 

The GitHub review tool more than 1 API to which this may refer. I will address your question to the best of my ability. Please let me know if you referred to a different API.

* The read-only `Metrics` property of the `Microsoft.ApplicationInsights.Metrics.MetricManager` class:

This provides access to all metrics in the manager's scope. A key use case is to allow users to manage existing metrics. For instance, consider an long-running application that manages containers with a limited life-time and tracks metrics about those containers. When a container life-time has finished, the business system can be sure that no further metrics about this container will be sent. However, the telemetry subsystem does not have this information. The `Metrics` collection allows a user to release no longer needed metrics and free up the corresponding memory.

* The `Microsoft.ApplicationInsights.Metrics.MetricSeries` class:

One of the primary requirements for the metrics APIs is to allow tracking 1000s of metric values per second with a small performance overhead. This requires the ability to obtain a handle to a specific metric series and to cache a reference to it so that it can be used to track values without the need for further look-ups. The `MetricSeries` class provides APIs for doing this without the need for any look-ups or object allocations.

This technique is discussed in detail in the samples 2 and 2a:
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/feature/PreAggMetrics.Master/Test/Microsoft.ApplicationInsights.Test/Shared/Metrics/MetricsExamples.cs

 Please let me know if this answers your question.
If you require adjustments to these APIs, please request *precisely* the changes you would like to see so that I can make them.
Thank you.

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
macrogreg: Thank you for this question.
The usage, the reason and the background for the constants in `MetricDimensionNames` are described in detail in the doc comments for that class, as well as in the samples 2 and 5:

https://github.com/Microsoft/ApplicationInsights-dotnet/blob/feature/PreAggMetrics.Master/Test/Microsoft.ApplicationInsights.Test/Shared/Metrics/MetricsExamples.cs

We have also previously discussed this, albeit briefly.

Please let me re-iterate some key background points to put the aforementioned resources in context:

A key feedback point for MetricManager v1 was that it did not allow to specify `TelemetryContext` properties. For example, if other kinds of telemetry are associated with a specific *session* or a specific *operation name*, they use the corresponding properties within the `TelemetryContext` to convey that information. However, metric aggregates produced by MetricManager v1 did allow access to the `TelemetryContext`, and such information needed to be stored as a normal dimensions. The inability to overcome this inconsistency was named as a release blocker by the SDK team at that time. 

Exposing `TelemetryContext` for metric series remains impractical because series represent aggregation scopes and require immutable identities that are fast to compare. So, to provide a solution for the described problem, we use well-defined dimension names. Dimensions with such names are treated as normal dimensions during pre-aggregation, but when the resulting aggregates are serialized for transport, the values of such dimensions are stored within the corresponding fields of the `TelemetryContext`.

Please let me know if this answers your question.
If you require adjustments to these APIs, please request *precisely* the changes you would like to see so that I can make them. Especially, if certain fields do no longer belong into the `TelemetryContext` and should be treated as normal dimensions, we should remove them from the constants list.
Thank you.

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
macrogreg: The review tool shows 3 APIs with return types and parameters including 5 types. Could you please specify what you refer to and what changes, if any, you require?

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
macrogreg: Could you please help me understand your question? Do you have concerns with the name "MetricIdentifier"? Do you have concerns with the usage of the type `MetricIdentifier`? Something else? The level of detail is to low for what specifically? What detail exactly do you refer to?

In order to provide as much context as possible, let me please describe the purpose and the usage of the `MetricIdentifier` class.

A metric is a grouping of one or more metric data series. Besides the actual data (aka the metric series) such a group has 2 attributes:

i) The *metric configuration*. It is a set of settings that define the behavior of the group and the contained data. For example, it specifies how values are aggregated, how many series are allowed in the group and so on. Metrics with the same configuration may contain different data, but they will exhibit the same  behavior. You can think of this as loosely corresponding to a "class type" for a metric.

ii) The *metric identifier*. It is a set of settings that define the identity of the group so that it can be differentiated from any other groups (aka from other metrics). The identifier includes the *metric name*, the *metric namespace*, as well as the number and the names of the *metric dimensions*. An identifier uniquely identifies a metric within the respective scope. You can think of this as loosely corresponding to a "class instance id" of a metric.

Please let me know if this answers your question.
If you require adjustments to these APIs, please request *precisely* the changes you would like to see so that I can make them.
Thank you.

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
macrogreg: > MetricConfiguration namespace looks empty - only two methods?

I am not sure what you mean by this comment. Please let me make some educated guesses about what you mean and try to provide as much information as possible.

* There is no namespace `MetricConfiguration`. There is a type with this name, but it has more than two members. I have described the usage and the purpose of that class in great details just a bit above in the response to your comment about `MetricIdentifier`.

* There is also an identifier with a similar name `MetricConfigurations`. Is is also not a namespace, but a static class. Its usage and purpose is described in the doc comments and in examples 1, 2, 3 and 5:
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/feature/PreAggMetrics.Master/Test/Microsoft.ApplicationInsights.Test/Shared/Metrics/MetricsExamples.cs

Please let me provide additional details here.
`MetricConfigurations` represents an easily discoverable (via Intellisense) point for pre-defined metric configurations:

The common `TelemetryClient.GetMetric(..)` method has several overloads expecting a parameter of type `MetricConfiguration`. When attempting to type this term, the user discovers the static instance `MetricConfigurations.Common` (in the same namespace). We use extension methods to define common metric configurations:
In this version, the base SDK defines `MetricConfigurations.Common.Measurement()`.
The *Metrics.Extensions* package is planned to add:
- `MetricConfigurations.Common.Accumulator()`
- `MetricConfigurations.Common.Gauge()`
- `MetricConfigurations.Common.NaiveDistinctCount()`

You have seen the complete ready and tested code for this additional configurations and you requested that we move it to a different repo. It is currently available in the SDK labs Repo.
Users may also add their own common configurations via the same extension method mechanism.

Please let me know if this answers your question.
If you require adjustments to these APIs, please request *precisely* the changes you would like to see so that I can make them.
Thank you.

Issue #686, opened 2017-12-27
StopOperation reports telemetry that is not in current scope


lmolkova:
see https://github.com/Microsoft/ApplicationInsights-dotnet/issues/644#issuecomment-353765515.

```C#
var firstOperation = telemetryClient.StartOperation<DependencyTelemetry>("task 1");
var firstTask = RunMyTaskAsync();

var secondOperation = telemetryClient.StartOperation<DependencyTelemetry>("task 2");
var secondTask = RunMyTaskAsync();

await firstTask;

// Must not be tracked, but it is
telemetryClient.StopOperation(firstOperation); 

await secondTask;
```

>  tried running the above example with firstOperation and secondOperation, and for me telemetry for task1 is still logged (after logging an error "Operation to stop does not match the current operation"). Looking at the code for OperationHolder.Dispose() it seems like the return statement on line 86 only returns from ActivityExtensions.TryRun(), but not from Dispose(). isActivityAvailable is true, so the second return is never called either, and the operation gets logged successfully. 

> Even if telemetry is reported, some of its properties may be initialized from the `System.Diagnostics.Activity` (.NET tracing context that is used by AppInsights) during `Track()` by SDK or custom TelemetryIntitializer. As the `Activity.Current` is no longer correct and represents another operation, so reported telemetry under some conditions will be incorrect.

MS-TimothyMothra:
This is affecting some customers;

> AI (Internal): Operation to stop does not match the current operation.


- #693

- [Unexpected Trace “AI (Internal): Operation to stop does not match the current operation”](https://social.msdn.microsoft.com/Forums/en-US/130c9038-8a08-4b8c-b9cf-7ab6b6caf4ea/unexpected-trace-ai-internal-operation-to-stop-does-not-match-the-current-operation?forum=ApplicationInsights)


Do we have a work around to provide customers?
@Dmitry-Matveev has suggested disabling [Diagnostics Module](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModule.cs).

lmolkova:
@MS-TimothyMothra 
the only known reason for it is when StartOperation/StopOperation are called incorrectly from the user code (e.g. see example in the issue description).
[...]

PR #736, opened 2018-03-22
Fix 

lmolkova:
Fix Issue https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/864 .

Original issue suggests that extra `TelemetryClient.Initialize` call could be removed from the `StartOperation`.
It does not seem to be possible without potentially significant refactoring in `RichPayloadEventSource` and its consumers as it relies on the presence of correlation ids and ikey in the telemetry when it's started.

We also have [documentation and examples](https://docs.microsoft.com/en-us/azure/application-insights/application-insights-custom-operations-tracking) that rely on the fact that correlation ids are set on the telemetery in StartOperation call. It helps users deal with Application Insights telemetery only without knowing anything about Activity.

TL;DR;
While it could be possible to prevent double-initialization (or optimize it to essential properties), it is a potentially breaking change that requires some level of refactoring in the SDK and potentially some consumers. It is a risky change and we should consider doing it when we have strong reasons to.


src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -82,6 +82,15 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
                 throw new ArgumentNullException("operationTelemetry cannot be null.");
             }
 
+            // We intialize telemetry here AND in Track method because of RichPayloadEventSource.
```
pharring: typo: initialize

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -224,6 +233,17 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
             activity.Start();
             T operationTelemetry = ActivityToTelemetry<T>(activity);
 
+            // We intialize telemetry here AND in Track method because of RichPayloadEventSource.
```
pharring: typo: initialize

lmolkova:
Also fixes #686

PR #737, opened 2018-03-22
New Config Property: Override ApplicationInsight's Service Endpoint Address

MS-TimothyMothra:
**Fix Issue**: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/853

Currently, we do not offer users a friendly way to override both 
`dc.services.visualstudio.com/v2/track` and `dc.services.visualstudio.com/api/profiles`
Used by our classes `InMemoryTelemetryChannel`, `ServerTelemetryChannel`, and `CorrelationIdLookupHelper`.

Add new root level config property **"ApplicationInsightsEndpointBaseAddress"**
This new root level config will allow users to override the BASE ENDPOINT `dc.services.visualstudio.com`.
If this has been overridden, our classes will append their relative paths to this config value.
The theory is that users should not be expected to know every possible service call we could make and override those individually.

To protect backward compatibility, the `ITelemetryChannel.EndpointAddress` property will remain unchanged and take precedence over this root config value.

Order of precedence will be: Prop > Config > Default

To discourage 3rd party TelemetryChannels from using this property, this config property has been given an "ApplicationInsights" name.

These documents will need to be updated to address this new config property:

- [Microsoft Docs: "IP Addressed used by Application Insights"](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-ip-addresses)
- [Microsoft Docs: "Application Insights - Proxy"](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-troubleshoot-faq#proxy)

----------

### DIFF
--- a/PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
+++ b/PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
@@ -1,3 +1,5 @@
+Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration.ApplicationInsightsEndpointBaseAddress.set -> void
+Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration.GetApplicationInsightsEndpointBaseUri() -> System.Uri
 Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.DiagnosticsTelemetryModule.AddHeartbeatProperty(string propertyName, string propertyValue, bool isHealthy) -> bool
 Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.DiagnosticsTelemetryModule.ExcludedHeartbeatProperties.get -> System.Collections.Generic.IList<string>
 Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.DiagnosticsTelemetryModule.ExcludedHeartbeatPropertyProviders.get -> System.Collections.Generic.IList<string>
--- a/src/Microsoft.ApplicationInsights/Channel/InMemoryTransmitter.cs
+++ b/src/Microsoft.ApplicationInsights/Channel/InMemoryTransmitter.cs
@@ -19,6 +19,10 @@ namespace Microsoft.ApplicationInsights.Channel
     /// </summary>
     internal class InMemoryTransmitter : IDisposable
     {
+        private const string TelemetryServiceRelativeEndpoint = "v2/track";
+        private const string TelemetryServiceEndpoint = "https://dc.services.visualstudio.com/v2/track";
+        private readonly Uri defaultTelemetryServiceUri = new Uri(TelemetryServiceEndpoint);
+
         private readonly TelemetryBuffer buffer;
 
         /// <summary>
@@ -34,8 +38,7 @@ namespace Microsoft.ApplicationInsights.Channel
         private int disposeCount = 0;
 
         private TimeSpan sendingInterval = TimeSpan.FromSeconds(30);
-        private Uri endpointAddress = new Uri(Constants.TelemetryServiceEndpoint);
-                
+
         internal InMemoryTransmitter(TelemetryBuffer buffer)
         {
             this.buffer = buffer;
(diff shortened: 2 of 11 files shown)


src/Microsoft.ApplicationInsights/Channel/InMemoryTransmitter.cs
```diff
@@ -52,10 +55,25 @@ internal InMemoryTransmitter(TelemetryBuffer buffer)
...
-            set { Property.Set(ref this.endpointAddress, value); }
+            get
+            {
+                if (this.InMemoryChannelEndpointAddress != null)
+                {
+                    return this.InMemoryChannelEndpointAddress;
+                }
+                else if (TelemetryConfiguration.Active.GetApplicationInsightsEndpointBaseUri() != null)
+                {
+                    return new Uri(TelemetryConfiguration.Active.GetApplicationInsightsEndpointBaseUri(), TelemetryServiceRelativeEndpoint);
```
Dmitry-Matveev: Minor: This code may get executed with every Send() in certain situations. It does not sounds like an overhead in case of batching, but it may lead to an overhead if items are sent individually. It's OK to leave as is to keep it simple unless you see some easy solution (besides heavy change "subscriber" models) to address this.

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -130,6 +132,23 @@ public string InstrumentationKey
             }
         }
 
+        /// <summary>
+        /// Sets an override to "https://dc.services.visualstudio.com".
+        /// All of our internal services will construct the full endpoint address starting with this base address.
```
Dmitry-Matveev: Except for QuickPulse, right?

MS-TimothyMothra: Yes, except for QuickPulse. Is there a more formal name for this service?

src/Microsoft.ApplicationInsights/Channel/InMemoryTransmitter.cs
MS-TimothyMothra: Yeah, I didn't see a way around this pattern, given that we expect the Config may change during runtime.

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
```diff
@@ -1,3 +1,5 @@
+Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration.ApplicationInsightsEndpointBaseAddress.set -> void
```
SergeyKanzhelev: We planned to remove EndpointAddress from `ITelemetryChannel` interface. Not every channel has an endpoint - some are ETW or has more parameters than just URL. So I don't think it's a good idea to add a base address here.

`app-id` resolution is needed by dependency telemetry module. Can this property be defined there instead?

MS-TimothyMothra: No. CorrelationIdLookupHelper is used by more than just dependency telemetry module.
The intent is to simplify our user configuration story by moving this property to the root of the config to allow all of our Microsoft services to inherit from this single property.

SergeyKanzhelev: The problem with endpoint approach is that this approach doesn't work for non-http-based extraction of app-id. What if we want app-id to be set from config? Or read from registry key that set by Agent?

I don't like the idea of implementation detail for our specific service creep out into the base library.

I agree though it is a good idea to have some identity resolution mechanism to be defined in here. Is there an issue with the design discussion?

MS-TimothyMothra: We currently don't have an interface for app-id resolution.
I feel like that would be a separate discussion from this change (which I'm open to having :) )

I agree with the hesitation towards implementation specific details in the BaseSdk.
But I feel like this approach is the best compromise to the alternatives. 
Lets discuss this in more detail.

MS-TimothyMothra:
Abandoning in favor of a more extensible implementation.

Issue #739, opened 2018-03-27
[New] Interface for ApplicationId Lookup (IApplicationIdProvider)


MS-TimothyMothra:
Proposing a new Interface for ~Correlation~ Application ID resolution in the BaseSDK.

**Why**:
This came about while working to resolve this issue: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/853

The current [`CorrelationIdLookupHelper`](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/Common/CorrelationIdLookupHelper.cs) was not designed with configuring the endpoint in mind.

Additionally, multiple modules have private instances of the `CorrelationIdLookupHelper` which prevents it from caching lookup results and preventing duplicate requests.

**What:**
`TelemetryConfiguration` will get a root level property:  `ApplicationIdProvider`
```
public IApplicationIdProvider ApplicationIdProvider { get; set; }
```

The new interface is proposed as such:
```
public interface IApplicationIdProvider
{
    bool TryGetApplicationId(string instrumentationKey, out string applicationId);
}
```

**Assumptions:**
- An app will only need one provider, not a collection of providers.
- an ikey(s) will be needed to lookup an app id(s)
- the provider will take an ikey and resolve an appid and may store that formatted
- Breeze applicaton id format: "cid-v1:{app id}"

**Delivering:**

- the new interface: `IApplicationIdProvider`
- new `TelemetryConfiguration` property `ApplicationIdProvider`
    - This property will be `null` by default in the BaseSDK. Users will be able to opt-in.
    - This property will be set to an ApplicationInsights specific implementation when the WebSDK is installed.
- three implementations:
    - ~StaticCorrelationIdProvider will allow users to configure a single app id.~
    - `DictionaryApplicationIdProvider` will allow users to configure a collection of ikey/appid pairs
    - `ApplicationInsightsApplicationIdPovider` is a copy of the `CorrelationIdLookupHelper` from the WebSDK. This will be our provider that supports our Breeze endpoint.

MS-TimothyMothra:
I pushed my working branch and created a PR so you can review my changes in the context of this conversation: #740

SergeyKanzhelev:
Do you think we will need `TryGetCorrelationIdAsync` method instead of sync version `TryGetCorrelationId`

SergeyKanzhelev:
[...]

PR #740, opened 2018-03-27
New Interface IApplicationIdProvider

MS-TimothyMothra:

- [x] Design discussion issue #739

The PR will trigger build, unit test, and functional tests automatically. If your PR was submitted from a fork, mention one of the committers to initiate the build for you.


src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/ApplicationInsightsCorrelationIdProvider.cs
```diff
@@ -0,0 +1,130 @@
...
+    using System;
+    using System.Collections.Concurrent;
+    using System.Threading.Tasks;
+    using Extensibility;
+
+    /// <summary>
+    /// A store for instrumentation App Ids. This makes sure we don't query the public endpoint to find an app Id for the same instrumentation key more than once.
+    /// </summary>
+    /// <remarks>
+    /// This was formerly the CorrelationIdhLookupHelper defined in the WebSdk
```
SergeyKanzhelev: this remark is unnecessary

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/CorrelationIdHelper.cs
```diff
@@ -0,0 +1,70 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.CorrelationLookup
+{
+    using System.Globalization;
+    using System.Text.RegularExpressions;
+
+    internal static class CorrelationIdHelper
+    {
+        private const string CorrelationIdFormat = "cid-v1:{0}";
```
SergeyKanzhelev: `cid-v1` is a special string that describes that ID was taken from AI backend. I haven't looked thru the code. Make sure it is not hardcoded for static and dictionary providers. So we will enable other types of correlation ID to be defined going forward.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/CorrelationLookupEventSource.cs
```diff
@@ -0,0 +1,83 @@
...
+    using System.Diagnostics.Tracing;
+    using System.Linq;
+    using System.Text;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// ETW EventSource tracing class.
+    /// </summary>
+    //// [EventSource(Name = "Microsoft-ApplicationInsights-Extensibility-CorrelationLookup")] - EVERY COMPONENT SHOULD DEFINE IT"S OWN NAME
+    internal sealed class CorrelationLookupEventSource : EventSource
```
SergeyKanzhelev: You can move those events into CoreEventSource

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/DictionaryCorrelationIdProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Correlation Provider that holds a dictionary of ikeys to correlationIds.
+    /// </summary>
+    public class DictionaryCorrelationIdProvider : ICorrelationIdProvider
+    {
+        /// <summary>
+        /// Gets or sets a dictionary of Ikeys to CorrelationIds
```
SergeyKanzhelev: Call it `InstrumentaitonKey`, not `Ikeys` to avoid possible confusions

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/DictionaryCorrelationIdProvider.cs
```diff
@@ -0,0 +1,32 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.CorrelationLookup
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Correlation Provider that holds a dictionary of ikeys to correlationIds.
+    /// </summary>
+    public class DictionaryCorrelationIdProvider : ICorrelationIdProvider
```
SergeyKanzhelev: this is cool! It enables config-based definition of correlation ID mapping. Love it

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/FailedRequestsManager.cs
```diff
@@ -0,0 +1,124 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.CorrelationLookup
+{
+    using System;
+    using System.Collections.Concurrent;
+    using System.Diagnostics;
+    using System.Net;
+    using Extensibility.Implementation.Tracing;
+
+    internal class FailedRequestsManager
```
SergeyKanzhelev: Maybe this class can be moved to "Internal" folder

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/FailedRequestsManager.cs
```diff
@@ -0,0 +1,124 @@
...
+                    this.failingInstrumentationKeys.TryRemove(instrumentationKey, out FailedResult value);
+                    return true;
+                }
+                else
+                {
+                    return false;
+                }
+            }
+
+            Debug.WriteLine("no failed result found");
```
SergeyKanzhelev: we do not do ` Debug.WriteLine` in our SDK. If you need it - create an verbose EventSource message. If not - please remove

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/CorrelationIdHelper.cs
MS-TimothyMothra: Thank you for the clarification!

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/FailedRequestsManager.cs
MS-TimothyMothra: Thanks for the clarification.
I need it right now while i'm investigating a bug in one of the failing unit tests.
I'll be careful to remove all uses.

lmolkova:
We use correlation term for two completely different things in the SDK: multi-ikey support and transaction/operation correlation.

It introduces some level of confusion (not extremely high though) for our users. I wonder, since we are already in the massive refactoring, could we avoid using  'correlation' word in the multi-ikey feature to avoid this confusion? e.g. rename it to IAppIdLookupHelper?

MS-TimothyMothra:
@lmolkova feel free to join the design conversation here! #739 
I will defer to you and @SergeyKanzhelev to tell me what the correct vocabulary should be :)

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/CorrelationIdHelper.cs
MS-TimothyMothra: @SergeyKanzhelev if the DictionaryProvider should return an unformatted ApplicationId, should the Interface method be renamed to TryGetApplicationId ?

Counter argument is that our ApplicationInsightsProvider **WILL** return a formatted id so it may be inappropriate to name it an ApplicationId.

Do we need a different vocabulary here?

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/ApplicationInsightsCorrelationIdProvider.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// Unit Test Only! Initializes a new instance of the <see cref="ApplicationInsightsCorrelationIdProvider" /> class and accepts mocks for fetching Application Id.
+        /// </summary>
+        internal ApplicationInsightsCorrelationIdProvider(ProfileServiceWrapper profileServiceWrapper)
+        {
+            this.applicationIdProvider = profileServiceWrapper;
+        }
+
+        /// <summary>
+        /// Gets or sets the endpoint that is to be used to get the Application Insights resource's profile (Application Id etc.). 
+        /// Default value is "https://dc.services.visualstudio.com/api/profiles/{0}/appId". If this is overwritten, MUST include the '{0}' for string replacement!
```
MS-TimothyMothra: @SergeyKanzhelev How do you feel about this?
The alternative would be to ask users to overwrite the base "https://dc.services.visualstudio.com" and then we append the relative URI ourselves.

Second question: if we keep this should we add an EventSource log if a user fails to include the string replacement token?

MS-TimothyMothra:
Remaining Todos:

- Investigate Async Interface method
- Investigate chaining Providers
- Investigate rename (CorrelationId -> ApplicationId)
- Decision about overwrite ProfileQueryEndpoint

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/CorrelationIdHelper.cs
SergeyKanzhelev: I think entire string `cid-v1:XXXXXXXX` is an application ID. It's like person identity can be established by `ssn:XXX` and `driver-lic-id:XXXX`.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/CorrelationLookup/ApplicationInsightsCorrelationIdProvider.cs
SergeyKanzhelev: I think override is a rare scenario. If user sophisticated enough to do it - replacement token should be specified. I wonder would named replacement tokens work here? Like https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated So it's clear that {0} is instrumentation key. Not sure though how it might work...

MS-TimothyMothra: Interpolation wouldn't work here because the string needs a reference to a specific object. It gets compiled as a `String.Format()`. (Note that in those documented examples the object already exists where the string is defined).

I'll leave it as is and make sure it's well documented. thanks!

src/Microsoft.ApplicationInsights/Extensibility/Implementation/ApplicationId/ApplicationIdHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+            if (string.IsNullOrWhiteSpace(input))
+            {
+                return input;
+            }
+
+            // US-ASCII characters (hex: 0x00 - 0x7F) (decimal: 0-127)
+            // ASCII Extended characters (hex: 0x80 - 0xFF) (decimal: 0-255) (NOT ALLOWED)
+            // Non-Printable ASCII characters are (hex: 0x00 - 0x1F) (decimal: 0-31) (NOT ALLOWED)
+            // Printable ASCII characters are (hex: 0x20 - 0xFF) (decimal: 32-255) 
+            return Regex.Replace(input, @"[^\u0020-\u007F]", string.Empty);
```
pharring: 👎 for using a Regex here -- considering that we expect replacement to be very rare because the input is almost certainly a GUID.

MS-TimothyMothra: that addition came from a bug fix. (i'll try to find the link)
if you have suggestions for an alternative I'd love to hear it but as of now that check is very necessary.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/ApplicationId/DictionaryApplicationIdProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// <summary>
+    /// Application Id Provider that holds a dictionary of Instrumentation Keys to Application Ids.
+    /// </summary>
+    public class DictionaryApplicationIdProvider : IApplicationIdProvider
+    {
+        /// <summary>
+        /// Gets or sets a dictionary of Instrumentation Keys to Application Ids
+        /// </summary>
+        public Dictionary<string, string> Defined { get; set; }
```
pharring: Consider typing this as `IReadOnlyDictionary<string,string>` which allows for more flexibility.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/ApplicationId/ApplicationIdHelper.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <summary>
+        /// special string which describes that ID was taken from Breeze
+        /// </summary>
+        private const string Format = "cid-v1:{0}";
+
+        /// <summary>
+        /// Max length of Application Id allowed in response from Breeze.
+        /// </summary>
+        private const int ApplicationIdMaxLengeth = 50;
+
```
SergeyKanzhelev: Length spelled incorrectly

src/Microsoft.ApplicationInsights/Extensibility/Implementation/ApplicationId/DictionaryApplicationIdProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// <returns>TRUE if Application Id was successfully retrieved, FALSE otherwise.</returns>
+        public bool TryGetApplicationId(string instrumentationKey, out string applicationId)
+        {
+            applicationId = null;
+
+            var found = this.Defined?.TryGetValue(instrumentationKey, out applicationId) ?? false;
+
+            if (!found)
+            {
+                found = this.Next?.TryGetApplicationId(instrumentationKey, out applicationId) ?? false;
```
SergeyKanzhelev: Is ?. Atomic? Ideally take it into local variable and only deal with it later. So nobody can change property in between the check and call

pharring: It is atomic in the sense that "this.Next" is evaluated exactly once. It is as if you had extracted to an unnamed local variable.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/ApplicationId/ApplicationIdHelper.cs
pharring: Frankly, even if we wanted to change from a GUID, I don't think we could. And even then, would we really use something that used non-printable characters?

I suggest:

Avoid the Regex unless you can't. i.e. First test to see if all chars are 'sane' and only sanitize if/when you find one.

If the rule is "any char outside the 20-7F range should be removed", then:
```c#
private static bool IsSane(char ch) => (uint)(ch - 0x20) <= (0x7F - 0x20);

private static string SanitizeString(string input)
{
    if (input == null)
    {
        return null;
    }

    foreach(var ch in input)
    {
        if (!IsSane(ch))
        {
            return Regex.Replace(input, @"[etc...
        }
    }

    return input;
}
```

src/Microsoft.ApplicationInsights/Extensibility/Implementation/ApplicationId/ProfileServiceWrapper.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public void Dispose()
+        {
+            this.httpClient.Dispose();
+        }
+
+        /// <summary>Send HttpRequest to get config id.</summary>
+        /// <remarks>This method is internal so it can be moq-ed in a unit test.</remarks>
+        internal virtual async Task<HttpResponseMessage> GetAsync(string instrumentationKey)
```
pharring: If you have an async method that has a simple "return await", then it should be transformed to a non-async, non-awaiting method. This avoids the async compiler-generated machinery (and the compiler doesn't do this for you).
i.e.
```C#
internal virtual Task<HttpResponseMessage> GetAsync(string instrumentationKey)
{
     Uri applicationIdEndpoint = this.GetApplicationIdEndPointUri(instrumentationKey.ToLowerInvariant());
     return this.httpClient.GetAsync(applicationIdEndpoint);
}
```

src/Microsoft.ApplicationInsights/Extensibility/Implementation/ApplicationId/ProfileServiceWrapper.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        /// <summary>
+        /// Strips off any relative path at the end of the base URI and then appends the known relative path to get the Application Id uri.
+        /// </summary>
+        /// <param name="instrumentationKey">AI resource's Instrumentation Key.</param>
+        /// <returns>Computed Uri.</returns>
+        private Uri GetApplicationIdEndPointUri(string instrumentationKey)
+        {
+            if (this.ProfileQueryEndpoint != null)
```
pharring: This method could be a one-liner using the `??` operator:
```C#
return new Uri(string.Format(CultureInfo.InvariantCulture, this.ProfileQueryEndpoint ?? Constants.ProfileQueryEndpoint, instrumentationKey));
```

src/Microsoft.ApplicationInsights/Extensibility/Implementation/ApplicationId/FailedRequestsManager.cs
```diff
@@ -0,0 +1,124 @@
...
+            /// </summary>
+            /// <param name="retryAfter">Time to wait before a retry</param>
+            /// <param name="httpStatusCode">Failure response code. Used to determine if we should retry requests.</param>
+            public FailedResult(TimeSpan retryAfter, HttpStatusCode httpStatusCode = HttpStatusCode.OK)
+            {
+                this.retryAfterTime = DateTime.UtcNow + retryAfter;
+
+                // Do not retry 4XX failures.
+                var failureCode = (int)httpStatusCode;
+                this.shouldRetry = !(failureCode >= 400 && failureCode < 500);
```
pharring: Ironically, 429 (too many requests) is usually a "retry later".

src/Microsoft.ApplicationInsights/Extensibility/Implementation/ApplicationId/ApplicationIdHelper.cs
MS-TimothyMothra: wow. good catch!

src/Microsoft.ApplicationInsights/Extensibility/Implementation/ApplicationId/ProfileServiceWrapper.cs
MS-TimothyMothra: Oh okay. Thanks for the tip!
Conventional wisdom was to sprinkle async/await everywhere and trust that the compiler would optimize. Can you recommend a source where I can learn more?

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/ApplicationId/ApplicationIdHelperTests.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        [TestMethod]
+        public void VerifyExtendedCharactersAreNotAllowed()
+        {
+            string testSet = new string(asciiExtendedCharacters.ToArray());
+
+            var actual = ApplicationIdHelper.SanitizeString(testSet);
+            Assert.AreEqual(string.Empty, actual);
+        }
+    }
```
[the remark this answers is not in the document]
MS-TimothyMothra: Thank you. Anything else? I'm ready to merge this :)

PR #715, opened 2018-02-28
Update schemas to have pageview.id

iusafaro:

PR #749, opened 2018-04-02
fix for DependencyTelemetry schema change

MS-TimothyMothra:
This amends a change introduced in #715 .

The DependencyTelemetry schema changed the base type of the Success property from a `Nullable<bool>` to a `bool`.
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/da86c509f487ae04d1621616b25ca50e9f0a601c/src/Microsoft.ApplicationInsights/Extensibility/Implementation/External/RemoteDependencyData_types.cs#L65

The `DependencyTelemetry` class made a change on how that base value was handled, wrapping it with an `isSet` flag.
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/da86c509f487ae04d1621616b25ca50e9f0a601c/src/Microsoft.ApplicationInsights/DataContracts/DependencyTelemetry.cs#L201-L211

This changed the effective default value of `DependenyTelemetry.Success` from `true` to `null`.
I am amending that change by adding `this.successFieldSet = true` to the constructor.
Doing this because the public api should not have been affected.

### DIFF
--- a/src/Microsoft.ApplicationInsights/DataContracts/DependencyTelemetry.cs
+++ b/src/Microsoft.ApplicationInsights/DataContracts/DependencyTelemetry.cs
@@ -30,6 +30,7 @@ namespace Microsoft.ApplicationInsights.DataContracts
         public DependencyTelemetry()
         {
             this.InternalData = new RemoteDependencyData();
+            this.successFieldSet = true;
             this.context = new TelemetryContext(this.InternalData.properties);
             this.GenerateId();
         }
--- a/Test/Microsoft.ApplicationInsights.Test/Shared/DataContracts/DependencyTelemetryTest.cs
+++ b/Test/Microsoft.ApplicationInsights.Test/Shared/DataContracts/DependencyTelemetryTest.cs
@@ -12,6 +12,17 @@
     [TestClass]
     public class DependencyTelemetryTest
     {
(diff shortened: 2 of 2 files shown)


src/Microsoft.ApplicationInsights/DataContracts/DependencyTelemetry.cs
```diff
@@ -30,6 +30,7 @@ public sealed class DependencyTelemetry : OperationTelemetry, ITelemetry, ISuppo
         public DependencyTelemetry()
         {
             this.InternalData = new RemoteDependencyData();
+            this.successFieldSet = true;
```
iusafaro: i thought we wanted to set it true if underlying value InternalData.success is set to true... this also works, but makes successFieldSet more meaningless...

MS-TimothyMothra: @Dmitry-Matveev and I shared a concern that if the public schema was changed to `false` by default, this would cause DependencyTelemetry to return `null` by default. Because the underlying type is not nullable, it seems that a decision was made that its base value was significant.

the `this.successFieldSet` still allows `Success` to be manually set to Null.

This is certainly open for further discussion.

Issue #755, opened 2018-04-06
(Minor Improvement) Guard TelemetryClient.Flush to not throw NullReference Exception


MichaConrad:
Hi,
today I ran into an odd behavior that on app shutdown of my ASP.NET Core service, one part of the application (the Serilog Sink) tries to `Flush` before it disposes.
But at that point in time, `this.configuration.TelemetryChannel` in `TelemetryClient` is already null.

And that causes `Flush` to throw a hard NullReference Exception [Code line 589](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Microsoft.ApplicationInsights/TelemetryClient.cs#L589)

see https://github.com/serilog/serilog-sinks-applicationinsights/issues/46

Could you just elvis the expression or at least null check it?

Thanks,
Michael

MichaCo:
@SergeyKanzhelev sure, just added the PR

PR #757, opened 2018-04-07
#755. adding null check to TelemetryClient.Flush.

MichaCo:
See Issue #755.

I added the null check to `TelemetryClient`.
I found that `ApplicationInsightsTelemetryPipeline` actually catches the exception, that's obsolete now and the try/catch can be removed.
Added a small unit test, too.


Test/Microsoft.ApplicationInsights.Test/Shared/TelemetryClientTest.cs
```diff
@@ -34,7 +34,26 @@ public void IsEnabledReturnsTrueIfTelemetryTrackingIsEnabledInConfiguration()
...
+
+            try
+            {
+                client.Flush();
+            }
+            catch (Exception ex)
+            {
+                Assert.IsNull(ex, ex.ToString());
+            }
+        }
```
MS-TimothyMothra: Should this test fail if an exception is not thrown?

Test/Microsoft.ApplicationInsights.Test/Shared/TelemetryClientTest.cs
```diff
@@ -34,7 +34,26 @@ public void IsEnabledReturnsTrueIfTelemetryTrackingIsEnabledInConfiguration()
...
+
+            configuration.Dispose();
+
+            try
+            {
+                client.Flush();
+            }
+            catch (Exception ex)
+            {
+                Assert.IsNull(ex, ex.ToString());
```
MS-TimothyMothra: nit: I think `Assert.Fail(ex.ToString())` is more appropriate here.

SergeyKanzhelev:
@MichaCo something fishy happening with build automation. Can you please push a change to PR, perhaps fixing what Timothy suggested in assert statement. So build will be restarted and merge unblocked. Thanks!

MS-TimothyMothra:
I merged Develop into his branch, which should have triggered a rebuild. Looks like that is still failing.

MichaCo:
@SergeyKanzhelev @MS-TimothyMothra I pushed the assert change, did that work with the build? ;) 
I don't see any status change

MS-TimothyMothra:
It looks like the build definition for external contributions is broken. :(
If we can't get this fixed in a couple of days, I can pull your branch and submit it as my own. 
Your changes look good, thankyou!

MichaCo:
ok thats fine, thanks ;)

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -586,7 +586,10 @@ public void TrackRequest(RequestTelemetry request)
         /// </remarks>
         public void Flush()
         {
-            this.configuration.TelemetryChannel.Flush();
+            if (this.configuration.TelemetryChannel != null)
```
macrogreg: I think, TelemetryClient is not s read-only property (please correct me if I am wrong).
If so, this is not a safe statement.
Please change to:

```
ITelemetryChannel telemetryChannel = this.configuration?.TelemetryChannel;
if (telemetryChannel != null)
{
    telemetryChannel.Flush();
}
```

Also - are you 100% sure that this is why the error was thrown and that id did not come out of the channel's Flush method?

macrogreg:
@SergeyKanzhelev this change touches the new metrics code. While it is certainly not necessary to consult me about such changes in the long term, I would love to be proactively notified and review during the first month (or two). This will give me c chance to learn about any issues or improvement opportunities people may find. If you notice more PRs that affect those files, could you please add me as a reviewer? Thanks!

(Same for @Dmitry-Matveev and @MS-TimothyMothra who I am sure review a lot of code :) )

src/Microsoft.ApplicationInsights/TelemetryClient.cs
macrogreg: Sorry, typo: ... _TelemetryClient_ is not _a_ read-only property... :)

MichaCo: Ok, agree with the change.

Re error  we got a NullReference exception on app shutdown (see linked issue in the [serilog repository](https://github.com/serilog/serilog-sinks-applicationinsights/issues/46)).
linke to [screenshot](https://user-images.githubusercontent.com/5837539/38416729-fa270aac-3995-11e8-8ddf-d4ef658d1377.png).
When I drilled into it, the channel was clearly null.
That happens when the configuration disposes, it sets it's channel property to null.

macrogreg: Thanks. 

> Also - are you 100% sure that this is why the error was thrown and that id did not come out of the channel's Flush method?

I totally believe you that the error was due to the reason you describe and that the change fixed that reason. What I am saying is that there could also be another race on dispose inside the channel, and so even if your null check succeeds, telemetryChannel.Flush() might still throw from within Flush. It may also be a non-issue, just want to make sure that you thought of it.

MichaCo: @macrogreg only looked at the `InMemoryChannel`, `transmitter` is read-only, shouldn't be null. And that `Flush` looks pretty guarded (try/catch at the end etc...).
Anyways, if we find more conditions, we can fix them later, too ;)
Letting NullReferences through to the consumer is never a good thing

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -586,7 +586,11 @@ public void TrackRequest(RequestTelemetry request)
         /// </remarks>
         public void Flush()
         {
-            this.configuration.TelemetryChannel.Flush();
+            var telemetryChannel = this.configuration?.TelemetryChannel;
```
macrogreg: Not sure what the convention of the SDK is, so you may dismiss this comment.
I was thinking, "var" should only be used in long LINQ statements or when the type is obvious from the same line, which is not the case here.

SergeyKanzhelev: Everything that not enforced is fine :-). We are using `var` widely in our repositories.

macrogreg: Ok

cijothomas:
Please add a one line in Changelog.md (https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/CHANGELOG.md) with link to the original issue. This helps us generate release notes easily.
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/755

MichaCo:
@cijothomas done

cijothomas:
Ready to merge, manually triggered CI build/tests and are passing.

Issue #759, opened 2018-04-10
Add Namespace to Metrics.


macrogreg:
Application Insights custom metrics are currently stored in a backend repository optimized for storing structured events from where they are made available for various user experiences in the Azure Portal (Ibiza), in the Analytics Portal and via the DRAFT APIs. These backends identify a metric within the scope of a resource by a 1-tuple [Metric Name].
In the future, custom metrics will be primarily stored in the Geneva pipeline (a backend repository that is optimized for storing timeseries) and made available via the Azure Monitoring APIs [[https://docs.microsoft.com/en-us/rest/api/monitor/metrics/list](https://docs.microsoft.com/en-us/rest/api/monitor/metrics/list)]. Those backends identify a metric within the scope of a resource by a 2-tuple [Metric Namespace, Metric Name]. In order to support these new backends we need to add _namespace_ as an additional attribute to Metric Telemetry Items as a sibling to _name_ wherever _name_ appears. Also, need to add API overloads that create metric items while specifying _namespace_.
 
Specifically:
 
1. Add an attribute `metricNamespace` to the bond spec of DataPoint next to `name`.
(We call it `metricNamespace` rather than `namespace` to avoid clashed with the frequently used keyword “namespace”)
 
1. Add a property `MetricNamespace` to the `MetricTelemetry` class.
 
1. For each ctor of `MetricTelemetry` class that includes a `name` parameter, add an overload that also expects a parameter `metricNamespace`.
 
1. Add an Obsolete Attribute to each ctor of the `MetricTelemetry` class that includes a `name` parameter, but _not_ a `metricNamespace` parameter.
 
1. Add / update unit tests according to the above changes.
 
Thank you, all, for your feedback about these upcoming changes!

Dmitry-Matveev:
`metricNamespace` is an optional attribute in the schema based on the PR above, will it be also optional in the new API to create MetricTelemetry? 

In case it's optional, it should not be an issue to mark old methods as obsolete unless behavior will change as well even with empty namespace. This will simplify initial migration. I'll let @SergeyKanzhelev to jump in on this if he disagrees :).

In case it's not optional (or maybe, in any case) we'd also need a good (but short) guidance for those unfamiliar with namespaces in metric telemetry. My guess is that larger part of the existing SDK customers simply used name with SDK metrics for a while without need or wonder about namespaces. Some good IntelliSense hints may help here.

Overall, looks good to me.

[...]

PR #780, opened 2018-04-17
Add Namespace to Metrics

macrogreg:
PR for changes proposed, discussed and reviewed in https://github.com/Microsoft/ApplicationInsights-dotnet/issues/759

### DIFF
--- a/src/Microsoft.ApplicationInsights/Metrics/MetricIdentifier.cs
+++ b/src/Microsoft.ApplicationInsights/Metrics/MetricIdentifier.cs
@@ -13,13 +13,15 @@
         /// <summary>@ToDo: Complete documentation before stable release. {369}</summary>
         public const int MaxDimensionsCount = 10;
 
+        private const string NoNamespaceIdentifierStringComponent = "<NoNamespace>";
+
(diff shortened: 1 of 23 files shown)


src/Microsoft.ApplicationInsights/Metrics/MetricIdentifier.cs
```diff
@@ -33,23 +35,33 @@ public static string DefaultMetricNamespace
 
             set
             {
-                ValidateLiteral(value, nameof(value));
+                ValidateLiteral(value, nameof(value), allowEmpty: true);
```
SergeyKanzhelev: I haven't noticed this class before. Basically it is a global singleton that allows to override the default namespace for the process?

It may not work well for aspnetcore apps which uses DI for per-app configuration. And multiple apps may be configured in a single process. Perhaps this setting can be moved to `telemetryconfiguration`?

macrogreg: Good point. Let's have a brief chat so that I can better understand why statics on Telemetry Config are better than here, and then I can get this moved accordingly.

PR #788, opened 2018-04-26
Add OperationDetails object bag to DependencyTelemtry for Initializers

michaelwstark:
Fix Issue # .
[Consolidated]
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/900

[Original]
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/820
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/747
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/587

- [X] Design discussion issue #
https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/587
- [X] Changes in public surface reviewed
- [X] CHANGELOG.md updated with one line description of the fix, and a link to the original issue.
- [X] The PR will trigger build, unit test, and functional tests automatically. If your PR was submitted from fork - mention one of committers to initiate the build for you.


michaelwstark:
FWIW.  I am not in love with the name of the property -- "Data" was already taken, and putting this on a "Data" field on OperationContext would have made this a much more significant change.  Open to suggestions (obviously).

pharring:
Consider making the Dictionary instantiation lazy to minimize the allocation penalty for code that doesn't need/use it. Also, it should probably be a ConcurrentDictionary (which I'm reluctant to point out because it is an allocation hog)

src/Microsoft.ApplicationInsights/DataContracts/DependencyTelemetry.cs
```diff
@@ -122,6 +125,14 @@ public override string Id
             set { this.InternalData.id = value; }
         }
 
+        /// <summary>
+        /// Gets the dependency operation details, if any.
+        /// </summary>
+        public IDictionary<string, object> OperationDetails
```
lmolkova: I'm concerned with two things here

1. confusion - we have too many things already.  I imagine users would not understand what info goes here and is it sent to AI backend.
2. abuse - users should not use this dict to pass something unrelated

One possible way to address at least first could be to
1. Make operation detail private field
2. Add getter and setter functions 
3. mark setter function as not browsable (like [TelemetryClient.Track](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Microsoft.ApplicationInsights/TelemetryClient.cs#L432))
4. keep getter public and browsable and give a detailed documentation summary of what it is.

Another way 
1. keep `OperationDetails` public, but not browsable
2. Add extension method in the DependencyCollector package like `TryGetValue(string key)`

Even more strict way could be make an enum with possible keys in the DependenycCollector package and allow to get by enum.

src/Microsoft.ApplicationInsights/DataContracts/DependencyTelemetry.cs
```diff
@@ -19,6 +22,7 @@ public sealed class DependencyTelemetry : OperationTelemetry, ITelemetry, ISuppo
 
         internal readonly RemoteDependencyData InternalData;
         private readonly TelemetryContext context;
+        private readonly Lazy<IDictionary<string, object>> operationDetails;
```
pharring: For compactness (fewer object allocations), I prefer not to use `Lazy<T>`.
Instead, make the field non-readonly `IDictionary<string, object>` and use `System.Threading.LazyInitializer.EnsureInitialized`.

The pattern is this:
```C#
using static System.Threading.LazyInitializer;

private IDictionary<string, object> operationDetails;

public IDictionary<string, object> OperationDetails => EnsureInitialized(ref operationDetails, () => new ConcurrentDictionary<string, object>());
```

src/Microsoft.ApplicationInsights/DataContracts/DependencyTelemetry.cs
```diff
@@ -281,6 +305,29 @@ public override ITelemetry DeepClone()
...
+        /// <summary>
+        /// In specific collectors, objects are added to the dependency telemetry which may be useful
+        /// to enhance DependencyTelemetry telemetry by <see cref="ITelemetryInitializer" /> implementations.
+        /// </summary>
+        /// <param name="key">The key of the value to get.</param>
+        /// <param name="detail">When this method returns, contains the object that has the specified key, or the default value of the type if the operation failed.</param>
+        /// <returns>true if the key was found; otherwise, false.</returns>
+        public bool TryGetOperationDetail(string key, out object detail)
+        {
+            return this.OperationDetails.TryGetValue(key, out detail);
```
pharring: Short-circuit if operationDetails (the backing field) has not been initialized (avoid creating an empty ConcurrentDictionary here).

CHANGELOG.md
```diff
@@ -3,8 +3,8 @@
 This changelog will be used to generate documentation on [release notes page](http://azure.microsoft.com/en-us/documentation/articles/app-insights-release-notes-dotnet/).
 
 ## Version 2.6.0
- -[Fix: changed namespace SamplingPercentageEstimatorSettings](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/727)
-
+- [Fix: changed namespace SamplingPercentageEstimatorSettings](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/727)
+- [NEW: Added TryGetOperationDetail to DependencyTelemetry to facilitate advanced ITelemetryInitializer scenarios] (https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/587)
```
cijothomas: this was meant for 2.6.0 stable or 2.7.0-beta1? @MS-TimothyMothra can help.

cijothomas: It'd be great to explicitly mention that these new fields are NOT meant to be serialized and sent to AI backend, they only exist for facilitate TelemetryInitializers.

michaelwstark: Let me know if the release the note is under needs to change -- I just put it under the latest one assuming that all pending changes should be put there.  If you'd rather not release something in a stable release that hasn't been in a beta release I understand.  I'll update the release note and the comment on the set method to explicitly state that the data is not serialized.

src/Microsoft.ApplicationInsights/DataContracts/DependencyTelemetry.cs
```diff
@@ -35,11 +40,11 @@ public DependencyTelemetry()
...
-        /// <summary>
-        /// Initializes a new instance of the <see cref="DependencyTelemetry"/> class with the given <paramref name="dependencyName"/>, <paramref name="data"/>,
-        /// <paramref name="startTime"/>, <paramref name="duration"/> and <paramref name="success"/> property values.
-        /// </summary>
-        [Obsolete("Use other constructors which allows to define dependency call with all the properties.")]
+    /// <summary>
+    /// Initializes a new instance of the <see cref="DependencyTelemetry"/> class with the given <paramref name="dependencyName"/>, <paramref name="data"/>,
+    /// <paramref name="startTime"/>, <paramref name="duration"/> and <paramref name="success"/> property values.
+    /// </summary>
+    [Obsolete("Use other constructors which allows to define dependency call with all the properties.")]
```
pharring: Lost some indenting here (lines 43-47). Please fix.

CHANGELOG.md
cijothomas: I think this changes would go into 2.7.0-beta1.
We had already shipped the last beta of 2.6.0 so only critical bugs would be released into 2.6.0-stable. So this change should be in beta1 of the next release, which is 2.7.0-beta1

lmolkova:
See comment from the Web SDK here: 

https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/897#issuecomment-385578860

------------------
The concern is that the feature would postpone GC for the requests/responses since reference would be kept in dependency telemetry until it is sent to the backend (that happens asynchronously).

I wonder, can we clean up the OperationDetails in TrackDependency call after initialize is called to address this concern?

Here

```csharp
        public void TrackDependency(DependencyTelemetry telemetry)
        {
            if (telemetry == null)
            {
                telemetry = new DependencyTelemetry();
            }

            this.Track(telemetry);

!!!         telemetry.OperationDetails.Clear();
        }
```

It totally makes sense to me, as Track happens syncronously with dependency call and request/response should not be accessed after that.

michaelwstark:
Seems reasonable.  I'll have to add some additional work to the test verification in the other repo, but shouldn't be to hard to grab these items prior to them being cleared out.

michaelwstark:
I expected my tests for the dependency collection package to start failing, and when they did not, I investigated and found TrackDependency() is never called -- Track() is called directly.  Is there a reason TrackDependency is not being called?

michaelwstark:
> Ok so the alternative could be to clean OperationDetail in the Web SDK (in dependency collector) where we assign them after the track call...

@lmolkova - I actually prefer the original suggestion (seems like less maintenance over time), and had already updated the web sdk repo to call TrackDependency(), I was just verifying that there was not a reason for this.  I noticed that Track() itself is not browsable.

cijothomas:
Manually triggered CI build/tests - all good and ready to merge

cijothomas:
@MS-TimothyMothra - Please merge this when branch is open for 2.7.0-beta1

MikeLarah:
Is there any way to retrieve the request body for HTTP dependency calls using the OperationDetails?

Following the links posted earlier, I find that I can retrieve the `HttpWebRequest` in a telemetry initializer like so:
```
if (!(telemetry is DependencyTelemetry dependency))
{
    return;
}

if (dependency.TryGetOperationDetail("HttpRequest", out var request)
    && request is HttpWebRequest httpWebRequest)
{
    if (httpWebRequest.Method == "POST" || httpWebRequest.Method == "PUT")
    {
        // get request body somehow?
        dependency.Properties.Add("RequestBody", body);
    }
}
```

But the stream on the request is closed at this point, so unable to be read.

Alternatively, is there any other way to capture the request body such that it is attached to the dependency telemetry?

lmolkova:
@MikeLarah no, there is no way to read closed stream. 
You may log the request body at the place where it is created and written to the request stream, you may log it with TelemetryClient.TrackTrace(). It is problematic to make it a child of the dependency, by default, it will be the child of the request

PR #790, opened 2018-04-27
nuget compliance

MS-TimothyMothra:
Updating some of the values for our packages.
Upgrade to Microbuild 0.3.0 to enable package signing.

Note: Due to a bug in Microbuild, we have to temporarily disable symbols package.


src/Microsoft.ApplicationInsights/Microsoft.ApplicationInsights.csproj
```diff
@@ -28,8 +28,8 @@
     <!--Copyright defined above-->
     <!--Copyright></Copyright-->
     <RequireLicenseAcceptance>true</RequireLicenseAcceptance>
-    <PackageLicenseUrl>http://go.microsoft.com/fwlink/?LinkID=510709</PackageLicenseUrl>
-    <PackageProjectUrl>http://go.microsoft.com/fwlink/?LinkId=392727</PackageProjectUrl>
+    <PackageLicenseUrl>https://go.microsoft.com/fwlink/?LinkID=510709</PackageLicenseUrl>
```
SergeyKanzhelev: I wonder should `owner` property refer to AppInsightsSDK as well now? Or Microsoft is sufficient?

MS-TimothyMothra: Author MUST be Microsoft only.
For Owner, it is recommended but not required that we use Microsoft and also the Nuget owner, AppInsightsSDK.

@SergeyKanzhelev Since this wasn't required, I was saving it for later but if you agree, I can add it now. :)

Issue #806, opened 2018-05-17
Move metrics extraction to the new APIs and remove MetricManager V1 & Co


macrogreg:
We have recent shipped new metric aggregation APIs.
However, standard metric extraction uses old internal APIs.
We need to 
a) Change the extractor to use the new APIs
b) Remove the old APIs.

This will not affect any public APIs or any public configuration options. However, customers who used reflection to access V1 MetricManager APIs will be broken.

MS-TimothyMothra:
Just to confirm, will V2 be public?
Do we need to publish any guidance for customers to migrate from V1 to V2?

macrogreg:
Yes, V2 is public, V1 is internal.
There is not guidance to migrate since V1 was internal. The extractors using V1 are public but transitioning them to V2 should not have any publicly observable effect, except a small performance improvement.

Dmitry-Matveev:
@macrogreg , I believe we can close this issue as it was successfully implemented?

PR #813, opened 2018-05-22
 Allow setting different valuesPerDimensionLimits for different dimensions in metric configuration

macrogreg:
This is 1st of 2 PRs addressing issue https://github.com/Microsoft/ApplicationInsights-dotnet/issues/806.

This PR adds some missing public overloads required to address the issue.
Specifically:
 - We now allow setting different dimension limits for different dimensions. This was done by V1 extractors.
 - We allow looking up metrics by their ID, not only by reference.


macrogreg:
Failing test "SamplingPercentageAdjustsAccordingToConstantHighProductionRate" is unrelated.

src/Microsoft.ApplicationInsights/Metric.cs
```diff
@@ -1200,17 +1200,20 @@ public bool TryTrackValue(object metricValue, string dimension1Value, string dim
...
 
-            if (isMultidimensional && configuration.ValuesPerDimensionLimit < 1)
+            for (int d = 0; d < dimensionsCount; d++)
             {
-                throw new ArgumentException("Multidimensional metrics must allow at least one dimension-value per dimesion"
-                                          + Invariant($" (but {configuration.ValuesPerDimensionLimit} was specified")
-                                          + Invariant($" in {nameof(configuration)}.{nameof(configuration.ValuesPerDimensionLimit)})."));
+                if (configuration.GetValuesPerDimensionLimit(d + 1) < 1)
+                {
+                    throw new ArgumentException("Multidimensional metrics must allow at least one dimension-value per dimesion"
```
SergeyKanzhelev: again - it will be great if you can find a way to not throw from the public API call. Maybe error message and fall back logic.

src/Microsoft.ApplicationInsights/Metrics/MetricsCollection.cs
```diff
@@ -63,6 +63,17 @@ public bool IsReadOnly
             return metric;
         }
 
+        /// <summary>@ToDo: Complete documentation before stable release. {799}</summary>
```
SergeyKanzhelev: Create a task for these so @MS-TimothyMothra can track it for the specific milestone. Also - please ensure documentation exists for `beta` as well, not only stable

macrogreg: Done:
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/817

src/Microsoft.ApplicationInsights/Metric.cs
macrogreg: I remember the prev conversation, just did not get to it yet. Note - the diff shows this as addition, but the code was there before.
It will be easier at some places and harder at others. It is not always possible to replace nonsensical parameters with reasonable defaults.

macrogreg:
I agree with @Dmitry-Matveev 's comments above. But I prefer not to over-complicate this. In 90% of use cases, if not more, a single valuesPerDimensionLimit will be used. And that is what I am optimizing for. This change is to support some rare cases like our default metrics extraction, where a configuration indeed relates to a specific metric.

PR #814, opened 2018-05-23
(WIP) Separate property bag of TelemetryContext and ISupportProperties

cijothomas:
To get early feedback only. This may be treated as breaking change and require Major version update.

Summary:
All the current Telemetry types defined in base sdk (eg: `RequestTelemetry`, `DependencyTelemetry`) implements `ISupportProperties`, which gives them a Properties bag (string,string). `ITelemetry` also has `TelemetryContext` which also has a Property bag. Currently these two property bags are internally referring to the same collection. ( so modifying one affects the other.)

This PR changes the implementation so that the `Properties` from `ISupportProperties` and `TelemetryContext` are now independent collection in each of the concrete `ITelemetry` implementations. Though not a breaking change, this has the effect that, if any customer depended on these two property collection being the same, they'll have undesired consequences.

Also, this does not change serialization logic for now. Both the property bags are serialized into the "properties" of the telemetry. But this may eventually change so as to 'promote' the properties from `TelemetryContext` to be outside the item itself.


lmolkova:
* I believe we have mixed usage of different properties in web SDK and I assume everywhere else. Is there any plan to ensure all our SDKs use the right thing?

* What is the motivation behind this change?
* It is actually confusing already and luckily they are the same and it does not matter, but if they become different things, I'd propose them to have different names

pharring:
I have the same question as Liudmila. Why?
Doesn't separating them makes it more confusing for consumers (and the rest of the SDK)?
i.e. Which one should I use? What if you write the same property with different values to each collection? Which one wins?

SergeyKanzhelev:
@cijothomas it would be great if you'll move discussion to the issue. PR are not as well searcheable.

cijothomas:
All - Sorry for not describing more context. Will open an issue with more details today :)

cijothomas:
@lmolkova @pharring @SergeyKanzhelev  Lets use #816 for higher level discussion, and the PR for code-comments.

cijothomas:
Abandoning in favor of #820 as it lets us do this without breaking customers.

PR #824, opened 2018-06-01
Metrics: Autocollected Metric Extraction now uses new aggregation APIs

macrogreg:
This is PR number 2 that addresses https://github.com/Microsoft/ApplicationInsights-dotnet/issues/806

One last PR is coming to remove the no longer used internal APIs.

### DIFF
--- a/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Metrics/DependencyMetricsExtractor.cs
+++ b/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Metrics/DependencyMetricsExtractor.cs
@@ -1,18 +1,17 @@
 ﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Metrics
 {
     using System;
-    using System.Collections.Concurrent;
-    using System.Collections.Generic;
-    using System.Runtime.ExceptionServices;
 
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.DataContracts;
-    using Microsoft.ApplicationInsights.Extensibility.Implementation;
-    using Microsoft.ApplicationInsights.Extensibility.Implementation.Metrics;
+    using Microsoft.ApplicationInsights.Metrics;
+    using Microsoft.ApplicationInsights.Metrics.Extensibility;
+    using static System.FormattableString;
 
     /// <summary>
     /// An instance of this class is contained within the <see cref="AutocollectedMetricsExtractor"/> telemetry processor.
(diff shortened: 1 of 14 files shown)


src/Microsoft.ApplicationInsights/Extensibility/Implementation/Metrics/DependencyMetricsExtractor.cs
```diff
@@ -88,19 +97,20 @@ public int MaxDependencyTypesToDiscover
...
         }
 
         /// <summary>
-        /// Initializes the internal metrics trackers based on settings.
+        /// Pre-initialize this extractor.
         /// </summary>
-        /// <param name="metricManager">The <c>MetricManager</c> to be used for creating and sending the metrics by this extractor.</param>
-        public void InitializeExtractor(MetricManagerV1 metricManager)
+        /// <param name="metricTelemetryClient">The <c>TelemetryClient</c> to be used for sending extracted metrics.</param>
+        public void InitializeExtractor(TelemetryClient metricTelemetryClient)
```
Dmitry-Matveev: This looks like a public API contract change from `void init(MetricManager)` to `void init(TelemetryClient)` - is this a breaking change or this class was hidden in 2.6?

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Metrics/RequestMetricsExtractor.cs
```diff
@@ -1,46 +1,48 @@
...
         {
         }
 
         public string ExtractorName { get; } = "Requests";
 
-        public string ExtractorVersion { get; } = "1.0";
+        public string ExtractorVersion { get; } = "1.1";
 
-        public void InitializeExtractor(MetricManagerV1 metricManager)
+        public void InitializeExtractor(TelemetryClient metricTelemetryClient)
```
Dmitry-Matveev: Same question here.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Metrics/DependencyMetricsExtractor.cs
macrogreg: The class is internal, so no breaking change. Internal classes have public methods if the method represents the contract of the class, but the class is not intended for public consumption.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Metrics/RequestMetricsExtractor.cs
macrogreg: No breaking change. Same answer. :)

PR #830, opened 2018-06-06
Fix NullReferenceException in telemetry sink after it was disposed

lmolkova:
Fix Issue https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/928.

Check if TelemterySink was disposed and write user actionable event source error.


src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -512,6 +512,16 @@ public void ApplicationIdProviderFetchApplicationIdFailedWithResponseCode(string
             this.WriteEvent(42, httpStatusCode, this.nameProvider.Name);
         }
 
+        [Event(
+            43,
+            Keywords = Keywords.UserActionable,
```
Dmitry-Matveev: It's marked as user actionable, does it mean user can alter the behavior by changing their approach to disposing somehow?

Dmitry-Matveev: How many of those do we expect? User Actionable error level telemetry is sent to the customer's ikey and should not significantly increase quota consumption.

lmolkova: > It's marked as user actionable, does it mean user can alter the behavior by changing their approach to disposing somehow?

There are many options here:
- Users who create/dispose TelemteryConfiguration manually could make sure they do it after all telemetry has been reported: i.e. configuration should be last to be disposed.
- AspNetCore dependency injection users have to shut down their services gracefully: i.e, finish all requests processing and dispose DI container after that.

In some cases it's not possible to prevent some telemetry from being reported after configuration is disposed because of async nature of dependency calls.

I can remove user actionable, but I think it makes sense in many cases.

Issue #498, opened 2017-03-31
Support for providing ExceptionTelemetry without actual/original exception object?


sopelt:
Hello, I was wondering if AI could provide support for tracking exceptions (creating an exception telemetry/exception details) without having the actual exception object instance. When using AI in different layers of an app (asp.net (core), JS ...) and adding a small IoT component it would be nice to unify the logs in a central system (AI).
This could work by serializing traces and exceptions, sending them to a blob/webapp/function/... and moving them to AI from there.
As ExceptionTelemetry requires an exception one could implement an exception type that just provides the serialized message and stacktrace but that loses the exception type.

Having a proper API for this use case (without exposing ExceptionDetails?) could be interesting if other people face this use-case as well.

SergeyKanzhelev:
Thanks for your proposal. We will likely implement it in the version after the next

dnelly:
I have a similar issue

The need:
My app emits exceptions and I want to have the exceptions recorded in Application Insights along with all attributes like inner exception objects and stack traces. However, I need to censor some of the values in the exceptions without dropping the attribute altogether (just change the value to something like XX-XX-XX).
 
The attempt:
I implemented a telemetry processor that applies the following pattern.
1.	Identify that the telemetry item is an ExceptionTelemetry item
2.	Serialize the ExceptionTelemetry to JSON
3.	Mutate the relevant values (without dropping the attributes altogether)
4.	Deserialize JSON back to an instance of an ExceptionTelemetry object and overwrite the original item
 
Once again, no attribute is dropped from the exception.
 
Strange result:
The mutated exceptions make it to Application Insights and he can also see them in Application Analytics, but very strangely, the mutated exceptions no longer have a message and stack trace attributes in the inner exception. If I disable the custom telemetry processor then the full exception makes it to AI without losing any attributes. Also, serialization/deserialization of the object works fine. This can also be observed from the output window in Visual Studio when in a F5 scenario.

dnelly:
Any updates on this?

SergeyKanzhelev:
@kunalspathak we are open for contributions. Otherwise I cannot give an ETA for the fix. The only workaround for now is to convert `ExceptionTelemetry` into `TraceTelemetry` by serializing exception into text manually.

kunalspathak:
[...]

PR #865, opened 2018-07-18
Add ExceptionDetailsInfo class

kunalspathak:
Fix Issue #498 .
Adds a class `ExceptionDetailsInfo` which acts as a wrapper for internal `ExceptionDetails` and is used to update the `TypeName` and `Message` fields of internal `ExceptionDetails` objects.

- [X] Design discussion issue #498 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This changelog will be used to generate documentation on [release notes page](ht
   Related: (https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/690)
 - Added method `ExceptionDetailsInfoList` on `ExceptionTelemetry` class that gives control to user to update exception
 message and exception type of underlying `System.Exception` object that user wants to send to telemetry. Related discussion is [here](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/498).
+- Added an option of creating ExceptionTelemetry object off of custom exception information rather than a System.Exception object.
 
 ## Version 2.7.0-beta2
 - [Fix: NullReferenceException if telemtery is tracked after TelemetryConfiguration is disposed](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/928)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,15 +2,16 @@
 
 This changelog will be used to generate documentation on [release notes page](http://azure.microsoft.com/documentation/articles/app-insights-release-notes-dotnet/).
 
+## Version 2.8.0-beta1
+- Added method `ExceptionDetailsInfoList` on `ExceptionTelemetry` class that gives control to user to update exception
+message and exception type of underlying `System.Exception` object that user wants to send to telemetry. Related discussion is [here](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/498).
+- Added an option of creating ExceptionTelemetry object off of custom exception information rather than a System.Exception object.
 
 ## Version 2.7.0-beta3
 - [Allow to set flags on event](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/844). It will be used in conjunction with the feature that will allow to keep IP addresses.
 - [Fix: SerializationException resolving Activity in cross app-domain calls](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/613)
 - [Make HttpClient instance static to avoid re-creating with every transmission. This had caused connection/memory leaks in .net core 2.1](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/594)
   Related: (https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/690)
-- Added method `ExceptionDetailsInfoList` on `ExceptionTelemetry` class that gives control to user to update exception
-message and exception type of underlying `System.Exception` object that user wants to send to telemetry. Related discussion is [here](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/498).
-- Added an option of creating ExceptionTelemetry object off of custom exception information rather than a System.Exception object.
 
 ## Version 2.7.0-beta2
 - [Fix: NullReferenceException if telemtery is tracked after TelemetryConfiguration is disposed](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/928)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,10 @@ This changelog will be used to generate documentation on [release notes page](ht
 ## Version 2.7.0-beta3
 - [Allow to set flags on event](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/844). It will be used in conjunction with the feature that will allow to keep IP addresses.
 - [Fix: SerializationException resolving Activity in cross app-domain calls](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/613)
-- [Make HttpClient instance static to avoid re-creating with every transmission. This had caused connection/memory leaks in .net core 2.1] (https://github.com/Microsoft/ApplicationInsights-dotnet/issues/594)
+- [Make HttpClient instance static to avoid re-creating with every transmission. This had caused connection/memory leaks in .net core 2.1](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/594)
   Related: (https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/690)
+- Added method `ExceptionDetailsInfoList` on `ExceptionTelemetry` class that gives control to user to update exception
+message and exception type of underlying `System.Exception` object that user wants to send to telemetry. Related discussion is [here](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/498).
 
 ## Version 2.7.0-beta2
 - [Fix: NullReferenceException if telemtery is tracked after TelemetryConfiguration is disposed](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/928)
--- a/PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Shipped.txt
+++ b/PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Shipped.txt
@@ -99,6 +99,11 @@ Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt
 Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt.Platform = 2 -> Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt
 Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt.Unhandled = 0 -> Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt
 Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt.UserCode = 1 -> Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt
+Microsoft.ApplicationInsights.DataContracts.ExceptionDetailsInfo
+Microsoft.ApplicationInsights.DataContracts.ExceptionDetailsInfo.TypeName.get -> string
+Microsoft.ApplicationInsights.DataContracts.ExceptionDetailsInfo.TypeName.set -> void
+Microsoft.ApplicationInsights.DataContracts.ExceptionDetailsInfo.Message.get -> string
+Microsoft.ApplicationInsights.DataContracts.ExceptionDetailsInfo.Message.set -> void
 Microsoft.ApplicationInsights.DataContracts.ExceptionTelemetry
 Microsoft.ApplicationInsights.DataContracts.ExceptionTelemetry.Context.get -> Microsoft.ApplicationInsights.DataContracts.TelemetryContext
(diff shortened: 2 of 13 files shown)


PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Shipped.txt
```diff
@@ -108,6 +113,7 @@ Microsoft.ApplicationInsights.DataContracts.ExceptionTelemetry.ExceptionTelemetr
 Microsoft.ApplicationInsights.DataContracts.ExceptionTelemetry.ExceptionTelemetry(System.Exception exception) -> void
 Microsoft.ApplicationInsights.DataContracts.ExceptionTelemetry.HandledAt.get -> Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt
 Microsoft.ApplicationInsights.DataContracts.ExceptionTelemetry.HandledAt.set -> void
+Microsoft.ApplicationInsights.DataContracts.ExceptionTelemetry.ExceptionDetailsInfoList.get -> System.Collections.Generic.IReadOnlyList<Microsoft.ApplicationInsights.DataContracts.ExceptionDetailsInfo>
```
SergeyKanzhelev: I'd suggest to make it a `list` so customer can add new exception details and remove ones they are not interested in

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Shipped.txt
```diff
@@ -99,6 +99,11 @@ Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt
 Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt.Platform = 2 -> Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt
 Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt.Unhandled = 0 -> Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt
 Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt.UserCode = 1 -> Microsoft.ApplicationInsights.DataContracts.ExceptionHandledAt
+Microsoft.ApplicationInsights.DataContracts.ExceptionDetailsInfo
```
SergeyKanzhelev: Should this collection define it's own collection inside to indicate parent-child relationship of exceptions?

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Shipped.txt
SergeyKanzhelev: Consider referencing only the first ExceptionDeails object that will have it's own children collection

kunalspathak: I thought about that, but intentionally didn't add for few reasons:
1. If user wants to add a new `ExceptionDetailsInfo` object, how would he/she create the internal `ExceptionDetails` object? Currently `ExceptionDetailsInfo` is purely designed to be invoked from within AI and not from outside.

2. `ExceptionDetails` hold the relation between parent-child exception objects using `id` and `outerId`. If we add or remove it, it will be tricky to reflect those changes on actual parent-child relation present in corresponding internal `ExceptionDetails` object. We don't want to mess up that relation by giving user too much control (at least thats what my thinking is).

Thoughts?

---
In reply to: [203859148](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/865#discussion_r203859148) [](ancestors = 203859148)

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Shipped.txt
kunalspathak: If we want to indicate parent-child relationship, I don't think there is any need to have own collection, but what we really want is to expose `id` and `outerId` fields of `ExceptionDetails`. Adding just `getters` for them doesn't add value, because user don't know how those ids were populated. If we add `setters` as well, then we will have to add extra logic to sync-up the relation reflected in `ExceptionDetailsInfo` into `ExceptionDetails` and vice-versa. I was not planning to do that (at least not in this PR).

 The problem that I was hoping to solve is if you have `System.Exception` and AI inner code flattens it, let us give opportunity to user to modify message name and type name before sending it over the wire.  @dnelly  and my scenario falls under this category. Alternatively, user can do it in his code, walks the exception objects tree and populate the flatten list of `ExceptionDetailsInfo` and just sets in `ExceptionTelemetry`. But I guess, AI already has that code, so opted for former option.

---
In reply to: [203866000](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/865#discussion_r203866000) [](ancestors = 203866000)

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Shipped.txt
kunalspathak: sorry, i didn't get what you are suggesting. Are you saying to have `IList<ExceptionDetails>` inside `ExceptionDetailsInfo` class? If yes, I didn't want it that way because then how would user change type and message of individual `ExceptionDetails`?

---
In reply to: [203866174](https://github.com/Microsoft/ApplicationInsights-dotnet/pull/865#discussion_r203866174) [](ancestors = 203866174)

src/Microsoft.ApplicationInsights/DataContracts/StackFrame.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace Microsoft.ApplicationInsights.DataContracts
+{
+    using System;
+
+    /// <summary>
+    /// Wrapper class for <see cref="Extensibility.Implementation.External.StackFrame"/> for API exposure.
+    /// </summary>
+    public sealed class StackFrame
```
kunalspathak: >StackFrame [](start = 24, length = 10)

is this class needed instead of using `System.Diagnostics.StackFrame`?

src/Microsoft.ApplicationInsights/DataContracts/ExceptionTelemetry.cs
```diff
@@ -286,32 +337,44 @@ private void ConvertExceptionTree(Exception exception, ExceptionDetails parentEx
             }
         }
 
-        private void UpdateExceptions(Exception exception)
+        private void UpdateData(Exception exception, ExceptionInfo exceptionInfo)
```
kunalspathak: >UpdateData [](start = 21, length = 10)

only one of the argument is valid for this method. Consider creating 2 separate methods `UpdateDataWithException()` and `UpdateDataWithExceptionInfo()` for this.

src/Microsoft.ApplicationInsights/DataContracts/ExceptionTelemetry.cs
```diff
@@ -286,32 +337,44 @@ private void ConvertExceptionTree(Exception exception, ExceptionDetails parentEx
...
-                        "The number of inner exceptions was {0} which is larger than {1}, the maximum number allowed during transmission. All but the first {1} have been dropped.",
-                        exceptions.Count,
-                        Constants.MaxExceptionCountToSave));
-
-                // remove all but the first N exceptions
-                exceptions.RemoveRange(Constants.MaxExceptionCountToSave, exceptions.Count - Constants.MaxExceptionCountToSave);
-
-                // we'll add our new exception and parent it to the root exception (first one in the list)
-                exceptions.Add(ExceptionConverter.ConvertToExceptionDetails(countExceededException, exceptions[0]));
+                this.Data = exceptionInfo ?? throw new ArgumentNullException(nameof(exceptionInfo));
```
kunalspathak: >Data [](start = 21, length = 4)

who sanitize the stack frames for this code path?

src/Microsoft.ApplicationInsights/DataContracts/StackFrame.cs
tokaplan: Yeah our StackFrame has some stuff that BCL's version doesn't, like "level" or "method"...

CHANGELOG.md
```diff
@@ -6,8 +6,11 @@ This changelog will be used to generate documentation on [release notes page](ht
 ## Version 2.7.0-beta3
 - [Allow to set flags on event](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/844). It will be used in conjunction with the feature that will allow to keep IP addresses.
 - [Fix: SerializationException resolving Activity in cross app-domain calls](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/613)
-- [Make HttpClient instance static to avoid re-creating with every transmission. This had caused connection/memory leaks in .net core 2.1] (https://github.com/Microsoft/ApplicationInsights-dotnet/issues/594)
+- [Make HttpClient instance static to avoid re-creating with every transmission. This had caused connection/memory leaks in .net core 2.1](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/594)
   Related: (https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/690)
+- Added method `ExceptionDetailsInfoList` on `ExceptionTelemetry` class that gives control to user to update exception
```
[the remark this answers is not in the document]
tokaplan: Got it.

src/Microsoft.ApplicationInsights/DataContracts/ExceptionTelemetry.cs
tokaplan: I don't want to sanitize it here. For my particular scenario, it's already sanitized. For a broader scenario where the customer is supplying custom data we don't want it to be changed in a pretty much random way (which is what sanitizing does right now). There is an argument to be made for excessive length, but I think it should be up to the customer to make that determination in this particular scenario.

src/Microsoft.ApplicationInsights/DataContracts/ExceptionTelemetry.cs
tokaplan: Good catch, thank you. It fell apart during refactoring and I never noticed :)

src/Microsoft.ApplicationInsights/DataContracts/ExceptionDetailsInfo.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// Wrapper class for <see cref="ExceptionDetails"/> that lets user gets/sets TypeName and Message.
+    /// </summary>
+    public sealed class ExceptionDetailsInfo
+    {
+        internal readonly ExceptionDetails InternalExceptionDetails = null;
+
+        /// <summary>
+        /// Constructs the instance of <see cref="ExceptionDetailsInfo"/>.
+        /// </summary>
+        public ExceptionDetailsInfo(int id, int outerId, string typeName, string message, bool hasFullStack,
```
SergeyKanzhelev: arguments descriptions please

src/Microsoft.ApplicationInsights/DataContracts/ExceptionInfo.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        internal ExceptionInfo(ExceptionData data)
+        {
+            this.data = data;
+        }
+
+        /// <summary>
+        /// Gets a list of <see cref="ExceptionDetailsInfo"/> to modify as needed.
+        /// </summary>
+        public IReadOnlyList<ExceptionDetailsInfo> ExceptionDetailsInfoList => this.data.exceptions.Select(ed => new ExceptionDetailsInfo(ed)).ToList().AsReadOnly();
```
SergeyKanzhelev: why do we need all these allocations on property get? Can we return raw list?

SergeyKanzhelev: why is it readonly? It is OK to allow people to remove unnecessary inner exceptions

SergeyKanzhelev:
I approved in assumption you will address my comments

src/Microsoft.ApplicationInsights/DataContracts/ExceptionDetailsInfo.cs
tokaplan: Done.

src/Microsoft.ApplicationInsights/DataContracts/ExceptionInfo.cs
tokaplan: The actual details that are being serialized are stored in a collection of ExceptionDetails, which is an internal type we don't want to expose here. So we have to wrap each element in ExceptionDetailsInfo (the external counterpart) before exposing. Hence the allocations.

tokaplan: See the comment above. Because we're wrapping objects into the publicly facing type, we'd have to jump through hoops in order to make this happen. Kunal's scenario doesn't require this, so I suggest until someone actually needs it we don't spend time here.

tokaplan:
Addressed comments (see above), merging.

src/Microsoft.ApplicationInsights/DataContracts/ExceptionInfo.cs
SergeyKanzhelev: I think this is an important scenario. It may save a lot of money to customer. Please create a separate issue to track it.

Issue #871, opened 2018-07-20
Strong typed extensibility for Telemetry types


cijothomas:
Application Insights today allows extensibility of existing Telemetry types via property bags of (string,string).
There is extensibility on the `TelemetryContext` as well via `Properties` and the newly introduced `GlobalProperties`. (https://github.com/Microsoft/ApplicationInsights-dotnet/issues/820)

While the above is sufficient for most use cases, there can be requirements to have strongly-typed extensibility to existing types. To support this, we are introducing the following changes.

`ITelemetry` (and hence all implementing Telemetry types like `RequestTelemetry`, `DependencyTelemetry` etc.) will define a new field, `Extension` of type `IExtension`. The interface `IExtension` defines the following methods.
`void Serialize(ISerializationWriter writer);`
`IExtension DeepClone();`

`ISerializationWriter` is another new interface which has methods enabling the author of `IExtension` implementations to emit serialization information about the new type. Its defined like follows. (Will be changing this api frequently during beta period based on feedback)

```
    public interface ISerializationWriter
    {
        void WriteProperty(string name, string value);        
        void WriteProperty(string name, double? value);
        void WriteProperty(string name, int? value);
        void WriteProperty(string name, bool? value);
        void WriteProperty(string name, TimeSpan? value);
        void WriteList(string name, IList<string> items);
        void WriteDictionary(string name, IDictionary<string, string> items);
        void WriteStartObject(string name);
        void WriteEndObject(string name);       
    }
```
These allow a user to extend any existing Telemetry types with new strongly-typed classes, provided the new class implements the interface `IExtension`. The channel which sends the data to backend, can supply its own `ISerializationWriter` to serialize the extension.

Please note that as of now, Application Insights back-end does not accept any new fields other than the current property bags. This strongly-typed extension is currently targeted for users who uses Application Insights SDK with custom non Application-Insights channel.

`ISerializationWriter ` based strongly-typed extensibility will be used for extending an existing type like `RequestTelemetry` and also to define completely new `ITelemetry` types. `ITelemetry` will be defining a new method  `void SerializeData(ISerializationWriter serializationWriter)`.

Users who define new telemetry type will have implement `ITelemetry` interface. The `SerializeData` method should be implemented by the author of this new type to help channels with serialization of the data from new class. Fields defined on `ITelemetry` itself, like `Extension`, `Context` etc. will be serialized by the channel itself. 
[...]

PR #875, opened 2018-07-26
Support for strongly-typed extension to ITelemetry

cijothomas:
Fix Issue #871 

### DIFF
--- a/src/Microsoft.ApplicationInsights/Extensibility/ISerializationWriter.cs
+++ b/src/Microsoft.ApplicationInsights/Extensibility/ISerializationWriter.cs
@@ -1,0 +1,29 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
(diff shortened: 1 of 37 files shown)


src/Microsoft.ApplicationInsights/Extensibility/ISerializationWriter.cs
```diff
@@ -0,0 +1,56 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// The interface for defining writers capable of serializing data into various formats.    
+    /// </summary>
+    public interface ISerializationWriter
```
Dmitry-Matveev: I assume we may want to support more methods in the future, did we currently scope the available methods down to the list agreed upon in some extensibility discussions? I had a question on long, List vs. Array and JSON serializer mapping, but those won't need to be answered if we simply expose the minimum in here. Is so, can/should we limit it even more, or it's already too few methods for the scenario?

PR #892, opened 2018-08-14
ITelemetry.SerializeData instead of Serialize

cijothomas:
Make ITelemetry implement SerializeData instead of Serialize, as only Data part need to be serialized. Rest of the content are known within ITelemetry itself and hence require no explicit serialization call.

So if user want to define a new type, implement ITelemetry. (which include SerializeData - to let channels know what fields are defined by the user. All channels should know how to serialzie the rest of the content as they are on `ITelemetry` itself.)

### DIFF
--- a/src/Microsoft.ApplicationInsights/Extensibility/Implementation/JsonSerializer.cs
+++ b/src/Microsoft.ApplicationInsights/Extensibility/Implementation/JsonSerializer.cs
@@ -235,7 +235,12 @@
 
                 SerializeHelper(telemetryItem, jsonSerializationWriter, availabilityTelemetry.BaseType, AvailabilityTelemetry.TelemetryName);
             }
-            
+            else
+            {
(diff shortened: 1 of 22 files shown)


src/Microsoft.ApplicationInsights/Extensibility/Implementation/JsonSerializer.cs
```diff
@@ -155,62 +155,176 @@ private static Stream CreateCompressedStream(Stream stream)
...
-        {                        
-                if (telemetryItem is EventTelemetry)
+        {
+            jsonSerializationWriter.WriteStartObject();
+
+            if (telemetryItem is EventTelemetry)
                 {
                     EventTelemetry eventTelemetry = telemetryItem as EventTelemetry;
                     Utils.CopyDictionary(telemetryItem.Context.GlobalProperties, eventTelemetry.Data.properties);
+
```
cijothomas: Will move duplicate code to a common method.

PR #918, opened 2018-09-15
Response code shouldn't be overwritten to 200 if not set

iusafaro:
Fix Issue # .
In our internal 'custom requests' tracking like for Service Bus implementation we don't set response code, but set success to false and as a result success is overwritten to true and response code set to 200. It shouldn't happen in case of StartOperation as well because response code doesn't apply there.

### DIFF
--- a/src/Microsoft.ApplicationInsights/DataContracts/RequestTelemetry.cs
+++ b/src/Microsoft.ApplicationInsights/DataContracts/RequestTelemetry.cs
@@ -256,15 +256,15 @@
             this.Data.id = Utils.PopulateRequiredStringValue(this.Data.id, "id", typeof(RequestTelemetry).FullName);
 
             // Required fields
-            if (string.IsNullOrEmpty(this.ResponseCode))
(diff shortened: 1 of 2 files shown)


src/Microsoft.ApplicationInsights/DataContracts/RequestTelemetry.cs
```diff
@@ -255,10 +255,14 @@ void ITelemetry.Sanitize()
...
 
-            // Required field
+            // Required fields
             if (string.IsNullOrEmpty(this.ResponseCode))
             {
-                this.ResponseCode = "200";
+                this.ResponseCode = string.Empty;
+            }
+
+            if (!this.Success.HasValue)
```
Dmitry-Matveev: This changes our current experience in UI where requests may now have <blank> value if we were not able to collect response code. However, it's correct to respect that success is set to false. Can we set request to 200 when success if true and set to success to true only if it was not set to false before?

PR #924, opened 2018-09-26
standardize nupkg attributes

MS-TimothyMothra:
copying the approach i'm using for WebSDK

### DIFF
--- a/Nupkg.props
+++ b/Nupkg.props
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- Including this file will generate both the *.nupkg and *.symbols.nupkg -->
+    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
+    <IncludeSymbols>True</IncludeSymbols>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- These are the common properties used when generating the nupkg -->
+    <!-- https://docs.microsoft.com/en-us/nuget/schema/msbuild-targets -->
+    <Company>Microsoft</Company>
+    <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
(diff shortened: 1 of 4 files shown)


Nupkg.props
```diff
@@ -0,0 +1,47 @@
...
+    <!-- <PackageVersion>Defined in GlobalStaticVersion.props</PackageVersion> -->
+    <Authors>Microsoft</Authors>
+    <Owners>Microsoft,AppInsightsSdk</Owners>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
+    <PackageLicenseUrl>https://go.microsoft.com/fwlink/?LinkID=510709</PackageLicenseUrl>
+    <PackageProjectUrl>https://go.microsoft.com/fwlink/?LinkId=392727</PackageProjectUrl>
+    <PackageIconUrl>http://appanacdn.blob.core.windows.net/cdn/icons/aic.png</PackageIconUrl>
+    <RepositoryUrl>https://github.com/Microsoft/ApplicationInsights-dotnet</RepositoryUrl>
+    <RepositoryType>Git</RepositoryType>
+    <PackageType>Dependency</PackageType>
```
SergeyKanzhelev: Good idea to define common Package tags here and then use this in project itself:

<PackageTags>$(PackageTags) newTag1 newTag2</PackageTags>

Nupkg.props
```diff
@@ -0,0 +1,47 @@
...
+    <PackageId>UNDEFINED</PackageId>
+    <Title>UNDEFINED</Title>
+    <Description>UNDEFINED</Description>
+    <PackageTags>UNDEFINED</PackageTags>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- Disable auto generation of package attributes. This resolves the 'Duplicate attribute' error. 
+         Common cause is a hidden AssemblyInfo.cs file in a project -->
+    <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
```
SergeyKanzhelev: we probably will need to get rid of those at some point... It is set to not clean up the custom code we created with props files before it was supported =)

PR #929, opened 2018-10-02
Metric Aggregation metadata is stored into a field instead of properties dictionary

### LINKED ISSUE #930: Store Metric Aggregation metadata into a field rather than properties dictionary to improve performance
AutoCollected MetricAggregator is configured as a TelemetryProcessor to be run _before_ Sampling processors, and every item processed by it is tagged with a custom property indicating the version of the Aggregator. Accessing the `Properties` collection is expensive as its implemented as a ConcurrentDictionary.
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Microsoft.ApplicationInsights/Metrics/Extensibility/AutocollectedMetricsExtraction/AutocollectedMetricsExtractor.cs#L216

Proposal here is to create a non-public field for DependencyTelemetry and RequestTelemetry to store  this meta data. So MetricAggregator will use this field, avoiding expensive Dictionary access. To avoid change in behaviour, Getter for Properties will copy the value from the internal field to the Properties Dictionary.  (during serialization properties are accessed and the field is copied over to the Dictionary, but this is happens only for those items which are sampled in)

## Version Info 
SDK Version :
.NET Version :	 						
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) :
OS :
Hosting Info (IIS/Azure WebApps/ etc) :

cijothomas:
Metric Aggregation metadata is stored into a field instead of properties dictionary, to improve performance.
When someone read the properties, the field value is copied to the properties, so change in behavior.

Since metric aggregator sits before Sampling, this avoids the Properties (ConcurrentDictionary) access until late in the pipeline (serialization). In effect, for items that are dropped after sampling, no Properties access.

Fix #930 

### DIFF
--- a/src/Microsoft.ApplicationInsights/Metrics/Extensibility/AutocollectedMetricsExtraction/AutocollectedMetricsExtractor.cs
+++ b/src/Microsoft.ApplicationInsights/Metrics/Extensibility/AutocollectedMetricsExtraction/AutocollectedMetricsExtractor.cs
@@ -191,29 +191,39 @@
         /// <param name="extractorInfo">The string to be added to the item's properties.</param>
         private static void AddExtractorInfo(ITelemetry item, string extractorInfo)
         {
-            var itemWithProperties = item as ISupportProperties;
-            if (itemWithProperties == null)
+            string extractionPipelineInfo = String.Empty;
+            if (item is RequestTelemetry)
             {
-                return;
-            }
-
-            string extractionPipelineInfo;            
-            bool hasPrevInfo = itemWithProperties.Properties.TryGetValue(MetricTerms.Extraction.ProcessedByExtractors.Moniker.Key, out extractionPipelineInfo);
-
-            if (false == hasPrevInfo)
-            {
-                extractionPipelineInfo = String.Empty;
-            }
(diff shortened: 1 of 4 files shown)


src/Microsoft.ApplicationInsights/Metrics/Extensibility/AutocollectedMetricsExtraction/AutocollectedMetricsExtractor.cs
```diff
@@ -190,30 +190,31 @@ private static string GetExtractorInfo(ISpecificAutocollectedMetricsExtractor ex
...
-                extractionPipelineInfo = String.Empty;
+                extractionPipelineInfo = extractionPipelineInfo + "; ";
             }
             else
             {
-                if (extractionPipelineInfo.Length > 0)
-                {
-                    extractionPipelineInfo = extractionPipelineInfo + "; ";
-                }
+                extractionPipelineInfo = String.Empty;
```
SergeyKanzhelev: is stylecop fine with `String` uppercased? I'm surprised

Issue #969, opened 2018-11-02
Fully convert ConcurrentDIctionary to lazy in all Telemetry 


cijothomas:
RequestTelemetry ctr() uses bond generated internal class which creates ConcurrentDictionary. It can be lazily constructed to improve performance. Similar to whats done here:
https://github.com/Microsoft/ApplicationInsights-dotnet/issues/934

Issue #965, opened 2018-10-30
Change Uri from string to Uri in autogenerated internal class to improve perf.


cijothomas:
RequestTelemetry has a public field Url of type Uri. This field is served by a string field leading to conversion between String and Uri throughout. This conversion need to happen only on serialization time. (i.e only on those items which are sampled in.) 

Similar to this: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/927

PR #970, opened 2018-11-02
Perf Improvements - RequestTelemetry

cijothomas:
Fix Issue #969 #965 
This removes the usage of bond generated internal class from RequestTelemetry. RequestTelemetry now servers all fields from within. During serialization, the RequestData class is generated for RichpayloadEventSource.

This led to removing URI to String conversions amd doing lazy instantiation of ConcurrentDictionary for the Properties field.

### DIFF
--- a/src/Microsoft.ApplicationInsights/DataContracts/RequestTelemetry.cs
+++ b/src/Microsoft.ApplicationInsights/DataContracts/RequestTelemetry.cs
@@ -31,7 +31,7 @@
         private IExtension extension;
         private double? samplingPercentage;
         private bool? success;
-        private IDictionary<string, double> measurements;
+        private IDictionary<string, double> measurementsValue;
 
         /// <summary>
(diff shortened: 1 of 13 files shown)


src/Microsoft.ApplicationInsights/DataContracts/RequestTelemetry.cs
```diff
@@ -244,6 +244,37 @@ internal string MetricExtractorInfo
             set;
         }
 
+        internal RequestData Data
+        {
+            get
+            {
+                return LazyInitializer.EnsureInitialized(ref this.dataPrivate,
```
SergeyKanzhelev: After initialized it might be requested again and will need to be updated. Unless you absolutely sure nobody will change any property between two calls to `Data`

SergeyKanzhelev: As a minimum leave a comment about this

PR #987, opened 2018-11-08
Mikp/perf optimizations

Mikhail-Pranovich:
Performance optimizations for various isolated places as identified by local profiling on a single core app.

* SamplingTelemetryProcessor: compare primitives first 
* multiple places: use compile-time nameof(), instead of runtime typeof().Name
* PlatformImplementation: GetMachineName - reduce usage of volatile-accessing LazyInitializers
* TelemetryClient: create TelemetryContext at construction, to remove unnecessary lazy initialization-related checks

### DIFF
--- a/src/Microsoft.ApplicationInsights/DataContracts/AvailabilityTelemetry.cs
+++ b/src/Microsoft.ApplicationInsights/DataContracts/AvailabilityTelemetry.cs
@@ -17,7 +17,7 @@
     {
         internal const string TelemetryName = "Availability";
 
-        internal readonly string BaseType = typeof(AvailabilityData).Name;
+        internal readonly string BaseType = nameof(AvailabilityData);
         internal readonly AvailabilityData Data;
         private readonly TelemetryContext context;
         private IExtension extension;
--- a/src/Microsoft.ApplicationInsights/DataContracts/AvailabilityTelemetry.cs
+++ b/src/Microsoft.ApplicationInsights/DataContracts/AvailabilityTelemetry.cs
@@ -17,7 +17,7 @@
     {
         internal const string TelemetryName = "Availability";
 
-        internal readonly string BaseType = nameof(AvailabilityData);
+        internal const string BaseType = nameof(AvailabilityData);
         internal readonly AvailabilityData Data;
         private readonly TelemetryContext context;
         private IExtension extension;
--- a/src/Microsoft.ApplicationInsights/TelemetryClient.cs
+++ b/src/Microsoft.ApplicationInsights/TelemetryClient.cs
@@ -24,7 +24,6 @@
     {
         private const string VersionPrefix = "dotnet:";
         private readonly TelemetryConfiguration configuration;
-        private TelemetryContext context;
         private string sdkVersion;
 
         /// <summary>
@@ -60,9 +59,11 @@
         /// </summary>
         public TelemetryContext Context
         {
-            get { return LazyInitializer.EnsureInitialized(ref this.context, () => new TelemetryContext()); }
-            internal set { this.context = value; }
+            get;
+            internal set;
         }
+
+        = new TelemetryContext();
 
         /// <summary>
         /// Gets or sets the default instrumentation key for all <see cref="ITelemetry"/> objects logged in this <see cref="TelemetryClient"/>.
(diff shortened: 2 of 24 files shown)


src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -60,10 +59,12 @@ public TelemetryClient(TelemetryConfiguration configuration)
         /// </summary>
         public TelemetryContext Context
         {
-            get { return LazyInitializer.EnsureInitialized(ref this.context, () => new TelemetryContext()); }
-            internal set { this.context = value; }
+            get;
```
SergeyKanzhelev: who will initialize it?

Mikhail-Pranovich: due to formatting rules, see few lines below.

src/Microsoft.ApplicationInsights/DataContracts/AvailabilityTelemetry.cs
```diff
@@ -17,7 +17,7 @@ public sealed class AvailabilityTelemetry : ITelemetry, ISupportProperties, ISup
     {
         internal const string TelemetryName = "Availability";
 
-        internal readonly string BaseType = typeof(AvailabilityData).Name;
+        internal readonly string BaseType = nameof(AvailabilityData);
```
pharring: Consider "const" instead of "readonly". Then there's no runtime cost.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -60,10 +59,12 @@ public TelemetryClient(TelemetryConfiguration configuration)
...
+            internal set;
         }
 
+        = new TelemetryContext();
```
pharring: I think you've traded one perf problem for another. From profiling, did you see lock contention or high CPU usage in the old Context getter?
For sure, you're now allocating a new TelemetryContext on each new TelemetryClient. There are some (anti-)patterns where new TelemetryClient instances are created frequently and I think this is what the LazyInitializer was trying to improve.

pharring:
What are the expected perf improvements from this set of changes?

1. I see some first-time cost improvements from replacing "typeof(T).Name" with nameof.
2. There's a short-circuiting change in the sampling test.
3. There's the replacement of LazyInitializer with non-interlocked initialization -- which is a win if you assume multi-threaded access is very rare and it's OK to potentially have multiple threads race to run the initializer.
4. The TelemetryContext property initializer doesn't look like an obvious win because of the guaranteed allocation -- but perhaps you've concluded that you can't do anything useful with TelemetryClient without also having a TelemetryContext.

Now, I haven't seen the data, but I do believe that most of the perf bottlenecks in the AI SDK come from allocations (and therefore induced GCs). I'm willing to believe that there may be some lock contention or CPU bottlenecks in some tight inner loops -- but I suspect these are **very** few (counted on the fingers of one hand).

src/Microsoft.ApplicationInsights/TelemetryClient.cs
Mikhail-Pranovich: The TelemetryContext is needed in most of the paths for TelemetryClient, thus it is always initialized. We are saving on additional checks and Volatile reads here: https://referencesource.microsoft.com/#mscorlib/system/threading/LazyInitializer.cs

Mikhail-Pranovich: Under what conditions TelemetryClients are created and what do they do that they don't need the TelemetryContext. Is the perf critical for such scenarios?

src/Microsoft.ApplicationInsights/DataContracts/AvailabilityTelemetry.cs
Mikhail-Pranovich: very valid feedback, now it can be a compile time const shared across instances.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
SergeyKanzhelev: @Mikhail-Pranovich we treat `TelemetryClient` as lightweight object that can be used like `(new TelemetryClient()).TrackEvent("Configuration was updated");`. We have quite a few customers who uses it this way.

It is not a use we have in default out of the box case.

So if this thread contention problem is critical for out of the box experience, it may be an OK tradeoff.

Mikhail-Pranovich: Those calls end up in Track, which access Context in case if enabled. If not-enabled high perf scenario is critical for those guys, we can advise on re-using telemetry client.

Mikhail-Pranovich:
> What are the expected perf improvements from this set of changes?
Reduction of overhead by ~8-11% for a single core app (9.6% -> 8.6% throughput penalty of using the SDK).
There will be more perf improvement changes coming.

Issue #988, opened 2018-11-09
Convert unknown telemetry into custom event


Dmitry-Matveev:
Currently, any telemetry item that implements `ITelemetry` but is not directly supported in the serialization pipeline is dropped.

The suggestion is to convert such telemetry item into `EventTelemetry` and serialize all values from this unknown telemetry item into the `Properties` of `Dictionary<string, string>` type.

This will allow delivery of custom telemetry items implementing ITelemetry interface and `public void SerializeData(ISerializationWriter serializationWriter)` method into Application Insights backend as Custom Events.

For insance, an event like this:
```
    public class DmitryEvent : ITelemetry
    {
        public DmitryEvent()
        {
            this.Timestamp = DateTimeOffset.Now;
            this.Context = new TelemetryContext();
        }

        public DateTimeOffset Timestamp { get; set; }

        public TelemetryContext Context { get; set; }

        public IExtension Extension { get; set; }

        public string Sequence { get; set; }

        public string DmitryField { get; set; }

        public ITelemetry DeepClone()
        {
            return new DmitryEvent() { DmitryField = this.DmitryField };
        }

        public void Sanitize() { }

        public void SerializeData(ISerializationWriter serializationWriter)
        {
            serializationWriter.WriteProperty("DmitryField", DmitryField);
        }
    }
```

Will be delivered as:
![image](https://user-images.githubusercontent.com/12158043/48235964-fcb17580-e374-11e8-9c5d-300a4c61cacb.png)

PR #989, opened 2018-11-09
Converting custom ITelemetry implementations into Event Telemetry with properties

Dmitry-Matveev:
Addressing enhancement #988.
Initial implementation of `DictionarySerializationWriter` to write a `Dictionary<string, string>` for all serialized fields in the given `ITelemetry`:
- Pending some perf optimizations;
- Note: most places with couple string concatenations are "+", more than couple is `Concat()`, no serial string concatenations to benefit from `StringBuilder`.
- Not sure on benefits of initializing `Stack` for each serialized item - something to think about...

- [ ] Design discussion issue #988


Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/IExtensionImplementations.cs
```diff
@@ -0,0 +1,83 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation
+{
+    using System;
+    using System.Linq;
+    using System.Collections.Generic;
+
+    public class MyTestExtension : IExtension
```
Dmitry-Matveev: Moved several existing similar test classes into one place..

src/Microsoft.ApplicationInsights/Extensibility/Implementation/DictionarySerializationWriter.cs
```diff
@@ -0,0 +1,327 @@
...
+
+            string key = this.GetKey(name);
+
+            if (!this.AccumulatedDictionary.ContainsKey(key))
+            {
+                this.AccumulatedDictionary[key] = value;
+            }
+        }
+
+        public void WriteProperty(string name, double? value)
```
Dmitry-Matveev: We may want to consider this as a metric, but there are several places and use-cases where it should be property but not metric... Storing it as Property for the time being.

MS-TimothyMothra:
Can we discuss this?
This might not be necessary with a simple modification to `ITelemetry`

Dmitry-Matveev:
Discussed offline with @MS-TimothyMothra, PR is still necessary :) :) :)

src/Microsoft.ApplicationInsights/Extensibility/Implementation/DictionarySerializationWriter.cs
```diff
@@ -0,0 +1,327 @@
...
+                    this.AccumulatedDictionary[key] = value.Value.ToString(CultureInfo.InvariantCulture);
+                }
+                else
+                {
+                    this.AccumulatedDictionary[key] = null;
+                }
+            }
+        }
+
+        public void WriteProperty(string name, int? value)
```
MS-TimothyMothra: I see that this method is copied several times based on the input parameter. could this accept a generic Nullable parameter?

Dmitry-Matveev: This is an implementation of `ISerializationWriter`, I would not be able to change this on implementing side. @cijothomas would know if we can change it on the Interface side. My expectation that we would not, it already requires a conditional compilation on `bool?` between .NET and .NET Core, so not all Nullables are treated the same.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/DictionarySerializationWriter.cs
```diff
@@ -0,0 +1,327 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+
+    internal class DictionarySerializationWriter : ISerializationWriter
```
cijothomas: comment the usage of this class for future devs :)

PR #991, opened 2018-11-09
Improve ITelemetry Serialization

MS-TimothyMothra:
Fix for #997
Alternative fix for #988.
~I think the unknown telemetry case could be eliminated.~

JsonSerializer is attempting several type casts.
- Because of the if statements, there are several failed casts before finding the correct matching type.

**Option 1**
~I propose to remove the serialization entirely by appending the `TelemetryName` and `BaseType` to the `ITelemetry` interface. This would completely eliminate any need for casting here.~
This isn't viable because ITelemetry is the foundation of the Common Schema and cannot take changes.

**Option 2**
If we don't want to add properties to `ITelemetry` we can instead introduce a new interface `ISerializableTelemetry`. With this there would be only one type cast (ITelemetry => ISerializableTelemetry).
This has the benefit of being applied only to our AI Telemetry objects and won't affect external types.


src/Microsoft.ApplicationInsights/Channel/IAiSerializeableTelemetry.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace Microsoft.ApplicationInsights.Channel
+{
+    using System.Collections.Generic;
+
+    public interface IAiSerializeableTelemetry
```
pharring: Spelling: Serializable does not have an 'e'
(Also change the filename)

MS-TimothyMothra: thanks

src/Microsoft.ApplicationInsights/Channel/IAiSerializableTelemetry.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace Microsoft.ApplicationInsights.Channel
+{
+    using System.Collections.Generic;
+
+    public interface IAiSerializableTelemetry
```
pharring: Shouldn't this be `internal`?

MS-TimothyMothra: Can we do that?
I thought it would be discouraged to have internal interfaces on public classes.

If we can do that, we absolutely should. :)

MS-TimothyMothra:
@SergeyKanzhelev Are you okay with this approach?

src/Microsoft.ApplicationInsights/Channel/IAiSerializableTelemetry.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Microsoft.ApplicationInsights.Channel
+{
+    /// <summary>
+    /// This interface defines Telemetry objects that are intended to be serialized for the Application Insights Breeze ingestion endpoint.
+    /// </summary>
```
cijothomas: Application Insights Ingestion Endpoint is good. remove breeze.

cijothomas:
@MS-TimothyMothra please do create a github issue with the problem description and tag it with perf. It'll be easy for discovery later :)

Issue #998, opened 2018-11-13
Improve RichPayloadEventSource


MS-TimothyMothra:
RichPayloadEventSource is attempting several type casts.
Because of the if statements, there are several failed casts before finding the correct matching type.

https://github.com/Microsoft/ApplicationInsights-dotnet/blob/52e32b5695f35caf8548eb4bd256a408dbb9bb72/src/Microsoft.ApplicationInsights/Extensibility/Implementation/RichPayloadEventSource.cs#L41-L64

This can be eliminated with an internal interface returning private properties.

MS-TimothyMothra:
I closed the PR but I don't want to close this issue.

My PR broke the EventSource and exposed a gap in our tests that we don't currently catch.
**Even if we don't attempt to improve RichPayloadEventSource we should try to improve test coverage.**

PR #999, opened 2018-11-13
[WIP] Improve RichPayloadEventSource

MS-TimothyMothra:
proposed changes for #998 

DO NOT MERGE
This PR causes problems that our unit tests currently do not catch

### DIFF
--- a/src/Microsoft.ApplicationInsights/DataContracts/DependencyTelemetry.cs
+++ b/src/Microsoft.ApplicationInsights/DataContracts/DependencyTelemetry.cs
@@ -4,6 +4,7 @@ namespace Microsoft.ApplicationInsights.DataContracts
     using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.ComponentModel;
+    using System.Diagnostics.Tracing;
     using System.Globalization;
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Extensibility;
@@ -16,7 +17,7 @@ namespace Microsoft.ApplicationInsights.DataContracts
     /// The class that represents information about the collected dependency.
     /// <a href="https://go.microsoft.com/fwlink/?linkid=839889">Learn more.</a>
     /// </summary>
-    public sealed class DependencyTelemetry : OperationTelemetry, ITelemetry, ISupportProperties, ISupportSampling, ISupportMetrics, IAiSerializableTelemetry
+    public sealed class DependencyTelemetry : OperationTelemetry, ITelemetry, ISupportProperties, ISupportSampling, ISupportMetrics, IAiSerializableTelemetry, ISupportRichPayloadEventSource
     {
         internal new const string TelemetryName = "RemoteDependency";
 
@@ -103,6 +104,12 @@ namespace Microsoft.ApplicationInsights.DataContracts
 
         /// <inheritdoc />
         string IAiSerializableTelemetry.BaseType => nameof(RemoteDependencyData);
+
+        EventKeywords ISupportRichPayloadEventSource.EventSourceKeyword => RichPayloadEventSource.Keywords.Dependencies;
+
+        object ISupportRichPayloadEventSource.Data => this.Data;
+
+        string ISupportRichPayloadEventSource.TelemetryName => TelemetryName;
 
         /// <summary>
         /// Gets or sets date and time when telemetry was recorded.
@@ -306,7 +313,7 @@ namespace Microsoft.ApplicationInsights.DataContracts
             get;
             set;
         }
-        
+
         /// <summary>
         /// Deeply clones a <see cref="DependencyTelemetry"/> object.
         /// </summary>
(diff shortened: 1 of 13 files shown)


src/Microsoft.ApplicationInsights/DataContracts/ISupportRichPayloadEventSource.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Microsoft.ApplicationInsights.DataContracts
+{
+    using System.Diagnostics.Tracing;
+
+    /// <summary>
+    ///
```
pharring: Fill in the summary text. Something like: "Marker interface to indicate that a telemetry item supports serializing to an event source."

pharring:
The change looks good to me, but we need to be pretty careful about changes to RichPayloadEventSource. The last round of changes caused a regression in AI Profiler due to a change in serialization of "Duration". In this PR, there's a subtle change in the WriteEvent method -- 'data' used to be strongly-typed, now it's just an object. I don't _think_ it'll be a problem, but I'm not 100% certain. 
I want to add some new unit tests in this repo to try and catch serialization changes -- I intended to do this already, but I've been busy with other things.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/RichPayloadEventSource.cs
```diff
@@ -40,209 +40,25 @@ public RichPayloadEventSource()
...
-                CopyGlobalPropertiesIfRequired(item, telemetryItem.Properties);
-                item.Sanitize();
-                this.WriteEvent(
-                    AvailabilityTelemetry.TelemetryName,
-                    telemetryItem.Context.InstrumentationKey,
-                    telemetryItem.Context.SanitizedTags,
-                    telemetryItem.Data,
-                    telemetryItem.Context.Flags,
-                    Keywords.Availability);
+                    eventName: RequestTelemetry.TelemetryName,
```
pharring: This should be `eventName: richPayloadTelemetryItem.TelemetryName`

cijothomas:
@pharring Please do add unit tests when you get a chance to ensure we dont break AI Profiler!

MS-TimothyMothra:
@pharring @cijothomas i can look into unit tests. I would like the confirmation that my changes don't break anything.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/RichPayloadEventSource.cs
MS-TimothyMothra: thank you

src/Microsoft.ApplicationInsights/DataContracts/DependencyTelemetry.cs
```diff
@@ -104,6 +105,12 @@ private DependencyTelemetry(DependencyTelemetry source)
         /// <inheritdoc />
         string IAiSerializableTelemetry.BaseType => nameof(RemoteDependencyData);
 
+        EventKeywords ISupportRichPayloadEventSource.EventSourceKeyword => RichPayloadEventSource.Keywords.Dependencies;
+
+        object ISupportRichPayloadEventSource.Data => this.Data;
```
cijothomas: This is not correct.. It should be this.InternalData

cijothomas: (in this case .Data is actually the field to store sqlcommand etc) name is confusing.

src/Microsoft.ApplicationInsights/DataContracts/ExceptionTelemetry.cs
```diff
@@ -100,6 +101,12 @@ private ExceptionTelemetry(ExceptionTelemetry source)
         /// <inheritdoc />
         string IAiSerializableTelemetry.BaseType => nameof(ExceptionData);
 
+        EventKeywords ISupportRichPayloadEventSource.EventSourceKeyword => RichPayloadEventSource.Keywords.Exceptions;
+
+        object ISupportRichPayloadEventSource.Data => this.Data;
```
cijothomas: this.Data.Data

alohaninja:
Trying to revive this thread - we are seeing intense performance issues (_2-5s per log request_) with `RichPayloadEventSource` - how can we turn this feature off or escalate a fix? This is a blocking call and causing significant performance delays for us using the serilog sink for AI. 

image

pharring:
@alohaninja are you sure you're interpreting that trace correctly? It looks like the time is spent in JIT compilation which should happen only once. The changes in this PR wouldn't, I think fix that.

For sure 2.5 seconds of JIT compilation time is high, but does it happen on _every_ request?

alohaninja:
@pharring - this issue does seem to appear with each request - not familiar with the action that's occurring when `RichPayloadEventSource` fires. Is it required to use AI? it does seem to be impacting multiple AI modules. 

image

PR #1006, opened 2018-11-16
Fix a typo in a method name

oco27:


oco27:
Any idea why check #20181116.3 failed ? I don't have access to the details.

PR #1007, opened 2018-11-16
Perf fix - return String.Empty for synthetic source, ip if they are Empty instead of returning nulls.

cijothomas:
This will help to modify the telemetry initializers for the same to check if the field is null or not, before doing the [expensive] work on populating the same. They will be modified to check for null (instead of IsNullOrEmpty), before doing the work.


cijothomas:
After more investigation, this was not necessary. because:
The synthetic telemetry initialzier wont be applicable for Dependencies.
For requests, we already fixed code sothat initializers are not called twice.

PR #1011, opened 2018-11-17
Revert "Perf fix - return String.Empty for synthetic source, ip if they are Empty instead of returning nulls."

cijothomas:
Reverts Microsoft/ApplicationInsights-dotnet#1007


cijothomas:
After more investigation, this was not necessary. because:
The synthetic telemetry initialzier wont be applicable for Dependencies.
For requests, we already fixed code sothat initializers are not called twice.

Issue #1012, opened 2018-11-18
How to remove System.Runtime.CompilerServices.TaskAwaiter from exception callstack?


mmaitre314:
More of a question or feature request than an issue. Is it possible to remove the lines with `System.Runtime.CompilerServices.TaskAwaiter.` in the exception callstacks? This is not 'show just my code' because I need functions outside of my codebase but it's to remove all the .NET async goo that makes callstack unreadable and seems to trigger callstack truncation in AppInsights.

If it's something I could implement on my end without an update to the AppInsight SDK, that would be fine too. I looked at the API surface and I'm not sure how to do that. Maybe calling `ExceptionTelemetry.SetParsedStack()`, but not sure.

Ex of callstacks currently logged:

```
Microsoft.Azure.Documents.RequestTimeoutException:
   at Microsoft.Azure.Documents.RntbdConnection+<RequestAsync>d__45.MoveNext (Microsoft.Azure.Documents.Client, Version=1.22.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35)
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089)
   at Microsoft.Azure.Documents.RntbdTransportClient+<InvokeStoreAsync>d__3.MoveNext (Microsoft.Azure.Documents.Client, Version=1.22.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35)
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089)
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd (mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089)
[...]

PR #1022, opened 2018-12-03
Extended SetParsedStack to handle stack traces of inner exceptions.

arturek:
Fix Issue Microsoft/ApplicationInsights-dotnet#1012.

The existing API allows to modify stack traces of logged exceptions in a very limited way. For example there's an option to replace the stack trace of the top level exception, but stack traces of inner exceptions cannot be modified at all. This simple change extends the existing SetParsedStack method to accept an additional parameter which specifies whether stack trace of the top most exception or it's selected inner exception should be replaced.

It would be nice to bring more of improvements described in dotnet/corefx#24627 but to do it a bigger change would be needed. Till then this modification should still help a lot.

### DIFF
--- a/src/Microsoft.ApplicationInsights/DataContracts/ExceptionTelemetry.cs
+++ b/src/Microsoft.ApplicationInsights/DataContracts/ExceptionTelemetry.cs
@@ -93,7 +93,7 @@
             {
                 this.exception = source.Exception;
             }
-            
+
             this.extension = source.extension?.DeepClone();
         }
 
@@ -303,7 +303,7 @@
         {
             if (this.Exceptions != null && this.Exceptions.Count > 0)
             {
-                if(exceptionLevel < 0)
+                if (exceptionLevel < 0)
                 {
                     throw new ArgumentOutOfRangeException(nameof(exceptionLevel), exceptionLevel, "The exceptionLevel value must not be negative.");
                 }
@@ -342,7 +342,7 @@
         /// </summary>
         public void SetParsedStack(System.Diagnostics.StackFrame[] frames)
         {
-            SetParsedStack(frames, 0);
+            this.SetParsedStack(frames, 0);
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/Microsoft.ApplicationInsights/DataContracts/ExceptionTelemetry.cs
```diff
@@ -293,16 +293,19 @@ public void SerializeData(ISerializationWriter serializationWriter)
...
-        public void SetParsedStack(System.Diagnostics.StackFrame[] frames)
+        /// <param name="frames">The new array of stack frames.</param>
+        /// <param name="exceptionLevel">Specifies which stack trace should be replaced.
+        /// 0 specifies top level exception, other value selects the respective inner exception.</param>
+        public void SetParsedStack(System.Diagnostics.StackFrame[] frames, int exceptionLevel)
         {
-            if (this.Exceptions != null && this.Exceptions.Count > 0)
+            if (this.Exceptions != null && this.Exceptions.Count > exceptionLevel)
             {
                 if (frames != null && frames.Length > 0)
```
mmaitre314: Check against `exceptionLevel` instead of 0? I am thinking of using that API and if I get the exception level wrong the code will crash instead of skip.

mmaitre314: Probably also ignore `exceptionLevel < 0`

SergeyKanzhelev:
CC: @tokaplan for review

lmolkova:
@tokaplan when you are back from holidays, could you please review?

lmolkova:
Also cc @cijothomas @MS-TimothyMothra

arturek:
@lmolkova I've made easy fixes and will add tests soon. The original SetParsedStack has no tests which made it slightly more difficult.
@tokaplan The scenario is described in the issue #1012 . Without this change only garbage from the top level exception can be removed. In case of inner exceptions AI still can show very long and hard to read stack traces. I was looking for other ways to do it but in all cases I either need to read internal data or write to them. Extending SetParsedStack seems to be the easiest way to fix the issue.

Issue #1026, opened 2018-12-06
Synchronous wrapping of asynchronous code


dennisbappert:
The AI codebase uses several synchronous wrappers for async code and exposes those APIs.
All flavors of Task.Wait, Task.WaitAll, GetAwaiter().GetResult(), .Result, etc.

## Repro Steps
1. Take a ASP.NET Core web application (no SynchronizationContext)
2. Use Application Insights
3. Put very high load on the application

##  Actual Behavior
Massive ThreadPool starvation until the application stops responding at all.

## Expected Behavior
No ThreadPool starvation.

## Version Info 
SDK Version : any
.NET Version : ASP.NET Core	 						
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) :
OS : any
Hosting Info (IIS/Azure WebApps/ etc) : any

macrogreg:
@dennisbappert : Does this repro without using the Nito.AsyncEx library?

dennisbappert:
Yes it does.

pharring:
Do we have a self-contained repro? One where the app runs successfully until Application Insights is added?
If so, could we get a dump of the process at the point of "massive ThreadPool starvation"? From that we should be able to see where the tasks are blocked.

PR #1028, opened 2018-12-08
Metrics: Use dedicated thread instead of the thread pool for the defa…

macrogreg:
This is in context of https://github.com/Microsoft/ApplicationInsights-dotnet/issues/1026.

Work in progress for now.

### DIFF
--- a/src/Microsoft.ApplicationInsights/Metrics/Implementation/DefaultAggregationPeriodCycle.cs
+++ b/src/Microsoft.ApplicationInsights/Metrics/Implementation/DefaultAggregationPeriodCycle.cs
@@ -16,26 +16,25 @@
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Naming Rules", "SA1310: C# Field must not contain an underscore", Justification = "By design: Structured name.")]
         private const int RunningState_Stopped = 2;
 
-        private readonly Action workerMethod;
-
         private readonly MetricAggregationManager aggregationManager;
         private readonly MetricManager metricManager;
 
+        private readonly TaskCompletionSource<bool> workerTaskCompletionControl;
+
         private int runningState;
-        private Task workerTask;
+        private Thread aggregationThread;
 
         public DefaultAggregationPeriodCycle(MetricAggregationManager aggregationManager, MetricManager metricManager)
         {
             Util.ValidateNotNull(aggregationManager, nameof(aggregationManager));
             Util.ValidateNotNull(metricManager, nameof(metricManager));
 
-            this.workerMethod = this.Run;
-
             this.aggregationManager = aggregationManager;
             this.metricManager = metricManager;
 
             this.runningState = RunningState_NotStarted;
-            this.workerTask = null;
+            this.workerTaskCompletionControl = new TaskCompletionSource<bool>();
+            this.aggregationThread = null;
         }
 
         ~DefaultAggregationPeriodCycle()
@@ -52,21 +51,28 @@
                 return false; // Was already running or stopped.
             }
 
-            this.workerTask = Task.Run(this.workerMethod)
-                              .ContinueWith(
-                                        (t) => { this.workerTask = null; },
-                                        TaskContinuationOptions.ExecuteSynchronously);
+            // We create a new thread rather than using the thread pool.
+            // This is because inside of the main loop in the Run() method we use a synchronous wait.
+            // The reason for that is to prevent aggregation from being affected by potential thread pool starvation.
+            // As a result, Run() is a very long running task that occupies a thread forever.
+            // If we were to schedule Run() on the thread pool it would be possible that the thread chosen by the
+            // pool had run user code before. Such user code may be doing an asynchronous wait scheduled to
+            // continue on the same thread(e.g. this can occur when using a custom synchronization context or a 
+            // custom task scheduler). If such case the waiting user code will never continue.
+            // By creating our own thread, we guarantee no interactions with potentially incorrectly written async user code.
+
+            this.aggregationThread = new Thread(this.Run);
+            this.aggregationThread.Name = nameof(DefaultAggregationPeriodCycle);
+
+            this.aggregationThread.Start();
             return true;
         }
 
         public Task StopAsync()
         {
             Interlocked.Exchange(ref this.runningState, RunningState_Stopped);
-            
-            // Benign race on being called very soon after start. Will miss a cycle but eventually complete correctly.
 
-            Task workerTask = this.workerTask;
-            return workerTask ?? Task.FromResult(true);
+            return this.workerTaskCompletionControl.Task;
         }
 
         public void FetchAndTrackMetrics()
@@ -139,11 +145,13 @@
                 DateTimeOffset now = DateTimeOffset.Now;
                 TimeSpan waitPeriod = GetNextCycleTargetTime(now) - now;
 
-                Task.Delay(waitPeriod).ConfigureAwait(continueOnCapturedContext: false).GetAwaiter().GetResult();
+                Thread.Sleep(waitPeriod);
 
                 int shouldBeRunning = Volatile.Read(ref this.runningState);
                 if (shouldBeRunning != RunningState_Running)
                 {
+                    this.aggregationThread = null;
+                    this.workerTaskCompletionControl.TrySetResult(true);
                     return;
                 }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.ApplicationInsights/Metrics/Implementation/DefaultAggregationPeriodCycle.cs
```diff
@@ -139,11 +145,13 @@ private void Run()
                 DateTimeOffset now = DateTimeOffset.Now;
                 TimeSpan waitPeriod = GetNextCycleTargetTime(now) - now;
 
-                Task.Delay(waitPeriod).ConfigureAwait(continueOnCapturedContext: false).GetAwaiter().GetResult();
+                Thread.Sleep(waitPeriod);
```
pharring: I think you want a cancelable wait. One way is to create a CancellationToken, get its WaitHandle and call WaitOne with the TimeSpan. The CancellationToken takes the place of the `shouldBeRunning` bool.

src/Microsoft.ApplicationInsights/Metrics/Implementation/DefaultAggregationPeriodCycle.cs
```diff
@@ -139,11 +145,13 @@ private void Run()
                 DateTimeOffset now = DateTimeOffset.Now;
```
pharring: Not part of this change, I know, but please use `UtcNow` instead of `Now` (also in `GetNextCycleTargetTime`)

src/Microsoft.ApplicationInsights/Metrics/Implementation/DefaultAggregationPeriodCycle.cs
macrogreg: Why would we want a cancelable wait at this point?

src/Microsoft.ApplicationInsights/Metrics/Implementation/DefaultAggregationPeriodCycle.cs
macrogreg: You are right. If I recall correctly, there may have been purpose to it - take a local TS so that the serializer can choose to UTC it or not. But thinking about it today - I agree that using UTC would be better - and if the serializer wanted to really do something local, it would do that consistently for all telemetry. Please feel free to file an issue.

src/Microsoft.ApplicationInsights/Metrics/Implementation/DefaultAggregationPeriodCycle.cs
pharring: For fast and graceful shutdown. Maybe I misunderstood the point of StopAsync, but I assume it's to be called at shutdown and the expectation is that shutdown is fast.

macrogreg: Got it. Right now we do not do another data dump on a cancelled cycle, but you are right: it would be a good improvement. On the other hand, dealing with WaitHandles here would mean dealing with Disposables. If we go there, I would prefer to do it as a separate change from addressing the issue at hand. Before the proposed change, this wait was not cancelable either.
Thoughts?

pharring: ok

pharring:
Perhaps a Timer (System.Threading.Timer) would work? It doesn't burn a thread. Callbacks happen on threadpool threads.

pharring:
Another idea. Go back to the original code (using `Task.Run`) and change `private void Run()` into `private async Task RunAsync(CancellationToken cancellationToken)` then you can use an asynchronous (and cancellable) wait:

```C#
private async Task RunAsync(CancellationToken cancellationToken)
{
    while (true)
    {
        DateTimeOffset now = DateTimeOffset.Now;
        TimeSpan waitPeriod = GetNextCycleTargetTime(now) - now;
        await Task.Delay(waitPeriod, cancellationToken).ConfigureAwait(false);
        this.FetchAndTrackMetrics();
    }
}
```

Start becomes:
```C#
public bool Start()
{
    if (this.workerTask != null)
    {
        return false; // Was already running or stopped.
    }

    this.workerTask = Task.Run(() => RunAsync(this.cancellationTokenSource.Token));
    return true;
}
```
StopAsync becomes:
```C#
public Task StopAsync()
{
    this.cancellationTokenSource.Cancel();
    return this.workerTask;
}
```

You don't need the runningState (by my reading it's impossible to re-start once StopAsync has been called) although you might want to protect Start against races with a sync object. The CancellationTokenSource should be created in the constructor.

macrogreg:
@pharring , Thanks for the feedback. Code comments added with this change describe how and why this is specifically targeted to not use the thread pool and thus the most obvious Task.Delay(..) approach. Do you have concerns with that reasoning?

In that context, if we wanted to make this wait cancelable, we would probably create a ManualResetEventSlim object in a try-finally block right before we actually perform the wait, and we would dispose the object in the same finally block after the wait. 
We would use the waiting period as the wait timeout and we would signal the event if a previous cancellation is requested, after appropriately making sure we have a valid, undisposed reference.

However, as mentioned, this change aims to address a customer issue in production. As such, I do not want not introduce any additional logic, however useful it might be in a separate change. This wait was not cancellable before and there is no need to make it so in the context of this fix.

However the signal logic described above would be a good improvement. Let's do it as a separate change.

Thanks! :)

src/Microsoft.ApplicationInsights/Metrics/Implementation/DefaultAggregationPeriodCycle.cs
```diff
@@ -16,26 +16,25 @@ internal class DefaultAggregationPeriodCycle
...
-        private readonly Action workerMethod;
-
         private readonly MetricAggregationManager aggregationManager;
         private readonly MetricManager metricManager;
 
+        private readonly TaskCompletionSource<bool> workerTaskCompletionControl;
+
         private int runningState;
-        private Task workerTask;
+        private Thread aggregationThread;
```
MS-TimothyMothra: Is there any value in making this a private static Thread?

I think there is a scenario where a customer is creating multiple TelemetryConfigurations, which would create multiple of these internal classes each with a private thread. Please consider if there is any value in making this thread globally unique.

pharring:
I have no concerns if the fix addresses the customer issue.

src/Microsoft.ApplicationInsights/Metrics/Implementation/DefaultAggregationPeriodCycle.cs
```diff
@@ -52,21 +51,28 @@ public bool Start()
...
-                                        TaskContinuationOptions.ExecuteSynchronously);
+            // We create a new thread rather than using the thread pool.
+            // This is because inside of the main loop in the Run() method we use a synchronous wait.
+            // The reason for that is to prevent aggregation from being affected by potential thread pool starvation.
+            // As a result, Run() is a very long running task that occupies a thread forever.
+            // If we were to schedule Run() on the thread pool it would be possible that the thread chosen by the
+            // pool had run user code before. Such user code may be doing an asynchronous wait scheduled to
+            // continue on the same thread(e.g. this can occur when using a custom synchronization context or a 
+            // custom task scheduler). If such case the waiting user code will never continue.
+            // By creating our own thread, we guarantee no interactions with potentially incorrectly written async user code.
```
pharring: Can I suggest we just say:
```C#
// We create a dedicated thread for aggregation. This is to prevent aggregation
// from being affected by potential thread pool starvation.
```

The rest is about the old implementation that is (IMO) best left unsaid because that implementation contributed to the very thread pool starvation we're trying to avoid! In addition, I think the "had run user code before" and "may be doing an asynchronous wait" parts are misleading and irrelevant to the crux of the problem which was the `.GetAwaiter().GetResult()` piece.

src/Microsoft.ApplicationInsights/Metrics/Implementation/DefaultAggregationPeriodCycle.cs
macrogreg: Yes, you could end up with several instances of DefaultAggregationPeriodCycle. Not only if you have several TelemetryConfigurations, but also if you have explicitly opted into aggregating at the scope of a telemetry client.
However, a shared thread would require shared cancellation and waiting logic, which is non-trivial. For a bug fix we should always err on the side of a minimal change. Here such change is to get off the thread pool.

But yes, now that we create threads explicitly, we can consider using one shared thread for all instances of DefaultAggregationPeriodCycle in the future. It would lighten the resource load.

lmolkova:
@macrogreg I've added missing reference. Please update changelog and we'll be good to go.

lmolkova:
I believe several improvements were discussed for this code. @macrogreg could you please create issues describing these improvements? We'll prioritize them accordingly.

I'd like to merge this PR and ship the fix in beta3.

macrogreg:
@lmolkova : The improvements discussed here are good suggestions for the future, but we should approach them separately from this fix. I fee you merged it today, so i am assuming we are on the same page. :)

lmolkova:
@macrogreg, please create a separate issues for these improvements so we don't forget about them

PR #1031, opened 2018-12-13
Fix: Start/StopOperation does not work when W3C is enabled

lmolkova:
When W3C is enabled in AspNetCore or Web SDK, [W3COperationCorrelationTelemetryInitializer](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/Common/W3C/W3COperationCorrelationTelemetryInitializer.cs#L135) changes telemetry.Id, so it no it no longer matches Activity.Id.

So StopOperation fails and telemetry is not tracked.

This change checks if W3C initializer added `ai_legacyRequestId` into the properties with former it. 
Unfortunately, it's another temporary change that will have to go away when Activity supports W3C natively.

### DIFF
--- a/Test/Microsoft.ApplicationInsights.Test/Shared/TelemetryClientExtensionTests.cs
+++ b/Test/Microsoft.ApplicationInsights.Test/Shared/TelemetryClientExtensionTests.cs
@@ -154,7 +154,7 @@
         }
 
         [TestMethod]
-        public void StopOperationDoesNotThrowExceptionIfParentOpertionIsStoppedBeforeChildOperation()
+        public void StopOperationDoesNotThrowExceptionIfParentOperationIsStoppedBeforeChildOperation()
         {
             using (var parentOperation = this.telemetryClient.StartOperation<DependencyTelemetry>("operationName"))
             {
@@ -191,7 +191,7 @@
         [TestMethod]
         public void DisposeOperationAppliesChangesOnActivityDoneAfterStart()
         {
-            this.telemetryClient.TelemetryConfiguration.TelemetryInitializers.Add(new ActivityTagsTelemetryIntitializer());
+            this.telemetryClient.TelemetryConfiguration.TelemetryInitializers.Add(new ActivityTagsTelemetryInitializer());
 
             DependencyTelemetry telemetry = null;
             using (var operation = this.telemetryClient.StartOperation<DependencyTelemetry>("TestOperationName"))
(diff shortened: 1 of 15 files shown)


Test/Microsoft.ApplicationInsights.Test/Shared/TelemetryClientExtensionTests.cs
```diff
@@ -282,12 +282,37 @@ public void StartOperationWithOperationTelemetrySetsOperationHolderTelemetry()
             Assert.AreEqual(requestTelemetry, this.sendItems[0]);
         }
 
+        [TestMethod]
+        public void StopOperationWhenTelemetryIdDoesNotMatchActivityId()
```
Dmitry-Matveev: May it break manual telemetry correlation that customer implemented on their own? We had multiple users doing manual setup of IDs/ParentIDs before. With us matching IDs to Activity, it looks like we may drop something user intended to track with custom IDs. Is it the case here or I'm overthinking this? :)

src/Microsoft.ApplicationInsights/Extensibility/Implementation/OperationHolder.cs
```diff
@@ -80,18 +80,26 @@ protected virtual void Dispose(bool disposing)
...
+                                // but keeps an original one in 'ai_legacyRequestId' property
+
+                                if (!operationTelemetry.Properties.TryGetValue("ai_legacyRequestId", out var legacyId) ||
+                                    legacyId != currentActivity?.Id)
+                                {
+                                    // this is for internal error reporting
+                                    CoreEventSource.Log.InvalidOperationToStopError();
+
+                                    // this are details with unique ids for debugging
+                                    CoreEventSource.Log.InvalidOperationToStopDetails(
```
Dmitry-Matveev: This message does not go into the default diagnostic traces, right? It seems to be "Warning", but not "Info"/"Verbose". Still not "Error" but enough to raise a question on it's relative severity.

Test/Microsoft.ApplicationInsights.Test/Shared/TelemetryClientExtensionTests.cs
lmolkova: it was the behavior we had for years. nothing has changed with this pr, except I added test that verifies it ;)

Also, while users play with operationids, and parents, this only checks for telemtery id.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/OperationHolder.cs
lmolkova: nothing new here too. There are two messages: 
1. AI Internal error that says that ids do not match - this shows up in users telemetry and our telemetry 
2. details that include ids - this is warning for debugging with perfview

PR #1064, opened 2019-01-18
Move common W3C classes from web to base

lmolkova:
W3C implementation is used by Web and AspNetCore SDKs.

We are going to enable W3C for some 1DS users that do not necessarily use Web SDK (and reference DependencyCollector that declares W3C implementation).

Without this change, we would require everyone to install base + dependency collector just to onboard W3C. With this change, it is still explicit choice to enable it.

This change is pure file move: no new functionality, namespace change, etc.


lmolkova:
@Dmitry-Matveev did you decide if we are allowed to do this?

Dmitry-Matveev:
@lmolkova , I'm addressing this offline with OSS team, will add you to an email thread.

lmolkova:
as per offline conversation with @Dmitry-Matveev  and OSS team:
- we create new APIs in base sdk (this PR)
- we keep old APIs in DependencyCollector, but deprecate them with error = true
- we forward calls from DepCollector.W3C apis to new APIs

no breaking changes. DepCollector W3C APIs will be removed in 3.0

lmolkova:
@Dmitry-Matveev are you ok with this? Can I merge?

Issue #1106, opened 2019-03-30
NullReferenceException in DiagnosticsEventListener.OnEventWritten


pharring:
This is the same as #341 (I can't re-open that one, though). Now I have a root cause and a fix ready to go.

## Repro Steps
Compile and run this console app on .NET Core referencing the latest ApplicationInsights (base) NuGet (actually it repros as far back as 2.4.0):
```csharp
using System;
using System.Threading.Tasks;
using Microsoft.ApplicationInsights.Channel;
using Microsoft.ApplicationInsights.Extensibility;

namespace AINullListenerRepro
{
    class Program
    {
        static void Main(string[] args)
        {
            var t1 = Task.Run(Race1);
            var t2 = Task.Run(Race2);
            Console.WriteLine("Running...");
            Console.ReadKey();
        }

        static async Task Race1()
        {
            while (true)
            {
                using (var channel = new InMemoryChannel())
                {
                    await Task.Yield();
                }
            }
        }

        static async Task Race2()
        {
            while (true)
            {
                using (var telemetryConfiguration = TelemetryConfiguration.Active)
                {
                    await Task.Yield();
                }
            }
        }
    }
}
```
##  Actual Behavior
After a few seconds:
NullReferenceException in DiagnosticsEventListener.OnEventWritten.
The `listener` field is null.

## Expected Behavior
No exception.

## Version Info 
SDK Version : 2.4.0 to 2.9.1
.NET Version : .NET Core 2.2 (although I think it repros on other platforms too)	 						
How Application was onboarded with SDK: VisualStudio
OS : Windows 10
Hosting Info (IIS/Azure WebApps/ etc) : N/A

This issue has been plaguing my unit tests for Snapshot Debugger for that last week or so. It's timing dependent.

Explanation for the repro:
1. Race1 just creates and disposes an InMemoryChannel over and over. That has the effect of reporting "No Telemetry Items passed to Enqueue" to the CoreEventSource over and over.
2. Race2 creates and disposes the TelemetryConfiguration.Active. A lot happens there, but the important thing for the repro is that it adds the DiagnosticsTelemtryModule which establishes the DiagnosticsEventListener.

[...]

PR #1107, opened 2019-03-30
Fix two bugs in DiagnosticsEventListener

pharring:
Fix Issue #1106 
Check for `this.listener == null` and exit early from OnEventWritten. It can be null if the constructor hasn't finished running when an event comes in.

Fix Issue #1108 
Record interesting EventSource instances in the OnEventSourceCreated callback while the base constructor is running. Replay them in DiagnosticEventListener's constructor, so that we can enable them with the appropriate level.

### DIFF
--- a/Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/Tracing/DiagnosticsListenerTest.cs
+++ b/Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/Tracing/DiagnosticsListenerTest.cs
@@ -70,10 +70,31 @@
                 Assert.AreEqual(1, senderMock.Messages.Count);
 
                 senderMock.Messages.Clear();
-                
+
                 listener.LogLevel = EventLevel.Error;
                 CoreEventSource.Log.LogError("Logging an error");
                 Assert.AreEqual(1, senderMock.Messages.Count);
+            }
+        }
+
+        [TestMethod]
+        public void TestEventSourceLogLevelWhenEventSourceIsAlreadyCreated()
+        {
+            using (var testEventSource = new EventSource("Microsoft-ApplicationInsights-" + nameof(TestEventSourceLogLevelWhenEventSourceIsAlreadyCreated)))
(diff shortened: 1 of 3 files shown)


pharring:
@cijothomas care to take a look?
Also, is the Linux build failure expected? Looks like a general problem with System.Net.Http missing on one of the netstandard builds; I "fixed" it locally by adding the nuget, but it's got nothing to do with the bugs I'm fixing here, so it's not included in the PR.

cijothomas:
@pharring Will get to this today. We internally talked about it yesterday but didn't get time to review.

pharring:
I see there is a genuine test case failure in Linux. I will investigate.

pharring:
That test case failure is tricky. We're asserting just one message, but we actually get a bunch. The extra messages seem to be "System.Reflection.TargetException: Non-static method requires a target" which is due to https://github.com/dotnet/corefx/issues/32606. But the real cause of that is that RichPayloadEventSource got enabled at Verbose level and Trace telemetry is being sent from the PortalDiagnosticsSender (because it's trying to log the error that the test generates). It's all quite difficult to unwind. Of course, the test doesn't fail when run alone; only when run in conjunction with other tests because the RichPayloadEventSource is a shared/static thing.

pharring:
Wow. Just wow. Investigating that unit test failure led me to all sorts of interesting places (like the DiagnosticsTelemetryModule can get added more than once -- no doubt leading to repeated telemetry -- I'll leave that for another bug). The fact that the item count was always 13 was highly significant -- it's the number of times the DiagnosticsTelemetryModule was leaked in previous tests. In the end I was able to fix the failure without touching product code; just making unit test changes. But boy, oh boy, that was an unexpected outcome of what I thought would be a localized bug fix.

Dmitry-Matveev:
@pharring , thanks a lot for doing this! I think that leaking diganostics module and verbose logging might've been related to the phantom test failures we had for a while!

pharring:
#readytomerge (I don't have permission to merge in this repo)

Issue #1111, opened 2019-04-06
DiagnosticsTelemetryModule can get added more than once


pharring:
## Repro Steps
1. Compile and run:
```csharp
using System;
using Microsoft.ApplicationInsights;
using Microsoft.ApplicationInsights.Extensibility;
using Microsoft.ApplicationInsights.Extensibility.Implementation;

namespace ConsoleApp1
{
    class Program
    {
        static void Main(string[] args)
        {
            TelemetryConfiguration.Active.Dispose();
            new TelemetryClient();
            Console.WriteLine(TelemetryModules.Instance.Modules.Count);
        }
    }
}
```

##  Actual Behavior
Outputs: 2

## Expected Behavior
Outputs: 1

## Version Info 
SDK Version : 2.10.0-beta3
.NET Version :	 net46, netcoreapp20					
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) : VisualStudio
OS : Windows 10
Hosting Info (IIS/Azure WebApps/ etc) : N/A

pharring:
What's going on here is that the `DiagnosticsTelemetryModule` is automatically injected into `TelemetryConfiguration.Active` by the `TelemetryConfigurationFactory`.
That module is held in a static instance of `TelemetryModules`. If you dispose of `TelemetryConfiguration.Active`, and re-create it (e.g. by creating a new `TelemetryClient`), then _another_ instance of `DiagnosticsTelemetryModule` is added. Once that happens, you've got multiple EventListeners each hooked up to a `PortalDiagnosticsSender` which leads to duplicate telemetry being sent for each diagnostics event.

The key here is calling Dispose on `TelemetryConfiguration.Active` allowing it to be re-created. That happens in unit tests. It's probably unlikely in customer code, but not impossible (as I've shown here).

PR #1112, opened 2019-04-06
Don't add DiagnosticsTelemetryModule more than once.

pharring:
Fix Issue #1111
Checks if `DiagnosticsTelemetryModule` is already in the modules list before adding it in `TelemetryConfigurationFactory.Initialize`


pharring:
@cijothomas done. #readytomerge

PR #1116, opened 2019-04-18
Use NuGet's V3 API during build

skofman1:
We strongly recommend moving to V3 API as it provides more reliable builds and [security improvements](https://blog.nuget.org/20180810/Introducing-Repository-Signatures.html).

The change is super simple:
Replace `https://www.nuget.org/api/v2/` with `https://api.nuget.org/v3/index.json`

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,6 +4,6 @@
     <!--  See the comments in Directory.Build.props for explanations.                         -->
   <packageSources>
     <clear/>
-    <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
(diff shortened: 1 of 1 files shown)


skofman1:
> Will you be doing it for all our repos or you just wanted to show an example with this repo?

Depends, how many repos do you have? 🤣

cijothomas:
https://github.com/Microsoft/ApplicationInsights-dotnet-server
https://github.com/Microsoft/ApplicationInsights-aspnetcore
https://github.com/Microsoft/ApplicationInsights-dotnet-logging
these are remaining repos.
we'll take care of it, if you dont get to them all.

skofman1:
@cijothomas it will be great if you can take care of it. We are working on reducing usage of our V2 API.

PR #1126, opened 2019-04-26
Correct telemetry timestamp precision on .NET Framework

lmolkova:
Fix Issue https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/1175 .

Timestamp is inaccurate on .NET Fx.

 The way to correct it is described here: https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.DateTime.netfx.cs

~~There is no reasonable way to test it - I tested manually.~~

### DIFF
--- a/src/Microsoft.ApplicationInsights/OperationTelemetryExtensions.cs
+++ b/src/Microsoft.ApplicationInsights/OperationTelemetryExtensions.cs
@@ -17,11 +17,6 @@
         private static readonly PreciseTimestamp Timestamp = PreciseTimestamp.Instance;
 
         /// <summary>
-        /// Multiplier to convert Stopwatch ticks to TimeSpan ticks.
-        /// </summary>
-        private static readonly double StopwatchTicksToTimeSpanTicks = (double)TimeSpan.TicksPerSecond / Stopwatch.Frequency;
-
-        /// <summary>
         /// An extension to telemetry item that starts the timer for the respective telemetry.
         /// </summary>
         /// <param name="telemetry">Telemetry item object that calls this extension method.</param>
@@ -98,7 +93,7 @@
         private static void StopImpl(OperationTelemetry telemetry, long timestamp)
         {
             long stopWatchTicksDiff = timestamp - telemetry.BeginTimeInTicks;
-            double durationInTicks = stopWatchTicksDiff * StopwatchTicksToTimeSpanTicks;
+            double durationInTicks = stopWatchTicksDiff * PreciseTimestamp.StopwatchTicksToTimeSpanTicks;
             StopImpl(telemetry, TimeSpan.FromTicks((long)Math.Round(durationInTicks)));
         }
 
--- a/src/Microsoft.ApplicationInsights/OperationTelemetryExtensions.cs
+++ b/src/Microsoft.ApplicationInsights/OperationTelemetryExtensions.cs
@@ -38,7 +38,7 @@
         /// <param name="timestamp">A high-resolution timestamp from <see cref="Stopwatch"/>.</param>
         public static void Start(this OperationTelemetry telemetry, long timestamp)
         {
-            telemetry.Timestamp = DateTimeOffset.UtcNow; 
+            telemetry.Timestamp = Timestamp.GetUtcNow(); 
 
             // Begin time is used internally for calculating duration of operation at the end call,
             // and hence is stored using higher precision Clock.
--- a/src/Microsoft.ApplicationInsights/OperationTelemetryExtensions.cs
+++ b/src/Microsoft.ApplicationInsights/OperationTelemetryExtensions.cs
@@ -12,6 +12,11 @@
     public static class OperationTelemetryExtensions
     {
         /// <summary>
+        /// Precise timestamp instance.
+        /// </summary>
+        private static readonly PreciseTimestamp Timestamp = PreciseTimestamp.Instance;
+
+        /// <summary>
         /// Multiplier to convert Stopwatch ticks to TimeSpan ticks.
         /// </summary>
         private static readonly double StopwatchTicksToTimeSpanTicks = (double)TimeSpan.TicksPerSecond / Stopwatch.Frequency;
@@ -33,7 +38,7 @@
         /// <param name="timestamp">A high-resolution timestamp from <see cref="Stopwatch"/>.</param>
         public static void Start(this OperationTelemetry telemetry, long timestamp)
         {
-            telemetry.Timestamp = DateTimeOffset.UtcNow;
+            telemetry.Timestamp = Timestamp.GetUtcNow();
 
             // Begin time is used internally for calculating duration of operation at the end call,
             // and hence is stored using higher precision Clock.
@@ -108,7 +113,7 @@
 
             if (telemetry.Timestamp == DateTimeOffset.MinValue)
             {
-                telemetry.Timestamp = DateTimeOffset.UtcNow;
+                telemetry.Timestamp = Timestamp.GetUtcNow();
             }
 
             RichPayloadEventSource.Log.ProcessOperationStop(telemetry);
--- a/src/Microsoft.ApplicationInsights/OperationTelemetryExtensions.cs
+++ b/src/Microsoft.ApplicationInsights/OperationTelemetryExtensions.cs
@@ -12,11 +12,6 @@
     public static class OperationTelemetryExtensions
     {
         /// <summary>
-        /// Precise timestamp instance.
-        /// </summary>
-        private static readonly PreciseTimestamp Timestamp = PreciseTimestamp.Instance;
-
-        /// <summary>
         /// An extension to telemetry item that starts the timer for the respective telemetry.
         /// </summary>
         /// <param name="telemetry">Telemetry item object that calls this extension method.</param>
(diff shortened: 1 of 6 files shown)


src/Microsoft.ApplicationInsights/PreciseTimestamp.cs
```diff
@@ -0,0 +1,109 @@
...
+        private static TimeSync timeSync;
+        private readonly Timer syncTimeUpdater;
+
+        private PreciseTimestamp()
+        {
+            this.syncTimeUpdater = InitializeSyncTimer();
+        }
+#endif
+
+        public static PreciseTimestamp Instance
```
pharring: This seems a bit heavy-handed. Can't you just use a static auto-prop?

src/Microsoft.ApplicationInsights/PreciseTimestamp.cs
```diff
@@ -0,0 +1,109 @@
...
+
+                return instance;
+            }
+        }
+
+        /// <summary>
+        /// Returns high resolution (1 DateTime tick) current UTC DateTime. 
+        /// </summary>
+        [SuppressMessage("Microsoft.Performance", "CA1822:MarkMembersAsStatic", Justification = "Enforcing instance fields initialization.")]
+        internal DateTimeOffset GetUtcNow()
```
pharring: Should probably be public (an internal method on an internal class is unnecessarily restrictive)

src/Microsoft.ApplicationInsights/PreciseTimestamp.cs
```diff
@@ -0,0 +1,109 @@
...
+            bool restoreFlow = false;
+            try
+            {
+                if (!ExecutionContext.IsFlowSuppressed())
+                {
+                    ExecutionContext.SuppressFlow();
+                    restoreFlow = true;
+                }
+
+                timer = new Timer(s => { Sync(); }, null, 0, 7200000);
```
pharring: What is the magic 7200000 constant?

src/Microsoft.ApplicationInsights/PreciseTimestamp.cs
pharring: Is there a reason to use `DateTimeOffset` over simply `DateTime`? 
DateTimeOffset is a larger struct and it seems like we really don't need to know the time-zone -- since it's always UTC.

src/Microsoft.ApplicationInsights/OperationTelemetryExtensions.cs
```diff
@@ -33,7 +38,7 @@ public static void Start(this OperationTelemetry telemetry)
         /// <param name="timestamp">A high-resolution timestamp from <see cref="Stopwatch"/>.</param>
         public static void Start(this OperationTelemetry telemetry, long timestamp)
         {
-            telemetry.Timestamp = DateTimeOffset.UtcNow;
+            telemetry.Timestamp = DateTimeOffset.UtcNow;
```
pharring: Shouldn't this also use the PreciseTimestamp ?

src/Microsoft.ApplicationInsights/PreciseTimestamp.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+#if NET45 || NET46
+            // DateTime.UtcNow accuracy on .NET Framework is ~16ms, this method 
+            // uses combination of Stopwatch and DateTime to calculate accurate UtcNow.
+
+            var tmp = timeSync;
+
+            // Timer ticks need to be converted to DateTime ticks
+            long dateTimeTicksDiff = (long)((Stopwatch.GetTimestamp() - tmp.SyncStopwatchTicks) * 10000000L /
+                                            (double)Stopwatch.Frequency);
```
pharring: This expression is captured in `StopwatchTicksToTimeSpanTicks` in `OperationTelemetryExtensions`. Can we share the two implementations?

src/Microsoft.ApplicationInsights/PreciseTimestamp.cs
```diff
@@ -0,0 +1,109 @@
...
+                }
+
+                return instance;
+            }
+        }
+
+        /// <summary>
+        /// Returns high resolution (1 DateTime tick) current UTC DateTime. 
+        /// </summary>
+        [SuppressMessage("Microsoft.Performance", "CA1822:MarkMembersAsStatic", Justification = "Enforcing instance fields initialization.")]
```
pharring: This 'smells' wrong. If it accesses only static fields, it should be a static method. "Enforcing instance fields" sounds like sub-optimal class design -- I think I know where it comes from: You have fields that are 'beforefieldinit'.
Suggestions:
The whole class ought to be static. There is no need for an "Instance" property. The one instance field (syncTimeUpdater) ought to be static too. Add a static constructor to ensure that it's created on first access.

src/Microsoft.ApplicationInsights/PreciseTimestamp.cs
lmolkova: No, I cannot. The problem is that I need to start a timer to correct stopwatch skew. 
Staring a timer without static constructor is a challenge - static field (unused) is optimized away. Static constructor has a perf hit, so I preferred to use singleton. 

Happy to learn if it could be avoided

src/Microsoft.ApplicationInsights/PreciseTimestamp.cs
lmolkova: 2 hours, will add a comment

src/Microsoft.ApplicationInsights/PreciseTimestamp.cs
lmolkova: because AppInsights uses offset everywhere

src/Microsoft.ApplicationInsights/OperationTelemetryExtensions.cs
lmolkova: good catch! this was part of my testing, forgot to return it back

src/Microsoft.ApplicationInsights/PreciseTimestamp.cs
lmolkova: I tried to explain my motivation above - static constructor would be worse for perf than a singleton in this case.

pharring: I've never heard of static constructor being a perf problem. I searched and found a stackoverflow post from 2010 that was quickly debunked. Do you have more information?
I admit I've never studied it, but that's probably because it's never shown up in any perf trace I've seen.

There _are_ surprises with beforefieldinit fields NOT being created in release bits if they're not actually touched by the JIT compiler. That can certainly be avoided with an explicit static constructor. So, I'd recommend that if we can settle the perf issue.

src/Microsoft.ApplicationInsights/PreciseTimestamp.cs
pharring: (I asked elsewhere, but I'll repeat it here.) Do you have more info on the perf hit for static ctors? I've never heard of one.

(I suppose we could measure, but it would likely take millions of iterations before a difference emerges)

lmolkova: I have never done benchmarking for static constructor, there are some sources like [Jon Skeet articles](https://csharpindepth.com/articles/BeforeFieldInit) or ~~apparetly~~ [FXcop docs](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1810-initialize-reference-type-static-fields-inline)

I also agree that if there is any impact, it is likely extremely low, I'll use static constructor.

pharring:
LGTM. Thanks for addressing my feedback.

pharring:
@lmolkova what's going on with all those "revert me" changes?

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -9,7 +10,29 @@ internal sealed class CoreEventSource : EventSource
...
+
+        private CoreEventSource()
+        {
+            //// We want to add application name to all events BUT
+            //// It is prohibited by EventSource rules to have more parameters in WriteEvent that in event source method
+            //// Parameter will be available in payload but in the next versions EventSource may 
+            //// start validating that number of parameters match
+            //// It is not allowed to call additional methods, only WriteEvent
+
+            try
```
pharring: I get that you want to set the applicationName just once in the constructor, but can't you still use ApplicationNameProvider instead of repeating the code?

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/ApplicationNameProvider.cs
```diff
@@ -5,35 +5,27 @@
...
             catch (Exception exp)
             {
-                name = "Undefined " + exp.Message ?? exp.ToString();
+                this.Name = "Undefined " + exp.Message ?? exp.ToString();
             }
-
-            return name;
         }
+
+        public string Name { get; private set; }
```
pharring: Since it's set only in the constructor, you can make it a "getter-only auto-property" by removing `private set;`

lmolkova:
@pharring I've been trying to make tests pass on this PR and found some bug that reproduces on build machine on netcoreapp2.0 only.
It seems to be a race-condition that makes applicationName not initialized. 
Here is the change that fixes this race-condition:  https://github.com/Microsoft/ApplicationInsights-dotnet/commit/0a2c4e0dafdb41204c7619f3b50db19657f9cb73#diff-38c3bb17d4df474170d2d57acf211a18R13

but I still don't understand why. I'm very interested in your input if you don't mind having a quick look there.

I will extract this into another PR - it has nothing to do with timestamp precision.  Sorry for the confusion.

Issue #1128, opened 2019-05-02
MethodsAreImplementedConsistentlyWithTheirAttributes does not test applicaiton name correctly and fails on .netcoreapp2.0 


lmolkova:
https://dev.azure.com/mseng/AppInsights/_build/results?buildId=9392652&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b

```
Starting test execution, please wait...
Failed   MethodsAreImplementedConsistentlyWithTheirAttributes
Error Message:
 Test method Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.CoreEventSourceTest.MethodsAreImplementedConsistentlyWithTheirAttributes threw exception: 
System.Exception: OperationIsNullWarning is implemented incorrectly: Application Name. expected: '', actual: 'testhost' ---> System.Exception: Application Name. expected: '', actual: 'testhost'
Stack Trace:
    at Microsoft.ApplicationInsights.TestFramework.EventSourceTest.AssertEqual[T](T expected, T actual, String message) in E:\A\_work\538\s\Test\TestFramework\Shared\EventSourceTest.cs:line 115
   at Microsoft.ApplicationInsights.TestFramework.EventSourceTest.VerifyMethodImplementation(EventSource eventSource, MethodInfo eventMethod) in E:\A\_work\538\s\Test\TestFramework\Shared\EventSourceTest.cs:line 37
--- End of inner exception stack trace ---
    at Microsoft.ApplicationInsights.TestFramework.EventSourceTest.VerifyMethodImplementation(EventSource eventSource, MethodInfo eventMethod) in E:\A\_work\538\s\Test\TestFramework\Shared\EventSourceTest.cs:line 41
   at Microsoft.ApplicationInsights.TestFramework.EventSourceTest.MethodsAreImplementedConsistentlyWithTheirAttributes(EventSource eventSource) in E:\A\_work\538\s\Test\TestFramework\Shared\EventSourceTest.cs:line 16
```

This happens because tests set `expectedApplicationName` to empty string for all .net core.
```csharp
#if !NETCOREAPP
            string expectedApplicationName = AppDomain.CurrentDomain.FriendlyName;
#else
            string expectedApplicationName = "";
#endif
            string actualApplicationName = actualEvent.Payload.Last().ToString();
            AssertEqual(expectedApplicationName, actualApplicationName, "Application Name");
```

[...]

PR #1129, opened 2019-05-02
Ignore EventSourceTests failures on .NET Core 2.0

lmolkova:
see #1128


cijothomas:
@lmolkova do we need this? We have green builds now

PR #1134, opened 2019-05-13
remove extra type casting in W3C TelemetryInitializer

MS-TimothyMothra:
`W3CActivityExtensions.UpdateTelemetry` is called from `W3COperationCorrelationTelemetryInitializer`.

This method is doing extra type casts (for every telemetry item) that I think are unnecessary. 
I moved the if bracket to `DependencyTelemetry` and added a method to the abstract class `OperationTelemetry`


lmolkova:
This is going to go away after we'll have W3C support in .NET platform, so unless it fixes some issue, I propose to keep it as is. 
Also, it is now disabled by default.

MS-TimothyMothra:
@lmolkova Can you re-review this?
Most of the changes are to the `UpdateTelemetry` method which I see is referenced frequently in the WebSDK.

lmolkova:
everything would be retired including the call from Web SDK

PR #1147, opened 2019-05-22
Fix SDL issue for System.Net.Security 4.3.0

lmolkova:
and clean up dependencies a bit


lmolkova:
proof: https://mseng.visualstudio.com/AppInsights/_build/results?buildId=9528768

PR #1148, opened 2019-05-23
Obsolete TelemetryConfiguration.Active singleton usage on .NET Core 

lmolkova:
We have a stream of issues and support requests coming from confusion around `TelemetryConfiguration.Active` on ASP.NET Core.

We DO NOT recommend using `TelemetryConfiguration.Active` - instead we recommend using DI from [Microsoft.Extensions.DependencyInjection](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.2). 

In the case of ASP.NET Core or generic host applications, here is an instruction on how to use/configure `TelemetryConfiguration`:  https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core

This change makes Active obsolete on .NET Core to further discourage its usage there.

### DIFF
--- a/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
+++ b/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
@@ -73,16 +73,16 @@
             this.telemetrySinks.Add(defaultSink);
         }
 
-#if NETSTANDARD1_3 ||  NETSTANDARD2_0
-        [Obsolete(@"We do not recommend using TelemetryConfiguration.Active on .NET Core.
-Instead we suggest using dependency injection extensions. 
-See https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.2 and https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core for more info")]
-#endif
         /// <summary>
         /// Gets the active <see cref="TelemetryConfiguration"/> instance loaded from the ApplicationInsights.config file. 
         /// If the configuration file does not exist, the active configuration instance is initialized with minimum defaults 
         /// needed to send telemetry to Application Insights.
         /// </summary>
+#if NETSTANDARD1_3 || NETSTANDARD2_0
+        [Obsolete(@"We do not recommend using TelemetryConfiguration.Active on .NET Core.
+Instead we suggest using dependency injection extensions. 
+See https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.2 and https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core for more info")]
+#endif 
         public static TelemetryConfiguration Active
         {
             get
--- a/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
+++ b/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
@@ -73,6 +73,11 @@
             this.telemetrySinks.Add(defaultSink);
         }
 
+#if NETSTANDARD1_3 ||  NETSTANDARD2_0
+        [Obsolete(@"We do not recommend using TelemetryConfiguration.Active on .NET Core.
+Instead we suggest using dependency injection extensions. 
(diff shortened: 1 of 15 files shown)


cijothomas:
@lmolkova I wonder what would be our suggestion for .NET Core Console Apps which don't use DI, so using .Active probably okay there?
How can we handle this in code? we cannot detect .NET Core vs Asp.Net Core, and apply Obsolete tag.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -29,13 +29,15 @@ public sealed class TelemetryClient
 #endif
         private readonly TelemetryConfiguration configuration;
         private string sdkVersion;
-        
+
+#pragma warning disable 612, 618
         /// <summary>
         /// Initializes a new instance of the <see cref="TelemetryClient" /> class. Send telemetry with the active configuration, usually loaded from ApplicationInsights.config.
         /// </summary>
         public TelemetryClient() : this(TelemetryConfiguration.Active)
```
pharring: Should we also obsolete this constructor (since it implicitly uses TC.Active) ?
i.e. DI users should always use an injected TC.

pharring: (by TC I mean TelemetryConfiguration)

lmolkova: yes! definitely!

lmolkova:
> I wonder what would be our suggestion for .NET Core Console Apps which don't use DI, so using .Active probably okay there?
> How can we handle this in code? we cannot detect .NET Core vs Asp.Net Core, and apply Obsolete tag.

Console applications should carry all objects explicitly. 
1. They should create the configuration with `TelemetryConfiguration.CreateDefault()` and carry it everywhere as any other object they create (e.g. DB client or Azure Storage client). 
2. They can install DI extensions from Microsoft.Extensions.DependencyInjection and use DI pattern or use any other DI framework. They also can explicitly pass configuration/client around. 
3. If they really want static singleton - it is their choice to store explicitly created instance in global static variable, but we should not promote it or do it ourselves.

If we decide to proceed with this change, we should update console app docs (and monitor other docs) to reflect this.
@Dmitry-Matveev @SergeyKanzhelev  I hope to hear your opinion on this.

Dmitry-Matveev:
@lmolkova , if docs are to be updated with the statements you outlined, I'm OK deprecating this in .NET Core branch.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -29,10 +29,18 @@ public sealed class TelemetryClient
...
+#pragma warning disable 612, 618 // TelemetryConfiguration.Active
         /// <summary>
         /// Initializes a new instance of the <see cref="TelemetryClient" /> class. Send telemetry with the active configuration, usually loaded from ApplicationInsights.config.
         /// </summary>
+#if NETSTANDARD1_3 || NETSTANDARD2_0
+        [Obsolete("TelemetryClient parameterless constructor uses TelemetryConfiguration.Active singleton. " +
+                  "We do not recommend using TelemetryConfiguration.Active on .NET Core. " + 
+                  "Instead we suggest using dependency injection pattern. " + 
+                  "See https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.2 " + 
+                  "and https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core for more info")]
```
cijothomas: its best to create a new issue in Github repo, and link to the above links from there. Main reason is - the above links are all 'asp.net core' specific, not so much about '.net core console'. So our Github issue can act as a wrapper, and have links to asp.net core with di, .net core with di, or just plain .net core console etc. (if needed).

Also it'd be good place for us to explain the rationale behind this decision in a github issue in this repo.

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -78,6 +78,12 @@ public TelemetryConfiguration(string instrumentationKey, ITelemetryChannel chann
         /// If the configuration file does not exist, the active configuration instance is initialized with minimum defaults 
         /// needed to send telemetry to Application Insights.
         /// </summary>
+#if NETSTANDARD1_3 || NETSTANDARD2_0
```
cijothomas: same comment as TelemetryClient() ctor.

PR #1158, opened 2019-06-20
Second wave of performance fixes

Dmitry-Matveev:
- Sampling Score Generator is moved to AI.dll, shortcut method to get sampling score off string value is provided;
- Telemetry Configuration adds two methods to Get and Set last observed sampling rate for the specific telemetry item type;
- SamplingTelemetryItemTypes flags are introduced to replace hashset operations that used to check each item for sampling applicability;
- ISupportSampling interface was extended (and later replaced with ISupportAdvancedSampling to preserve public API surface and 2.0 major version):
  - Sampling item type is now provided;
  - IsProactivelySampledOut field to mark item as proactively sampled out is provided;
  - SupportsProactiveSampling field to mark items that do support proactive sampling approach (currently, only requests, but the code is ready to handle other types as well);
- W3C Utliities no longer uses Guid.NewGuid() to generate ID, but leverages two instances of WeakConcurrentRandom instead;
- Telemetry Client skips Initializers if item was proactively sampled out already (e.g. in AspNetCore Diagnostics Listener) or skips initializers after the OperationId Initializer if sampling score is higher than the sample rate;
- Sampling Telemetry processor no longer uses Hash Sets;
- Sampling Telemetry processor calculates overall item count of proactively sampled out items if it decides it should've sampled them in (e.g. if sampling rate was increased while item was going through initializers), this item count is later added to the next item that is sampled in of this type to correct miscalc;

In places code looks ugly as its optimized for perf. I'm super-open to make it better while preserving perf gains, please do let me know your suggestions, we'll check perf and take those. (E.g. Interlocked caused me to have Array vs. Dictionary in one place and switch-case in the other).

The overall plan is to make proactive sampling a feature flag when I merge with develop that supports feature flags. App Service integration will enable this feature flag just like in case with the deferred QuickPulse GetURI() calculation.

One caveat for Public API surface: If someone had implemented a Telemetry Initializer / Processor that samples the item in, so this change (proactive sampling part) technically cannot be part of SDK itself, only of codeless way of attaching SDKs. That's something we'd need to discuss.

P.S. Sorry for the PR size, mostly API/Tests/DataContracts/Interface changes that spawned across multiple item types.

- [x] I ran Unit Tests locally
- [ ] Changes in public surface reviewed (reviewing in this PR)
- [ ] CHANGELOG.md updated with one line description of the fix, and a link to the original issue (Will update when we have a final scope)


src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -513,44 +510,94 @@ public void Initialize(ITelemetry telemetry)
...
+                            CoreEventSource.Log.LogError(string.Format(
+                                                            CultureInfo.InvariantCulture,
+                                                            "Exception while initializing {0}, exception message - {1}",
+                                                            this.configuration.TelemetryInitializers[index].GetType().FullName,
+                                                            exception));
+                        }
+                    }                    
                 }
-                catch (Exception exception)
+                else
```
MS-TimothyMothra: i think this `else` can be removed.

I propose introducing a new class `AdvancedSamplingHelper` that does the evaluation above in the IF and could internalize some extra variables.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -513,44 +510,94 @@ public void Initialize(ITelemetry telemetry)
...
-            // set RoleInstance to the machine name if it's not initialized yet
-            if (string.IsNullOrEmpty(telemetry.Context.Cloud.RoleInstance))
-            {
-                telemetry.Context.Cloud.RoleInstance = PlatformSingleton.Current.GetMachineName();
+                // set RoleInstance to the machine name if it's not initialized yet
+                if (string.IsNullOrEmpty(telemetry.Context.Cloud.RoleInstance))
+                {
+                    telemetry.Context.Cloud.RoleInstance = PlatformSingleton.Current.GetMachineName();
+                }
             }
```
MS-TimothyMothra: You're sitting next to me right now and told me that we need to review QuickPulse because these might be required.
Adding comment here so we don't forget :)

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -3,32 +3,44 @@
...
+        private const string ExceptionTelemetryName = "EXCEPTION";
+        private const string PageViewTelemetryName = "PAGEVIEW";
+        private const string RequestTelemetryName = "REQUEST";
+        private const string TraceTelemetryName = "TRACE";
 
         private readonly char[] listSeparators = { ';' };
         private readonly IDictionary<string, Type> allowedTypes;
 
-        private HashSet<Type> excludedTypesHashSet;
+        // Interlocked operations support an array indexer but not a dictionary property accessor
```
MS-TimothyMothra: I propose encapsulating this array and Dictionary into a new class like you've done with `SamplingRateStore`

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -113,20 +144,41 @@ public string IncludedTypes
...
+
                 if (!string.IsNullOrEmpty(value))
                 {
                     string[] splitList = value.Split(this.listSeparators, StringSplitOptions.RemoveEmptyEntries);
                     foreach (string item in splitList)
                     {
                         if (this.allowedTypes.ContainsKey(item))
                         {
-                            newIncludedTypesHashSet.Add(this.allowedTypes[item]);
+                            switch (item.ToUpperInvariant())
```
MS-TimothyMothra: nit: Duplicate code

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -513,44 +510,94 @@ public void Initialize(ITelemetry telemetry)
...
+            // Telemetry can be already sampled out if that decision was made before calling Track()
+            bool sampledOut = telemetryWithSampling?.IsProactivelySampledOut ?? false;
+
+            // If telemetry was not sampled out, we need to check its sampling score after we learn its Operation Id
+            bool shoudlCheckSamplingScore = !sampledOut;
+
+            if (!sampledOut)
             {
-                try
+                if (telemetryWithSampling?.SupportsProactiveSampling ?? false)
```
lmolkova: hm, it seems it is already checked with first `if (!sampledOut)`

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -513,44 +510,94 @@ public void Initialize(ITelemetry telemetry)
...
             telemetry.Context.Initialize(this.Context, instrumentationKey);
-            foreach (ITelemetryInitializer initializer in this.configuration.TelemetryInitializers)
+
+            ISupportAdvancedSampling telemetryWithSampling = telemetry as ISupportAdvancedSampling;
+
+            // Telemetry can be already sampled out if that decision was made before calling Track()
+            bool sampledOut = telemetryWithSampling?.IsProactivelySampledOut ?? false;
+
+            // If telemetry was not sampled out, we need to check its sampling score after we learn its Operation Id
+            bool shoudlCheckSamplingScore = !sampledOut;
```
lmolkova: typo: shoudlCheckSamplingScore

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -3,32 +3,44 @@
...
     /// </summary>
     public sealed class SamplingTelemetryProcessor : ITelemetryProcessor
     {
-        private const string DependencyTelemetryName = "Dependency";
-        private const string EventTelemetryName = "Event";
-        private const string ExceptionTelemetryName = "Exception";
-        private const string PageViewTelemetryName = "PageView";
-        private const string RequestTelemetryName = "Request";
-        private const string TraceTelemetryName = "Trace";
+        private const string DependencyTelemetryName = "DEPENDENCY";
```
lmolkova: what happened with CAPSLOCK? :)

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,59 @@
+Microsoft.ApplicationInsights.Channel.SamplingScoreGenerator
```
SergeyKanzhelev: Please hide all methods into extension method like this:
https://github.com/microsoft/ApplicationInsights-dotnet/blob/39a5ef23d834777eefdd72149de705a016eb06b0/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryContextExtensions.cs

Most of them are not designed for end user to use

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,59 @@
...
+Microsoft.ApplicationInsights.DataContracts.PageViewTelemetry.IsProactivelySampledOut.set -> void
+Microsoft.ApplicationInsights.DataContracts.RequestTelemetry.IsProactivelySampledOut.set -> void
+Microsoft.ApplicationInsights.DataContracts.ISupportAdvancedSampling.SupportsProactiveSampling.get -> bool
+Microsoft.ApplicationInsights.DataContracts.ISupportAdvancedSampling.SupportsProactiveSampling.set -> void
+Microsoft.ApplicationInsights.DataContracts.DependencyTelemetry.SupportsProactiveSampling.get -> bool
+Microsoft.ApplicationInsights.DataContracts.EventTelemetry.SupportsProactiveSampling.get -> bool
+Microsoft.ApplicationInsights.DataContracts.TraceTelemetry.SupportsProactiveSampling.get -> bool
+Microsoft.ApplicationInsights.DataContracts.ExceptionTelemetry.SupportsProactiveSampling.get -> bool
+Microsoft.ApplicationInsights.DataContracts.PageViewPerformanceTelemetry.SupportsProactiveSampling.get -> bool
+Microsoft.ApplicationInsights.DataContracts.PageViewTelemetry.SupportsProactiveSampling.get -> bool
```
SergeyKanzhelev: Proactive -> Head. It's how it's typically called (head based sampling vs. tail based sampling)

src/Microsoft.ApplicationInsights/TelemetryClient.cs
Dmitry-Matveev: `if (!sampledOut)` checks for case when telemetry was already sampled out and we do not need to run initializers.
`if (supportsProactiveSampling)` checks whether current telemetry type supports proactive sampling and we may skip initializers after we know its operation ID or we need to run all initializers.

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
Dmitry-Matveev: `ToUpper()` is a requirement for switch cases as per Code Analysis. So I can `ToUpper()` here on lower-case constants or can use a chance to SCREAM in my code :)

src/Microsoft.ApplicationInsights/TelemetryClient.cs
Dmitry-Matveev: if (!sampledOut) checks for case when telemetry was already sampled out and we do not need to run initializers at all.
if (supportsProactiveSampling) checks whether current telemetry type supports proactive sampling and we may skip initializers after we know its operation ID (if branch) or we need to run all initializers (else branch).

I'll take a look at moving those to a class / function.

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
Dmitry-Matveev: Hid Telemetry Configuration methods, not sure on SamplingScoreGenerator as it's in Channel namespace and has two static methods.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
MS-TimothyMothra: I think this would be equivalent:

```suggestion
                if (!sampledOut)
            {
                for (int index = 0; index < this.configuration.TelemetryInitializers.Count; index++)
                {
                    if (telemetryWithSampling?.SupportsProactiveSampling ?? false)
                    {
                        if (shoudlCheckSamplingScore && !string.IsNullOrEmpty(telemetry.Context.Operation.Id))
                        {
                            if (SamplingScoreGenerator.GetSamplingScore(telemetry.Context.Operation.Id) >= this.configuration.GetLastObservedSamplingPercentage(telemetryWithSampling.ItemTypeFlag))
                            {
                                telemetryWithSampling.IsProactivelySampledOut = true;
                                sampledOut = true;
                                break;
                            }
                            else
                            {
                                shoudlCheckSamplingScore = false;
                            }
                        }
                    }


                    try
                    {
                        this.configuration.TelemetryInitializers[index].Initialize(telemetry);
                    }
                    catch (Exception exception)
                    {
                        CoreEventSource.Log.LogError(string.Format(
                                                        CultureInfo.InvariantCulture,
                                                        "Exception while initializing {0}, exception message - {1}",
                                                        this.configuration.TelemetryInitializers[index].GetType().FullName,
                                                        exception));
                    }
                }
            }
```


This moves the `if (telemetryWithSampling?.SupportsProactiveSampling ?? false)` and the `if (shoudlCheckSamplingScore && !string.IsNullOrEmpty(telemetry.Context.Operation.Id))` both inside the `for` loop, which removes the need for the `Else`.

You could still reduce it further by moving the Sampling code into an external class/function.

MS-TimothyMothra: Ignore the last comment, I think this would be cleaner:

Can separate the sampling check from calling TelemetryInitializers.
Can remove an extra boolean.

```suggestion
            ISupportAdvancedSampling telemetryWithSampling = telemetry as ISupportAdvancedSampling;

            // Telemetry can be already sampled out if that decision was made before calling Track()
            bool sampledOut = telemetryWithSampling?.IsProactivelySampledOut ?? false;

            // If telemetry was not sampled out, we need to check its sampling score after we learn its Operation Id
            if (!sampledOut)
            {
                if (telemetryWithSampling?.SupportsProactiveSampling ?? false && !string.IsNullOrEmpty(telemetry.Context.Operation.Id))
                {
                    if (SamplingScoreGenerator.GetSamplingScore(telemetry.Context.Operation.Id) >= this.configuration.GetLastObservedSamplingPercentage(telemetryWithSampling.ItemTypeFlag))
                    {
                        telemetryWithSampling.IsProactivelySampledOut = true;
                        sampledOut = true;
                    }
                }
            }

            // Invoke TelemetryInitializers
            if (!sampledOut)
            {
                for (int index = 0; index < this.configuration.TelemetryInitializers.Count; index++)
                {
                    try
                    {
                        this.configuration.TelemetryInitializers[index].Initialize(telemetry);
                    }
                    catch (Exception exception)
                    {
                        CoreEventSource.Log.LogError(string.Format(
                                                        CultureInfo.InvariantCulture,
                                                        "Exception while initializing {0}, exception message - {1}",
                                                        this.configuration.TelemetryInitializers[index].GetType().FullName,
                                                        exception));
                    }
                }
            }
```

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -3,32 +3,34 @@
...
         private readonly IDictionary<string, Type> allowedTypes;
 
-        private HashSet<Type> excludedTypesHashSet;
+        private readonly AtomicSampledItemsCounter proactivelySampledOutCounters = new AtomicSampledItemsCounter();
+
+        private SamplingTelemetryItemTypes excludedTypesFlags;
         private string excludedTypesString;
 
-        private HashSet<Type> includedTypesHashSet;
+        private SamplingTelemetryItemTypes includedTypesFlags;
```
MS-TimothyMothra: Why do we need both `included` and `excluded` flags?

src/Microsoft.ApplicationInsights/TelemetryClient.cs
Dmitry-Matveev: Won't work unfortunately, we need `GetSamplingScore(telemetry.Context.Operation.Id)` after OperationCorrelationTelemetryInitializer was run and we cannot guarantee it's first in line and no one added other initiliazers before it. So this check is bound to be inside of the cycle. I used to have the implementation like you suggested about a couple of weeks ago and it didn't operate well.

If we separate Booleans while having this check inside the cycle, we will run this check on every step of the cycle if item is sampled it (check is expensive, so we can't). So, it means if one is not possible, the other is canceled out as well.

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
Dmitry-Matveev: We use both in the default configuration - we set up one sampling with excluded events, another with only events included.

MS-TimothyMothra: In the default config, I thought those were two separate instances of SamplingTelemetryProcessor.

Is it ever valid for one Processor to have a set of both Included and Excluded?

Dmitry-Matveev: Correct, those are separate in the default config. So, it's more of a homage to the previous implementation rather than a requirement, I guess. They are currently both exposed as public properties on the configuration, so that part would stay the same. We can cook up internal implementation to only construct included list. Due to no perf gain at this (we remove one atomic check on `excluded.HasFlag`), not sure if I want this as part of this PR.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
Dmitry-Matveev: Yep, manual test still remains. Upon code investigation, it looks like we'll need to use "deferRequestProperties" flag in addition to "proactiveSampling" flag in .NET Core App Services to compensate for the absence of URL on sampled requests just like you do in dotnet-server repo.

PublicAPI/Microsoft.ApplicationInsights.dll/net45/PublicAPI.Unshipped.txt
Dmitry-Matveev: It does work nicely with `SupportsHeadSampling`, but it does not play well with the following:
- `IsProactivelySampledOut`? -> `IsHeadSampled`? -> `IsSampledOutAtHead`?

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -39,6 +39,7 @@ public sealed class TelemetryClient
 #endif
         public TelemetryClient() : this(TelemetryConfiguration.Active)
         {
+            this.proactiveSamplingEnabled = TelemetryConfiguration.Active.EvaluateExperimentalFeature("proactiveSampling");
```
cijothomas: would be nice to move this to a constant. "proactiveSampling"

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -516,44 +518,95 @@ public void Initialize(ITelemetry telemetry)
...
+                        {
+                            if (SamplingScoreGenerator.GetSamplingScore(telemetry.Context.Operation.Id) >= this.configuration.GetLastObservedSamplingPercentage(telemetryWithSampling.ItemTypeFlag))
+                            {
+                                telemetryWithSampling.IsProactivelySampledOut = true;
+                                sampledOut = true;
+                                break;
+                            }
+                            else
+                            {
+                                shouldCheckSamplingScore = false;
```
cijothomas: is there an implicit assumption that sampling will be based on the operation.Id set by the first telemetryinitializer, and that operation.id set by any later initializers will not be taken into consideration for sampling calculations?

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -516,44 +518,95 @@ public void Initialize(ITelemetry telemetry)
...
+            ISupportAdvancedSampling telemetryWithSampling = telemetry as ISupportAdvancedSampling;
+
+            // Telemetry can be already sampled out if that decision was made before calling Track()
+            bool sampledOut = telemetryWithSampling?.IsProactivelySampledOut ?? false;
+            bool shouldTryHeadSampling = this.proactiveSamplingEnabled && (telemetryWithSampling?.SupportsProactiveSampling ?? false);
+
+            if (!sampledOut)
             {
-                try
+                if (shouldTryHeadSampling)
```
cijothomas: suggest to stick to consistent naming - proactivesampling or headsampling.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -516,44 +518,95 @@ public void Initialize(ITelemetry telemetry)
...
                 {
-                    initializer.Initialize(telemetry);
+                    // If telemetry was not sampled out, we need to check its sampling score after we learn its Operation Id
```
cijothomas: can we be more aggresive and move lines 492 to 518 to inside - if(shouldTryHeadSampling) ? We don't want to pay the price of doing ConcurrentDictionary instantiation and copying, if item is to be proactively sampled out.

Test/ServerTelemetryChannel.Test/Shared.Tests/SamplingTelemetryProcessorTest.cs
```diff
@@ -15,6 +15,8 @@
     [TestClass]
     public class SamplingTelemetryProcessorTest
     {
+        Random random = new Random();
+
```
cijothomas: Given that SamplingProcessor is modified to parse included/excluded types flags from user input- please ensure unit tests already cover this, or add new tests.

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -3,32 +3,34 @@
...
     using System.Threading;
-    
+
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.DataContracts;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
     using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation;
-    
+
```
cijothomas: need to update public comment for this public class, as this now handles pro-active sampling as well.

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -217,17 +200,86 @@ public void Process(ITelemetry item)
...
+            if (advancedSamplingSupportingTelemetry != null)
+            {
+                if (advancedSamplingSupportingTelemetry.IsProactivelySampledOut)
+                {
+                    this.proactivelySampledOutCounters.AddItems(advancedSamplingSupportingTelemetry.ItemTypeFlag, Convert.ToInt64(100 / currentSamplingPercentage));
+
+                    if (TelemetryChannelEventSource.IsVerboseEnabled)
+                    {
+                        TelemetryChannelEventSource.Log.ItemSampledOut(item.ToString());
```
cijothomas: Do a different log here to indicate item was sampled out because of Proactive sampling set. The decision to drop item was made elsewhere, and logs should hint this class did not make this decision.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -516,44 +518,95 @@ public void Initialize(ITelemetry telemetry)
...
+                                                            "Exception while initializing {0}, exception message - {1}",
+                                                            this.configuration.TelemetryInitializers[index].GetType().FullName,
+                                                            exception));
+                        }
+                    }
                 }
             }
 
-            if (telemetry.Timestamp == default(DateTimeOffset))
+            if (!sampledOut)
```
cijothomas: if (sampledOut) - then log that rest of Initialize is skipped.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -516,44 +518,95 @@ public void Initialize(ITelemetry telemetry)
...
+                    // If telemetry was not sampled out, we need to check its sampling score after we learn its Operation Id                    
+                    bool shouldCheckSamplingScore = true;
+
+                    for (int index = 0; index < this.configuration.TelemetryInitializers.Count; index++)
+                    {
+                        if (shouldCheckSamplingScore && !string.IsNullOrEmpty(telemetry.Context.Operation.Id))
+                        {
+                            if (SamplingScoreGenerator.GetSamplingScore(telemetry.Context.Operation.Id) >= this.configuration.GetLastObservedSamplingPercentage(telemetryWithSampling.ItemTypeFlag))
+                            {
+                                telemetryWithSampling.IsProactivelySampledOut = true;
```
cijothomas: please log that we are setting this item to be pro-act-sampled. Would be good to mention the 'index' in the logs as well, so we can track which initializers from the list will be skipped.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -55,6 +56,7 @@ public TelemetryClient(TelemetryConfiguration configuration)
             }
 
             this.configuration = configuration;
+            this.proactiveSamplingEnabled = this.configuration.EvaluateExperimentalFeature("proactiveSampling");
```
cijothomas: logInformation that TC is created with this featureflag ON.

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
Dmitry-Matveev: As per Sergey's request, this experimental feature flag is currently only built for internal consumption. I can mention something like "This processor respects experimental feature flags related to sampling."

src/Microsoft.ApplicationInsights/TelemetryClient.cs
cijothomas: Its possible to use telemetryClient, with no built-in telemetry processors. so its best to log who is setting IsProactivelySampledout = true. Since its possible that other auto-collector could also be setting this to true.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
cijothomas: Ok - if not already done, can you make it obvious in the PR description so there is a trail of decisions made.. (or maybe a comment in the code itself)

src/Microsoft.ApplicationInsights/TelemetryClient.cs
cijothomas: purely for investigation purposes if something behaves unexpectedly.. like we are seeing telemetry with unknown timestamp or sdkversion etc, and and to isolate it has nothing to do with the new proactive sampling.

src/Microsoft.ApplicationInsights/TelemetryClient.cs
cijothomas: It didn't show up in perf trace because the scenario was not triggered.
The following copies all context from TelemetryClitn to individual telemetry, instantiating a concurrent dictionary if needed. This can be avoided if item is being sampled out.
Utils.CopyDictionary(this.Context.Properties, telemetry.Context.Properties);

(there is nothing preventing an autocollector (eg: aspnet core request collection) to use telemetryclient.context)

Let me know if this make sense/not.

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
cijothomas: Yea something like that would do. I'd prefer if the word pro-active sampling is mentioned as well. (its not really mentioned in official docs which SDK is used by code-less scenarios, so many people assume its the same sdk with same config.. ) May avoid some confusion.

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -217,17 +200,86 @@ public void Process(ITelemetry item)
...
+
+            if (advancedSamplingSupportingTelemetry != null)
+            {
+                if (advancedSamplingSupportingTelemetry.IsProactivelySampledOut)
+                {
+                    this.proactivelySampledOutCounters.AddItems(advancedSamplingSupportingTelemetry.ItemTypeFlag, Convert.ToInt64(100 / currentSamplingPercentage));
+
+                    if (TelemetryChannelEventSource.IsVerboseEnabled)
+                    {
+                        TelemetryChannelEventSource.Log.ItemProactivelySampledOut(item.ToString());
```
Dmitry-Matveev: Add separate log message

src/ServerTelemetryChannel/SamplingTelemetryProcessor.cs
```diff
@@ -217,17 +201,86 @@ public void Process(ITelemetry item)
...
+
+        private void HandlePossibleProactiveSampling(ITelemetry item, double currentSamplingPercentage, ISupportAdvancedSampling samplingSupportingTelemetry = null)
+        {
+            var advancedSamplingSupportingTelemetry = samplingSupportingTelemetry ?? item as ISupportAdvancedSampling;
+
+            if (advancedSamplingSupportingTelemetry != null)
+            {
+                if (advancedSamplingSupportingTelemetry.IsSampledOutAtHead)
+                {
+                    this.proactivelySampledOutCounters.AddItems(advancedSamplingSupportingTelemetry.ItemTypeFlag, Convert.ToInt64(100 / currentSamplingPercentage));
```
cijothomas: it'd be really good to have a comment explaining the logic. I did see comments in unit tests, but i recommend an explanation here as well considering this is a bit complex scenario.

Dmitry-Matveev: Adding...

PR #1170, opened 2019-07-15
Minor perf improvements..

cijothomas:
Fix Issue # .

### DIFF
--- a/src/Microsoft.ApplicationInsights/TelemetryClient.cs
+++ b/src/Microsoft.ApplicationInsights/TelemetryClient.cs
@@ -478,13 +478,6 @@
         [EditorBrowsable(EditorBrowsableState.Never)]
         public void Initialize(ITelemetry telemetry)
         {
-            string instrumentationKey = this.Context.InstrumentationKey;
-
-            if (string.IsNullOrEmpty(instrumentationKey))
-            {
-                instrumentationKey = this.configuration.InstrumentationKey;
-            }
-
             ISupportAdvancedSampling telemetryWithSampling = telemetry as ISupportAdvancedSampling;
 
             // Telemetry can be already sampled out if that decision was made before calling Track()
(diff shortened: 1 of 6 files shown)


src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -478,20 +478,20 @@ public void InitializeInstrumentationKey(ITelemetry telemetry)
...
-            }
-
             ISupportAdvancedSampling telemetryWithSampling = telemetry as ISupportAdvancedSampling;
 
             // Telemetry can be already sampled out if that decision was made before calling Track()
             bool sampledOut = telemetryWithSampling?.IsSampledOutAtHead ?? false;
 
             if (!sampledOut)
             {
+                string instrumentationKey = this.Context.InstrumentationKey;
```
MS-TimothyMothra: nit: Consider moving this down to line 523 where the instrumentationKey is being used.

cijothomas: agree

src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -520,6 +513,13 @@ public void Initialize(ITelemetry telemetry)
                     Utils.CopyDictionary(this.Context.GlobalProperties, telemetry.Context.GlobalProperties);
                 }
 
+                string instrumentationKey = this.Context.InstrumentationKey;
+
+                if (string.IsNullOrEmpty(instrumentationKey))
+                {
+                    instrumentationKey = this.configuration.InstrumentationKey;
```
Dmitry-Matveev: I remember keeping this assignment outside to make sure that proactively sampled out item won't get "Invalid Ikey" or won't get dropped before it arrives to sampling for gain up calculation in Sampling Telemetry Processor. If we are sure that items make it all the way without IKey - this change is OK. There was a case where setting empty IKey prevented all telemetry from leaving SDK (I may have false memory here, though).

cijothomas: The lines i moved into the if block will not have helped with the issue. It just assigns ikey to a local variable, which is ONLY used inside the if block. I just moved the assignment closer to actual usage, in the same if block.

cijothomas: https://github.com/microsoft/ApplicationInsights-aspnetcore/blob/develop/src/Microsoft.ApplicationInsights.AspNetCore/DiagnosticListeners/Implementation/HostingDiagnosticListener.cs#L513 This initialization is probably what you are referring to. That assignment is outside of any Sampling If-Else statements, and is correctly done so,

Issue #1173, opened 2019-07-19
Bug: Both TelemetryChannels do not correctly handle AggregateException


MS-TimothyMothra:
If you are reporting bug/issue, please provide detailed Repro instructions.

## Repro Steps
**still investigating, don't have a repro yet**

Breeze fixed their issues. Can test using this: https://badssl.com/

##  Actual Behavior

We aren't handling the `AggregateException`.
Instead, we log this message: 
> "Failed to send: One or more errors occurred.."

https://github.com/microsoft/ApplicationInsights-dotnet/blob/f181abbcdf6a6a2a4aea821ce1d72e98b0c68fbe/src/Microsoft.ApplicationInsights/Channel/InMemoryTransmitter.cs#L136-L144

Send method returns a Task
https://github.com/microsoft/ApplicationInsights-dotnet/blob/f181abbcdf6a6a2a4aea821ce1d72e98b0c68fbe/src/Microsoft.ApplicationInsights/Channel/InMemoryTransmitter.cs#L151

## Expected Behavior

We need to correctly handle the `AggregateException`.

## Version Info 
SDK Version : all
.NET Version :	all						
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) :
OS : nuget
Hosting Info (IIS/Azure WebApps/ etc) : onprem

MS-TimothyMothra:
`ServerTelemetryChannel` might have the same bug. Investigating...

cijothomas:
@MS-TimothyMothra Great catch! We really want to log most accurate error message if things are not working well, and fixing this further improves supportability.
Can you confirm that this issue is not affecting reliability (we are still doing retries as expected), but only our logging, and hence suportability is affected?

pharring:
@MS-TimothyMothra I don't think ServerTelemetryChannel has the bug because it does an `await` which unpacks the Exception.

MS-TimothyMothra:
@cijothomas No behavior is affected, we're only logging incorrect or incomplete exceptions

PR #1174, opened 2019-07-20
fix: InMemoryChannel does not log HttpClient exceptions

MS-TimothyMothra:
Fix Issue #1173 

Unpack `AggregateException`.
If `HttpRequestException`, concat all InnerException messages.

**This is the old message:**
>  Failed to send: One or more errors occurred..

**This is the new message:**  :)
> Failed to send: Type: 'System.Net.Http.HttpRequestException' Message: 'An error occurred while sending the request.;The underlying connection was closed: Could not establish trust relationship for the SSL/TLS secure channel.;The remote certificate is invalid according to the validation procedure.'.
----

### DIFF
--- a/PublicAPI/Microsoft.ApplicationInsights.dll/net46/PublicAPI.Unshipped.txt
+++ b/PublicAPI/Microsoft.ApplicationInsights.dll/net46/PublicAPI.Unshipped.txt
@@ -54,6 +54,3 @@ Microsoft.ApplicationInsights.Extensibility.SequencePropertyInitializer.Sequence
 Microsoft.ApplicationInsights.Extensibility.Implementation.OperationTelemetry.OperationTelemetry() -> void
 Microsoft.ApplicationInsights.Extensibility.OperationCorrelationTelemetryInitializer.OperationCorrelationTelemetryInitializer() -> void
 Microsoft.ApplicationInsights.Extensibility.Implementation.ApplicationId.DictionaryApplicationIdProvider.DictionaryApplicationIdProvider() -> void
-Microsoft.ApplicationInsights.Extensibility.Implementation.Internal.ExceptionExtensions
-static Microsoft.ApplicationInsights.Extensibility.Implementation.Internal.ExceptionExtensions.FlattenMessages(this System.Exception ex) -> string
-static Microsoft.ApplicationInsights.Extensibility.Implementation.Internal.ExceptionExtensions.ToLogString(this System.Exception ex) -> string
--- a/PublicAPI/Microsoft.ApplicationInsights.dll/net46/PublicAPI.Unshipped.txt
+++ b/PublicAPI/Microsoft.ApplicationInsights.dll/net46/PublicAPI.Unshipped.txt
@@ -54,3 +54,6 @@ Microsoft.ApplicationInsights.Extensibility.SequencePropertyInitializer.Sequence
 Microsoft.ApplicationInsights.Extensibility.Implementation.OperationTelemetry.OperationTelemetry() -> void
 Microsoft.ApplicationInsights.Extensibility.OperationCorrelationTelemetryInitializer.OperationCorrelationTelemetryInitializer() -> void
 Microsoft.ApplicationInsights.Extensibility.Implementation.ApplicationId.DictionaryApplicationIdProvider.DictionaryApplicationIdProvider() -> void
+Microsoft.ApplicationInsights.Extensibility.Implementation.Internal.ExceptionExtensions
+static Microsoft.ApplicationInsights.Extensibility.Implementation.Internal.ExceptionExtensions.FlattenMessages(this System.Exception ex) -> string
+static Microsoft.ApplicationInsights.Extensibility.Implementation.Internal.ExceptionExtensions.ToLogString(this System.Exception ex) -> string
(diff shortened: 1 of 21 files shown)


src/Microsoft.ApplicationInsights/Channel/InMemoryTransmitter.cs
```diff
@@ -138,9 +140,27 @@ private void DequeueAndSend(TimeSpan timeout)
                     // send request
                     this.Send(telemetryItems, timeout).Wait();
                 }
-                catch (Exception e)
+                catch (AggregateException aex)
```
cijothomas: https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/Test/ServerTelemetryChannel.Test/Shared.Tests/ServerTelemetryChannelE2ETests.cs#L162

There are few tests in SeverChannel E2e - they are not run in CI due to not stable when run parallel.

But you can run locally, and modiy it to validate the new exception unwinding based message. its okay to run/validate this when the cert issue is still occuring.

src/Microsoft.ApplicationInsights/Channel/InMemoryTransmitter.cs
```diff
@@ -138,9 +140,27 @@ private void DequeueAndSend(TimeSpan timeout)
                     // send request
                     this.Send(telemetryItems, timeout).Wait();
```
pharring: Consider:
```csharp
this.Send(telemetryItems, timeout).GetAwaiter().GetResult();
```

Then the rest of the code can stay the same (you won't have to unpack the AggregateException)

src/Microsoft.ApplicationInsights/Extensions/ExceptionExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public static string ToLogString(this Exception ex, bool includeStackTrace = false)
+        {
+            string msg = "Type: '{0}' Message: '{1}'";
+
+            var log = string.Format(CultureInfo.InvariantCulture, msg, ex.GetType().ToString(), ex.FlattenMessages());
+
+            if (includeStackTrace)
+            {
+                log = string.Concat(log, "StackTrace: ", ex.StackTrace);
```
Dmitry-Matveev: Do you need Environment.NewLine or something similar to separate it a bit from the messages?

src/Microsoft.ApplicationInsights/Properties/AssemblyInfo.cs
```diff
@@ -10,6 +10,8 @@
 [assembly: InternalsVisibleTo("Microsoft.ApplicationInsights.netcoreapp11.Tests" + AssemblyInfo.PublicKey)]
 [assembly: InternalsVisibleTo("Microsoft.ApplicationInsights.netcoreapp20.Tests" + AssemblyInfo.PublicKey)]
 
+
+[assembly: InternalsVisibleTo("Microsoft.AI.ServerTelemetryChannel" + AssemblyInfo.PublicKey)]
```
Dmitry-Matveev: We typically try to avoid InternalsVisibleTo if possible, let's continue our offline discussion to see if it's still necessary.

src/Microsoft.ApplicationInsights/Extensions/ExceptionExtensions.cs
MS-TimothyMothra: I don't think we have any rich text display for ETW messages. 
Could also make an argument is the StackTrace necessary to collect here?
I needed it initially to narrow down the exception last week before i started collecting the InnerExceptions.

What do you think?

Dmitry-Matveev: Since, it's sent to customer IKey, it's typically better to have smaller amount of bytes, while sufficient enough to diagnose the problem. Stack does not seem super-necessary if you show all inner messages already.

PublicAPI/Microsoft.ApplicationInsights.dll/net46/PublicAPI.Unshipped.txt
```diff
@@ -53,4 +53,7 @@ Microsoft.ApplicationInsights.Extensibility.Implementation.HttpWebResponseWrappe
 Microsoft.ApplicationInsights.Extensibility.SequencePropertyInitializer.SequencePropertyInitializer() -> void
 Microsoft.ApplicationInsights.Extensibility.Implementation.OperationTelemetry.OperationTelemetry() -> void
 Microsoft.ApplicationInsights.Extensibility.OperationCorrelationTelemetryInitializer.OperationCorrelationTelemetryInitializer() -> void
-Microsoft.ApplicationInsights.Extensibility.Implementation.ApplicationId.DictionaryApplicationIdProvider.DictionaryApplicationIdProvider() -> void
\ No newline at end of file
+Microsoft.ApplicationInsights.Extensibility.Implementation.ApplicationId.DictionaryApplicationIdProvider.DictionaryApplicationIdProvider() -> void
```
lmolkova: why do they have to be public APIs? If they are needed in ServerTelemteryChannel, could they be defined there instead?

MS-TimothyMothra: I talked to Dmitry offline about this. I'm not happy adding extension methods to the public api and I'm open to alternatives.

We have two channels and they're in different projects. Both channels need access to these extension methods.

lmolkova: copy-paste is one option. shared file compiled into both project is another one.

PR #1183, opened 2019-07-29
Updated version of System.Diagnostics.DiagnosticSource to 4.6.0-preview7

cijothomas:
Updated version of System.Diagnostics.DiagnosticSource to 4.6.0-preview7.19362.9. This will be updated to 4.6.0 stable when it releases, and before 2.11.0 stable release.

This is a pre-req to  leveraging new W3C capabilities in 4.6.0.

This also removed sdk from being cls compliant as the dependency (System.Diagnostics.DiagnosticSource) is no longer compliant to cls.

### DIFF
--- a/Test/Microsoft.ApplicationInsights.Test/Net45/Microsoft.ApplicationInsights.Net45.Tests.csproj
+++ b/Test/Microsoft.ApplicationInsights.Test/Net45/Microsoft.ApplicationInsights.Net45.Tests.csproj
@@ -31,7 +31,7 @@
     <PackageReference Include="MSTest.TestFramework" Version="1.4.0" />
     <PackageReference Include="CompareNETObjects" Version="4.59.0" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.5.0" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.6.0-preview7.19362.9" />
     <PackageReference Include="Moq" Version="4.10.0" />
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System" />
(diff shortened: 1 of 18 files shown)


src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CActivityExtensions.cs
```diff
@@ -29,7 +29,11 @@ public static class W3CActivityExtensions
         /// <param name="activity">Activity to generate W3C context on.</param>
         /// <returns>The same Activity for chaining.</returns>
         [EditorBrowsable(EditorBrowsableState.Never)]
+#pragma warning disable CS3001 // Argument type is not CLS-compliant
```
lmolkova: nit (and you don't have to fix it): you can list all warnings in one line separated by comma
`#pragma warning disable CS3001, CS3002`

Test/Microsoft.ApplicationInsights.Test/Net45/Microsoft.ApplicationInsights.Net45.Tests.csproj
```diff
@@ -31,7 +31,7 @@
     <PackageReference Include="MSTest.TestFramework" Version="1.4.0" />
     <PackageReference Include="CompareNETObjects" Version="4.59.0" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.5.0" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.6.0-preview7.19362.9" />
```
lmolkova: all these projects except Ms.AppInsights.csproj don't need a direct reference to DiagnosticSource (they'll get it with Ms.AppInsights dependency implicitly). Maybe we should remove the reference instead of updating it?

PR #1187, opened 2019-08-02
SDK connection string

TimothyMothra:
**Ruleset**
1. check for explicit endpoint (location is ignored)
2. check for endpoint suffix (location is optional)
3. use classic endpoint (location is ignored)

**Design**
- New property: `TelemetryConfiguration.ConnectionString`
- New method: `TelemetryConfiguration.SetConnectionString())`
Connection string needs to be parsed (via `EndpointProvider`) and validated.
This method will set `TelemetryConfiguration.InstrumentationKey`.
Resulting endpoints are set to `TelemetryConfiguration.Endpoints` (`EndpointContainer`).
This method will also update `TelemetryChannel.Endpoint`
- Contains about 220 new Tests.


src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/Constants.cs
```diff
@@ -0,0 +1,32 @@
...
+        internal const string BreezeEndpoint = "https://dc.services.visualstudio.com/";
+
+        /// <summary>Default endpoint for Live Metrics (aka QuickPulse).</summary>
+        internal const string LiveMetricsEndpoint = "https://rt.services.visualstudio.com/";
+
+        /// <summary>Default endpoint for Profiler.</summary>
+        internal const string ProfilerEndpoint = "https://agent.azureserviceprofiler.net/";
+
+        /// <summary>Default endpoint for Snapshot Debugger.</summary>
+        internal const string SnapshotEndpoint = "https://ppe.azureserviceprofiler.net/";
```
pharring: Please use agent.azureserviceprofiler.net (yes, it's the same as ProfilerEndpoint)

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,142 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Format: "location.prefix.suffix".
+        /// Example: "https:// westus2.dc.applicationinsights.azure.cn/".
+        /// </remarks>
+        /// <returns>Returns a <see cref="Uri"/> built from the inputs.</returns>
+        internal static Uri BuildUri(string prefix, string suffix, string location = null)
+        {
+            // Location and Host are user input fields and need to be checked for extra periods.
+            var trimPeriod = new char[] { '.' };
```
pharring: Make it a static readonly field to reduce allocations.

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/Endpoints/EndpointProviderTests.cs
```diff
@@ -0,0 +1,84 @@
...
+            Assert.AreEqual(expectedLiveMetricsEndpoint, liveMetricsTest.AbsoluteUri);
+
+            var profilerTest = endpoint.GetEndpoint(EndpointName.Profiler);
+            Assert.AreEqual(expectedProfilerEndpoint, profilerTest.AbsoluteUri);
+
+            var snapshotTest = endpoint.GetEndpoint(EndpointName.Snapshot);
+            Assert.AreEqual(expectedSnapshotEndpoint, snapshotTest.AbsoluteUri);
+        }
+    }
+}
```
pharring: Would be wise to test error paths too (badly-formed connection strings). e.g. no InstrumentationKey, duplicate keys, illegal characters.

pharring:
@jemodyms @clguimanMSFT FYI

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/Constants.cs
MS-TimothyMothra: Sure. Can you also reach out to Tomasz to update his spec?
All the other SDKs are coding off of that document.

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/Endpoints/EndpointProviderTests.cs
MS-TimothyMothra: Gladly :)
I've got a few already. I'm going to sync up with the other SDK devs and share test cases.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,178 @@
...
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                return new Dictionary<string, string>(0);
+            }
+
+            try
+            {
+                return value
+                    .Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries)
```
juliusl: >new[] { ';' } [](start = 27, length = 13)

Nit: I would make this a constant/static field.

src/ServerTelemetryChannel/ServerTelemetryChannel.cs
```diff
@@ -304,6 +304,8 @@ public void Initialize(TelemetryConfiguration configuration)
         {
             this.Transmitter.Initialize();
 
+            this.EndpointAddress = configuration.Endpoint.Ingestion.AbsoluteUri + "v2/track";
```
juliusl: >AbsoluteUri [](start = 68, length = 11)

Are you guaranteed a trailing slash?

juliusl: Here is a suggestion. Maybe I'm being too careful but since this is SDK code I'm reviewing to be cautious.

```
if (Uri.TryCreate(configuration.Endpoint.Ingestion, "v2/track", out Uri endpointAddress) {
  this.EndpointAddress = endpointAddress.AbsoluteUri;
}
else {
..
}
```

---
In reply to: [313673505](https://github.com/microsoft/ApplicationInsights-dotnet/pull/1187#discussion_r313673505) [](ancestors = 313673505)

TimothyMothra: In my testing, it appears that `(new Uri(someString)).AbsoluteUri` always appends the trailing slash to the input string. I can do some more testing on specifically this case to confirm. Thanks :)

TimothyMothra: > For hierarchical URIs, if the host is not terminated with a forward slash (/), one is added.

https://docs.microsoft.com/en-us/dotnet/api/system.uri?view=netframework-4.8

TimothyMothra: I wrote a boat load of tests to validate the expected behavior.
I think here it'll be safest to create a new Uri and then save the resulting AbsoluteUri. 
The Uri class will give us validation.

```
    [TestClass]
    public class UriTests
    {
        [TestMethod]
        public void TestMethod1()
        {
            var expected = "https://dc.services.visualstudio.com/";

            var test1 = new Uri("https://dc.services.visualstudio.com");
            var test2 = new Uri("https://dc.services.visualstudio.com/");

            Assert.AreEqual(expected, test1.AbsoluteUri);
            Assert.AreEqual(expected, test2.AbsoluteUri);
        }

        [TestMethod]
        public void TestMethod2()
        {
            var expected = "https://dc.services.visualstudio.com/v2/track";

            var host = new Uri("https://dc.services.visualstudio.com");

            var part1 = new Uri("v2/track", UriKind.Relative);
            var part2 = new Uri("/v2/track", UriKind.Relative);

            var test1 = new Uri(host, part1);
            var test2 = new Uri(host, part2);

            Assert.AreEqual(expected, test1.AbsoluteUri);
            Assert.AreEqual(expected, test2.AbsoluteUri);
        }

        [TestMethod]
        public void TestMethod3()
        {
            var expected = "https://dc.services.visualstudio.com/v2/track";

            var host = new Uri("https://dc.services.visualstudio.com");

            var test1 = new Uri(host, "v2/track");
            var test2 = new Uri(host, "/v2/track");

            Assert.AreEqual(expected, test1.AbsoluteUri);
            Assert.AreEqual(expected, test2.AbsoluteUri);
        }
    }
```

reyang:
@lzchen please take a look.

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationFactoryTest.cs
```diff
@@ -22,11 +22,13 @@
     public class TelemetryConfigurationFactoryTest
     {
         private const string EnvironmentVariableName = "APPINSIGHTS_INSTRUMENTATIONKEY";
+        private const string EnvironmentVariableConnectionString = "APPLICATIONINSIGHTS_CONNECTION_STRING";
```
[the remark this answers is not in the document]
TimothyMothra: LOL. we had a long email thread about this. :)
Microsoft doesn't have the trademark for "AppInsights" so we have to stop using that name.

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -288,6 +288,41 @@ internal TelemetryProcessorChain TelemetryProcessorChain
...
+                    ConnectionString = connectionString,
+                };
+
+                this.InstrumentationKey = endpointProvider.GetInstrumentationKey();
+
+                this.Endpoint = new EndpointContainer(endpointProvider);
+
+                foreach (var tSink in this.TelemetrySinks)
+                {
+                    tSink.TelemetryChannel.EndpointAddress = this.Endpoint.Ingestion.AbsoluteUri + "v2/track";
```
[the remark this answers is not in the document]
TimothyMothra: The context here is that the `Ingestion` needs to be used to construct both "v2/track" and also "api/profiles/{0}/appId" uris.
I made the decision to append that string only where it was needed. 

I'm investigating the AppId stuff today and just got confirmation today that we need to continue to support it. (It's due to be retired in the near future).
I think your suggestion about making these uri paths also configurable is justified. I need to sync up with the other SDK devs on this subject.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/ConnectionStringDuplicateKeyException.cs
```diff
@@ -0,0 +1,51 @@
...
+    using System;
+#if !NETSTANDARD1_3
+    using System.Runtime.Serialization;
+#endif
+
+    /// <summary>
+    /// This exception is used to notify the user that their Connection String has duplicate keys.
+    /// </summary>
+#if !NETSTANDARD1_3
+    [Serializable]
```
lmolkova: why should they be serializable?

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+
+            try
+            {
+                return value
+                    .Split(SplitSemicolon, StringSplitOptions.RemoveEmptyEntries)
+                    .Select(part => part.Split('='))
+                    .ToDictionary(split => split[0], split => split[1], StringComparer.OrdinalIgnoreCase);
+            }
+            catch (ArgumentException ex) when (ex.Message.StartsWith("An item with the same key has already been added.", StringComparison.Ordinal))
```
lmolkova: does custom exception type worth this logic?

If it does, I suggest to avoid creating particular exception based on the dictionary exception message and fill  dictionary without LINQ and with Contains checks.

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -288,6 +288,41 @@ internal TelemetryProcessorChain TelemetryProcessorChain
...
 
+        /// <summary>
+        /// This method will set the connection string, set the instrumentation key, validate the endpoints, and set the TelemetryChannel.Endpoint.
+        /// </summary>
+        /// <param name="connectionString">Users unique connection string. This must include the Instrumentation Key. Format: "key1=value1;key2=value2;key3=value3".</param>
+        /// <exception cref="ConnectionStringDuplicateKeyException">Thrown when the connection string has duplicate keys.</exception>
+        /// <exception cref="ConnectionStringInvalidDelimiterException">Thrown when the connection string is formatted incorrectly.</exception>
+        /// <exception cref="ConnectionStringInvalidEndpointException">Thrown when the endpoint that was provided is invalid. Exception contains more information.</exception>
+        /// <exception cref="ConnectionStringMissingInstrumentationKeyException">Thrown when the connection string does not contain an instrumentation key (required).</exception>
+        public void SetConnectionString(string connectionString)
```
lmolkova: it could be a setter on `ConnectionString` property - why does it have to be a separate method?

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/ConnectionStringDuplicateKeyException.cs
TimothyMothra: FxCop insisted on this. Our other custom exceptions have the same thing.

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -288,6 +288,41 @@ internal TelemetryProcessorChain TelemetryProcessorChain
...
+                    ConnectionString = connectionString,
+                };
+
+                this.InstrumentationKey = endpointProvider.GetInstrumentationKey();
+
+                this.Endpoint = new EndpointContainer(endpointProvider);
+
+                foreach (var tSink in this.TelemetrySinks)
+                {
+                    tSink.TelemetryChannel.EndpointAddress = new Uri(this.Endpoint.Ingestion, "v2/track").AbsoluteUri;
```
lmolkova: why all sinks should get the same endpoint? My guess is sinks ARE intended to send data to different places, e.g.: one to AppInsights, another to Geneva. (check out [this](https://1dsdocs.azurewebsites.net/articles/concepts/telemetry-sink.html) or [this](https://1dsdocs.azurewebsites.net/getting-started/csharp-getting_started.html#geneva-channel) )

 I'd assume only default one should have AI endpoint at all or only our channels - InMem and Server should get it.

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
TimothyMothra: 1. In my opinion, it feels wrong that a property setter would have so many side effects.
I thought that an explicit method call would be more appropriate in this case. 

2. In the TelemetryConfigurationFactory, we use reflection to set all the properties on the TelemetryConfiguration object.
In our requirements, a config set ConnectionString should always overwrite a config set InstrumentationKey.
These properties are set in the order that they are discovered, so it's possible that the config IKey could overwrite the value from the config ConnString.
In my opinion, having an explicit method to call makes it more obvious that the ConnectionString would overwrite all other properties.
(See TelemetryConfigurationFactory line 482)

src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
```diff
@@ -465,5 +462,31 @@ private static IEnumerable<XElement> GetPropertyDefinitions(XElement instanceDef
...
+            else if (PlatformSingleton.Current.TryGetEnvironmentVariable(InstrumentationKeyEnvironmentVariable, out string instrumentationKeyEnVar))
+            {
+                // TODO: LOG TO ETW INFORMATION: Instrumentation Key Environment Variable detected. 
+                configuration.InstrumentationKey = instrumentationKeyEnVar;
+            }
+            else if (configuration.ConnectionString != null)
+            {
+                // A set ConnectionString will have come from parsing the config file.
+                // Call Set() to force the connection string to be parsed.
+                configuration.SetConnectionString(configuration.ConnectionString);
```
lmolkova: is it setting the same value that is already set?

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
lmolkova: Well, check out .NET. e.g. HTTP client https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/HttpClient.cs#L97 or https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/HttpClientHandler.Windows.cs#L207

the side effects of  setters are fine. 
Having two ways to set ikey IS confusing, but I don't see how separate method helps with it. 
It's even more confusing to have inconsistent ways to set ikey vs connection string.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
TimothyMothra: This is needed for the case where Reflection overwrites the Ikey.

Requirements state that a config Connection String should always overwrite the config Instrumentation Key.

If the customer's config has IKey last, it will overwrite the Connection String and I need this to set it back.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/ConnectionStringDuplicateKeyException.cs
lmolkova: The one recommendation I found on MSDN is [this](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229064(v=vs.100)) "Do make exceptions serializable. An exception must be serializable to work correctly across application domain and remoting boundaries." 

which makes no sense - we do not cross-app-domian and I hope nobody does remoting in 2019.

But I guess this is fine to keep it for now and consider relaxing out fxcop rules so it does not require us to do meaningless things :)

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
pharring: I really shouldn't have to say this: Exception message can be localized. Don't filter exceptions based on their message!

src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
lmolkova: But I mean you take the value as it as set it as is. If this is needed, it does not look right

```csharp
            else if (configuration.ConnectionString != null)
                configuration.SetConnectionString(configuration.ConnectionString);
```

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,180 @@
...
+        /// Will also attempt to sanitize user input.
+        /// </summary>
+        /// <remarks>
+        /// Format: "location.prefix.suffix".
+        /// Example: "https:// westus2.dc.applicationinsights.azure.cn/".
+        /// </remarks>
+        /// <returns>Returns a <see cref="Uri"/> built from the inputs.</returns>
+        internal static Uri BuildUri(string prefix, string suffix, string location = null)
+        {
+            // Location and Host are user input fields and need to be checked for extra periods.
```
pharring: It would also be bad to have a slash in the prefix or the location (and probably also the suffix)

src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
lmolkova: Ah, I guess if ConnectionString setter is the same as SetConnectionString, it will be handled there

TimothyMothra: I agree this doesn't look right, but it's very much needed even if i move to setters!

The edge case I'm solving is that a customer can specify either property in any order in their config file.
Whichever one is last will overwrite the first.
But our requirements are that the Connection String should always take precedence, regardless of the order specified in the config file.

I've captured this in these test cases: https://github.com/microsoft/ApplicationInsights-dotnet/blob/12cff074bb047b8ebbf416ca7bf400de51784c40/Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationFactoryTest.cs#L171-L207

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
TimothyMothra: Crap, I didn't consider localization. :( 
I'll replace this tomorrow.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/ConnectionStringDuplicateKeyException.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <param name="message">The message that describes the error.</param>
+        public ConnectionStringDuplicateKeyException(string message) : base(message)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ConnectionStringDuplicateKeyException"/> class.
+        /// </summary>
+        /// <param name="message">The message that describes the error.</param>
+        /// <param name = "innerException" > The exception that is the cause of the current exception, or a null reference(Nothing in Visual Basic) if no inner exception is specified.</param>
```
reyang:
```suggestion
        /// <param name="innerException">The exception that is the cause of the current exception, or a null reference (Nothing in Visual Basic) if no inner exception is specified.</param>
```

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/ConnectionStringDuplicateKeyException.cs
```diff
@@ -0,0 +1,51 @@
...
+    using System.Runtime.Serialization;
+#endif
+
+    /// <summary>
+    /// This exception is used to notify the user that their Connection String has duplicate keys.
+    /// </summary>
+#if !NETSTANDARD1_3
+    [Serializable]
+#endif
+    public class ConnectionStringDuplicateKeyException : Exception
```
reyang: Curious, what's the benefit of having concrete exception types deriving from `Exception`?
What's the pros / cons of using one exception type for all the connection string validation failures?

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/Constants.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>Default endpoint for Live Metrics (aka QuickPulse).</summary>
+        internal const string LiveMetricsEndpoint = "https://rt.services.visualstudio.com/";
+
+        /// <summary>Default endpoint for Profiler.</summary>
+        internal const string ProfilerEndpoint = "https://agent.azureserviceprofiler.net/";
+
+        /// <summary>Default endpoint for Snapshot Debugger.</summary>
+        internal const string SnapshotEndpoint = "https://agent.azureserviceprofiler.net/";
+
+        /// <summary>Sub-domain for Breeze endpoint. (https:// dc.applicationinsights.azure.com/).</summary>
```
reyang:
```suggestion
        /// <summary>Sub-domain for Breeze endpoint. (https://dc.applicationinsights.azure.com/).</summary>
```

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/Constants.cs
```diff
@@ -0,0 +1,32 @@
...
+        internal const string BreezePrefix = "dc";
+
+        /// <summary>Sub-domain for Breeze endpoint. (https:// live.applicationinsights.azure.com/).</summary>
```
reyang: similar extra space in the other comments.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/Constants.cs
```diff
@@ -0,0 +1,32 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Endpoints
+{
+    /// <summary>
+    /// Endpoint Constants.
+    /// </summary>
+    internal static class Constants
+    {
+        /// <summary>Default endpoint for Breeze (aka Ingestion).</summary>
+        internal const string BreezeEndpoint = "https://dc.services.visualstudio.com/";
```
reyang: We probably want to call it "IngestionEndpoint" as Breeze is the code name.
No strong opinion though, up to you.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,180 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Endpoints
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    using Microsoft.ApplicationInsights.Common.Extensions;
+
+    /// <summary>
+    /// This class encapsulates parsing a connection string and returning an Endpoint's Uri.
```
reyang:
```suggestion
    /// This class encapsulates parsing a connection string and returning an Endpoint's URI.
```

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,180 @@
...
+        public string ConnectionString
+        {
+            get
+            {
+                return this.connectionString;
+            }
+
+            set
+            {
+                if (value != null && value.Length > ConnectionStringMaxLength)
```
reyang: minor suggestion on the flow:
```csharp
if (value == null)
{
    throw new ArgumentNullException(nameof(this.ConnectionString));
}
if (value.length > ConnectionStringMaxLength)
{
    ...
}
...
```

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,180 @@
...
+                return value;
+            }
+            else
+            {
+                throw new ConnectionStringMissingInstrumentationKeyException();
+            }
+        }
+
+        /// <summary>
+        /// Parse a string connection string and return a Dictionary.
```
reyang:
```suggestion
        /// Parse a connection string and return a Dictionary.
```

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
reyang: Even if there is no localization, determine the exception type by parsing exception message is a very dangerous idea - the fact that you're relying on the actual implementation behavior rather than the interface.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,180 @@
...
+            {
+                return new Dictionary<string, string>(0);
+            }
+
+            try
+            {
+                return value
+                    .Split(SplitSemicolon, StringSplitOptions.RemoveEmptyEntries)
+                    .Select(part => part.Split('='))
+                    .ToDictionary(split => split[0], split => split[1], StringComparer.OrdinalIgnoreCase);
```
reyang: This is buggy, what if the connection string is `key=foo=bar`?

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,180 @@
...
+            else
+            {
+                throw new ConnectionStringMissingInstrumentationKeyException();
+            }
+        }
+
+        /// <summary>
+        /// Parse a string connection string and return a Dictionary.
+        /// </summary>
+        /// <remarks>Example: "key1=value1;key2=value2;key3=value3".</remarks>
```
reyang: Do have ABNF definition for the connection string? Would like to see it in the doc string.

reyang: A general question around the connection string format across Azure - it will be great to follow the connection string convention - for example, if other Azure services allow `OWS` (Optional White Space), we should also allow it here.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Construct a Uri from the possible parts.
+        /// Will also attempt to sanitize user input.
+        /// </summary>
+        /// <remarks>
+        /// Format: "location.prefix.suffix".
+        /// Example: "https:// westus2.dc.applicationinsights.azure.cn/".
```
reyang:
```suggestion
        /// Example: "https://westus2.dc.applicationinsights.azure.cn/".
```

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
reyang: If there are existing functions/methods to do the validation, we should use them.
If there is no existing function, we should follow the RFC and perhaps use regular expression to match the ABNF.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/IEndpointProvider.cs
```diff
@@ -0,0 +1,13 @@
...
+
+    internal interface IEndpointProvider
+    {
+        string ConnectionString { get; set; }
+
+        Uri GetEndpoint(EndpointName endpointName);
+
+        string GetInstrumentationKey();
+    }
+}
```
reyang: nit, put empty line before EOF.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
lmolkova: > The edge case I'm solving is that a customer can specify either property in any order in their config file. Whichever one is last will overwrite the first.

May be fix this instead and do not allow setting ikey if ConnectionString is set?

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
TimothyMothra: The `Uri` class will do validation which is why I want to use it here.
If there's something easy to fix like a trailing slash or period I would like to fix it on the user's behalf.
But there's an argument to be made for failing fast and explicitly when there's bad input.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
TimothyMothra: I've got a dozen test cases on this method specifically, but I did not have THIS example. Thank you!

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/ConnectionStringDuplicateKeyException.cs
TimothyMothra: My intention here was to be explicit with what exception is being thrown.
Also In the unit tests, this lets me catch the type instead of evaluating the message.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,207 @@
...
+
+            var endpointMeta = EndpointMetaAttribute.GetAttribute(endpointName);
+
+            try
+            {
+                if (this.connectionStringParsed.TryGetValue(endpointMeta.ExplicitName, out string explicitEndpoint))
+                {
+                    try
+                    {
+                        return new Uri(explicitEndpoint);
```
lmolkova:
```csharp
if (Uri.TryParse(explicitEndpoint, out var uri)) 
{
  return uri;
}
else
{
  throw
}
```

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
```diff
@@ -0,0 +1,207 @@
...
+        /// <summary>
+        /// Construct a Uri from the possible parts.
+        /// Will also attempt to sanitize user input.
+        /// </summary>
+        /// <remarks>
+        /// Format: "location.prefix.suffix".
+        /// Example: "https://westus2.dc.applicationinsights.azure.cn/".
+        /// </remarks>
+        /// <returns>Returns a <see cref="Uri"/> built from the inputs.</returns>
+        internal static Uri BuildUri(string prefix, string suffix, string location = null)
```
lmolkova: bool TryBuildUri(..., out var uri) and rather return false. And above, where you call this method check the return value

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
lmolkova: tryparse is MUCH faster than try/catch and more readable and clean

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
lmolkova: I still have the same concern :)

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/ConnectionStringDuplicateKeyException.cs
reyang: What's the benefit of telling the concrete error from unit test?
For example, we could potentially do:
1. feed the unit test with a valid connection string, expect a valid result.
2. feed the unit test with an invalid connection string, expect an `ArgumentException`.

TimothyMothra: It's no benefit to the unit test.
The test is validating that a meaningful exception was emitted. 

I can rework this to use `ArgumentException`.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointProvider.cs
TimothyMothra: I spoke with Reiley offline. We're not going to block the PR for this, but we will deliver an ABNF definition in the final documentation. I added this to the OneNote.

reyang: Yep, this is not blocking. Thanks!

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationTest.cs
```diff
@@ -285,6 +288,210 @@ public void InstrumentationKeyCanBeSetToProgrammaticallyDefineInstrumentationKey
...
+        [TestCategory("ConnectionString")]
+        public void E2E_DefaultScenario_WithConnectionString()
+        {
+            var tConfig = new TelemetryConfiguration
+            {
+                ConnectionString = "InstrumentationKey=00000000-0000-0000-0000-000000000000;IngestionEndpoint=https://127.0.0.1/"
+            };
+
+            Assert.AreEqual("https://127.0.0.1/v2/track", tConfig.DefaultTelemetrySink.TelemetryChannel.EndpointAddress);
+        }
```
TimothyMothra: @lmolkova , @cijothomas , @Dmitry-Matveev 
I've got 3 "E2E" examples here of creating a TelemetryConfiguration and setting the TelemteryChannel.
Can you think of any scenarios that I might have missed?

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationTest.cs
```diff
@@ -285,6 +288,210 @@ public void InstrumentationKeyCanBeSetToProgrammaticallyDefineInstrumentationKey
...
+            Assert.AreEqual("https://dc.services.visualstudio.com/v2/track", tConfig.DefaultTelemetrySink.TelemetryChannel.EndpointAddress);
+        }
+
+        [TestMethod]
+        [TestCategory("ConnectionString")]
+        public void E2E_DefaultScenario_ConfigurationConstructor()
+        {
+            var tConfig = new TelemetryConfiguration("00000000-0000-0000-0000-000000000000", new InMemoryChannel());
+
+            Assert.AreEqual("https://dc.services.visualstudio.com/v2/track", tConfig.DefaultTelemetrySink.TelemetryChannel.EndpointAddress);
```
cijothomas: should you validate ikey as well?

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationTest.cs
```diff
@@ -285,6 +288,210 @@ public void InstrumentationKeyCanBeSetToProgrammaticallyDefineInstrumentationKey
...
+        [TestMethod]
+        [TestCategory("ConnectionString")]
+        public void E2E_DefaultScenario_WithConnectionString()
+        {
+            var tConfig = new TelemetryConfiguration
+            {
+                ConnectionString = "InstrumentationKey=00000000-0000-0000-0000-000000000000;IngestionEndpoint=https://127.0.0.1/"
+            };
+
+            Assert.AreEqual("https://127.0.0.1/v2/track", tConfig.DefaultTelemetrySink.TelemetryChannel.EndpointAddress);
```
cijothomas: validate ikey?

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationTest.cs
TimothyMothra: thanks. i was focused on confirming that the endpoint is propagated correctly. 
this is done.

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationTest.cs
TimothyMothra: done.

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationTest.cs
```diff
@@ -285,6 +288,221 @@ public void InstrumentationKeyCanBeSetToProgrammaticallyDefineInstrumentationKey
...
+            Assert.AreEqual("https://dc.services.visualstudio.com/", configuration.Endpoint.Ingestion.AbsoluteUri);
+            Assert.AreEqual("https://dc.services.visualstudio.com/v2/track", channel.EndpointAddress);
+        }
+
+        [TestMethod]
+        [TestCategory("ConnectionString")]
+        public void VerifySetConnectionString_SetsChannelExpliticEndpoint()
+        {
+            var explicitEndpoint = "https://127.0.0.1/";
+            var connectionString = $"InstrumentationKey=00000000-0000-0000-0000-000000000000;IngestionEndpoint={explicitEndpoint}";
```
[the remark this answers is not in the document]
TimothyMothra: fixed.

TimothyMothra:
@lmolkova Can you please re-review? You're currently blocking this PR. :)

src/ServerTelemetryChannel/ServerTelemetryChannel.cs
```diff
@@ -304,6 +304,8 @@ public void Initialize(TelemetryConfiguration configuration)
         {
             this.Transmitter.Initialize();
 
+            this.EndpointAddress = new Uri(configuration.Endpoint.Ingestion, "v2/track").AbsoluteUri;
```
cijothomas: why construct a new Uri and back to string?  Config.EndPoint.Ingestion.AbsoluteUrl + "v2/track" would do. Validation for ingestion would have already happened .

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -258,7 +249,67 @@ public ITelemetryChannel TelemetryChannel
...
+
+            set
+            {
+                this.applicationIdProvider = value;
+                SetApplicationIdEndpoint(this.applicationIdProvider, this.Endpoint.FormattedApplicationIdEndpoint);
+            }
+        }
+
+        /// <summary>
+        /// Gets the Endpoint Controller responsible for making service endpoints available.
```
cijothomas: Controller -> Container

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -258,7 +249,67 @@ public ITelemetryChannel TelemetryChannel
...
+        /// </summary>
+        public EndpointContainer Endpoint { get; private set; } = new EndpointContainer(new EndpointProvider());
```
cijothomas: rename variable to say EndPointContainer

src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
```diff
@@ -469,5 +466,26 @@ private static IEnumerable<XElement> GetPropertyDefinitions(XElement instanceDef
...
+            {
+                CoreEventSource.Log.TelemetryConfigurationFactoryFoundInstrumentationKeyEnvironmentVariable(variableName: InstrumentationKeyEnvironmentVariable);
+                configuration.InstrumentationKey = instrumentationKeyEnVar;
+            }
+
+            // SANITY CHECK
+            if (configuration.InstrumentationKey == null)
+            {
+                // LOG TO ETW WARNING: No Instrumentation Key found. Needs to be manually set.
+                CoreEventSource.Log.TelemetryConfigurationFactoryNoInstrumentationKey();
```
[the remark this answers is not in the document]
TimothyMothra: I thought Warning would be appropriate here. This is the factory code that would get executed before any customer code which might be setting the Ikey.

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -371,6 +422,47 @@ internal MetricManager GetMetricManager(bool createIfNotExists)
...
+            }
+        }
+
+        /// <summary>
+        /// This will check the TelemetryChannel and attempt to set the endpoint.
+        /// This only supports our first party providers <see cref="InMemoryChannel"/> and ServerTelemetryChannel.
+        /// </summary>
+        /// <param name="channel">TelemetryChannel to set.</param>
+        /// <param name="endpoint">Endpoint value to set.</param>
+        private static void SetTelemetryChannelEndpoint(ITelemetryChannel channel, string endpoint)
```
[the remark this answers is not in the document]
TimothyMothra: @lmolkova specifically did not want me doing that.
I was told it is valid for a GenevaChannel to send to their own unique endpoint.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointContainer.cs
```diff
@@ -0,0 +1,36 @@
...
+        public Uri Live { get; private set; }
+
+        /// <summary>Gets the endpoint for the Profiler service.</summary>
+        public Uri Profiler { get; private set; }
+
+        /// <summary>Gets the endpoint for the Snapshot service.</summary>
+        public Uri Snapshot { get; private set; }
+
+        /// <summary>Gets the fully formatted endpoint for the ingestion service.</summary>
+        internal string FormattedIngestionEndpoint => new Uri(this.Ingestion, "v2/track").AbsoluteUri;
```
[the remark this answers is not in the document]
TimothyMothra: I'm relying on the Uri library to validate endpoints so I'm intentionally initializing a Uri first.

src/ServerTelemetryChannel/ServerTelemetryChannel.cs
TimothyMothra: I'm relying on the Uri library to validate endpoints so I'm intentionally initializing a Uri first.

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -258,7 +249,67 @@ public ITelemetryChannel TelemetryChannel
...
+                    this.connectionString = value ?? throw new ArgumentNullException(nameof(this.ConnectionString));
+
+                    var endpointProvider = new EndpointProvider
+                    {
+                        ConnectionString = value,
+                    };
+
+                    this.InstrumentationKey = endpointProvider.GetInstrumentationKey();
+
+                    this.Endpoint = new EndpointContainer(endpointProvider);
```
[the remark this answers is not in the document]
reyang: https://docs.microsoft.com/en-us/dotnet/api/system.servicemodel.description.serviceendpoint

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
TimothyMothra: Can you explain why?
I don't have a strong opinion, but I felt that "Endpoint" would keep it simpler

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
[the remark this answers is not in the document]
TimothyMothra: Can you help me understand why?
I don't have a strong opinion, but I felt that "Endpoint" would keep it simpler

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
TimothyMothra: done

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
TimothyMothra: done

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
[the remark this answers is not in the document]
TimothyMothra: done.

src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
cijothomas: so that variable name hints about its type.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointContainer.cs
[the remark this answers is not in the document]
cijothomas: this.Ingestion is already validated as Uri.
Are you ensuring that appending "v2/track" to a valid uri make it invalid? If yes, do the same for ApplicationIDEndpoint as well.

src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
```diff
@@ -469,5 +466,26 @@ private static IEnumerable<XElement> GetPropertyDefinitions(XElement instanceDef
...
+                configuration.ConnectionString = connectionStringEnVar;
+            }
+            else if (PlatformSingleton.Current.TryGetEnvironmentVariable(InstrumentationKeyEnvironmentVariable, out string instrumentationKeyEnVar))
+            {
+                CoreEventSource.Log.TelemetryConfigurationFactoryFoundInstrumentationKeyEnvironmentVariable(variableName: InstrumentationKeyEnvironmentVariable);
+                configuration.InstrumentationKey = instrumentationKeyEnVar;
+            }
+
+            // SANITY CHECK
+            if (configuration.InstrumentationKey == null)
```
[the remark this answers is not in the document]
TimothyMothra: I don't think this is necessary, The Connection String is optional, but the Ikey is required.
If this sanity check fails, that means neither were set but only the Ikey was required.

Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationTest.cs
```diff
@@ -285,6 +288,242 @@ public void InstrumentationKeyCanBeSetToProgrammaticallyDefineInstrumentationKey
...
+            };
+
+            Assert.AreEqual("00000000-0000-0000-0000-000000000000", configuration.InstrumentationKey);
+            Assert.AreEqual(explicitEndpoint, configuration.Endpoint.Ingestion.AbsoluteUri);
+            Assert.AreEqual(explicitEndpoint + "v2/track", channel.EndpointAddress);
+        }
+
+        [TestMethod]
+        [TestCategory("ConnectionString")]
+        public void E2E_DefaultScenario()
```
cijothomas: minor - these tests are just validating TelemetryConfiguration class, and is no way an E2E test. (in my opinion, E2E would involve TelemetryClient instance and calling Track on it .

src/Microsoft.ApplicationInsights/Extensibility/Implementation/Endpoints/EndpointContainer.cs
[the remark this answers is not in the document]
TimothyMothra: Oh, thank you. ApplicationIdEndpoint should absolutely be doing the same thing.

PR #1193, opened 2019-08-16
Make BASE SDK Use W3C Correlation by default.

cijothomas:
Application Insights used Request-ID based (aka Hierarchical) model to propagate correlation headers between services. i.e it parsed the header 'Request-Id' for incoming correlation ids, and populated the same header for propagating correlations ids to outbound services.

https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/HttpCorrelationProtocol.md

While application insights SDKs also contained the capability to use the new W3C Standard for Distributed Tracing (https://www.w3.org/TR/trace-context/), but this was not enabled by default. Also the implementation had to work around the fact the the .NET correlation primitive `Activity` did not natively support W3C based correlation. (it only supported Hierrachial model) So SDK used Activity.Tags for storing W3C concepts like TraceID, Spand ID, etc which was inefficient.

Starting with System.Diagnostics.DiagnosticSource package 4.6.0, which this SDK is taking dependency on, Activity now natively support W3C Tracing Distributed Tracing. Activity now has methods like TraceID (https://www.w3.org/TR/trace-context/#trace-id), SpanID(https://www.w3.org/TR/trace-context/#parent-id) which make it fully W3C Tracing aware.

Though Activity added native support for W3C based correlation, the default format is still Hierarchical.

**This PR introduces the following changes:**

1. Introduce a new public field on TelemetryConfiguration named EnableW3CCorrelation which will be set to true by default. True on this field sets the default ID Format used on System.Diagnostics.Activity to be W3C.
Users can set the above to be false, to make SDK fall-back to the old way of correlation using Hierarchical ids.
_**If any issues are faced with new format, use this to fall-back to original behavior**_

2. OperationCorrelationTelemetryInitializer, which does the in-proc correlation of telemetry is now made W3C-Tracing aware.  (Described more at the end of this)

3. Few classes and methods like W3CCorrelationTelemetryInitializer, W3CActivityExtensions, which were used previously to support W3C tracing are no longer needed and are marked obsolete. Some of them methods are not relevant and are made no-ops.  
 
**OperationCorrelationTelemetryInitializer**
This class was responsible for correlating telemetry items to parent telemetry item (either Request Telemetry or Dependency Telemetry). This is done by setting their operation.id and operation.parentid from Activity.Current. More specifically, operation.id was set to current Activity's RootID, following (https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/HttpCorrelationProtocol.md), and operation.ParentId was set to current Activity's ID.

With the new changes in this PR, the default behavior is to set IDs in the W3C Tracing format. Specifically, this means operation.ID will be set to current Activity's TraceID, and operation.ParentID will be set to {!TraceID.SpanId.} from current Activity. This format for parent id is done to ensure backward compatibility with older SDKs excepting ids in Hierarchical format.

**TelemetryClient extension methods StartOperation**
These methods provided a way for users to track a custom operation telemetry (as opposed to auto collected Request/Dependency). This will now generate IDs in the new W3C compliant format. 
If you used to provide your own operationid for the StartOperation() overload which accepts operationid, it will be used only if they are compatible with W3C Ids. (16 byte hex encoded string) Otherwisde a new compatible ID is generated, and the provided id will be stored in custom property "ai_legacyRootId". ApplicationInsights UI (app map/e2e transaction views) are aware of this custom field, and hence there should be no degradation in the final experience.

Read more about W3C Standard For Distributed Tracing here - https://www.w3.org/TR/trace-context/

There will be related changes in other components of application insights SDK - in the Request Collection Module, and in DependencyCollection Modules.

Additional notes:
All changes are backward compatible - components instrumented with this new SDK will continue to work with components instrumented with older version of the SDK.


src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CUtilities.cs
```diff
@@ -12,6 +13,27 @@
...
     {
         private static readonly uint[] Lookup32 = CreateLookup32();
+        private static readonly Regex TraceIdRegex = new Regex("^[a-f0-9]{32}$", RegexOptions.Compiled);
+
+        /// <summary>
+        /// Constructs a Telemetry ID from given traceid and span id in the format |traceid.spanid.
+        /// This is the format used by Application Insights.        
+        /// </summary>
+        /// <returns>constructed Telemetry ID.</returns>
+        public static string FormatTelemetryId(string traceId, string spanId)
```
lmolkova: don't make it public - anyone can copy-paste this code

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CUtilities.cs
```diff
@@ -12,6 +13,27 @@
...
+        {
+            return string.Concat("|", traceId, ".", spanId, ".");
+        }
+
+        /// <summary>
+        /// Checks if the given string is a valid trace-id as per W3C Specs.
+        /// https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md#trace-id .
+        /// </summary>
+        /// <returns>true if valid w3c trace id, otherwise false.</returns>
+        public static bool IsCompatibleW3CTraceID(string traceId)
```
lmolkova: don't make it public. copy-paste is better. also, it should be more performant to check it without regex (each char in value range and length is 32)

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -133,11 +138,29 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
                 if (parentActivity == null)
                 {
                     // telemetryContext.Id is always set: if it was null, it is set to opTelemetry.Id and opTelemetry.Id is never null
-                    operationActivity.SetParentId(telemetryContext.Id);
+                    if (Activity.DefaultIdFormat == ActivityIdFormat.W3C)
+                    {
+                        if(W3CUtilities.IsCompatibleW3CTraceID(telemetryContext.Id))
+                        {
+                            operationActivity.SetParentId(ActivityTraceId.CreateFromString(telemetryContext.Id.AsSpan()), ActivitySpanId.CreateRandom());
```
lmolkova: don't create random parent span id - use invalid 00000... so we can later know there was no valid parent.

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -133,11 +138,29 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
+                    {
+                        operationActivity.SetParentId(telemetryContext.Id);
+                    }
                 }
 
                 operationActivity.Start();
-                operationTelemetry.Id = operationActivity.Id;
+                if (operationActivity.IdFormat == ActivityIdFormat.W3C)
+                {
+                    operationTelemetry.Id = W3CUtilities.FormatTelemetryId(operationActivity.TraceId.ToHexString(), operationActivity.SpanId.ToHexString());
```
lmolkova: cache traceId.ToHexString()

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CActivityExtensions.cs
```diff
@@ -1,385 +0,0 @@
-﻿namespace Microsoft.ApplicationInsights.Extensibility.W3C
```
lmolkova: we cannot remove it - it's public. It could proxy calls to real Activity API when make sense and be noop in other cases

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CActivityExtensions.cs
```diff
@@ -1,385 +0,0 @@
...
-    using System.Text.RegularExpressions;
-    using Microsoft.ApplicationInsights.Channel;
-    using Microsoft.ApplicationInsights.DataContracts;
-    using Microsoft.ApplicationInsights.Extensibility.Implementation;
-
-    /// <summary>
-    /// Extends Activity to support W3C distributed tracing standard.
-    /// </summary>
-    [EditorBrowsable(EditorBrowsableState.Never)]
-    public static class W3CActivityExtensions
```
lmolkova: we cannot remove it - it's public. It could proxy calls to real Activity API when make sense and be noop in other cases

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CUtilities.cs
cijothomas: sticking with Regex now, as this is not a hot-path code. (only used in StartOperation(with operationid) which is not common). let me know if this is a common use case.

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CActivityExtensions.cs
```diff
@@ -48,9 +37,10 @@ public static Activity GenerateW3CContext(this Activity activity)
         /// <param name="activity">Activity to check.</param>
         /// <returns>True if Activity has W3C properties, false otherwise.</returns>
         [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("Activity from System.Diagnostics.DiagnosticSource 4.6.0 onwards natively support W3C.")]
         public static bool IsW3CActivity(this Activity activity)
-        {
-            return activity != null && activity.Tags.Any(t => t.Key == W3CConstants.TraceIdTag);
+        {            
+            return activity != null && Activity.DefaultIdFormat.Equals(ActivityIdFormat.W3C);
```
lmolkova: nit: check the format on the Activity instance rather than static. not really important but may help one day

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CActivityExtensions.cs
```diff
@@ -79,34 +62,22 @@ public static Activity UpdateContextOnActivity(this Activity activity)
...
-                    case W3CConstants.VersionTag:
-                        version = tag.Value;
-                        break;
-                    case W3CConstants.SampledTag:
-                        sampled = tag.Value;
-                        break;
-                }
-            }
+            // no-op as there is no native way to obtain TraceParent from Activity. 
+            return string.Empty;
```
lmolkova:
```suggestion
            return activity.Id
```

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CUtilities.cs
lmolkova: this will be common in web/aspnet core, but there we can implement more efficient check.

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -53,7 +54,33 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
+                        if (W3CUtilities.IsCompatibleW3CTraceID(operationId))
+                        {
+                            // If the user provided operationid is W3C Compatible, use it.
+                            operationTelemetry.Context.Operation.Id = operationId;
+                        }
+                        else
+                        {
+                            // If user provided operationid is not W3C compatible, generate a new one instead.
+                            // and store supplied value inside customproperty.
+                            operationTelemetry.Context.Operation.Id = W3CUtilities.GenerateTraceId();
```
lmolkova:
```suggestion
                            operationTelemetry.Context.Operation.Id = ActivityTraceId.CreateRandom().ToHexString()
```

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -132,12 +159,28 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
                 if (parentActivity == null)
                 {
-                    // telemetryContext.Id is always set: if it was null, it is set to opTelemetry.Id and opTelemetry.Id is never null
-                    operationActivity.SetParentId(telemetryContext.Id);
+                    // telemetryContext.Id is always set: if it was null, it is set to newly generated W3C compatible TraceID
+                    if (Activity.DefaultIdFormat == ActivityIdFormat.W3C)
+                    {
+                        // There is no need of checking if TelemetryContext.ID is W3C Compatible. It is always set to 
+                        // W3C compatible id. Even user supplied non-compatible ID is ignored.                                                
+                        operationActivity.SetParentId(string.Join("-", W3CConstants.DefaultVersion, telemetryContext.Id, W3CConstants.InvalidSpanID, W3CConstants.DefaultTraceFlag));
```
lmolkova: to make it cleaner you can do this:
`SetParentId(ActivityTraceId.CreateFromString(telemetryContext.Id.AsSpan()), default, ActivityTraceFlags.None);`

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -132,12 +159,28 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
+                    {
+                        operationActivity.SetParentId(telemetryContext.Id);
+                    }
                 }
 
                 operationActivity.Start();
-                operationTelemetry.Id = operationActivity.Id;
+                if (operationActivity.IdFormat == ActivityIdFormat.W3C)
+                {
+                    operationTelemetry.Id = W3CUtilities.FormatTelemetryId(operationActivity.TraceId.ToHexString(), operationActivity.SpanId.ToHexString());
```
lmolkova:
```suggestion
                    operationTelemetry.Id = W3CUtilities.FormatTelemetryId(telemetryContext.Id, operationActivity.SpanId.ToHexString());
```

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -255,10 +298,18 @@ private static T ActivityToTelemetry<T>(Activity activity) where T : OperationTe
...
             OperationContext operationContext = telemetry.Context.Operation;
-            operationContext.Name = activity.GetOperationName();
-            operationContext.Id = activity.RootId;
+            operationContext.Name = activity.GetOperationName();            
             operationContext.ParentId = activity.ParentId;
-            telemetry.Id = activity.Id;
+            if (activity.IdFormat == ActivityIdFormat.W3C)
+            {
+                operationContext.Id = activity.TraceId.ToHexString();                
+                telemetry.Id = W3CUtilities.FormatTelemetryId(activity.TraceId.ToHexString(), activity.SpanId.ToHexString());
```
lmolkova:
```suggestion
                telemetry.Id = W3CUtilities.FormatTelemetryId(operationContext.Id.ToHexString(), activity.SpanId.ToHexString());
```

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CUtilities.cs
```diff
@@ -27,6 +29,26 @@ public static string GenerateTraceId()
...
+        /// <summary>
+        /// Checks if the given string is a valid trace-id as per W3C Specs.
+        /// https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md#trace-id .
+        /// </summary>
+        /// <returns>true if valid w3c trace id, otherwise false.</returns>
+        internal static bool IsCompatibleW3CTraceID(string traceId)
+        {
+            return TraceIdRegex.IsMatch(traceId);
+        }
+
```
lmolkova: please update GenerateId methods below to use ActivityTraceId/SpanId.CreateRandom()

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
cijothomas: Will modify the W3Utilities method to internally use ActivityTraceId.CreateRandom() if Activity is available. Else fallback to current implementation.

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
cijothomas: I opted for this because I cannot find a way to create a Default SpanID. as SpanId.CreateFromString will throw for all 0s. An alternate is to do new Activity().SpanId, which gives a SpanId with all zeroes, and cache this. Will discuss in person about this.

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CUtilities.cs
cijothomas: We need to keep all code so as to fall back if Activity is not available. (standalone use case)

cijothomas: generateid methods modified to leverage CreateRandom() if available.

PublicAPI/Microsoft.ApplicationInsights.dll/netstandard2.0/PublicAPI.Shipped.txt
```diff
@@ -1,3 +1,5 @@
+Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration.EnableW3CCorrelation.set -> void
+Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration.EnableW3CCorrelation.get -> bool
```
TimothyMothra: shouldn't this be the Unshipped doc?

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
lmolkova: please deprecate the old one and use new one here

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -103,10 +130,10 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
             }
 
             // If the operation is not executing in the context of any other operation
-            // set its name and id as a context (root) operation name and id
+            // set its name and id as a context (root) operation name and generate new W3C compatible id
             if (string.IsNullOrEmpty(telemetryContext.Id))
             {
-                telemetryContext.Id = operationTelemetry.Id;
+                telemetryContext.Id = W3CUtilities.GenerateTraceId();
```
lmolkova: please use ActivityTraceId.CreateRandom()

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -132,12 +159,31 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
                 if (parentActivity == null)
                 {
-                    // telemetryContext.Id is always set: if it was null, it is set to opTelemetry.Id and opTelemetry.Id is never null
-                    operationActivity.SetParentId(telemetryContext.Id);
+                    // telemetryContext.Id is always set: if it was null, it is set to newly generated W3C compatible TraceID
+                    if (Activity.DefaultIdFormat == ActivityIdFormat.W3C)
+                    {
+                        // There is no need of checking if TelemetryContext.ID is W3C Compatible. It is always set to 
+                        // W3C compatible id. Even user supplied non-compatible ID is ignored.                                                
+                        operationActivity.SetParentId(string.Join("-", W3CConstants.DefaultVersion, telemetryContext.Id, W3CConstants.InvalidSpanID, W3CConstants.DefaultTraceFlag));
```
lmolkova: still SetParentId(ActivityTraceId.CreateFromString(telemetryContext.Id.AsSpan()), default, ActivityTraceFlags.None) is cleaner and does not throw, please use that one

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -53,7 +54,33 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
 
             if (string.IsNullOrEmpty(operationTelemetry.Context.Operation.Id) && !string.IsNullOrEmpty(operationId))
             {
-                operationTelemetry.Context.Operation.Id = operationId;
+                var isActivityAvailable = ActivityExtensions.TryRun(() =>
```
lmolkova: we never bothered with making StartOperation work without Activity (it does not work today) - I don't think we should start now. 

The motivation is: from C# noone expects to call AI SDK without installing nugets and depednencies. From script like powershell (which was one scenario that we wanted to address wihtout DiagnosticSource.dll) it's extremely hard to call generic methods, so the expectation is: noone will call startoperation from powershell and no reason to bother to make it resilent for invalid usage,

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CUtilities.cs
```diff
@@ -12,6 +14,7 @@
     public static class W3CUtilities
```
lmolkova: please deperecate this one (or at least all methods except FormatId.. ) and don't bother with missing DiagnosticSource.dll here - it should not be handled here

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
lmolkova: ah, ok, we bothered in some scenarios. I feel quite bad that we overcomplicate code so much just to satisfy some illegal scenario....

cijothomas: The standalone Unittest was doing tc.StartOperation...

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CUtilities.cs
cijothomas: GenerateTraceId() is the only public method. Marked it obsolete. Rest all removed/retained as needed, but are internal only.

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CActivityExtensions.cs
```diff
@@ -48,9 +37,10 @@ public static Activity GenerateW3CContext(this Activity activity)
         /// <param name="activity">Activity to check.</param>
         /// <returns>True if Activity has W3C properties, false otherwise.</returns>
         [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("Activity from System.Diagnostics.DiagnosticSource 4.6.0 onwards natively support W3C.")]
         public static bool IsW3CActivity(this Activity activity)
         {
-            return activity != null && activity.Tags.Any(t => t.Key == W3CConstants.TraceIdTag);
+            return activity != null && activity.IdFormat.Equals(ActivityIdFormat.W3C);
```
lmolkova: Id foramt is enum
```suggestion
            return activity != null && activity.IdFormat == ActivityIdFormat.W3C;
```

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CActivityExtensions.cs
```diff
@@ -79,34 +62,11 @@ public static Activity UpdateContextOnActivity(this Activity activity)
...
-                }
-            }
-
-            if (traceId == null || spanId == null || version == null || sampled == null)
-            {
-                return null;
-            }
-
-            return string.Join("-", version, traceId, spanId, sampled);
+            // returning activity id as there is no native way to obtain exact TraceParent header from Activity.
```
lmolkova: comment is not correct, activity.Id returns id formatted as  traceparent - it's a native way to format it

src/Microsoft.ApplicationInsights/Extensibility/W3C/W3CConstants.cs
```diff
@@ -1,60 +1,13 @@
...
-        /// <summary>
-        /// Recorded and requested sampled flag value.
-        /// </summary>
-        internal const string TraceFlagRecordedAndRequested = "03";
-
-        /// <summary>
-        /// Requested trace flag.
-        /// </summary>
-        internal const byte RequestedTraceFlag = 1;
-
```
lmolkova: is `LegacyRequestIdProperty` still used anywhere? I guess no and it can be removed

Issue #1205, opened 2019-09-06
When StartOperation is called in scope of Activity, but user provides different operationId, parentId, custom ids are ignored


lmolkova:
https://github.com/microsoft/ApplicationInsights-dotnet/blob/c46ebc61365f97527f79c1acf0a958b67907db28/src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs#L169

PR #1206, opened 2019-09-07
Fix StartOperation when user provides custom Ids in scope of unrelated Activity

lmolkova:
Fix Issue #1205

### DIFF
--- a/src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
+++ b/src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
@@ -157,11 +157,8 @@
 
                 if (idsAssignedByUser)
                 {
-                    // telemetryContext.Id is always set: if it was null, it is set to newly generated W3C compatible TraceID
                     if (Activity.DefaultIdFormat == ActivityIdFormat.W3C)
                     {
-                        // There is no need of checking if TelemetryContext.ID is W3C Compatible. It is always set to 
-                        // W3C compatible id. Even user supplied non-compatible ID is ignored.
                         if (W3CUtilities.IsCompatibleW3CTraceId(telemetryContext.Id))
                         {
                             // If the user provided operationId is W3C Compatible, use it.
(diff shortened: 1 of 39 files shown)


src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -159,14 +155,15 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
...
-                if (parentActivity == null)
+                if (idsAssignedByUser)
                 {
                     // telemetryContext.Id is always set: if it was null, it is set to newly generated W3C compatible TraceID
                     if (Activity.DefaultIdFormat == ActivityIdFormat.W3C)
                     {
                         // There is no need of checking if TelemetryContext.ID is W3C Compatible. It is always set to 
                         // W3C compatible id. Even user supplied non-compatible ID is ignored.
-                        operationActivity.SetParentId(ActivityTraceId.CreateFromString(telemetryContext.Id.AsSpan()), default(ActivitySpanId), ActivityTraceFlags.None);
+                        operationActivity.SetParentId(ActivityTraceId.CreateFromString(telemetryContext.Id.AsSpan()),
```
cijothomas: StartOperation<T>(this TelemetryClient telemetryClient, T operationTelemetry) is public, so we need to validate if operationTelemetry.Context.Operation.ID is W3c compatible before taking it without which exception is thrown by activity. Need similar check as done for other overload of startoperation.

src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs
```diff
@@ -159,14 +155,26 @@ public static IOperationHolder<T> StartOperation<T>(this TelemetryClient telemet
                     operationActivity.SetOperationName(operationName);
                 }
 
-                if (parentActivity == null)
+                if (idsAssignedByUser)
                 {
                     // telemetryContext.Id is always set: if it was null, it is set to newly generated W3C compatible TraceID
                     if (Activity.DefaultIdFormat == ActivityIdFormat.W3C)
                     {
                         // There is no need of checking if TelemetryContext.ID is W3C Compatible. It is always set to
```
cijothomas: please remove this comment as it is no longer relevant.

PR #1211, opened 2019-09-10
Set tracestate and sampling decision even if operation id is set

lmolkova:
`OperationCorrelationTelemteryInitializer` should set tracestate and sampling decision even when operation id is provided on telemtery item already. Auto-collectors always set operation id, but they should not duplicate code for tracestate/sampling


TimothyMothra:
@lmolkova Do you need this in the new beta release?

PR #1237, opened 2019-10-09
fix Transmission Tests

TimothyMothra:
**SetsEndpointAddressPropertyToSpecifiedValue**.

This test has failed 44 times in the last 30 days, but the code is literally verifying that the constructor sets a public property!!!!????

This test failed two days ago:
_“Error calling Test Cleanup method for test class Microsoft.ApplicationInsights.Channel.TransmissionTest+Constructor: System.Net.Http.HttpRequestException: Error while copying content to a stream. ---> System.IO.IOException: Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host. ---> System.Net.Sockets.SocketException: An existing connection was forcibly closed by the remote host”_

This is suspicious because this test doesn’t invoke any network operations.
Taking a closer look, this test class inherits some base classes. Reading the comments these seem to only exist to support testing Async operations in .Net 3.5 !!!!
Soooo….. I’m going to remove some of these complicated test dependencies.

**Changes**
- I removed the dependency on the subclasses. This test is literally only testing the constructor. All this extra stuff is not needed.
- I replaced the test Uri with localhost. The Uri class does some non-trivial validation so only creating one should reduce some of the work. Also, IF for some reason this tries to make an outbound test, it should only ping the local host and fail fast.
- I removed the dependenc8y on the AssertEx utility class. This is unnecessary, the MSTest framework is more than capable of catching expected exceptions.

### DIFF
--- a/Test/Microsoft.ApplicationInsights.Test/Shared/Channel/TransmissionTest.cs
+++ b/Test/Microsoft.ApplicationInsights.Test/Shared/Channel/TransmissionTest.cs
@@ -29,62 +29,57 @@
         }
 
         [TestClass]
-        public class Constructor : TransmissionTest
+        public class Constructor
         {
+            private readonly Uri testUri = new Uri("https://127.0.0.1/");
+
             [TestMethod]
             public void SetsEndpointAddressPropertyToSpecifiedValue()
             {
-                var expectedAddress = new Uri("expected://uri");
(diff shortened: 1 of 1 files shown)


TimothyMothra:
Unrelated Test failed: ProactivelySampledInTelemetryCapturedWhenProactiveSamplingRateIsHigherThanTarget

PR #1243, opened 2019-10-11
Metrics: Support auto-fallback to magic dim value when dim-cap is reached.

macrogreg:
Hey @cijothomas 

After we spoke I realized that you were right - one would likely need to touch the Cube to add support for auto-fallback to a special dimension name when the cap is reached.

I looked over the code and I think it would look something like this.

I created this PR for your convenience - please feel free to use it or not as you see fit. If it's useful, please feel free to take over the Metrics.DimCapAPI branch and to add whatever else you feel is necessary to ship this.. Please let me know if you need a code review or any other kind of feedback.

Thanks!

### DIFF
--- a/src/Microsoft.ApplicationInsights/Metrics/Implementation/ConcurrentDatastructures/MultidimensionalCube2.cs
+++ b/src/Microsoft.ApplicationInsights/Metrics/Implementation/ConcurrentDatastructures/MultidimensionalCube2.cs
@@ -348,6 +348,20 @@
                 return result;
             }
 
+            // If we have fallen back to the default dimension value at any time, we may have an existing point:
+
+            if (appliedUnsafeDimensionValuesCapFallback)
+            {
+                pointMoniker = BuildPointMoniker(coordinates);
+
+                hasPoint = this.points.TryGetValue(pointMoniker, out point);
+                if (hasPoint)
+                {
+                    var result = new MultidimensionalPointResult<TPoint>(MultidimensionalPointResultCodes.Success_ExistingPointRetrieved, point);
+                    return result;
+                }
+            }
+
             // Create new point:
 
             try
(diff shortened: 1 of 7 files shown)


macrogreg:
@cijothomas, I noticed 3 test failures in the PR that I sent. They are for the relevant code, so must be related. However they are very weird: without the debugger i get 3 tests failing. Under the debugger only one of them fails, and in a different spot.
So this is either some intricate timing bug in my change (most likely) or it is an issue with the SDK's different behavior under the debugger (also possible, but less likely since I did not touch it). Could the the interaction of these too.

macrogreg:
Interesting...
Looks like I get the same tests failing when running local tests on DEVELOP. So maybe some other recent change? But then - how did it get into develop? :)

I have only limited time to look at this until mid next week.
Hopefully, the pointer on what to change is useful until then. We can sync up mid next week and look at things together, or if this is time-sensitive, perhaps you could get a head start.

src/Microsoft.ApplicationInsights/Metrics/Implementation/ConcurrentDatastructures/MultidimensionalCube2.cs
```diff
@@ -342,6 +370,11 @@ private MultidimensionalPointResult<TPoint> TryCreatePoint(string[] coordinates,
             }
 
             { 
+                if (appliedUnsafeDimensionValuesCapFallback)
+                {
+                    pointMoniker = BuildPointMoniker(coordinates);
```
cijothomas: this should be done outside of the createpoint call, as "other' may have already created.

cijothomas:
@macrogreg I am picking #1245 over this. Closing this.

src/Microsoft.ApplicationInsights/Metrics/Implementation/ConcurrentDatastructures/MultidimensionalCube2.cs
macrogreg: Good catch. However, not outside, but a little further up, while you are under the same lock. PR updated.

cijothomas:
closing as i am reusing code into new pr. due to repo structure changes, it'll tough to hangle merge conflicts.

Issue #1244, opened 2019-10-11
Enable dim cap to Metrics API


cijothomas:
**Problem Statement**
The Metrics API currently does not offer a setting to apply dimension capping. Instead, when dimension values reach the limit, TrackValue/TryGetSeries returns false, and the value is not tracked. The expectation is that the caller is responsible for handling this situation.

Consider the following example with a metric "AnimalsSold" with 2 dimensions "Dimension1" and "Dimension". The config sets a limit of 2 to each dimensions. 

```
var metConfig = new MetricConfiguration(seriesCountLimit: 100, valuesPerDimensionLimit:2,
                new MetricSeriesConfigurationForMeasurement(restrictToUInt32Values: false));

Metric animalsSold = client.GetMetric("AnimalsSold", "Dimension1", "Dimension2", metConfig);

// Start tracking.
animalsSold.TrackValue(100, "Dim1Value1", "Dim2Value1");
animalsSold.TrackValue(100, "Dim1Value1", "Dim2Value2");

// the following call gives 3rd unique value for dimension2, which is above the limit of 2.
animalsSold.TrackValue(100, "Dim1Value1", "Dim2Value3");
//The above call does not track the metric, and returns false.
```

There only recommended action for clients to do when the cap is hit is to revisit the dimension value and ensure their cardinality is limited. This prevents future issues, but the tracked values are lost, affecting Metric charts etc. 
Because of this, the recommended approach of tracking metric involve checking the return value and doing something (like sending a telemetry) to indicate caps are hit.  

```

if (! animalsSold.TrackValue(100, "Dim1Value1", "Dim2Value3"))
            {
                telemetryClient.TrackTrace("Metric value not tracked as value of one of the dimension exceeded the cap. Revisit the dimensions to ensure they are within the limits",
                                           SeverityLevel.Error);
            }
```

Instead of losing data when a cap is hit, some users may prefer to instead keep tracking the value, but with a special dimension value to which all the dimension values encountered after cap is hit is rolled into. In the above case, the dimension value "Dim2Value3" could be replaced with something like "Other". This approach may sound an easy solution, but it severely affects data quality, as SDK Metric System is only aware of dimension values from its current execution. Thus values rolled into "Other" will vary across application instances, across service instances or even across restarts of the same instance. In the above example, in one instance "Dim2Value3" may be rolled into "Other", but in another instance, "Dim2Value2" could be the one rolled into "Other".
[...]

PR #1245, opened 2019-10-11
Enable dim cap to Metrics API

cijothomas:
Fix Issue #1244  .


macrogreg:
Having discussed this, I am surprised - it seems exactly the opposite of what we discussed:

* You are making this very dangerous functionality public. Why? The customers can and will misunderstand this API. The problem you are trying to solve exists only for standard metrics. Do not create a back-compat burden by making this API public.

* You are baking in a non-localizable magic dimension value string into the core of the general aggregation.
  
* You are not offering a way to configure this only for some dimensions. You mentioned that you were doing it for simplicity, but in fact the change is more complex than the alternative.

* In addition to these points you are also going through a lock/release cycle in a loop. At the very least it is a pref issue under load. We should also look very carefully at whether this is correct.

I strongly urge you to consider https://github.com/microsoft/ApplicationInsights-dotnet/pull/1243 over this. That change is smaller, simpler, more flexible, more performant and does not have the concurrency challenge. If you opt for not specifying flags per dimension, please address the other points form above.

Thanks!

Issue #1214, opened 2019-09-19
Consolidating Application Insights Repos


TimothyMothra:
## Currently 
We are shipping 17 packages across 4 repos:
- Base https://github.com/microsoft/ApplicationInsights-dotnet
- Web https://github.com/microsoft/ApplicationInsights-dotnet-server
- Logging https://github.com/microsoft/ApplicationInsights-dotnet-logging
- AspNetCore https://github.com/microsoft/ApplicationInsights-aspnetcore

## Challenges

- This creates additional work to synchronize changes across each repo.
Typically work begins in this (Base) repo and requires a signed build to test those changes in another repo.
This slows down the dev cycle.
- We take this cost again whenever we do a release. Releases must occur in stages (Base > Web > Logging > Core) and involves a lot of waiting for automation to complete.
A full release typically costs us a week. In the last 2 years, we've never released one repo without also releasing all others.
- This creates confusion for customers. Most of our customers use "application insights" and don't know which repo to open issues.
- This creates confusion for contributors. It takes some effort to determine the correct repo to contribute for a specific feature.

## Proposal
Over the next month, we will merge all 4 repos into this one.

AspNetCore did this exercise [last year](https://github.com/dotnet/coreclr/issues/26175) and CoreClr is doing this [now](https://github.com/dotnet/coreclr/issues/26175).

## Timeframe

I'll be starting this work in October and I expect this can be completed within a month, finishing in early to mid November.

## Next Steps

- All PRs will be tagged with this issue for tracking.
- I'll start opening prs to reorganize files before the migration to help prevent conflicts.
- I'll migrate one repo at a time and confirm that builds are successful to prevent blocking any contributor. We want to preserve the entire commit history.
- After the migration, I'll lock the source repo.
- I expect there will several follow up PRs as I attempt to consolidate code.
- I don't know the correct mechanism to migrate issues but I will make every attempt to do so.

# Detailed Plan

## In Scope
- 4 Repos (Base, Web, AspNetCore, Logging)
- 4 PR Build definitions
- 4 Signing Definitions
- 4 Release Pipelines

## End Goal
- Move all code to Base SDK repo, and preserve commit history
- Shared build definitions for PRs, Signing, and Releases

### Prep work
1. **DONE** Reorganize individual repos. This is to ensure that there are no conflicting directories or file names and to guarantee no conflicts when migrating source. This will require some minor changes to existing build definitions.

### Migrate source
[...]

PR #1248, opened 2019-10-14
Reorg: move files from root to Base directory

TimothyMothra:
https://github.com/microsoft/ApplicationInsights-dotnet/issues/1214

Step 1. Reorganize individual repos. This is to ensure that there are no conflicting directories or file names and to guarantee no conflicts when migrating source. This will require some minor changes to existing build definitions.

**Changes**
Update Directory
- EnlistmentRoot stays in root directory
- Add SourceRoot
- Update Directory.Build.props

Update directory in build tasks:
- Clean script
- StrongName Verification script (enable & disable)
- Build directory "dirs.proj"
- GenerateReleaseMetadata script
- SDL script files


BASE/README.md
```diff
@@ -0,0 +1,121 @@
+[![Build Status](https://mseng.visualstudio.com/AppInsights/_apis/build/status/ChuckNorris/AI-BaseSDK-develop-linux?branchName=develop)](https://mseng.visualstudio.com/AppInsights/_build/latest?definitionId=6237?branchName=develop)
```
cijothomas: why is this appearing as a new file instead of simple renaming?

BASE/src/Microsoft.ApplicationInsights/GlobalSuppressions.cs
```diff
@@ -1,30 +1,30 @@
...
-
-[assembly:SuppressMessage(
-            category:       "Microsoft.Usage",
-            checkId:        "CA2213: Disposable Fields Should Be Disposed", 
-            MessageId =     "telemetryChannel", 
-            Scope =         "member", 
-            Target =        "Microsoft.ApplicationInsights.Extensibility.TelemetrySink.#Dispose()", 
-            Justification = "telemetryChannel is being disposed using ?. syntax")]
-
+﻿// <copyright file="GlobalSuppressions.cs" company="Microsoft">
```
cijothomas: this should also simply appear as "File renamed without changes"

BASE/README.md
TimothyMothra: i did copy the readme. i wanted to make sure i preserve a copy in each sub folder, but i still needed one in the root for github to display

BASE/src/Microsoft.ApplicationInsights/GlobalSuppressions.cs
TimothyMothra: i honestly have no idea about this one. the contents appear to be identical and no tests are failing (i confirmed the # of tests are the same)

PR #1257, opened 2019-10-22
Stabilize ProactivelySampledInTelemetryCapturedWhenProactiveSamplingRateIsHigherThanTarget

lmolkova:
Adaptive sampling is designed in a way that eventually, the sampling rate will get to configured one under the scale. 

Our tests start with a rate that is far from expected and they need time to adjust.
If we produce a few telemtery items, sampling algorithm results will be skewed by statistical errors (not enough data).  If we produce a lot of items, we are skewed by test machines capacity and virtualization issues. 

So it is expected that sampling tests require some tolerance.

We discovered that on Linux test agents `ProactivelySampledInTelemetryCapturedWhenProactiveSamplingRateIsHigherThanTarget` fail too frequently. And since we don't have any linux-specific code in id generation or sampling, I blame test infra for skewed data and I'm simply increasing test tolerance to make tests fail less.

### DIFF
--- a/Test/Microsoft.ApplicationInsights.Test/Net45/Microsoft.ApplicationInsights.Net45.Tests.csproj
+++ b/Test/Microsoft.ApplicationInsights.Test/Net45/Microsoft.ApplicationInsights.Net45.Tests.csproj
@@ -44,6 +44,10 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


lmolkova:
test failures are not related to this change and it seems are expected as we are working on major change in build infra, right @TimothyMothra  ?

TimothyMothra:
yup, these failures are unrelated to your change. I'll watch this and get it merged today. Thank you

Issue #1233, opened 2019-10-03
Improving pre-aggregated Standard Metrics in SDK


cijothomas:
Background read: https://docs.microsoft.com/en-us/azure/azure-monitor/app/pre-aggregated-metrics-log-metrics

The UI screenshots in this post are from Metric Explorer.

The default configuration of SDK (when used with Web, AspNetCore or WorkerService nugets) adds a special telemetry processor called `AutocollectedMetricsExtractor` before sampling processors. This processor inspects every `Request` and `Dependency` telemetry items, and aggregates them with key dimension. This pre-agg metric is sent to Application Insights once every min, and are available to query under StandardMetrics->Server in metric explorer.

As this metric is calculated _before_ sampling, these metrics are superior in terms of accuracy.
There are 5 standard metrics which are populated from the above.
![image](https://user-images.githubusercontent.com/5232798/66161552-67008380-e5e1-11e9-8835-1b04f605ecb2.png)

They are:
Dependency Calls - Count of dep calls.
Dependency duration - Duration of dep calls
Server request - Count of incoming Server request
Server response time - Duration of Server requests
Server request rate - Derived metric from server request count showing requests/sec.

As explained in the docs, these counters are sent with certain key dimensions only. (instead of entire dimensions available in log-based metrics). In UI, these dimensions allow splitting the metric chart, and Alerting allows alerting based on these dimensions.

For Requests, currently the only dimension populated by SDK is 'successful request', even though the UI lists 6 dimensions.
![image](https://user-images.githubusercontent.com/5232798/66161871-150c2d80-e5e2-11e9-8901-ab28f71a9286.png)

For Dependencies, SDK populates dimensions Dependency type and Sucessful call, though UI lists 6 dimensions.
![image](https://user-images.githubusercontent.com/5232798/66161815-fad24f80-e5e1-11e9-99e5-8bb063247891.png)

As SDK pre-aggregation did not add all the dimensions supported in UI, it was not possible to split the charts using these dimensions. (eg: it is not possible to split based on cloudRoleInstance)

We plan to improve the metric experience for the Standard Metrics, by adding more dimension to the SDK pre-aggregators.

This will be done in several steps:
1 Add dimensions which are low/fixed cardinality like  CloudRoleName, CloudRoleInstance, Dependency/Request performance, real/synthetic. Users will be allowed to pick a max limit to the number of values for these. (similar to `MaxDependencyTypesToDiscover`)

[...]

PR #1259, opened 2019-10-23
Enable DimensionCapping API internally

cijothomas:
Internal at this time for unblocking #1233 

#1244 has more discussions.


BASE/src/Microsoft.ApplicationInsights/Metrics/Implementation/AutocollectedMetricsExtraction/MetricTerms.cs
```diff
@@ -4,7 +4,8 @@
 
     internal static class MetricTerms
     {
-        private const string MetricPropertiesNamePrefix = "_MS";
+        internal const string DimensionCapFallbackValue = "DIMENSION_CAPPED";
```
Dmitry-Matveev: This PR does not seem to consume `DimensionCapFallbackValue`, should we update `dimensionCapValue: ` definition above?

cijothomas: This was not part of this PR :D it was from my upcoming PR with standard metric. it just got accidentaly spilled into this..

cijothomas:
@TimothyMothra Merge when you think this is safe to merge.

Issue #1260, opened 2019-10-24
Improve supportability regarding Invalid Instrumentation Keys


TimothyMothra:
**Problem**
Customers are affected when the SDK does not report invalid instrumentation keys.

**Out of Scope**
- Note that the SDK cannot validate an ikey. The SDK is able to send to multiple ingestion services so it must be the responsibility of those ingestion services to validate keys. Http response codes are emitted in ETW logs.

**Changes**
- The TelemetryConfiguration factory will now report if an Ikey was not found in either the config file or environment variables. **DONE** https://github.com/microsoft/ApplicationInsights-dotnet/pull/1187 (This was shipped in 2.12-Beta1)
https://github.com/microsoft/ApplicationInsights-dotnet/blob/7633ae849edc826a8547745b6bf9f3174715d4bd/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs#L470-L489

- The `TelemetryConfiguration.InstrumentationKey` currently throws an exception when set to null. 
This will be changed to include empty strings and illegal characters. 
https://github.com/microsoft/ApplicationInsights-dotnet/blob/7633ae849edc826a8547745b6bf9f3174715d4bd/BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs#L145-L150

phatcher:
@TimothyMothra Regarding making empty string invalid, how does this affect the use case of running locally with no key so Visual Studio enables telemetry but doesn't forward it to an AI instance?

TimothyMothra:
Hi @phatcher, thanks for the reminder. I'll test this scenario.

My change here only throws an exception if a user tries to explicitly set an empty string. if you're leaving the key property empty by default, nothing should change. :)

SergeyKanzhelev:
@TimothyMothra the question remains - what would be the customer experience look like? One may want to disable telemetry when running locally and do it programmatically.

SergeyKanzhelev:
Also, if ikey is set from config file and config file is broken - is it OK to just crash startup of an app?

TimothyMothra:
for 2.12, I'm expanding ETW logs to report when an Instrumentation Key is missing or misconfigured.

for 2.13, I'm considering an Environment Variable driven "diagnostic mode" that would auto-inject our FileWritter that saves ETW events to text file. 
[...]

PR #1261, opened 2019-10-24
validate ikey

TimothyMothra:
#1260 

As we discussed, InstrumentationKey will be checked for invalid characters when set.


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/InstrumentationKeyValidation.cs
```diff
@@ -0,0 +1,40 @@
...
+            if (instrumentationKey == null)
+            {
+                throw new ArgumentNullException(nameof(instrumentationKey));
+            }
+            else if (string.IsNullOrWhiteSpace(instrumentationKey))
+            {
+                throw new ArgumentException("Instrumentation Key can not be empty.");
+            }
+
+            Regex allowedCharacters = new Regex("^[a-zA-Z0-9-:]*$");
```
Dmitry-Matveev: This excludes Unicode alphanumeric character and only includes English alphanumeric - do we expect IKey string to be non-English in some cases? If yes, we can change the Regex to account for that.

cijothomas:
@TimothyMothra Can you describe the validation behavior if ikey is configured via other options? eg: environment variable? And asp.net core offers additional places for provide instrumentation key. (aioptions)
How about ikey provided inside connectionstrings?

TimothyMothra:
@cijothomas 
Environment Variable and ConnectionString are both used to set the telemetryconfiguration.instrumentationkey, so they are covered by these same validation rules.
IIRC aioptions also maps to the telemetryconfiguration.

cijothomas:
> @cijothomas
> Environment Variable and ConnectionString are both used to set the telemetryconfiguration.instrumentationkey, so they are covered by these same validation rules.
> IIRC aioptions also maps to the telemetryconfiguration.

If you can do a unit test asserting these, it'd be great.

TimothyMothra:
Abandoning this, we are going in a different direction.

PR #1269, opened 2019-10-28
consolidating dependencies

TimothyMothra:
#1214


TimothyMothra:
Base and Core builds passed on second run, they didn't update here.
Since this has no product code changes, I'm going to force merge

PR #1274, opened 2019-10-29
update common props

TimothyMothra:
#1214 

This is an update to the common props file.
These are some of the most common properties shared by all our projects.


TimothyMothra:
I am force merging this PR. This is editing .props and .targets files.
Do production projects are importing these props yet.

PR #1275, opened 2019-10-29
Tilee/everything

TimothyMothra:
#1214 
Step 6. All projects should be merged into a single directory. This can be done as follow up PRs.

**Changes**
- Remove nuget dependencies from projects in favor of project references.
- Added a new `Everything.sln`. This solution includes everything. :)
- Update scoped solutions with new project references.
- I had to rename some of the Shared projects so these could co-exist. Eventually, each of these will be consolidated too.
- Our builds use filename/directory matching patterns to run tests against their required framework. I had to rename some projects to match their actual compiled framework.


cijothomas:
Can you take care of project references in the DependencyCollector/E2E test apps as well? Its okay to do it as separate PR.

Also, I prefer ApplicationInsights.All.sln instead of Everything.

Dmitry-Matveev:
I'd say I'd prefer SuperMegaGodzilla.sln (it's in the line with Godzilla Vs Mothra adaptation that I have an opportunity to enjoy every day in the office watching poor Mothra fighting this unreasonably big thing), but I do not think you'd take me seriously, so ApplicationInsights.FreeForAll.sln should do.

TimothyMothra:
I'm going to close this so the team stops getting emails. 
I'll re-open when I'm closer to being finished.

Issue #1221, opened 2019-09-24
SDK Connection Strings


TimothyMothra:
Application Insights will start to support Connection Strings.
This will simplify the endpoint configuration necessary for some of our customers, specifically for customers that have a business need to connect to specific Azure Regions.
The Azure Portal will start displaying Connection Strings this week.
There will be an official blog post at a later date.

![image](https://user-images.githubusercontent.com/28785781/67131682-7c1d0b00-f1ba-11e9-960d-616880cbb41c.png)

**Ruleset**
1. check for explicit endpoint (location is ignored)
2. check for endpoint suffix (location is optional)
3. use classic endpoint (location is ignored)

**Design**
- New property: `TelemetryConfiguration.ConnectionString`
- New method: `TelemetryConfiguration.SetConnectionString())`
Connection string needs to be parsed (via `EndpointProvider`) and validated.
This method will set `TelemetryConfiguration.InstrumentationKey`.
Resulting endpoints are set to `TelemetryConfiguration.Endpoints` (`EndpointContainer`).
This method will also update `TelemetryChannel.Endpoint`

**TODO**
- Need to integrate Javascript SDK with AspNetCore SDK. Waiting for new Javascript SDK to be released with connection string support.

PR #1276, opened 2019-10-30
javascript connection string

TimothyMothra:
#1221 

**Changes**
- Replace JS Snippet with updated version
- Replace the instrumentation key injection to choose between connection string or ikey

### DIFF
--- a/NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/JavaScriptSnippet.cs
+++ b/NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/JavaScriptSnippet.cs
@@ -87,6 +87,7 @@
                     }
 
                     // Return full snippet
+                    // Developer Note: If you recently updated the snippet and are now getting "FormatException: Input string was not in a correct format." you need to escape all the curly braces; '{' => '{{' and '}' => '}}'. 
                     return string.Format(CultureInfo.InvariantCulture, Snippet, insertConfig, insertAuthUserContext);
                 }
                 else
--- a/NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/JavaScriptSnippet.cs
+++ b/NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/JavaScriptSnippet.cs
@@ -59,6 +59,7 @@
             {
                 if (!this.telemetryConfiguration.DisableTelemetry)
                 {
+                    // Config JS SDK
                     string insertConfig;
                     if (!string.IsNullOrEmpty(this.telemetryConfiguration.ConnectionString))
                     {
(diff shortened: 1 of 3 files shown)


TimothyMothra:
@markwolff, Please review

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/JavaScriptSnippet.cs
```diff
@@ -55,23 +55,37 @@ public class JavaScriptSnippet : IJavaScriptSnippet
...
+                    {
+                        insertConfig =  string.Format(CultureInfo.InvariantCulture, "connectionString: '{0}'", this.telemetryConfiguration.ConnectionString);
+                    }
+                    else if (!string.IsNullOrEmpty(this.telemetryConfiguration.InstrumentationKey))
+                    {
+                        insertConfig = string.Format(CultureInfo.InvariantCulture, "instrumentationKey: '{0}'", this.telemetryConfiguration.InstrumentationKey);
+                    }
+                    else
+                    {
+                        return string.Empty;
```
markwolff: Just curious, what happens when this is reached? Would it be different from the old behavior? Looks like previously passes ikey as empty string and doesn't "crash". Breeze will reject the telemetry, but I think this is okay. If neither is provided, then we "crash".

TimothyMothra: I preserved the same behavior as before: 
- previously, there was the single if statement: If Disabled & ikey empty, returns `string.Empty`
- now with the tested ifs, if Disabled & ikey empty & connString empty, returns `string.Empty'

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Resources.resx
```diff
@@ -122,15 +122,15 @@
...
...ppInsights";var aiName=window[sdkInstance],aisdk=window[aiName]||function(e){function n(e){t[e]=function(){var n=arguments;t.queue.push(function(){t[e].apply(t,n)})}}var t={config:e};t.initialize=!0;var i=document,a=window;setTimeout(function(){var n=i.createElement("script");n.src=e.url||"https://az416426.vo.msecnd.net/scripts/b/ai.2.min.js",i.getElementsByTagName("script")[0].parentNode.appendChild(n)});try{t.cookie=i.cookie}catch(e){}t.queue=[],t.version=2;for(var r=["Event","PageView","Exception","Trace","DependencyData","Metric","PageViewPerformance"];r.length;)n("track"+r.pop());n("startTrackPage"),n("stopTrackPage");var s="Track"+r[0];if(n("start"+s),n("stop"+s),n("addTelemetryInitializer"),n("setAuthenticatedUserContext"),n("clearAuthenticatedUserContext"),n("flush"),!(!0===e.disableExceptionTracking||e.extensionConfig&amp;&amp;e.extensionConfig.ApplicationInsightsAnalytics&amp;&amp;!0===e.extensionConfig.ApplicationInsightsAnalytics.disableExceptionTracking)){n("_"+(r="onerror"));var o=a[r];a[r]=function(e,n,i,a,s){var c=o&amp;&amp;o(e,n,i,a,s);return!0!==c&amp;&amp;t["_"+r]({message:e,url:n,lineNumber:i,columnNumber:a,error:s}),c},e.autoExceptionInstrumented=!0}return t}(
```
cijothomas: https://az416426.vo.msecnd.net -- is this constant for all clouds? or separate cdn for non public azure?

cijothomas:
https://github.com/microsoft/ApplicationInsights-aspnetcore/issues/866 - Is this issue being resolved here?

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Resources.resx
markwolff: It would be a different CDN. I wonder if this is something that JS SDK connection string should support?

cijothomas: This snippet need to point to a different CDN, then the CDN value should come from connectionstring. @TimothyMothra Can you check if this can be incorporated to connectionstring? Without this, I don't think javascript support can be enabled in non public regions.

TimothyMothra:
@cijothomas, Yes THIS (https://github.com/microsoft/ApplicationInsights-aspnetcore/issues/866) issue will be solved by connection string. Channel Endpoint is included in Connection String :)

Regarding CDN, I'm going to have to discuss that with Mark to understand how the correct values for CDN.
Thank you for raising this concern.

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Resources.resx
TimothyMothra: I just spoke to @markwolff , a CDN for private regions does not exist yet. So We will proceed with this change as is and will have to add support at a later date. I'll discuss this with Reiley so he's aware.

PR #1285, opened 2019-11-06
InjectXmlLanguage: replace CodeTaskFactory with RoslynCodeTaskFactory

TimothyMothra:
CodeTaskFactory has been deprecated and the recommended replacement is RoslynCodeTaskFactory
https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-roslyncodetaskfactory?view=vs-2019


malkia:
Where is the official statement that it has been deprecated, because this documentation here - https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-roslyncodetaskfactory?view=vs-2019 does not say so.

Thanks!

TimothyMothra:
You're right, I can't find this message anymore.
I think I was getting this message in the output from our build system. 
Switching to the newer task factory fixed all my issues. :)

PR #1286, opened 2019-11-06
rename shared projects

TimothyMothra:
#1214 
Step 6. All projects should be merged into a single directory. This can be done as follow up PRs.

**Changes**
In preparation for consolidating projects, I've renamed each repo's "Common" shared projects.
The "everything" solution cannot have multiple projects with identical names.

This PR has no functional changes.


TimothyMothra:
I audited all the projects that I renamed.
I think they still follow the naming conventions:

> Shared: code for multiple targets
> Cammon: code for multiple libraries

 As we proceed, I would like to revisit this. I see several projects that fit both definitions but were inconsistently named one or the other. Including one project which was named both. :)

PR #1296, opened 2019-11-11
Tilee/everything signing

TimothyMothra:
#1214 

Step 7. The newer single directory source will require consolidated build definitions. This can be validated against a feature branch before merging.

**Changes**
- Add new solution for signing projects.
- Remove unused scripts


TimothyMothra:
I'm force merging. This PR has no product changes.

PR #1299, opened 2019-11-12
Tilee/fix running tests

TimothyMothra:
#1214 

Here I've updated all the test dependencies that the projects would let me.
This introduced a new xunit analyzer that required some test classes to be modified.


TimothyMothra:
these changes seem to have caused more problems than they solved.
closing/reopening to try once more. 
If it happens again, I will cancel this PR and try upgrading a few dependencies at a time at a later date

PR #1300, opened 2019-11-13
Add support for collecting Track 2 Azure SDK client activities

pakrym:
cc @lmolkova @SergeyKanzhelev


WEB/Src/DependencyCollector/Shared.Tests/Implementation/AzureSdkDiagnosticListenerTest.cs
```diff
@@ -0,0 +1,147 @@
...
+                Assert.IsTrue(telemetry.Properties.TryGetValue("tracestate", out var tracestate));
+                Assert.AreEqual("state=some", tracestate);
+            }
+        }
+
+        [TestMethod]
+        public void AzureClientSpansAreMarkedAsFailed()
+        {
+            using (var listener = new DiagnosticListener("Azure.SomeClient"))
+            using (var module = new DependencyTrackingTelemetryModule())
```
cijothomas: Can you add one test with EnableAzureSdkTelemetryListener = false?

pakrym: Done

pakrym:
Output from my sample app:

image

lmolkova:
Looking into the screenshot, there are several things we need to fix:

1. HTTP requests - we need to dedup double-instrumentation for HTTP from azure SDK and in AppInsights. In existing listeners (for core and desktop), we should back-off if traceparent header is present.

2. Dependency types: OTHER - we'll need to fix them.   
  - EventHub processor should be request-telemetry (not dependency) All server span should translate  into requests
  - Internal spans (convenience layer) should be dependency telemetry with kind InProc
  - Client spans should translate into Dependency with the appropriate type. EventHubs is the only SDK that emits such spans now and properties on the dependency should match what we report for existing EventHub (Track1) SDK in the existing eventhub listener. It includes dependency type,  data and target.

I can help with this.

pakrym:
> I can help with this.

I would appreciate the help.

lmolkova:
@pakrym can you please also update changelog

pakrym:
> @pakrym can you please also update changelog

Done

lmolkova:
@pakrym please also merge https://github.com/pakrym/ApplicationInsights-dotnet/pull/2 - I broke some tests....

pakrym:
> @pakrym please also merge pakrym#2 - I broke some tests....

Done

lmolkova:
last call for comments. Going to merge in 2 hours.

Issue #1328, opened 2019-11-25
AI (Internal): Unknown error, message System.UriFormatException: Invalid URI: The hostname could not be parsed.


TimothyMothra:
**Error:**

> AI (Internal): [Microsoft-ApplicationInsights-Extensibility-Web] Unknown error, message System.UriFormatException: Invalid URI: The hostname could not be parsed.
>    at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
>    at System.Web.Util.UriUtil.BuildUriImpl(String scheme, String serverName, String port, String path, String queryString, Boolean useLegacyRequestUrlGeneration)
>    at System.Web.HttpRequest.BuildUrl(Func1 pathAccessor)
>    at System.Web.UnvalidatedRequestValues.get_Url()
>    at Microsoft.ApplicationInsights.Common.RequestTrackingUtilities.UpdateRequestTelemetryFromRequest(RequestTelemetry requestTelemetry, HttpRequest request, IApplicationIdProvider applicationIdProvider)
>    at Microsoft.ApplicationInsights.Web.RequestTrackingTelemetryModule.OnEndRequest(HttpContext context)
>    at Microsoft.ApplicationInsights.Web.AspNetDiagnosticTelemetryModule.AspNetEventObserver.OnNext(KeyValuePair2 value)

**Source:**
https://github.com/microsoft/ApplicationInsights-dotnet/blob/35e4bb2624bcbe2af9fb5d1e724b28ffd49c460c/WEB/Src/Common/RequestTrackingUtilities.cs#L20-L30

The getter for the Unvalidated Url is failing to parse and create a Uri object.
Should we wrap this in a try/catch and ignore the Exception?

**Internal Call Stack:**
https://referencesource.microsoft.com/#System.Web/UnvalidatedRequestValues.cs,144

>     public sealed class UnvalidatedRequestValues {
>         private Uri _url;
>         public Uri Url {
>             get {
>                 if (_url == null) {
>                     _url = _request.BuildUrl(() => Path);
>                 }
>                 return _url;
>             }
>         }

https://referencesource.microsoft.com/#System/net/System/UriExt.cs,32

>         private void CreateThis(string uri, bool dontEscape, UriKind uriKind)
>         {
>             // if (!Enum.IsDefined(typeof(UriKind), uriKind)) -- We currently believe that Enum.IsDefined() is too slow 
>             // to be used here.
[...]

PR #1329, opened 2019-11-25
fix UpdateRequestTelemetryFromRequest throwing UriFormatException

TimothyMothra:
#1328 

RequestTrackingUtilities was throwing an exception when trying to read an "invalid" URI from HttpRequest.
This adds a try/catch and ETW logging around that exception.

Also, I moved the class RequestTrackingUtilities from the Common sharedproject to the Web project.
No other projects were referencing this class, and this class needed access to Web's EventSource class.

### DIFF
--- a/WEB/Src/Common/RequestTrackingUtilities.cs
+++ b/WEB/Src/Common/RequestTrackingUtilities.cs
@@ -26,7 +26,15 @@ namespace Microsoft.ApplicationInsights.Common
 
             if (requestTelemetry.Url == null)
             {
-                requestTelemetry.Url = request.Unvalidated.Url;
+                try
+                {
+                    requestTelemetry.Url = request.Unvalidated.Url;
+                }
+                catch
+                {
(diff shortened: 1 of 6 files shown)


WEB/Src/Common/RequestTrackingUtilities.cs
```diff
@@ -26,7 +26,15 @@ internal static void UpdateRequestTelemetryFromRequest(RequestTelemetry requestT
...
             if (requestTelemetry.Url == null)
             {
-                requestTelemetry.Url = request.Unvalidated.Url;
+                try
+                {
+                    requestTelemetry.Url = request.Unvalidated.Url;
+                }
+                catch
+                {
+                    // "AI (Internal): Unknown error, message System.UriFormatException: Invalid URI: The hostname could not be parsed."
```
lmolkova: we can log it with Info/Warning verbosiry

PR #1331, opened 2019-11-27
add etw log for missing instrumentation key

TimothyMothra:
#1260 

If I telemetry item has made it all the way to the TelemetryChannel without an Ikey, it is safe to assume than an Ikey was never configured.
Here I want to emit a UserActionable ETW Error. 

Note that ETW Event 56 is a config time warning, whereas ETW Event 57 is a channel time error.

### DIFF
--- a/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TransmissionProcessor.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TransmissionProcessor.cs
@@ -3,6 +3,7 @@
     using System;
 
     using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
 
     /// <summary>
     /// An <see cref="ITelemetryProcessor"/> that act as a proxy to the Transmission of telemetry"/>.
@@ -28,6 +29,11 @@
         public void Process(ITelemetry item)
         {
             TelemetryDebugWriter.WriteTelemetry(item);
+
+            if (string.IsNullOrWhiteSpace(item.Context.InstrumentationKey))
+            {
+                CoreEventSource.Log.TransmissionProcessorNoInstrumentationKey();
+            }
 
             this.sink.TelemetryChannel.Send(item);
         }
--- a/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TransmissionProcessor.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TransmissionProcessor.cs
@@ -3,7 +3,6 @@
     using System;
 
     using Microsoft.ApplicationInsights.Channel;
-    using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
 
     /// <summary>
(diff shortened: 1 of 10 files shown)


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TransmissionProcessor.cs
```diff
@@ -29,6 +30,11 @@ public void Process(ITelemetry item)
         {
             TelemetryDebugWriter.WriteTelemetry(item);
 
+            if (string.IsNullOrWhiteSpace(item.Context.InstrumentationKey))
+            {
+                CoreEventSource.Log.TransmissionProcessorNoInstrumentationKey();
```
Dmitry-Matveev: Just to keep in mind - some custom channel implementations maybe OK with no instrumentation key, such as custom file writer to save telemetry locally.

Also, AI VS Output / AI VS Integration UX may not need IKey to power up its experiences - will we emit this message for those scenarios? These scenarios might be fairly common, we can put #if !debug here to avoid this message in VS/Debug.

TimothyMothra: The `#if !debug` would be a compile time switch. I don't think we can detect when a consumer is in debug mode

cijothomas: Even if we do it, we should Log it once maybe, rather than for every item? Or maybe once in few seconds..

TimothyMothra: @Dmitry-Matveev, I don't know how to balance the concern that an empty ikey could be valid. I don't know enough about these scenarios. 
Do you think it's valid to optimize for the most common/expected scenarios?

@cijothomas, I agree for every item is too much, but I'm reluctant to log only once because I don't want the log to get missed. I can ask Raj for a second opinion on how frequent he thinks the log should be to get recognized.

Dmitry-Matveev: Something like `Debugger.IsAttached` can be a better pick here - does not execute locally in VS, does execute when deployed to PROD. One notable issue - log won't be published is you debug in production, e.g. with WinDbg or remote debugging, but that should not matter already since you are capturing everything you need in debugger.

TimothyMothra: @Dmitry-Matveev's comment about channels got me thinking about what if we only set this on our 1st party channels.
And I just discovered the log I'm adding already exists on both!
However, both channels restrict this to Verbose logs only. 
I'm not sure if our troubleshooting guides are collecting this. I will investigate...

https://github.com/microsoft/ApplicationInsights-dotnet/blob/7633ae849edc826a8547745b6bf9f3174715d4bd/BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs#L136-L144

https://github.com/microsoft/ApplicationInsights-dotnet/blob/4ff3b7fb22274321671058e47a65738c61e78438/BASE/src/ServerTelemetryChannel/ServerTelemetryChannel.cs#L266-L274

TimothyMothra:
#1483

PR #1340, opened 2019-12-04
remove httpcontext from rolename initializer

TimothyMothra:


TimothyMothra:
@Dmitry-Matveev, we need to re-test performance after this merges

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Implementation/RoleNameContainer.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        private static string roleName = string.Empty;
+
+        public static string RoleName
+        {
+            get => roleName;
+
+            set
+            {
+                if (value != roleName)
```
lmolkova: why not using Interlocked.CompareExchange for comparison and exchange?

TimothyMothra: Simple answer is, i didn't know the correct way to use CompareExchange.
Now that I have better test coverage, i'll take another look at this. :)

TimothyMothra: If you know something I don't, please share. :)

From what I'm reading, CompareExchange will always replace the `ref variable` with an incoming value.
In my use case, I only need to replace the `ref variable` when the incoming value does not match.
Specifically, when a VIP SWAP occurs, we need to update the role name.
My understanding is, replacing this with a CompareExchange will always replace the `ref variable` with the incoming value regardless of if it's new.

I believe CompareExchange doesn't help me in this use case.
But please correct me if i'm wrong.

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
```diff
@@ -72,7 +49,16 @@ public AzureAppServiceRoleNameFromHostNameHeaderInitializer(IHttpContextAccessor
...
         /// </summary>
-        public string WebAppSuffix { get; set; }
+        public string WebAppSuffix
+        {
+            get => RoleNameContainer.HostNameSuffix;
+
+            set
+            {
+                RoleNameContainer.HostNameSuffix = value;
+                RoleNameContainer.SetFromEnvironmentVariable(out _);
```
Dmitry-Matveev: It looks like `out bool isAzureWebApp` is used as a control leading to `out _` in other place. Is there a way to provide this control outside of `SetFromEnvironmentVariable` method to make operation mode selected separately?

TimothyMothra: I refactored this to initialize from EnvironmentVariable in a constructor.
This required some additional changes, but it makes both use and testing scenarios cleaner. :)

Issue #1491, opened 2019-12-09
Javascript snippet does not include Sub Resource Integrity parameters


tcarabine:
If you are reporting bug/issue, please provide detailed Repro instructions.

## Repro Steps
1.  Add the Microsoft.ApplicationInsights.AspNetCore package to a web application
2. Follow the steps in https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core#enable-client-side-telemetry-for-web-applications to add client side telemetery
3. Run the application and inspect the injected javascript

##  Actual Behaviour
A script tag is added without SRI fields set

## Expected Behaviour
Sub Resource Integrity fields to be set on the script tag

## Version Info 
.NET Version : NetCore 2.2				
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) :
OS : Any
Hosting Info (IIS/Azure WebApps/ etc) : Any

TimothyMothra:
@tcarabine, Thank you for your PR!
I'm reluctant to take this PR because our team does not own the "javascript snippet". We get updates from https://github.com/microsoft/ApplicationInsights-js and if I take your bug fix here, we'll lose it the next time we update.

@Markwolff can I transfer this issue to your repo?

markwolff:
Hi @tcarabine we do not include any SRI in the snippet because we update the underlying file on the CDN to enable auto updating. You can welcome to include it for your own implementation, but it will likely break when we roll out updates. To get around this, you can modify the url to the CDN in the snippet to a hardcoded version, e.g. locate `ai.0.js` in the snippet and change it to `ai.1.0.20.0.js`

TimothyMothra:
I'm closing this issue. 

if you still have questions, feel free to reopen.

PR #1492, opened 2019-12-09
Adding sri to js

tcarabine:
Fix Issue #1491. Added Integrity and CORS tags to Javascript snippet in the resources file.
This will need updating in-step with the deployed version of the Javascript on the CDN.

- [ x] I ran Unit Tests locally.
- [ x] CHANGELOG.md updated with one line description of the fix, and a link to the original issue if available.

Notes for reviewers:


TimothyMothra:
Per a conversation with @markwolff, we can't enable SRI in this snippet because it would disable the auto-update that we're taking advantage of.

Following Mark's recommendation, if this is something you need, you can disable the auto-insertion of the snippet and manually insert the snippet into your application and hardcode a specific version and the SRI.

PR #1494, opened 2019-12-09
Fix exception correlation with Request-Id and customError=Off

### LINKED ISSUE #1493: Exceptions are not correlated to requests in some cases
## Repro steps

0. Create WebAPI or MVC ASP.NET app
1. Configure customerErrors=off
2. Throw exception in controller/view
3. Send request with Request-Id header with W3C-compatible value like `Request-Id: |1234db4ce72f482698722d4f51beb518.0`

## Expected
Exception and request telemetries are reported. Exception is correlated to request

## Actual
Exception and request telemetries are reported. Exception is NOT correlated to request

Bug introduced in 2.11

lmolkova:
Fixes #1493

When W3C-compatible Request-Id is received, ASP.NET TelemetryCorrelation reads it and keeps Request-Id in Activity.ParentId, but creates W3C-activity (trace-id and span-id) regardless of the Request-Id value. This is intended Activity + ASP.NET behavior when W3C is forced.

ApplicationInsights in this case (for back-compat reasons) creates another Activity for request tracking that has Trace-Id parsed from Request-Id.

This activity is stored in the same key in HttpContext.Items as Asp.Net Telemetry Correlation module.
Telemetry Correlation module takes care of restoring this Activity if it is lost using one in HttpContext.Items.  Telemtery Correlation module makes sure it stops its own Activity (not some user's orphaned one)

So what happens is
1. Telemetry Correlation starts activity1 and notifies AppInsights 
2. AppInsights starts another activity2 in OnStart event
    * and stores it in HttpContext.Items["activity"]
3. Once the activity1 is started and events are sent, Telemetry Correlation stores activity1 in HttpContext.Items["activity"] (same key). 
    * AI activity2 is still current, but if it's lost in native-managed hops, Telemetry Correlation will restore activity1.
4. When request ends
    * TelemetryCorrelation makes sure it stops activity1: reads it from HttpContext.Items and makes sure it's current. If not - updates current, so activity1 becomes current here in all cases.
   * Then it notifies AppInsights 
5. AppInsights checks if there are errors during request execution (it only applies to customErrors=off mode)
    * tracks all exceptions from errors: which are initialized from activity1 (BUG)
    * tracks requests (which are initialized from activiy1 in step 2)

Fix:
- store AppInsights specific activity in a separate key in HttpContext.Items (so TelCor never overwrites it)
- make AppInsightsHttpModule restore this activity whenever possible.


TimothyMothra:
Changelog please :)

TimothyMothra:
I see that this is a bug since 2.11.
We shipped 2.12.0 this morning.
Does this meet the bar for a hotfix release (2.12.1) or can this wait until the next major release (2.13) ETA Feb 2020?

lmolkova:
> Does this meet the bar for a hotfix release (2.12.1) or can this wait until the next major release (2.13) ETA Feb 2020?

I think this is rather a big change for a hotfix and some parts of this issue (not restoring AI activity never really worked). Also, customErrors=Off is a rare case for production services (not a default setting). I'd suggest waiting for the next stable to ship it.

WEB/Src/Web/Web.Shared.Net/ApplicationInsightsHttpModule.cs
```diff
@@ -82,15 +98,34 @@ public ApplicationInsightsHttpModule()
         /// <summary>
         /// Initializes module for a given application.
         /// </summary>
-        /// <param name="context">HttpApplication instance.</param>
+        /// <param name="application">HttpApplication instance.</param>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1062:Validate arguments of public methods", MessageId = "0", Justification = "Context cannot be null")]
-        public void Init(HttpApplication context)
+        public void Init(HttpApplication application)
```
TimothyMothra: I agree with this change, but it might break customers who use named parameters.

PR #1498, opened 2019-12-11
Trim W3C ids to parent/span Id

lmolkova:
In W3C mode we report parentId and Id in the format of `|traceId.parentSpanId.` and `|traceId.spanId.` respectively.

This is not needed anymore even for backward compatibility reasons. Compatibility with old versions of Application Insights SDK is handled by ingestion.

**Describe the solution you would like**

**When SDK receives a valid incoming request** with
1. `traceparent = 00-traceId-parentId-00`, it should track request telemetry
  * request.Context.Operation.Id = traceId // no changes
  * request.Context.Operation.ParentId = parentId  // new, was |traceId.parentId.
  * request.Id = new span Id // new, was |traceId.spanId.

2. no header, it should track request telemetry
  * request.Context.Operation.Id = new traceId  // no changes
  * request.Context.Operation.ParentId = null // no changes
  * request.Id = new span Id// new, was |traceId.spanId.

3. legacy Request-Id header, it should track request telemetry
  * request.Context.Operation.Id = if request-id root part is compatible with trace-id, reuse it, otherwise generate a new one
  * request.Context.Operation.ParentId = headers[Request-Id] // no changes
  * request.Id = new span Id // new, was |traceId.spanId.
  * properties["ai_LegacyRootId"] = if request-id root part is NOT compatible with trace-id, put it here, otherwise keep empty // no changes

**Similarly for dependencies**
-  dependency.Context.Operation.Id = traceId,
-  dependency.Context.Operation.ParentId = parent span id, // new, was |trace.parent.
-  dependency.Id = new span id, // new, was |trace.span.

**When sending Request-Id in outgoing requests it still MUST be formatted in the |TraceId.SpanId. format!**

### DIFF
--- a/BASE/Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/OperationHolderTests.cs
+++ b/BASE/Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/Implementation/OperationHolderTests.cs
@@ -56,7 +56,7 @@
             var operation = new OperationHolder<DependencyTelemetry>(client, new DependencyTelemetry(), originalActivity);
 
             var newActivity = new Activity("new").SetParentId("detached-parent").Start();
-            operation.Telemetry.Id = $"|{newActivity.TraceId.ToHexString()}.{newActivity.SpanId.ToHexString()}.";
+            operation.Telemetry.Id = newActivity.SpanId.ToHexString();
 
             operation.Dispose();
             Assert.AreEqual(Activity.Current, originalActivity);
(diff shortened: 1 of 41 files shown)


TimothyMothra:
FxCop: 
>   DiagnosticListeners\Implementation\HostingDiagnosticListener.cs(578,94): error CA1801: Parameter operationId of method GetParentId is never used. Remove the parameter or use it in the method body. [E:\A\_work\20\s\NETCORE\src\Microsoft.ApplicationInsights.AspNetCore\Microsoft.ApplicationInsights.AspNetCore.csproj]

lmolkova:
@TimothyMothra  needless to say how much I love fxcop

TimothyMothra:
Sorry. FxCop is an SDL requirement.
If you can review and confirm that this change was intentional, I would be more than happy to fix it for you. :)

You're also getting errors about a missing XML document. This is a new check that I added yesterday. These shouldn't be running against the Test projects and it's not actually failing your build.
It's safe to ignore this error and I'll get this fixed today.

PR #1523, opened 2020-01-02
refactor props

TimothyMothra:
Remove analyzers from Test projects that are not passing. We may re-add these at a later date.

Fix Issue #1214 .

**Changes**
- move all analyzer properties to a shared props file.
- remove Analyzers from Test projects (not passing).
- something is wrong with our published symbols. i've added the symbols to all nupkgs until we have more time to investigate.

Notes for reviewers:


TimothyMothra:
#1527

PR #1526, opened 2020-01-03
cleanup props

TimothyMothra:
Fix Issue #1214 

**Changes**
- remove stylecop from test project
- remove unused props file 
- remove explicit reference to AddXmlLanguage, now included with Nupkg.props

Notes for reviewers:


.props/_Nupkg.props
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <!-- This targets file will add an English localization tag to our xml documentation file. -->
+  <Import Project="$(TargetsRoot)\AddXmlLanguage.targets" />
```
reyang: Do we need `$(TargetsRoot)` or just using the `.props` folder for all `*.props` file?

TimothyMothra: I honestly hadn't considered that, but I think it makes sense. I'm testing that change now.

Issue #1527, opened 2020-01-04
Regarding Package Symbols


TimothyMothra:
## Problems
- I recently discovered a problem with our SDK symbols unavailable on Microsoft's Symbol Server. @lmolkova helped me verify.

- As part of the larger repo consolidation #1214, I'm investigating and consolidating our *.props.
I discovered that half our products were shipping `DebugType Full` and the other half was shipping `DebugType PdbOnly`
For context:
In 2018 we investigated changing to the `DebugType Portable`, but Microsoft's Symbol Server wasn't ready for this change.
In 2019 I tried to onboard to the Nuget Symbol Server, but they require the `DebugType Portable`.

## TODO
- We need to investigate if we can change to Portable now as recommended here: https://github.com/dotnet/sdk/issues/2679#issuecomment-441098067
- Need to investigate what happened to Microsoft Symbol Server.
- Need to finish onboarding to Nuget Symbol Server.

## Workaround
- Changing all SDKs to use `DebugType Full`. https://github.com/microsoft/ApplicationInsights-dotnet/pull/1528/files#diff-958413da6065a421fb6ced33222f648d
- Changing all NUPKGs to include symbols. https://github.com/microsoft/ApplicationInsights-dotnet/pull/1523/files#diff-775eb2dabbb1d381756ae13c69421241

PR #1528, opened 2020-01-04
refactor common props and delete empty props

TimothyMothra:
Fix Issue #1214 
Fix Issue #1527 

**Changes**
- Move Common.props to the common props in PropsRoot.
- Note that half our products were shipping with FULL symbols, and the other half were shipping with PDBONLY. I've changed all projects to use FULL, until we have time to investigate if we can switch to the newer PORTABLE.

Notes for reviewers:


.scripts/release_NupkgAudit.ps1
```diff
@@ -165,6 +165,15 @@ function Get-DoesXmlDocContainsLang ([string]$dllPath) {
     Test-Condition ($result) $message $requirement;
 }
 
+function Get-DoesPdbExist ([string]$dllPath) {
+    # CONFIRM .PDB SYMBOLS EXIST WITH EACH DLL
+    [string]$docFile = $dllPath -replace ".dll", ".pdb";
```
reyang: If `$dllPath` contains `.dll` in the file name or folder name, they will be replaced, too.

TimothyMothra: This is currently not an issue. 
In this scenario, i've just unpacked a NUPKG and am inspecting that each sub directory contains one of each (dll, xml, pdb).
Since i'm unzipping the NUPKG, I have high control of the file path and NUPKGs themselves don't contain ".dll" in their paths.

I think it's safe to ignore this for now. 
If someone makes a change to our build system or paths it would be a breaking change at that time.

PR #1649, opened 2020-01-28
New Flush API - FlushAsync

rajkumar-rangaraj:
Fix Issue # 

FlushAsync is a new task based Flush API in TelemetryChannel with below features.

- Waits for response from backend, to understand if the ingestion is successful
- Uses existing policies, to signal back failures
- Ensures Transmissions are flushed from process buffer to local disk storage before application exit

## Scenarios

**Scenario 1** If FlushAsync call has a successful ingestion, it just returns success and does not work on buffer or storage
**Scenario 2** If there are HTTP response failure from backend, this API relies on existing policies to move items from buffer to storage and sends success flag. If there an unrecoverable error which cannot be retried we set the flush as failure
**Scenario 3** If flush could not send transmissions due to capacity issues, then it moves the buffer content to local storage to ensure we don't miss transmissions on application exit

This API, considers flush as success if transmission either reaches the backend or all items are moved from buffer to local disk storage.

## Changes

New FlushAsync API is introduced in ITelemetryChannel and same is implemented in ServerTelemetryChannel and InMemoryChannel

**Task<bool> FlushAsync()**

In ServerTelemetryChannel, aysnc version of Flush is achieved with minimal code change and by using TaskCompletionSource.

TaskCompletionSource task is introduced in TelemetrySerializer. On flush completion, either Transmitter or TransmissionSender will set a completion signal to TelemetrySerializer. TaskCompletionSource  is implemented with a timeout of 100 seconds, if no response from backed it cancels a task. On TaskCompletionSource task completion it will send the status of the flush to awaiting task and result is made available as boolean value.

InMemoryChannel FlushAsync is made very easy by just wrapping it Task.Run on earlier implementation. 

### Checklist

- [ ] Design discussion issue # Flushdoesn't return anything indicating whether flush was successful or not

### Notes for authors:

### Notes for reviewers:


BASE/Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationFactoryTest.cs
```diff
@@ -1939,6 +1940,8 @@ private class StubTelemetryChannel2 : ITelemetryChannel, ITelemetryModule
 
             public void Flush() { }
 
+            public Task<bool> FlushAsync() { return Task.Run(() => true); }
```
TimothyMothra: Please change this to `return Task.FromResult(true);`

BASE/Test/Microsoft.ApplicationInsights.Test/Shared/Metrics/MetricsExamples.cs
```diff
@@ -1020,6 +1021,11 @@ public void Dispose()
         public void Flush()
         {
         }
+
+        public Task<bool> FlushAsync()
+        {
+            return Task.Run(() => true);
```
TimothyMothra: Please change this to `return Task.FromResult(true);`

BASE/Test/Microsoft.ApplicationInsights.Test/Shared/Metrics/TestUtility/StubApplicationInsightsTelemetryChannel.cs
```diff
@@ -73,5 +73,13 @@ public void Flush()
         {
             OnFlush();
         }
+
+        /// <summary>
+        /// Implements  the <see cref="ITelemetryChannel.FlushAsync" /> method.
+        /// </summary>
+        public Task<bool> FlushAsync()
+        {
+            return Task.Run(() => true);
```
TimothyMothra: Please change this to `return Task.FromResult(true);`

BASE/Test/TestFramework/Shared/StubTelemetryChannel.cs
```diff
@@ -76,5 +76,13 @@ public void Flush()
         {
             this.OnFlush();
         }
+
+        /// <summary>
+        /// Implements  the <see cref="ITelemetryChannel.FlushAsync" /> method.
+        /// </summary>
+        public Task<bool> FlushAsync()
+        {
+            return Task.Run(()=>true);
```
TimothyMothra: Please change this to `return Task.FromResult(true);`

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -674,6 +675,32 @@ public void Flush()
...
+            }
+
+            TelemetryConfiguration pipeline = this.configuration;
+            if (pipeline != null)
+            {
+                MetricManager sharedMetricManager = pipeline.GetMetricManager(createIfNotExists: false);
+                sharedMetricManager?.Flush(flushDownstreamPipeline: false);
+
+                ITelemetryChannel channel = pipeline.TelemetryChannel;
+                return await (channel?.FlushAsync()).ConfigureAwait(false);
```
TimothyMothra: what happens here if the channel is null?

BASE/Test/Microsoft.ApplicationInsights.Test/Shared/Extensibility/TelemetryConfigurationFactoryTest.cs
rajkumar-rangaraj: Fixed

BASE/Test/Microsoft.ApplicationInsights.Test/Shared/Metrics/TestUtility/StubApplicationInsightsTelemetryChannel.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
-
+using System.Threading.Tasks;
```
reyang: Looks like the coding style is trying to put a space between `System` and vendor namespaces.

BASE/Test/Microsoft.ApplicationInsights.Test/Shared/Metrics/TestUtility/StubApplicationInsightsTelemetryChannel.cs
```diff
@@ -73,5 +73,13 @@ public void Flush()
         {
             OnFlush();
         }
+
+        /// <summary>
+        /// Implements  the <see cref="ITelemetryChannel.FlushAsync" /> method.
```
reyang:
```suggestion
        /// Implements the <see cref="ITelemetryChannel.FlushAsync" /> method.
```
It'll be helpful to describe the purpose/intention of the API.

Dmitry-Matveev:
@SergeyKanzhelev would need to take a look as well, this change modifies the public API.

BASE/src/ServerTelemetryChannel/Implementation/TelemetrySerializer.cs
```diff
@@ -27,6 +29,33 @@ public Uri EndpointAddress
...
+            this.HandleTelemetryException(items);
+
+            var transmission = new Transmission(this.EndpointAddress, items, manualFlushAsync);
+            
+            using (var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(100)))
+            {
+                using (cancellationTokenSource.Token.Register(() =>
+                {
+                    transmission.SetFlushTaskCompletionSource(false);
+                }))
```
TimothyMothra: I think this can be simplified as
```suggestion
using (cancellationTokenSource.Token.Register(() =>transmission.SetFlushTaskCompletionSource(false);))
```

reyang:
Would you help to put some high level description in the PR?
1. What is the result of FlushAsync? Does it flush the in-memory data to disk, does it flush both in-memory and on-disk data to ingestion, does it flush on-disk data to ingestion?
2. What is the behavior of FlushAsync? Does it run indefinitely or there is a time bound. What's the behavior if there is error happened, how would the caller handle the error?
3. It'll be great to put some simple examples.

BASE/src/ServerTelemetryChannel/Implementation/TelemetrySerializer.cs
```diff
@@ -27,6 +29,33 @@ public Uri EndpointAddress
...
 
         public virtual void Serialize(ICollection<ITelemetry> items)
+        {
+            this.HandleTelemetryException(items);
+
+            var transmission = new Transmission(this.EndpointAddress, items);
+            this.Transmitter.Enqueue(transmission);
+        }
+
+        public virtual async Task<bool> Serialize(ICollection<ITelemetry> items, bool manualFlushAsync)
```
pharring: Task-returning methods should have "Async" suffix. i.e `SerializeAsync`.
Also, please add a doc-comment explaining the meaning of `manualFlushAsync`.

BASE/src/ServerTelemetryChannel/Implementation/TelemetrySerializer.cs
```diff
@@ -27,6 +29,33 @@ public Uri EndpointAddress
...
+        {
+            this.HandleTelemetryException(items);
+
+            var transmission = new Transmission(this.EndpointAddress, items, manualFlushAsync);
+            
+            using (var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(100)))
```
pharring: I would recommend adding a `CancellationToken` parameter to the method (i.e. have the caller decide when to cancel).

BASE/src/ServerTelemetryChannel/Implementation/TelemetryBuffer.cs
```diff
@@ -203,6 +210,26 @@ IEnumerator IEnumerable.GetEnumerator()
             return this.GetEnumerator();
         }
 
+        private List<ITelemetry> MoveTelemeteryBufferForFlush()
+        {
+            List<ITelemetry> telemetryToFlush = null;
+            if (this.itemBuffer.Count > 0)
+            {
+                lock (this)
```
pharring: I know this was in the original but, instead of `lock(this)`, we should use a dedicated syncObject for synchronization.

BASE/src/ServerTelemetryChannel/Implementation/TelemetryBuffer.cs
```diff
@@ -189,8 +194,10 @@ public virtual async Task FlushAsync()
                 // Flush on thread pull to offload the rest of the channel logic from the customer's thread.
                 // This also works around the problem in ASP.NET 4.0, does not support await and SynchronizationContext correctly.
                 // See also: http://www.bing.com/search?q=UseTaskFriendlySynchronizationContext
-                await Task.Run(() => this.serializer.Serialize(telemetryToFlush)).ConfigureAwait(false);
+                return await this.serializer.Serialize(telemetryToFlush, true).ConfigureAwait(false);
```
pharring: Tip: async methods with some pre-amble and ending with a single "return await" still cause the compiler to emit the async state machine code. This can be simplified by removing the `async` from the method signature and returning tasks directly. It also means you don't need ConfigureAwait. i.e.
```csharp
public virtual Task<bool> ManualFlushAsync()
{
    List<ITelemetry> telemetryToFlush = this.MoveTelemeteryBufferForFlush();
    return telemetryToFlush != null ? this.serializer.Serialize(telemetryToFlush, true) : Task.FromResult(false);
}
```

BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs
```diff
@@ -186,6 +187,16 @@ public void Flush(TimeSpan timeout)
             }
         }
 
+        /// <summary>
+        /// Will send all the telemetry items stored in the memory asynchronously.
+        /// </summary>
+        public async Task<bool> FlushAsync()
+        {
+            return await Task.Run(() => this.Flush(default(TimeSpan)))
```
pharring: I don't understand the point of the ContinueWith here.
Consider this (note: no async, no await):
```csharp
public Task<bool> FlushAsync()
{
    return Task.Run(() => this.Flush(default(TimeSpan)));
}
```

BASE/src/Microsoft.ApplicationInsights/Channel/ITelemetryChannel.cs
```diff
@@ -31,5 +31,10 @@ public interface ITelemetryChannel : IDisposable
         /// Flushes the in-memory buffer.
         /// </summary>
         void Flush();
+
+        /// <summary>
+        /// Flushes the in-memory buffer asynchronously.
+        /// </summary>
+        Task<bool> FlushAsync();
```
pharring: Two things:
1. Please document the return value.
2. Please add a `CancellationToken cancellationToken` parameter. That way callers can decide if/when to cancel (e.g. Ctrl+C, or critical shutdown, or based on a timeout)

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -133,6 +162,18 @@ public ICollection<ITelemetry> TelemetryItems
             get; private set;
         }
 
+        /// <summary>
+        /// Send signal to Flush Async.
+        /// </summary>
+        public void SetFlushTaskCompletionSource(bool manualFlushAsyncFlag)
```
pharring: Why have a bool parameter when the only valid value is `true`?
I see there are places in this PR where we call `SetFlushTaskCompletionSource(false)`. What does that do?

BASE/src/Microsoft.ApplicationInsights/Channel/ITelemetryChannel.cs
pharring: Also, is the intent to allow `FlushAsync` at any time, or only when the channel is shutting down? For example, what happens if you try to send more telemetry through the channel after `FlushAsync` completes successfully? Should you be able to call `FlushAsync` again? Or are those decisions up to each channel implementation?

pharring:
Please add unit tests.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -125,6 +136,24 @@ public string Id
...
+        public bool ManualFlushAsyncFlag
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// Gets Test to support async Flush.
+        /// </summary>
+        public TaskCompletionSource<bool> FlushTaskCompletionSource
```
pharring: Is this really supposed to be public? It feels like implementation detail. Should it be internal?
Lower down, there's also a public `SetFlushTaskCompletionSource` method which is probably misnamed -- suggest: "SetFlushTaskCompletionSourceResult".

There's probably a narrower API we can expose from Transmission to support manual flushing.

rajkumar-rangaraj: Changed name to SetFlushTaskCompletionSourceResult. Transmission is a part of Microsoft.ApplicationInsights project. ServerTelemetryChannel is a part of different project. Most properties in this class are exposed as public.

BASE/src/Microsoft.ApplicationInsights/Channel/ITelemetryChannel.cs
rajkumar-rangaraj: Added CancellationToken

Main intention of this FlushAsync is to get called when the channel is shutting down. But, this can be used in other scenarios to confirm a successful transmission.

BASE/src/Microsoft.ApplicationInsights/Channel/ITelemetryChannel.cs
```diff
@@ -31,5 +32,11 @@ public interface ITelemetryChannel : IDisposable
         /// Flushes the in-memory buffer.
         /// </summary>
         void Flush();
+
+        /// <summary>
+        /// Flushes the in-memory buffer asynchronously.
+        /// </summary>
+        /// <returns>The task to await.</returns>
```
pharring: What is the meaning of the bool value? Does it indicate whether the Flush was successful? If it returns false, is there something else that the caller should do?

BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs
```diff
@@ -186,6 +188,17 @@ public void Flush(TimeSpan timeout)
...
 
+        /// <summary>
+        /// Will send all the telemetry items stored in the memory asynchronously.
+        /// </summary>
+        /// <returns>The task to await. Task has true set on successful flush.</returns>
+        public async Task<bool> FlushAsync(CancellationToken cancellationToken)
+        {
+            var taskResult = Task.Run(() => this.Flush(default(TimeSpan)), cancellationToken);
+            await taskResult.ConfigureAwait(false);
+            return taskResult.IsCompleted;
```
pharring: I still don't understand this. Can you explain?
Note: with the `await` in the line above, is there any way to get here without `IsCompleted` being true?

BASE/src/ServerTelemetryChannel/Implementation/ErrorHandlingTransmissionPolicy.cs
```diff
@@ -122,6 +122,11 @@ private void HandleTransmissionSentEvent(object sender, TransmissionProcessedEve
                     TelemetryChannelEventSource.Log.TransmissionDataLossError(e.Transmission.Id,
                         "Unknown Exception Message");
                 }
+
+                if (e.Transmission.ManualFlushAsyncFlag)
```
pharring: This check is unnecessary since it's done inside `SetFlushTaskCompletionSourceResult`

BASE/src/Microsoft.ApplicationInsights/Channel/ITelemetryChannel.cs
```diff
@@ -31,5 +32,11 @@ public interface ITelemetryChannel : IDisposable
...
+        Task<bool> FlushAsync(CancellationToken cancellationToken);
```
lmolkova: adding method to the interface is a [breaking change](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes#binary-breaking-change) and needs major version upgrade.

pharring: Perhaps it could/should be in a new interface. e.g. `ITelemetryChannel2` or `IAsyncFlushable`

BASE/src/ServerTelemetryChannel/Implementation/ErrorHandlingTransmissionPolicy.cs
rajkumar-rangaraj: Wanted to call SetFlushTaskCompletionSourceResult only if we have a special flag set in Transmission. As SetFlushTaskCompletionSourceResult is a public method and does want consumer to send a completion on Null object.

BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs
rajkumar-rangaraj: Incase of InMemoryChannel Flush. I want to just wrap existing Flush on a Task.Run. If Task is complete want to return success to caller.

return Task.Run(this.Flush(default(TimeSpan)), cancellationToken); ==> This link of code does not return bool or Task<bool>

rajkumar-rangaraj: Your comment in another section will solve this issue. Creating a new Interface like ITelemetryChannel2 or IAsyncFlushable and not touching the InMemoryChannel at all. Creating Async in InMemeoryChannel will not create any difference.

BASE/src/Microsoft.ApplicationInsights/Channel/IAsyncFlushable.cs
```diff
@@ -0,0 +1,22 @@
...
+
+    /// <summary>
+    /// Represents asynchronous flush for  to application insights.
+    /// </summary>
+    public interface IAsyncFlushable : IDisposable
+    {
+        /// <summary>
+        /// Flushes the in-memory buffer asynchronously.
+        /// </summary>
+        /// <returns>The task to await.</returns>
```
pharring: I think we still need to close on the meaning of the task result. It looks like it can be false if there were no items to flush. Is that the intent? Is the idea that callers should call it in a loop until it returns false?
What I'm getting at is what are clients supposed to do with the returned value? My guess is there's nothing reasonable they can or need to do.
Should this simply return `Task` instead of `Task<bool>`?

reyang: +1 on having a clear meaning of the task result (and what could a caller do with it).

This is what I'm thinking about the **semantic** - a `SUCCESS` return from the flush API indicates that the ownership of telemetry data emitted before the flush invocation has transferred outside the process; a `FAILED` return (or should we just throw exception rather than return?) indicates a failure to transfer the ownership (so the process still owns all or part of the telemetry).

For example, if the SDK cannot send the telemetry data to the ingestion (e.g. due to network unavailable) AND cannot persist the data locally (e.g. due to disk full or no access), the data ownership is still tight to the process life, flush should return failure (or throw).

BASE/src/ServerTelemetryChannel/Implementation/ErrorHandlingTransmissionPolicy.cs
pharring: I don't understand. Let me restate: The body of `SetFlushTaskCompletionSourceResult` already has the line `if (this.ManualFlushAsyncFlag)` and does nothing if the flag is false, then there's no point to testing the flag here. Just call `SetFlushTaskCompletionSourceResult` unconditionally.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
pharring: I recommend **not** exposing the FlushTaskCompletionSource. Make it a private field. Instead, expose a `FlushTask` property which returns the `.Task` of that private field.

BASE/src/ServerTelemetryChannel/Implementation/TelemetrySerializer.cs
```diff
@@ -27,6 +29,27 @@ public Uri EndpointAddress
...
+
+        public virtual Task<bool> SerializeAsync(ICollection<ITelemetry> items, CancellationToken cancellationToken)
+        {
+            this.HandleTelemetryException(items);
+
+            var transmission = new Transmission(this.EndpointAddress, items, true);
+
+            using (cancellationToken.Register(() => transmission.SetFlushTaskCompletionSourceResult(false)))
+            {
+                Task.Run(() => this.Transmitter.Enqueue(transmission));
```
pharring: I'd prefer to see cancellation result in a canceled task (not a completed one with a 'false' return).
That would imply that you need a method like `CancelFlushTask()` that calls `SetCanceled()` on the TCS.
That way, await-int the task will throw a TaskCanceledException as expected.
You might also want to pass the cancellationToken into the Task.Run call.
You should also add a `cancellationToken.ThrowIfCanceled()` call at the top of this method.
I'd also like to see unit tests that cover this case.
(Sorry for all the demands)

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -674,6 +676,37 @@ public void Flush()
...
+
+            TelemetryConfiguration pipeline = this.configuration;
+            if (pipeline != null)
+            {
+                MetricManager sharedMetricManager = pipeline.GetMetricManager(createIfNotExists: false);
+                sharedMetricManager?.Flush(flushDownstreamPipeline: false);
+
+                ITelemetryChannel channel = pipeline.TelemetryChannel;
+
+                if (channel is IAsyncFlushable)
```
pharring: Use pattern matching to avoid the extra cast and null check. i.e.
```csharp
if (channel is IAsyncFlushable asyncFlushableChannel)
{
    return asyncFlushableChannel.FlushAsync(cancellationToken);
}
```

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -674,6 +676,37 @@ public void Flush()
...
+                {
+                    return ((IAsyncFlushable)channel)?.FlushAsync(cancellationToken);
+                }
+            }
+
+            return Task.FromResult(false);
```
pharring: We should also check for cancellation here:
```csharp
return cancellationToken.IsCancellationRequested ? Task.FromCanceled<bool>(cancellationToken) : Task.FromResult(false);
```

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -133,6 +162,18 @@ public ICollection<ITelemetry> TelemetryItems
             get; private set;
         }
 
+        /// <summary>
+        /// Send signal to Flush Async.
+        /// </summary>
+        public void SetFlushTaskCompletionSourceResult(bool status)
```
pharring: Having looked at the usage, I think this is better expressed with three separate methods:
1. `CompleteFlush()`
2. `CancelFlush()`
3. `FailFlush(Exception ex)`

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -125,6 +136,24 @@ public string Id
             get; private set;
         }
 
+        /// <summary>
+        /// Gets a value indicating whether flush is async.
+        /// </summary>
+        public bool ManualFlushAsyncFlag
```
pharring: Boolean properties should be named `Is...` or `Has...`
e.g. `IsFlushAsync` or `HasFlushTask`

I also think this can be implemented completely over the `TaskCompletionSource` without using a separate (backing) field.

i.e.
```csharp
public bool HasFlushTask =>
    FlushTaskCompletionSource != null && !FlushTaskCompletionSource.Task.IsCompleted;
```
Then, in the `SetFlushTaskCompletionSourceResult()` method, you don't set this property to false, you just SetResult/SetCancel/SetException as necessary.

BASE/Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryBufferTest.cs
```diff
@@ -379,6 +380,137 @@ public void DoesNotContinueOnCapturedSynchronizationContextToImprovePerformance(
...
+            {
+                bool telemetrySerialized = false;
+                var serializer = new StubTelemetrySerializer();
+                serializer.OnSerialize = telemetry =>
+                {
+                    telemetrySerialized = true;
+                };
+                var telemetryBuffer = new TelemetryBuffer(serializer, new StubApplicationLifecycle());
+
+                telemetryBuffer.ManualFlushAsync(new CancellationToken()).GetAwaiter().GetResult();
```
pharring: Nit: use `CancellationToken.None` or, if you're allowed to use newer C# langauge features, just `default`

BASE/Test/ServerTelemetryChannel.Test/Shared.Tests/Implementation/TelemetryBufferTest.cs
```diff
@@ -379,6 +380,137 @@ public void DoesNotContinueOnCapturedSynchronizationContextToImprovePerformance(
...
+                    Assert.IsFalse(postedBack);
+                }
+            }
+
+            [TestMethod]
+            public void SetCancellationTokenOnManualFlushAsync()
+            {
+                var telemetryBuffer = new TelemetryBuffer(new StubTelemetrySerializer(), new StubApplicationLifecycle());
+                var task = telemetryBuffer.ManualFlushAsync(new CancellationToken(true));
+                Thread.Sleep(TimeSpan.FromMilliseconds(100));
```
pharring: This is fragile (subject to timing variance).
Can you "await" the task and catch OperationCanceledException?

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -674,6 +676,44 @@ public void Flush()
...
+                {
+                    return asyncFlushableChannel.FlushAsync(cancellationToken);
+                }
+            }
+
+            return cancellationToken.IsCancellationRequested ? Task.Factory.StartNew(() =>
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                return false;
+            }, cancellationToken) : Task.FromResult(false);
```
pharring: I presume this block of code is a substitute for `Task.FromCanceled(cancellationToken)`. Can you move it to a static utility class with `#ifdef`s so that targets that _do_ have Task.FromCanceled (all except NET45) can benefit from that built-in? I think we used to have a TaskEx class for things like that, but I can't see it in the sources.

TimothyMothra:
@rajkumar-rangaraj your latest commit deleted a test class "TelemetryConfigurationFactoryTest.cs "
That's a very important test class, I can't let you delete it without a replacement.
Please sync with me offline so I can help you resolve any conflicts.

BASE/src/Microsoft.ApplicationInsights/Channel/IAsyncFlushable.cs
rajkumar-rangaraj: Added detailed documentation for return type

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
rajkumar-rangaraj: Added support for Task.Canceled and simplified return.

BASE/src/ServerTelemetryChannel/ServerTelemetryChannel.cs
```diff
@@ -310,6 +311,29 @@ public void Flush()
...
+                TelemetryChannelEventSource.Log.StorageNotInitializedError();
+            }
+
+            TelemetryChannelEventSource.Log.TelemetryChannelFlushAsync();
+
+#if NET45
+            return cancellationToken.IsCancellationRequested ? Task.Factory.StartNew(() => false, cancellationToken) : this.TelemetryBuffer.ManualFlushAsync(cancellationToken);
+#else
+            return cancellationToken.IsCancellationRequested ? Task.FromCanceled<bool>(cancellationToken) : this.TelemetryBuffer.ManualFlushAsync(cancellationToken);
+#endif
```
pharring: This pattern appears 3 times in the PR. It's not exactly the same each time, but it is all for the lack of `Task.FromCanceled` in NET45.

I know I'm repeating myself, but I suggest creating a helper method (or maybe 2 if we need the non-generic version too) that matches the signature of `Task.FromCanceled` and can be called from here and the other 2 places without using conditional compilation (`#if / #else / #endif`).

For example:
```csharp
internal static class TaskEx
{
    public static Task FromCanceled(CancellationToken cancellationToken)
    {
#if NET45
        // TODO: Equivalent for .NET 4.5
#else
        return Task.FromCanceled(cancellationToken);
#endif  
    }

    public static Task<TResult> FromCanceled<TResult>(CancellationToken cancellationToken)
    {
#if NET45
        // TODO: Equivalent for .NET 4.5
#else
        return Task.FromCanceled<TResult>(cancellationToken);
#endif  
    }
}
```

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -133,6 +134,47 @@ public ICollection<ITelemetry> TelemetryItems
             get; private set;
         }
 
+        /// <summary>
+        /// Gets a value indicating whether flush is async.
+        /// </summary>
+        public bool HasFlushTask => this.flushTaskCompletionSource != null && !this.flushTaskCompletionSource.Task.IsCompleted;
```
debugerr: Would this syntax work here: 

`!this.flushTaskCompletionSource?.Task.IsCompleted;`

?

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -133,6 +134,47 @@ public ICollection<ITelemetry> TelemetryItems
...
+
+        /// <summary>
+        /// Gets Test to support async Flush.
+        /// </summary>
+        public Task<bool> GetFlushTask(CancellationToken cancellationToken)
+        {
+            if (this.flushTaskCompletionSource == null)
+            {
+                this.flushTaskCompletionSource = new TaskCompletionSource<bool>();
```
debugerr: I think if this is called by multiple threads you have a potential for race and double initialization + double registration in .Register.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -133,6 +134,47 @@ public ICollection<ITelemetry> TelemetryItems
...
+        }
+
+        /// <summary>
+        /// Send signal to flush async task on completion.
+        /// </summary>
+        public void CompleteFlushTask(bool success)
+        {
+            if (this.HasFlushTask)
+            {
+                this.flushTaskCompletionSource.TrySetResult(success);
```
debugerr: `this.flushTaskCompletionSource?.TrySetResult(success);` 

might work here as well?

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -674,6 +676,43 @@ public void Flush()
...
+
+                if (channel is IAsyncFlushable asyncFlushableChannel && !cancellationToken.IsCancellationRequested)
+                {
+                    return asyncFlushableChannel.FlushAsync(cancellationToken); 
+                }
+            }
+#if NET45
+            return cancellationToken.IsCancellationRequested ? Task.Factory.StartNew(() => false, cancellationToken) : Task.FromResult(false);
+#else
+            return cancellationToken.IsCancellationRequested ? Task.FromCanceled<bool>(cancellationToken) : Task.FromResult(false);
```
debugerr: if we had a valid privateMetricManager on line 693 and flushed those out, we would still return false here, but we actually flushed data. Would we want to return true in that case?

BASE/src/ServerTelemetryChannel/Implementation/TelemetrySerializer.cs
```diff
@@ -27,6 +29,29 @@ public Uri EndpointAddress
...
+
+            var transmission = new Transmission(this.EndpointAddress, items);
+            Task<bool> resultTask = transmission.GetFlushTask(cancellationToken);
+
+            if (!cancellationToken.IsCancellationRequested)
+            {
+                Task.Run(() => this.Transmitter.Enqueue(transmission), cancellationToken);
+            }
+
+            return resultTask;
```
debugerr: If the cancellationToken was canceled, when is the resultTask canceled?

debugerr: I think it will be cancelled when GetFlushTask is called, as there is the registration for the cancellation in there.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
debugerr: It could be problematic, for the first task that would get initialized. Unless it would be cancelled, it would never complete.

rajkumar-rangaraj: This gets called only when client calls FlushAsync API and not meant to be called as standalone. Each transmission will have its own TaskCompletionSource. TelemetryChannel and Microsoft.ApplicationInsights are two different project, we have to make this method as public for TelemetryChannel to consume it.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
rajkumar-rangaraj: HasFlushTask is present to avoid setting flushTaskCompletionSource for the second time. It should be set only once.

BASE/src/ServerTelemetryChannel/Implementation/TelemetrySerializer.cs
rajkumar-rangaraj: correct!

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
rajkumar-rangaraj: It gives syntax error as we have .NET45 build

BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Tests/Implementation/TelemetryBufferTest.cs
```diff
@@ -379,6 +380,145 @@ public void DoesNotContinueOnCapturedSynchronizationContextToImprovePerformance(
...
+            {
+                bool telemetrySerialized = false;
+                var serializer = new StubTelemetrySerializer();
+                serializer.OnSerialize = telemetry =>
+                {
+                    telemetrySerialized = true;
+                };
+                var telemetryBuffer = new TelemetryBuffer(serializer, new StubApplicationLifecycle());
+
+                telemetryBuffer.FlushAsync(CancellationToken.None).GetAwaiter().GetResult();
```
pharring: :bulb: You can use `await` in test methods. Just make the signature `public async Task DoesntSerialize...`

pharring: :bulb: Also check the result of the FlushAsync task. It should be 'true', right?

BASE/src/Microsoft.ApplicationInsights/Channel/IAsyncFlushable.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// <summary>
+    /// Represents asynchronous flush for  to application insights.
+    /// </summary>
+    public interface IAsyncFlushable : IDisposable
+    {
+        /// <summary>
+        /// Flushes the in-memory buffer asynchronously.
+        /// </summary>
+        /// <returns>
+        /// True indicates telemetry data ownership is transferred out of process, that are emitted before the flush invocation.
```
cijothomas: Lets elaborate more on whats meant by out-of-process - "that data may either be sent all the way to Azure service, or stored in local disk."

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -133,6 +134,47 @@ public ICollection<ITelemetry> TelemetryItems
             get; private set;
         }
 
+        /// <summary>
+        /// Gets a value indicating whether flush is async.
```
cijothomas: this comment seems incorrect. Bool indicates whether flush task is completed or not.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
cijothomas: Editorbrowsable can be set to false, if this was not intended for consumers to use.

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
```diff
@@ -161,6 +162,34 @@ internal virtual void Enqueue(Transmission transmission)
 
             if (this.Buffer.Enqueue(transmissionGetter))
             {
+                if (transmission.HasFlushTask)
```
cijothomas: Please add comments to explain why Transmitter is forcing items out of buffer.

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
```diff
@@ -170,7 +199,10 @@ internal virtual void Enqueue(Transmission transmission)
             if (!this.Storage.Enqueue(transmissionGetter))
             {
                 TelemetryChannelEventSource.Log.TransmitterStorageSkipped(transmission.Id);
+                transmission.CompleteFlushTask(false);
             }
+
+            transmission.CompleteFlushTask(true);
```
cijothomas: please add comment.

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
```diff
@@ -161,6 +162,34 @@ internal virtual void Enqueue(Transmission transmission)
...
+                        int temp_maxSenderCapacity = this.maxSenderCapacity;
+                        int temp_maxBufferCapacity = this.maxBufferCapacity;
+                        this.maxSenderCapacity = 0;
+                        this.maxBufferCapacity = 0;
+                        this.Storage.IsEnqueueSuccess = false;
+
+                        // Move buffer to storage
+                        this.ApplyPolicies();
+
+                        if (this.Storage.IsEnqueueSuccess)
```
cijothomas: Can we rely on IsEnqueueSuccess as it stores the last state?

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
```diff
@@ -161,6 +162,34 @@ internal virtual void Enqueue(Transmission transmission)
...
+                {
+                    lock (this.lockObj)
+                    {
+                        int temp_maxSenderCapacity = this.maxSenderCapacity;
+                        int temp_maxBufferCapacity = this.maxBufferCapacity;
+                        this.maxSenderCapacity = 0;
+                        this.maxBufferCapacity = 0;
+                        this.Storage.IsEnqueueSuccess = false;
+
+                        // Move buffer to storage
```
cijothomas: if the intention is to move everything to storage, that can done directly, without calling ApplyPolicies and setting capacity to 0.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionSender.cs
```diff
@@ -205,6 +205,7 @@ private async Task StartSending(Transmission transmission)
                     }
 
                     this.OnTransmissionSent(new TransmissionProcessedEventArgs(acceptedTransmission, exception, responseContent));
+                    transmission.CompleteFlushTask(true);
```
cijothomas: add comments.

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -674,6 +676,40 @@ public void Flush()
...
+        /// Asynchronously Flushes the in-memory buffer and any metrics being pre-aggregated.
+        /// </summary>
+        /// <returns>
+        /// True indicates telemetry data ownership is transferred out of process, that are emitted before the flush invocation.
+        /// False indicates transfer of telemetry data has failed, the process still owns all or part of the telemetry.
+        /// </returns>
+        /// <remarks>
+        /// <a href="https://go.microsoft.com/fwlink/?linkid=525722#flushing-data">Learn more</a>
+        /// </remarks>
+        public Task<bool> FlushAsync(CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
rajkumar-rangaraj: FlushAsync is added only for ServerTelemetryChannel.

Issue #1653, opened 2020-01-29
SDK Diagnostics Mode, emit ETW logs without redeploying an application.


TimothyMothra:
This is a continuation of #1059 

We want to introduce a **DiagnosticsMode** into the SDK.
This would be controlled by an environment variable.
A customer could set this instead of redeploying an application, although this would require restarting an application.

### First Version (2.13 milestone)
- Diagnostic Mode would enable the `FileDiagnosticTelemetryModule` to collect ETW logs emitted by the Application Insights SDK.
- The application would need to be restarted for this change to take effect.

### Future suggested features
- To enable without restarting an app.
- To allow the SDK to throw internal exceptions.
- Set the verbosity level of logs to collect.
- Possibly include this the config file to control this on a per-application level. Possibly simpler than manually configuring modules themselves. (Scenario-based config values)

### Technical Spec

Naming is up for debate. Suggestions are "Diagnostic(s)", "DiagnosticMode", "SelfDiagnostics"

- Environment Variable: `APPLICATIONINSIGHTS_DIAGNOSTICS`
    - DataType: String
    - Value: `true/enabled` to turn on. `false/disabled` or delete the variable to turn off. Provides flexibility for new keywords in the future.
    - If not set, SDK default is `false`.
- Environment Variable: `APPLICATIONINSIGHTS_DIAGNOSTICS_LOG_DIRECTORY`
    - DataType: String
    - Value: full path of the directory where the log files would be stored. 
    - If not set, SDK default is `%TEMP%`.

### How to enable

- Windows Command Line: `set APPLICATIONINSIGHTS_DIAGNOSTICS=true`
- PowerShell: `$env:APPLICATIONINSIGHTS_DIAGNOSTICS = 'true'`

TimothyMothra:
@reyang - please review.

@trask, @littleaj - Can you please provide feedback? 
My goal is that this should be generic enough to be implemented by other SDKs so we can have a consistent supportability story. I would love to get your opinions/perspectives.

reyang:
Adding @maxgolov who has similar requirements for our C++ SDKs.

Is there a reason that we prefer the SDK to collect ETW logs generated by itself and pipe it to a file?
Have we considered providing a separate tool which interacts with SDK internal ETW logs?

reyang:
What is the expected behavior if the environment variable is set to something else (e.g. "foo")?

TimothyMothra:
[...]

PR #1658, opened 2020-01-31
[WIP] sdk diagnostics mode

TimothyMothra:
Fix Issue #1653  .

## Changes
- Modified TelemetryConfigurationFactory to evaluate new Diagnostics environment variables

### Checklist

### Notes for authors:

### Notes for reviewers:


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
```diff
@@ -112,6 +119,33 @@ public virtual void Initialize(TelemetryConfiguration configuration, TelemetryMo
...
 
+        [SuppressMessage("Microsoft.Reliability", "CA2000:DisposeObjectsBeforeLosingScope", Justification = "We want objects created in this method to live for the life of the application.")]
+        [SuppressMessage("Microsoft.Usage", "CA1801:ReviewUnusedParameters", Justification = "This value is not used in NetStandard1.3 but is used for all other frameworks..")]
+        internal void EvaluateDiagnosticsMode(TelemetryModules modules)
+        {
+#if !NETSTANDARD1_3
+            // TODO: NEED TO EMIT ETW LOGS
+            if (PlatformSingleton.Current.TryGetEnvironmentVariable(DiagnosticsEnvironmentVariable, out string diagnosticsValue) && bool.TryParse(diagnosticsValue, out bool boolValue) && boolValue)
+            {
+                if (modules != null && !modules.Modules.Any(module => module is FileDiagnosticsTelemetryModule))
```
Dmitry-Matveev: We may want to make Environment variables (Diag=true/false and LogDir=///) win over configuration file for the same parameters, so that these values can be affected outside of the process, just like connection string variables are. In the current implementation we seem to ignore LogDir variable if FileDiagnosticsTelemetryModule is already configured.

That also brings a question on what to do if FileDiagnostics Module is mentioned in both config and variables - do we keep both module or only register first/discard the second or something else?

TimothyMothra: If i inject a Module via EnVar, AND a Module exists in the customer's config... currently the TCFactory will override the first (EnvVar) with the one from the config.
THIS is a problem because the new module will have lost all of the SDK initialization ETW traces.

Note for self: 
On Monday:
Add a new derived private class  `SelfFileDiagnosticsmodule : FileDiagnosticsModule`.
We will inject the private one via EnVar and this should not conflict with the one from the config file.

TimothyMothra:
Abandoning this PR. will be reworking to meet new requirements.

Issue #1702, opened 2020-02-21
AspNetCore AddApplicationInsightsSettings() and MissingMethodException


TimothyMothra:
This was brought to my attention by @brettsam

----

In  AppInsights.AspNetCore 2.12.0 I changed the method signature of `AddApplicationInsightsSettings` by adding the optional parameter `connectionString`.
https://github.com/microsoft/ApplicationInsights-dotnet/blob/0076e0fac81a6a7fadb36b390a863fc2024c1e84/NETCORE/src/Shared/Extensions/ApplicationInsightsExtensions.cs#L163-L168

This can cause a **MissingMethodException** for users upgrading from an older version of our SDK.
(Explanation of MissingMethodException and default parameters: https://stackoverflow.com/a/9884700/1466768)

## Next Steps
I think this will be a straightforward fix, I should be able to add a method overload matching the signature of the original method.
I'm setting up a repro right now so I can verify a fix.
**The plan is to release a patch for versions 2.12, 2.13, and 2.14**

## Next Next Steps
This highlights the need to add a PublicApiAnalyzer to our other projects.
We currently only have this analyzer on 2/17 of our projects.

TimothyMothra:
I have a repro and I'm verifying the fix.

I'm trying to keep both method signatures (old and new), but this can cause **AmbiguousMatchException**. 

I think the solution is going to look like this:
```
        public static IConfigurationBuilder AddApplicationInsightsSettings(
            this IConfigurationBuilder configurationSourceRoot,  
            bool? developerMode = null, 
            string endpointAddress = null, 
            string instrumentationKey = null)
            => configurationSourceRoot.AddApplicationInsightsSettings(developerMode, endpointAddress, instrumentationKey);

        public static IConfigurationBuilder AddApplicationInsightsSettings(
            this IConfigurationBuilder configurationSourceRoot,
            string connectionString,
            bool? developerMode = null,
            string endpointAddress = null,
            string instrumentationKey = null)
        {
```

Note that `connectionString` can't be an optional parameter because that causes the Ambiguous exception.
The only fix is going to be moving it to the first parameter, which will be a breaking change for anybody who took a dependency on it being the last. :(
Fun times.

TimothyMothra:
The fix has been published to NuGet and should be generally available within an hour.

- https://www.nuget.org/packages/Microsoft.ApplicationInsights.AspNetCore/2.12.2
[...]

PR #1703, opened 2020-02-21
fix for AspNetCore AddApplicationInsightsSettings() and MissingMethodException

TimothyMothra:
Fix Issue #1702.

## Changes
- Restore the removed method signature, now an overload.
- Note that this brakes the method signature for anyone who took a dependency on the order of parameters instead of using named parameters.

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/NETCORE/src/Shared/Extensions/ApplicationInsightsExtensions.cs
+++ b/NETCORE/src/Shared/Extensions/ApplicationInsightsExtensions.cs
@@ -158,14 +158,25 @@
         /// <param name="developerMode">Enables or disables developer mode.</param>
         /// <param name="endpointAddress">Sets telemetry endpoint address.</param>
         /// <param name="instrumentationKey">Sets instrumentation key.</param>
-        /// <param name="connectionString">Sets connection string.</param>
         /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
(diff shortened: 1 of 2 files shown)


NETCORE/src/Shared/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -158,14 +158,25 @@ public static IServiceCollection ConfigureTelemetryModule<T>(this IServiceCollec
         /// <param name="developerMode">Enables or disables developer mode.</param>
         /// <param name="endpointAddress">Sets telemetry endpoint address.</param>
         /// <param name="instrumentationKey">Sets instrumentation key.</param>
-        /// <param name="connectionString">Sets connection string.</param>
         /// <returns>The <see cref="IConfigurationBuilder"/>.</returns>
+        public static IConfigurationBuilder AddApplicationInsightsSettings(this IConfigurationBuilder configurationSourceRoot,  bool? developerMode = null, string endpointAddress = null, string instrumentationKey = null)
+            => configurationSourceRoot.AddApplicationInsightsSettings(developerMode, endpointAddress, instrumentationKey);
```
Dmitry-Matveev: This method call does not seem to use `connectionString` parameter or assign it to null - can you take a look?

TimothyMothra: wow, so I was blindly trusting unit tests to catch something like this.
what actually happened was that it is an infinite loop, which crashes the test assembly and no errors get reported. 
learning lots of things today

PR #1723, opened 2020-03-10
Removing the SQL command text from the dependency name for Framework 

stebet:
Removing the SQL command text from the dependency name for Framework processing to reduce event bloat.

## Changes
When using the [Microsoft.Data.SqlClient](https://www.nuget.org/packages/Microsoft.Data.SqlClient/) NuGet package for SQL statements, which now provides the SQL Command Text, the SQL dependency telemetry are all bloated with the SQL Command text for ad-hoc queries when running under .NET Framework.

For comparison, the .NET Core SQL dependency collector does not use the command text in the dependency name, and just relies on the `Data` property to provide that information, so this should bring more parity between .NET Framework and .NET Core SQL dependencies, and reduce the amount of data sent when running under the .NET Framework and executing ad-hoc statements (like when using Entity Framework).

### Checklist


stebet:
Here is an example of what the dependency telemetry currently looks like in the Performance->Dependencies blade in AppInsights. The long blurred out segments are the SQL Command Text, which should only be needed in the "Data" property and shouldn't be a part of the name. This might be handy for stored procs, but it is dreadful for long ad-hoc queries.

image

stebet:
Is this something that can be looked at @TimothyMothra? Don't know if I'm supposed to add a reviewer or if that's all on your end :)

TimothyMothra:
Hi @stebet !
Our work schedules have been interrupted and we fell behind in reviews. Thanks for the reminder. 
This looks straight forward to me, I'll queue the builds now.

TimothyMothra:
@cijothomas I'm okay with this change. I just wanted to double-check with you... do you know if anything was dependent on this behavior?

cijothomas:
Thanks. The full sql text was supposed to be part of .Data only . Thanks for fixing this.

Issue #1727, opened 2020-03-12
Application Insights Profiler stopped working with MS.AI.AspNetCore 2.14.0-beta3


xiaomi7732:
.NET Core Web API (2.1) + MS.AI.AspNetCore 2.14.0-beta3 + AI.Profiler from site extension.

* SDK version: 2.14.0-beta3
  * This does **NOT** repro with previous versions of SDK - 2.14.0-beta2 or 2.13.1.
* .NET Core 2.1 LTS, .NET Core 3.1 LTS, target netcoreapp2.1 / 3.1 accordingly.
* Hosted on Azure Web Apps
* Platform and OS version: Windows

Starting with 2.14.0-beta3, the profiler can't get request events.

Still looking into what is the cause. At the same time, @TimothyMothra @SergeyKanzhelev any direction?

xiaomi7732:
Loop in @Dmitry-Matveev for awareness.

cijothomas:
https://github.com/microsoft/ApplicationInsights-dotnet/pull/1715/files#diff-532acfbbd4b63e26096cb70d12ef859dR25 The envelope name changed cause this?

This changed json format over wire too.
{"name":"Microsoft.ApplicationInsights.Dev.79a19363f02545b7a3f4b2eb317a4db7.Request",  
 - beta2

{"name":"AppRequests", - beta2

Dmitry-Matveev:
@cijothomas , yes, we are covering/debugging it offline with @xiaomi7732 , added you to the thread as well.

xiaomi7732:
Verified working with 2.14.0-beta5. Close this one.

PR #1730, opened 2020-03-13
Revert envelope names for ETW channel

Dmitry-Matveev:
Fix Issue #1727  .

## Changes
Reverted to previous envelope names for all telemetry types on ETW channel.

### Checklist
- [x] I ran Unit Tests locally. __NET45/NET46 didn't run locally again, will fix the tests in PR__

### Notes for reviewers:


BASE/src/Microsoft.ApplicationInsights/DataContracts/AvailabilityTelemetry.cs
```diff
@@ -15,6 +15,7 @@
     /// </summary>
     public sealed class AvailabilityTelemetry : ITelemetry, ISupportProperties, ISupportMetrics, IAiSerializableTelemetry
     {
+        internal const string EtwEnvelopeName = "Availability";
```
SergeyKanzhelev: it may be the best to move these strings to the ETW channel itself.

Dmitry-Matveev: I see. My initial idea was to reduce the confusion for anyone looking into the code and assuming that new name is for everything. Const inside the telemetry type immediately lets the person know that there is more to envelope name than meets the eye.

It's used in two places: RichPayloadEventSource and RichPayloadEventSource.TelemetryHandler, so new home will still be something shared between these two, e.g. a separate constants file in the RPES namespace.

If you think we still need to proceed with moving out of telemetry classes, it's 5-min change, just let me know in response.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/RichPayloadEventSourceTest.cs
```diff
@@ -297,8 +316,8 @@ public void RichPayloadEventSourceOperationStartStopTest()
                     // Expect exactly two events (start and stop)
                     var actualEvents = listener.Messages.Where(m => m.Keywords.HasFlag(RichPayloadEventSource.Keywords.Operations)).Take(2).ToArray();
 
-                    VerifyOperationEvent(requestTelemetry, RequestTelemetry.TelemetryName, EventOpcode.Start, actualEvents[0]);
-                    VerifyOperationEvent(requestTelemetry, RequestTelemetry.TelemetryName, EventOpcode.Stop, actualEvents[1]);
+                    VerifyOperationEvent(requestTelemetry, ServiceProfilerNameContracts[requestTelemetry.GetType()], EventOpcode.Start, actualEvents[0]);
```
Dmitry-Matveev: @xiaomi7732, @SergeyKanzhelev , this is new behavior here and on lines 357-360 below. Tests passed in PR but failed locally on these checks. It looks like tests expected that when Request and Dependency are used in StartOperation API they always yield "operation" as event name instead of "Request" and "RemoteDependency".

However, somewhere along our changes (unknown ATM, tests on PRs were passing regardless...) we changed this behavior and now if outside operation is RequestTelemetry its event name is still "Request" in ETW and not "Operation". Please let me know how important it is and what it the expected behavior here.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/RichPayloadEventSourceTest.cs
```diff
@@ -23,6 +23,25 @@
     [TestCategory("WindowsOnly")] // do not run these tests on linux builds
     public class RichPayloadEventSourceTest
     {
+        public Dictionary<Type, string> ServiceProfilerNameContracts = new Dictionary<Type, string>
```
Dmitry-Matveev: @xiaomi7732, added contract tests as per our conversation.

BASE/src/Microsoft.ApplicationInsights/DataContracts/AvailabilityTelemetry.cs
Dmitry-Matveev: @SergeyKanzhelev , did you have a change to take a look at my response?

Issue #1761, opened 2020-04-06
Fix issue around checking whether the Activity type is available in ActivityExtensions.


macrogreg:
There are two issues in the **Microsoft.ApplicationInsights.ActivityExtensions** class: a correctness bug around concurrency, and a potential issue around reflection. The referenced PR addresses both problems (https://github.com/microsoft/ApplicationInsights-dotnet/pull/1762).

Details:

1) The TryRun/Initialize code has a concurrency bug where `isInitialized` is assigned a value before `isAvailable`. The former is assigned a value inside the [Initialize() method](https://github.com/microsoft/ApplicationInsights-dotnet/blob/a185e8fb05ce914af48eff8739e63b5918b50f03/BASE/src/Microsoft.ApplicationInsights/ActivityExtensions.cs#L71) and the later only after that method [returns](https://github.com/microsoft/ApplicationInsights-dotnet/blob/a185e8fb05ce914af48eff8739e63b5918b50f03/BASE/src/Microsoft.ApplicationInsights/ActivityExtensions.cs#L26). As a result, the default value of `isAvailable` may be used (`false`) where `true` would have been appropriate.

2) The `Initialize()` method uses AssemblyLoad(..) to [see if the Activity type is available](https://github.com/microsoft/ApplicationInsights-dotnet/blob/a185e8fb05ce914af48eff8739e63b5918b50f03/BASE/src/Microsoft.ApplicationInsights/ActivityExtensions.cs#L55). This attempts to load a specific assembly with a specific version, and relies on the assumption that if it successes, then the required type is available. This makes an unvalidated assumption about the location of that type. The assumption may be incorrect: If the hosting environment performed non-trivial type redirection; or in the future, if the type should move. Instead of loading the assembly, we can try loading the type directly. This will follow any potential type redirects and avoid any problems related to explicit assembly loads.

TimothyMothra:
close because we're removing this feature

PR #1762, opened 2020-04-06
Fix issue around checking whether the Activity type is available in ActivityExtensions.

macrogreg:
Fix issue around checking whether the Acitity type is available in ActivityExtensions (https://github.com/microsoft/ApplicationInsights-dotnet/issues/1761).

### DIFF
--- a/BASE/src/Microsoft.ApplicationInsights/ActivityExtensions.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/ActivityExtensions.cs
@@ -3,14 +3,18 @@
     using System;
     using System.Diagnostics;
     using System.Linq;
-    using System.Reflection;
     using System.Runtime.CompilerServices;
+    using System.Threading;
 
     internal static class ActivityExtensions
     {
         private const string OperationNameTag = "OperationName";
-        private static bool isInitialized;
-        private static bool isAvailable;
+
+        private const int ActivityAvailabilityUnknown = 0;
+        private const int ActivityAvailable = 1;
+        private const int ActivityNotAvailable = 2;
+
+        private static int activityTypeAvailability = ActivityAvailabilityUnknown;
 
         /// <summary>
         /// Executes action if Activity is available (DiagnosticSource DLL is available).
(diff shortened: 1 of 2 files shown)


cijothomas:
Thanks @macrogreg . I'll wait to hear from @SergeyKanzhelev if this TryRun approach can be removed entirely.

I believe the whole idea of TryRun is to enable scenario where AI.dll can run standalone without System.DiagnosticsSource.dll. I wonder if this is even a valid scenario? For anyone installing application insights via nuget, System.DiagnosticsSource will be installed by nuget. 
The only scenario I can think of is - if someone manually copies the AI.DLL file for use from a powershell module?

@SergeyKanzhelev Do you think if we should still support this? Every time we upgrade System.DiagnosticSource package version, we keep updating this file.

BASE/src/Microsoft.ApplicationInsights/ActivityExtensions.cs
```diff
@@ -47,29 +50,62 @@ internal static void SetOperationName(this Activity activity, string operationNa
...
+                return false;
             }
         }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static bool LoadActivityType()
+        {
+            // At this point the Activity type has already been loaded, or - if not - the runtime will attempt to load the assembly from which it is referenced.
+            // While doing so, it will respect type redirects potentially set up by the execution environment. This is preferable to doing Assembly-Load directly.
+            return typeof(System.Diagnostics.Activity) != null;
```
cijothomas, line 108: SDK must have a particular version of the assembly, as it relies of features not available in older versions. (like w3c format etc.) This was probably the motivation behind original code trying to load the exact same version.
With this code, we may break else where, if an Activity of older version is available, and we attempt to call methods not available in that version.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## VNext
 
 ## Version 2.14.0-beta4
+- [Fix: https://github.com/microsoft/ApplicationInsights-dotnet/issues/1761](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1761)
```
TimothyMothra, line 6: Please move this to the VNext section. Beta4 has already been compiled and will be released soon.

cijothomas:
Closing this PR. Had synced with Sergey and decided that we wont need to support the standalone usage of ai.dll. That makes this PR irrelevant, as SDK would no longer attempt to load DiagnosticSource.

Issue #1820, opened 2020-04-28
Remove support for running applicationinsights.dll standalone


cijothomas:
**Is your feature request related to a problem?**
The SDK currently attempts to load `System.Diagnostics.DiagnosticSource` to determine if it is available. All code related to `Activity` is wrapped around a logic to see if DiagnosticSource dll is available.
https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/BASE/src/Microsoft.ApplicationInsights/ActivityExtensions.cs

This has several issues:
1. Everytime we update DiagnosticSouce package, this logic needs update.
2. There is no valid scenario where this support is required.

**Describe the solution you'd like.**
Simply cut support for this scenario. Ai.dll cannot be used standalone - it requires DiagnosticSouce, and it is listed as dependency in nuspec.

**Additional context.**
The original reason why this was supported was for powershell scripts which simply copy "ai.dll" and used it.

PR #1823, opened 2020-04-28
Removing TryRun method from ActivityExtensions

eddynaka:
Fix Issue #1820.

## Changes
Removed TryRun and Initialize from ActivityExtensions
Maintained the logic and changed the tryRun with a TryCatch. If an exception occurs, we change to false and everything will be the same. A log will be created.

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/OperationHolder.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/OperationHolder.cs
@@ -74,12 +74,12 @@
                         var operationTelemetry = this.Telemetry;
                         operationTelemetry.Stop();
 
-                        bool isActivityAvailable = false;
-                        isActivityAvailable = ActivityExtensions.TryRun(() =>
+                        bool isActivityAvailable = true;
+                        try
                         {
                             var currentActivity = Activity.Current;
-                            if (currentActivity == null 
-                            || (Activity.DefaultIdFormat != ActivityIdFormat.W3C && operationTelemetry.Id != currentActivity.Id) 
+                            if (currentActivity == null
+                            || (Activity.DefaultIdFormat != ActivityIdFormat.W3C && operationTelemetry.Id != currentActivity.Id)
                             || (Activity.DefaultIdFormat == ActivityIdFormat.W3C && operationTelemetry.Id != currentActivity.SpanId.ToHexString()))
                             {
                                 // W3COperationCorrelationTelemetryInitializer changes Id
@@ -107,13 +107,18 @@
 
                             currentActivity?.Stop();
 
-                            if (this.originalActivity != null && 
-                                Activity.Current != this.originalActivity && 
+                            if (this.originalActivity != null &&
+                                Activity.Current != this.originalActivity &&
                                 this.originalActivity is Activity activity)
                             {
                                 Activity.Current = activity;
                             }
-                        });
+                        }
+                        catch (Exception exc)
+                        {
+                            CoreEventSource.Log.ActivityNotAvailable(exc.ToInvariantString());
+                            isActivityAvailable = false;
+                        }
 
                         if (!isActivityAvailable)
                         {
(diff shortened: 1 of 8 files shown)


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -634,6 +634,9 @@ public void InitializationIsSkippedForSampledItem(string appDomainName = "Incorr
         [Event(67, Message = "Backend has responded with {0} status code in {1}ms.", Level = EventLevel.Informational)]
         public void IngestionResponseTime(int responseCode, float responseDurationInMs, string appDomainName = "Incorrect") => this.WriteEvent(67, responseCode, responseDurationInMs, this.nameProvider.Name);
 
+        [Event(68, Message = "Activity isn't available.", Level = EventLevel.Error)]
+        public void ActivityNotAvailable() => this.WriteEvent(68, this.nameProvider.Name);
```
TimothyMothra, line 638: You're missing this default paramater: `string appDomainName = "Incorrect"`

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/OperationHolder.cs
```diff
@@ -105,13 +105,18 @@ protected virtual void Dispose(bool disposing)
...
+                                Activity.Current != this.originalActivity &&
                                 this.originalActivity is Activity activity)
                             {
                                 Activity.Current = activity;
                             }
-                        });
+                        }
+                        catch (Exception)
+                        {
+                            CoreEventSource.Log.ActivityNotAvailable();
```
TimothyMothra, line 117: I expect that we should log the exception here as well.

eddynaka: thanks for reviewing. just added the parameter to save it!

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -634,6 +634,9 @@ public void InitializationIsSkippedForSampledItem(string appDomainName = "Incorr
         [Event(67, Message = "Backend has responded with {0} status code in {1}ms.", Level = EventLevel.Informational)]
         public void IngestionResponseTime(int responseCode, float responseDurationInMs, string appDomainName = "Incorrect") => this.WriteEvent(67, responseCode, responseDurationInMs, this.nameProvider.Name);
 
+        [Event(68, Message = "Activity isn't available.", Level = EventLevel.Error)]
+        public void ActivityNotAvailable(string exception, string appDomainName = "Incorrect") => this.WriteEvent(68, exception, this.nameProvider.Name);
```
TimothyMothra, line 638: This won't work. check out some of the other examples of passing an exception to the EventSource.

eddynaka: @TimothyMothra , we have some patters, so i got confused:
- some places use exception.ToString()
- some places use exception.Message
- some places use exception.ToInvariantString()

After that, we have places that do (in CoreEventSource.cs):
- WriteEvent(number, error, this.nameProvider.name)
- WriteEvent(number, error ?? string.Empty, this.nameProvider.name)

TimothyMothra: where the exception is caught, you should call ToInvariantString().
This will pass a string into the EventSource message which needs to be included in the Message attribute.

eddynaka: ok! thanks for the update!

TimothyMothra:
@eddynaka This PR is perfect, but we decided that because this is a breaking change, we can't take it until the 3.0 major release.

Issue #1767, opened 2020-04-07
Remove NetStandard 1.x


TimothyMothra:
Parent: #1160 

Net framework compatibility: https://docs.microsoft.com/en-us/dotnet/standard/net-standard

Note that all Test projects should target NetCore2.1 and NetCore3.1. #1769 

#### Remove
- NetStandard 1.1
- NetStandard 1.3
- NetStandard 1.6

#### Replace With
- NetStandard 2.0
- NetStandard 2.1

#### Affected Projects (11 of 17)
- BASE
    - Microsoft.ApplicationInsights
    - ServerTelemetryChannel
- WEB
    - DependencyCollector
    - PerformanceCollector
    - WindowsServer
- LOGGING
    - DiagnosticSourceListener
    - EventSourceListener
    - Log4NetAppender
    - NLogTarget
    - TraceListener
- NETCORE
    - Microsoft.ApplicationInsights.AspNetCore

#### Affected Builds
- Gated build (ADO)
- Linux builds (yml)

#### Action Plan
- First, remove TargetFrameworks from Product. This immediately stops compiling the NetStandard 1.x **Done**
- Second, remove TargetFrameworks from Tests.  **Done**
- Third, cleanup code. Remove preprocessors and delete code forks. It's important to do this one project at a time so we can thoroughly review. (this can be done at any time by anyone on the team)
    - **PerformanceCollector** is fractured into 6 different shared projects. (NetFull, NetStandard, NetStandard.Stubs, NetStandard16.Stubs, NetStandard20, NetStandard20Net45). After deleting NetStandard1.x, this needs to be refactored into only 2 shared projects (Perf.Shared.NetFull, Perf.Shared.NetStandard).

#### Risks
- I noticed right away that some projects currently ONLY compile to netstandard 1.x.

TimothyMothra:
@eddynaka, i merged your first PR.
When you start working again, please work on PerformanceCollector by itself so we can thoroughly review. 
Thanks :)

TimothyMothra:
@eddynaka Next, can you remove the #IF NETCOREAPP1 code branches? These are mostly in our test classes

eddynaka:
Sure!

PR #1845, opened 2020-05-01
Remove not supported framework

eddynaka:
Fix Issue #1767

## Changes
- removing netcoreapp1.0 / 1.1 from targetframework
- removing netstandard1.3/1.6 and replace for netstandard2.0

### Checklist

### Notes for authors:

### Notes for reviewers:


TimothyMothra:
I would ask that you revert your changes to Perf and lets address those in a follow-up PR.
That effort should be different from changing the target frameworks.

eddynaka:
In the build pipeline yml, we have this:
https://github.com/microsoft/ApplicationInsights-dotnet/blob/3ac8c0b14466247263271a70ee5cbf3f54437145/NETCORE/.vsts/linux-build.yml#L26-L32

I couldn't find that project. Was that renamed or removed?

cijothomas:
> In the build pipeline yml, we have this:
> https://github.com/microsoft/ApplicationInsights-dotnet/blob/3ac8c0b14466247263271a70ee5cbf3f54437145/NETCORE/.vsts/linux-build.yml#L26-L32
> 
> I couldn't find that project. Was that renamed or removed?

Yes, renamed - https://github.com/microsoft/ApplicationInsights-dotnet/pull/1824

BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Nuget.Tests/TelemetryChannel.Nuget.Tests.csproj
```diff
@@ -20,7 +20,7 @@
 
   <PropertyGroup Condition="$(OS) != 'Windows_NT'">
```
cijothomas, line 21: this whole section need to be removed. This test is only required for net45. To avoid running them in linux, we can rename project to .Win.Tests.csproj, and use "*Win.Tests.proj" as a filter in Linux build yml file.
This would apply to any other test project which are only supposed to targe full framework like .NET45.
(This can be done separately from this PR)

eddynaka:
> > In the build pipeline yml, we have this:
> > https://github.com/microsoft/ApplicationInsights-dotnet/blob/3ac8c0b14466247263271a70ee5cbf3f54437145/NETCORE/.vsts/linux-build.yml#L26-L32
> > 
> > I couldn't find that project. Was that renamed or removed?
> 
> Yes, renamed - #1824

Thanks. Just updated the pipeline!

BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Nuget.Tests/TelemetryChannel.Nuget.Tests.csproj
eddynaka: I removed that section. and later, as you suggested, we can check a new pipeline in windows.
Thanks for reviewing

TimothyMothra:
> > > In the build pipeline yml, we have this:
> > > https://github.com/microsoft/ApplicationInsights-dotnet/blob/3ac8c0b14466247263271a70ee5cbf3f54437145/NETCORE/.vsts/linux-build.yml#L26-L32
> > > 
> > > I couldn't find that project. Was that renamed or removed?
> > 
> > 
> > Yes, renamed - #1824
> 
> Thanks. Just updated the pipeline!

Eddy, this same file:
https://github.com/microsoft/ApplicationInsights-dotnet/blob/3ac8c0b14466247263271a70ee5cbf3f54437145/NETCORE/.vsts/linux-build.yml#L62

This is another test that we renamed. Can you please fix this here too and then confirm in the build log that it runs tests. If you look at the log now you'll see that it's empty

BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Nuget.Tests/TelemetryChannel.Nuget.Tests.csproj
```diff
@@ -18,11 +18,6 @@
     <DebugSymbols>true</DebugSymbols> 
   </PropertyGroup>
 
-  <PropertyGroup Condition="$(OS) != 'Windows_NT'">
-    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
-    <TargetFrameworks>netstandard1.3</TargetFrameworks>
-  </PropertyGroup>
-  
   <ItemGroup Condition="$(OS) == 'Windows_NT'">
```
TimothyMothra, line 21: I agree with this change, but it's causing the build to fail.
For now, I think you could revert it and just add a comment inside the PropertyGroup:
`<!-- This can be removed. Requires update to .yml -->`

You can create a github issue to investigate a fix later.

eddynaka:
> Eddy, this same file:
> https://github.com/microsoft/ApplicationInsights-dotnet/blob/3ac8c0b14466247263271a70ee5cbf3f54437145/NETCORE/.vsts/linux-build.yml#L62
> 
> This is another test that we renamed. Can you please fix this here too and then confirm in the build log that it runs tests. If you look at the log now you'll see that it's empty

Sure, I can change that as well. For the part of 20, we was running .netcoreapp2.x, right now, our two integrationtests (normal and web one) targets netcoreapp3.1. Should i:
- target 2.1 as well
- remove that part

What about integrationTests.Web, should I add here as well?

thanks for reviewing

BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Nuget.Tests/TelemetryChannel.Nuget.Tests.csproj
eddynaka: ok, done: https://github.com/microsoft/ApplicationInsights-dotnet/issues/1847

Issue #1792, opened 2020-04-15
Telemetry loss on network/server errors in non-windows environment


rajkumar-rangaraj:
## Issue

`ServerTelemetryChannel` channel retries sending telemetry if transient errors occur. This channel uses local disk storage to keep items on disk during network outages or high telemetry volumes. In Windows, either `%LOCALAPPDATA%` or `%TEMP%` is used if no custom path is specified explicitly. In environments other than Windows, custom folder location should be provided or telemetry won't be stored to local disk. 

Below trace information is logged to Application Insights, if storage location is not specified in non-windows environments like Linux or MacOS
```
AI: Local storage access has resulted in an error (User: ) (CustomFolder: ). If you want Application Insights SDK to store telemetry locally on disk in case of transient network issues please give the process access to %LOCALAPPDATA% or %TEMP% folder. If application is running in non-windows platform, create StorageFolder yourself, and set ServerTelemetryChannel.StorageFolder to the custom folder name. After you gave access to the folder you need to restart the process. Currently monitoring will continue but if telemetry cannot be sent it will be dropped. Error message:
```
In Windows, SDK creates a storage folder in `%LOCALAPPDATA%` or `%TEMP%` and restricts access to `Current User` and `Administrators`. Restricting permissions in non-windows environment won’t work as there are no .NET APIs available to change permissions in Linux or MacOS. SDK does not add permission when custom folder name is used, hence providing customer folder name is the only solution to use StorageFolder in Linux or MacOS environment. Providing custom storage location requires code change and not a viable option for production applications.

Related Issue: #1067 
Microsoft Doc: [StorageFolder - Linux/MacOS](https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core#if-i-run-my-application-in-linux-are-all-features-supported)

## Proposal

Modify ` ServerTelemetryChannel` to read `StorageFolder` from an environment variable named `APPLICATIONINSIGHTS_SERVERCHANNEL_STORAGEFOLDER`. When this specific environment variable is present, consider the case as same of custom folder and avoid setting permission on it.

reyang:
Some principles I could think of:
1. We don't want to use temp folders (`%TEMP%` or `/tmp`) in this scenario, temp folders are designed to store files that can be deleted by the OS during reboot or cleanup.
[...]

PR #1860, opened 2020-05-06
ServerTelemetryChannel stores telemetry data in default folder during transient errors in non-windows environments]

rajkumar-rangaraj:
Fix Issue #1792 

### Proposal
For the application running in Linux or MacOS, `ServerTelemetryChannel` doesn’t automatically create a local storage folder to keep telemetry temporarily if there are network issues. Channel should be configured to use local folder to avoid telemetry data loss.

To fix this issue in non-windows environment, following folder locations have been identified to create a storage folder automatically to store telemetry data.

- `${TMPDIR}`, if `${TMPDIR}` environment variable is set; otherwise,
- `/var/tmp` if that exists; otherwise,
- `/tmp` if that exists; otherwise no storage folder is used.

In Windows, SDK creates a storage folder in %LOCALAPPDATA% or %TEMP% and restricts access to Current User and Administrators. 

 **But in Non-Windows environment, we don’t manage permissions for storage folder. If your application logs confidential information that cannot be stored in above mentioned location, we strongly recommend using custom folder and restring permission to process identity of an application.** 

### Checklist

### Notes for authors:

### Notes for reviewers:


BASE/src/ServerTelemetryChannel/Managed/Shared/Implementation/NonWindowsIdentityProvider.cs
```diff
@@ -18,15 +18,25 @@ public NonWindowsIdentityProvider(IDictionary environment)
...
             {
                 return this.environment["USER"].ToString();
-            }            
-            else
+            }
+
+            if (this.environment["USERNAME"] != null)
             {
-                return string.Empty;
+                return this.environment["USERNAME"].ToString();
```
cijothomas, line 36: Can we use Environment.UserName all the time and not check for these environment variables ourselves?

cijothomas: Environment.UserName should be avauilable in both net20 and net45

BASE/src/ServerTelemetryChannel/Implementation/ApplicationFolderProvider.cs
```diff
@@ -56,22 +60,47 @@ public IPlatformFolder GetApplicationFolder()
             var errors = new List<string>(this.environment.Count + 1);
 
             var result = this.CreateAndValidateApplicationFolder(this.customFolderName, createSubFolder: false, errors: errors);
-            
-            if (result == null)
+
+            // In the constructor, for windows environment we set the value of ApplySecurityToDirectory to SetSecurityPermissionsToAdminAndCurrentUserWindows.
```
cijothomas, line 64: we can do better here by leveraging Environment.OSVersion to detect the OS. THe current code does thing based on exception being thrown or not - as that was the only way in .netstandard1.3 - Can you refactor this to use Environment.OSVersion to make this decision?

BASE/src/ServerTelemetryChannel/Implementation/ApplicationFolderProvider.cs
```diff
@@ -136,6 +162,22 @@ private static string GetSHA256Hash(string input)
             return hashString.ToString();
         }
 
+        private static bool IsWindowsOperatingSystem()
```
cijothomas, line 165: if you make this internal, then the test code can use this, and not duplicate.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 - [Read all properties of ApplicationInsightsServiceOptions from IConfiguration](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1882)
 - [End support for NetStandard 1.x, Add support for NetStandard 2.0](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1160)
 - [Add support for SourceLink.Github to all SDKs.](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1760)
-
+- [Fix: ServerTelemetryChannel stores telemetry data in default folder during transient errors in all environments](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1792)
```
cijothomas, line 7: Lets rewrite this to say something like:
ServerTelemetryChannel by default uses local disk storage in non Windows, to store telemetry during transient errors.

PR #1886, opened 2020-06-02
NetCore remove TC.Active and fix NetCore Heartbeat

TimothyMothra:
#1953, #1954

NetCore had a dependency on TelemetryConfiguration.Active to support DiagnosticTelemetryModule and Heartbeats.
This change removes that dependency and introduces a feature flag to support TelemetryConfiguration.Active.

## Changes
- remove TelemetryConfiguration.Active
- remove TelemetryModules.Instance
- Refactoring Heartbeats for NetCore scenarios previously dependent on TC.Active

### Checklist

### Notes for authors:

### Notes for reviewers:


cijothomas:
Lets try to do this in steps 
1st remove .Active (with a config option to enable it for some customers who use it)
Add all modules including DiagModule via DI.
Fix heartbeat modules like AppServiceMetadata, AzureMetadata etc.

TimothyMothra:
@cijothomas i don't think that will work because I think these things are too tightly coupled. Please respond if I'm misunderstanding...

- First, supporting both .Active and not would be some additional work that I don't know if it's worth the effort.
- Second, I can't support the "not .Active" config without first fixing the DiagModule and Heartbeat Modules.
- Lastly, after I've fixed all the modules, I don't think there's any value in continuing to support .Active.

cijothomas:
> @cijothomas i don't think that will work because I think these things are too tightly coupled. Please respond if I'm misunderstanding...
> 
> * First, supporting both .Active and not would be some additional work that I don't know if it's worth the effort.
> * Second, I can't support the "not .Active" config without first fixing the DiagModule and Heartbeat Modules.
> * Lastly, after I've fixed all the modules, I don't think there's any value in continuing to support .Active.

If there are users somehow tied to use .Active, then we want to have a way to support them. For majority of cases, we can ask to switch to DI model, but there are cases like some logging adapters for which there is no better alternate than TC.Active.

TimothyMothra:
Hey @cijothomas, this test is failing at line 154:
I'm debugging it but I'm a little lost.
The `server.Listener` did not catch the `TrackTrace` from line 152.

Can you think of anything that I should investigate?

I confirmed that I can run this test locally on the `develop` branch, but I do get the same failure by replacing `TC.Active` with `TC.CreateDefault()`.

https://github.com/microsoft/ApplicationInsights-dotnet/blob/d2e6e7b0ed975485a483f2b67f9be7b02186f9e5/NETCORE/test/FunctionalTests.WebApi.Tests/FunctionalTest/MultipleWebHostsTests.cs#L124-L161

NETCORE/src/Shared/Implementation/TelemetryConfigurationOptionsSetup.cs
```diff
@@ -149,6 +153,15 @@ public void Configure(TelemetryConfiguration configuration)
                     // So this approach of adding all modules to DI, but selectively
                     // disable those modules which user has disabled is chosen.
 
+                    if (module is DiagnosticsTelemetryModule)
+                    {
+                        if (!this.applicationInsightsServiceOptions.EnableDiagnosticsTelemetryModule)
+                        {
+                            DisposeIfDisposable(module);
```
cijothomas, line 160: The heartbeatmanager is extraced and enabled independent of this, which likely means they operate on disposed heartbeatmanager, when one does EnableDiagnosticsTelemetryModule=false.

WEB/Src/WindowsServer/WindowsServer/Implementation/HeartbeatPropertyManagerProvider.cs
```diff
@@ -14,7 +14,7 @@ public static IHeartbeatPropertyManager GetHeartbeatPropertyManager()
             {
                 try
                 {
-                    foreach (var module in telemetryModules.Modules)
```
cijothomas, line 17: qn - why do we still access .Modules here?

.publicApi/Microsoft.ApplicationInsights.AspNetCore.dll/net451/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,4 @@
+Microsoft.ApplicationInsights.AspNetCore.Extensions.ApplicationInsightsServiceOptions.EnableDiagnosticsTelemetryModule.get -> bool
+Microsoft.ApplicationInsights.AspNetCore.Extensions.ApplicationInsightsServiceOptions.EnableDiagnosticsTelemetryModule.set -> void
+Microsoft.ApplicationInsights.AspNetCore.Extensions.ApplicationInsightsServiceOptions.EnableTelemetryConfigurationActiveBackwardsCompatibility.get -> bool
```
cijothomas, line 3: Consider naming this to something like `ConfigureTelemetryConfigurationActive`. There is no need of 'backwardcompatibility' in field name in my opinion.

NETCORE/src/Shared/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -316,6 +316,13 @@ private static void AddCommonInitializers(IServiceCollection services)
 
         private static void AddCommonTelemetryModules(IServiceCollection services)
         {
+            // Previously users were encouraged to manually add the DiagnosticsTelemetryModule.
+            // They could have added this either as an INSTANCE or as a TYPE. 
+            if (!services.Any(o => typeof(IHeartbeatPropertyManager).IsAssignableFrom(o.ImplementationType ?? o.ImplementationInstance.GetType())))
```
cijothomas, line 321: Logically, adding `DiagnosticsTelemetryModule` should check if `DiagnosticsTelemetryModule` is already present or not. Why checking IHeartbeatPropertyManager?

Please have a comment explaining this.

NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
```diff
@@ -116,11 +116,25 @@ public class ApplicationInsightsServiceOptions
         public bool EnableAuthenticationTrackingJavaScript { get; set; } = false;
 #endif
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the <see cref="Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.DiagnosticsTelemetryModule"/> should be enabled.
+        /// </summary>
+        public bool EnableDiagnosticsTelemetryModule { get; set; } = true;
```
cijothomas, line 122: there is a `EnableHeartbeat` field in this class. What would happen to it. Do we allow heartbeat to be turned on, while turning off diagnostictelemetrymodule? or diagnosticmodule must be turned on as a pre-req for heartbeat?

NETCORE/src/Shared/Implementation/TelemetryConfigurationOptionsSetup.cs
```diff
@@ -194,23 +207,31 @@ public void Configure(TelemetryConfiguration configuration)
...
+                    if (module is AppServicesHeartbeatTelemetryModule appServicesHeartbeatTelemetryModule)
                     {
                         if (!this.applicationInsightsServiceOptions.EnableAppServicesHeartbeatTelemetryModule)
                         {
                             DisposeIfDisposable(module);
                             continue;
                         }
+                        else
+                        {
+                            HeartbeatHelper.SetHeartbeatPropertyManager(appServicesHeartbeatTelemetryModule, heartbeatPropertyManager);
```
cijothomas, line 219: I'd use reflection only if every other option is exhausted. Did we try alternate options instead of reflection hack?
The Asp.Net Core/Worker knows about heartbeat, appservice module etc - so can we modify the appservice module to take the heartmanager via constructor?

NETCORE/src/Shared/Internals/ApplicationNameProvider.cs
```diff
@@ -25,7 +25,17 @@ private static string GetApplicationName()
         {
             try
             {
-                return Assembly.GetEntryAssembly().GetName().Name;
+                var assembly = Assembly.GetEntryAssembly();
```
cijothomas, line 28: Is this related to heartbeat?

TimothyMothra: no, not related. This throws NullRefExceptions and was making it difficult to debug.

NETCORE/src/Shared/Implementation/TelemetryConfigurationOptionsSetup.cs
TimothyMothra: I didn't investigate that because of a few unknowns.
- DI can have conflicts when adding overload constructors. 
- We're registering DiagnosticTelemetryModule as ITelemetry, but would need to use IHeartbeatPropertyManager in a constructor.

WEB/Src/WindowsServer/WindowsServer/Implementation/HeartbeatPropertyManagerProvider.cs
TimothyMothra: this is a helper class for net full scenarios. 
looks like I changed the whitespace here. I'll remove it.

NETCORE/src/Shared/Extensions/ApplicationInsightsExtensions.cs
TimothyMothra: I fixed this and moved the logic to an extension method (with tests) to clean it up.

.publicApi/Microsoft.ApplicationInsights.AspNetCore.dll/net451/PublicAPI.Unshipped.txt
TimothyMothra: I would prefer if we don't use "Configure" twice.
How do you feel about "EnableTelemetryConfigurationActive" or "EnableLegacyConfiguration"?

NETCORE/src/Shared/Implementation/TelemetryConfigurationOptionsSetup.cs
TimothyMothra: So I looked into this and I think we can make it work, but I want to review this with you before I make some major changes.

I'll have to add a new constructor to both AzureInstanceMetadataTelemetryModule and AppServiceHeartbeatTelemetryModule.
```
        public AzureInstanceMetadataTelemetryModule(IHeartbeatPropertyManager heartbeatPropertyManager)
        {
            this.HeartbeatPropertyManager = heartbeatPropertyManager;
        }
```

but to make this work, we HAVE to register DiagnosticTelemetryModule with the other interface.
```
        services.AddSingleton<IHeartbeatPropertyManager, DiagnosticsTelemetryModule>();
```

This is SIGNIFICANTLY cleaner than the reflection approach.
But I don't know how this will affect customers who register an instance of the Module.
We've been instructing users to do this:
```
        services.AddSingleton<ITelemetryModule>(new DiagnosticsTelemetryModule());
```

I don't know how to make this work with that, it'll take some extra testing.
What do you think?

TimothyMothra: This article suggests some workarounds.
https://andrewlock.net/how-to-register-a-service-with-multiple-interfaces-for-in-asp-net-core-di/

I'll test these tomorrow
```
    services.AddSingleton<Foo>(); // We must explicitly register Foo
    services.AddSingleton<IFoo>(x => x.GetRequiredService<Foo>()); // Forward requests to Foo
    services.AddSingleton<IBar>(x => x.GetRequiredService<Foo>()); // Forward requests to Foo
```

.publicApi/Microsoft.ApplicationInsights.AspNetCore.dll/net451/PublicAPI.Unshipped.txt
cijothomas: Not sure if "legacy" is correct. TelemetryConfiguration itself is not legacy! only the usage of .Active is.

EnableTelemetryConfigurationActive - this is misleading. Even with this changes, TC.Active is still available. Its just that it wont be configured correctly.

How does `EnableActiveTelemetryConfigurationSetup` sound?

.publicApi/Microsoft.ApplicationInsights.AspNetCore.dll/netstandard1.6/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,2 @@
+Microsoft.ApplicationInsights.AspNetCore.Extensions.ApplicationInsightsServiceOptions.EnableDiagnosticsTelemetryModule.get -> bool
```
cijothomas, line 1: we removed net1.6 target right? so this file wont be needed anymore.

.publicApi/Microsoft.ApplicationInsights.WorkerService.dll/netstandard2.0/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,4 @@
+Microsoft.ApplicationInsights.WorkerService.ApplicationInsightsServiceOptions.EnableDiagnosticsTelemetryModule.get -> bool
+Microsoft.ApplicationInsights.WorkerService.ApplicationInsightsServiceOptions.EnableDiagnosticsTelemetryModule.set -> void
+Microsoft.ApplicationInsights.WorkerService.ApplicationInsightsServiceOptions.EnableTelemetryConfigurationActiveBackwardsCompatibility.get -> bool
```
cijothomas, line 3: Shouldn't need this as WorkerService never populated Active, and no need to introduce that behavior heere.

NETCORE/src/Shared/Implementation/TelemetryConfigurationOptionsSetup.cs
TimothyMothra: solved

NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
TimothyMothra: I think DiagnosticModule should still be allowed to turn on when Heartbeat is turned off.

cijothomas: I think DiagnosticModule should still be allowed to turn on when Heartbeat is turned off. - this is correct.

My question was other way around - If we disable DiagnosticModule, should HeartBeat be still allowed?

TimothyMothra: Gotcha. I took a look at the implementation and there's not a straight forward way to Disable the internal EventListener.

https://github.com/microsoft/ApplicationInsights-dotnet/blob/0fd652cc2bdc98ce54bda0b293c18af194a1b7a0/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModule.cs#L34

I could introduce a new boolean `IsDiagnosticsEnable` and then conditionally set the EventListener to NULL. 
I think this is doable, but I don't know what other side effects this may cause. 
I would ask that we consider this an enhancement and postpone this specific change.

What do you think?

cijothomas: why not just dispose it if not enabled?

TimothyMothra: Should I change `options.EnableDiagnosticsTelemetryModule` to `options.EnableDiagnostics`?
I would also introduce a new property `DiagnosticsTelemetryModule.IsDiagnosticsEnabled`.

This would match the implementation of `options.EnableHeartbeat` and `DiagnosticsTelemetryModule.IsHeartbeatEnabled`.

Or would you expect that users could also remove the `DiagnosticsTelemetryModule` completely?

NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
```diff
@@ -27,14 +31,24 @@ public class ApplicationInsightsServiceOptions
         /// <summary>
         /// Gets or sets a value indicating whether AppServicesHeartbeatTelemetryModule should be enabled.
         /// Defaults to <value>true</value>.
+        /// IMPORTANT: This setting will be disabled if <see cref="EnableDiagnosticsTelemetryModule"/> is set to false.
         /// </summary>
-        public bool EnableAppServicesHeartbeatTelemetryModule { get; set; } = true;
+        public bool EnableAppServicesHeartbeatTelemetryModule
+        {
+            get => this.enableAppServicesHeartbeatTelemetryModule && this.EnableDiagnosticsTelemetryModule;
```
cijothomas, line 38: Don't think we need to couple this setting with other. i.e
If a customer disables DiagnosticModule, hearbeats won't flow anyway. But `EnableAppServicesHeartbeatTelemetryModule` is still true.

WEB/Src/WindowsServer/WindowsServer/AzureInstanceMetadataTelemetryModule.cs
```diff
@@ -15,17 +15,22 @@ public sealed class AzureInstanceMetadataTelemetryModule : ITelemetryModule
...
 
         /// <summary>
         /// Gets or sets an instance of IHeartbeatPropertyManager. 
         /// </summary>
         /// <remarks>
         /// This is expected to be an instance of <see cref="DiagnosticsTelemetryModule"/>.
-        /// Note that tests can also override the heartbeat manager.
         /// </remarks>
-        internal IHeartbeatPropertyManager HeartbeatPropertyManager
+        public IHeartbeatPropertyManager HeartbeatPropertyManager
```
cijothomas, line 33: what are your thoughts on making this not Editorbrowseable? Its public, but we dont expect customers to every change it themselves right?

NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
cijothomas: If one user adds 
```
service.Add(DiagnosticModule)
service.ConfigureTelemetryModule(diagmodule.Enableheartbeat = false)
```

The above also results in no heartbeats. I think there is no need to link the Setter/Getter fields. They can be independent. The comment should still say the net effect which is - if diagmodule is disabled, then heartbeats are also not sent, including appservice/azure specific heartbeats.

WEB/Src/WindowsServer/WindowsServer/AzureInstanceMetadataTelemetryModule.cs
TimothyMothra: I think that's an excellent idea.

NETCORE/src/Shared/Implementation/TelemetryConfigurationOptionsSetup.cs
```diff
@@ -255,7 +255,12 @@ public void Configure(TelemetryConfiguration configuration)
                     }
                     catch (Exception ex)
                     {
-                        throw new Exception($"Failed to initialize module '{module.GetType()}'", ex);
+                        var tempEx = new Exception($"Failed to initialize module '{module.GetType()}'", ex);
+#if AI_ASPNETCORE_WEB
+                        AspNetCoreEventSource.Instance.TelemetryConfigurationSetupFailure(tempEx.ToInvariantString());
```
cijothomas, line 260: let add a new method `TelemetryModuleInitialziationSetupFailure` with name of module as an argument. That way we can easily isolate issues if one module goes rogue.

cijothomas: An error has occurred while initializing {modulename}. Functionalities offered by this module will be affected. -- something like this

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Extensibility/Implementation/Tracing/AspNetCoreEventSource.cs
```diff
@@ -250,6 +250,19 @@ public void LogAzureAppServiceRoleNameFromHostNameHeaderInitializerWarning(strin
             this.WriteEvent(24, exception, this.applicationNameProvider.Name);
         }
 
+        /// <summary>
+        /// Logs an event when TelemetryConfiguration configure has failed.
```
cijothomas, line 254: not: wrong comment.

NETCORE/src/Microsoft.ApplicationInsights.WorkerService/Implementation/Tracing/WorkerServiceEventSource.cs
```diff
@@ -80,6 +80,19 @@ public void TelemetryConfigurationSetupFailure(string errorMessage, string appDo
             this.WriteEvent(5, errorMessage, this.applicationNameProvider.Name);
         }
 
+        /// <summary>
+        /// Logs an event when TelemetryConfiguration configure has failed.
```
cijothomas, line 84: not: wrong comment.

NETCORE/src/Microsoft.ApplicationInsights.WorkerService/Implementation/Tracing/WorkerServiceEventSource.cs
```diff
@@ -80,6 +80,19 @@ public void TelemetryConfigurationSetupFailure(string errorMessage, string appDo
...
+        /// </summary>
+        [Event(
+            6,
+            Keywords = Keywords.Diagnostics,
+            Message = "An error has occurred while initializing a TelemetryModule: '{0}'. Error message: '{1}' ",
```
cijothomas, line 89: initiazliing the TelemetryModule:

PR #1887, opened 2020-06-02
Ingestion service data delivery status

rajkumar-rangaraj:
## Ingestion service delivery status
Users could track ingestion service response by subscribing to transmission event. Notification is sent to subscriber whenever channel makes an attempt to send telemetry to the Ingestion service. The notification event includes information about the response, and also the telemetry items which were part of the transmission. Subscriber could extract information from ingestion response. For example, following metrics could be extracted. 

- How much traffic was accepted/rejected by ingestion service?
- What instrumentation keys were associated with the traffic?

## Changes
- Event handler `TransmissionStatusEvent` in `Transmission` could be wired up through `ServerTelemetryChannel`. Subscriber will utilize a public property in `ServerTelemetryChannel` to attach an event

``` csharp
// Sample
class Program
    {
        static void Main(string[] args)
        {
            var config = TelemetryConfiguration.CreateDefault();

            ServerTelemetryChannel channel = new ServerTelemetryChannel();
            channel.Initialize(config);
            config.TelemetryChannel = channel;
            config.InstrumentationKey = "IKEY";
            channel.TransmissionStatusEvent += delegate (object sender, TransmissionStatusEventArgs transmissionStatusEventArgs)
            {
                var transmission = sender as Transmission;

                if (transmissionStatusEventArgs.Response.StatusCode != (int)HttpStatusCode.PartialContent)
                {
                    Console.WriteLine($"Status Code: {transmissionStatusEventArgs.Response.StatusCode}");
                }
                else
                {
                    BackendResponse backendResponse = GetBackendResponse(transmissionStatusEventArgs.Response.Content);

                    Console.WriteLine($"ItemsAccepted: {backendResponse.ItemsAccepted}");
                    Console.WriteLine($"ItemsReceived: {backendResponse.ItemsReceived}");
                    
                    foreach (var error in backendResponse.Errors)
                    {
                        Console.WriteLine($"Error Index: {error.Index}");
                        Console.WriteLine($"Error Message: {error.Message}");
                    }
                }

                var iKeyGroup = transmission.TelemetryItems.GroupBy(x => x.Context.InstrumentationKey);

                foreach (var telemetry in iKeyGroup)
                {
                    Console.WriteLine($"Ikey: {telemetry.Key}, Count: {telemetry.Count()}");
                }

            };

            var client = new TelemetryClient(config);
            SendTelemetryToApplicationInsights(client);
            Console.ReadLine();
        }

        private static void SendTelemetryToApplicationInsights(TelemetryClient client)
        {
            EventTelemetry eventTelemetry1 = new EventTelemetry("Event1");
            eventTelemetry1.Context.InstrumentationKey = "IKEY_1";
            client.TrackEvent(eventTelemetry1);

            EventTelemetry eventTelemetry2 = new EventTelemetry("Event2");
            eventTelemetry2.Context.InstrumentationKey = "IKEY_2";
            client.TrackEvent(eventTelemetry2);

            EventTelemetry eventTelemetry3 = new EventTelemetry("Event3");
            eventTelemetry3.Context.InstrumentationKey = "6c49c07c-e95c-48fe-8a7b-eff230955cc5";
            client.TrackEvent(eventTelemetry3);

            EventTelemetry eventTelemetry4 = new EventTelemetry("Event3");
            eventTelemetry4.Context.InstrumentationKey = "IKEY_2";
            client.TrackEvent(eventTelemetry4);

            EventTelemetry eventTelemetry5 = new EventTelemetry("Event5");
            eventTelemetry5.Context.InstrumentationKey = "IKEY_1";
            client.TrackEvent(eventTelemetry5);

            client.Flush();
        }

        [DataContract]
        private class BackendResponse
        {
            [DataMember(Name = "itemsReceived", IsRequired = true)]
            public int ItemsReceived { get; set; }

            [DataMember(Name = "itemsAccepted", IsRequired = true)]
            public int ItemsAccepted { get; set; }

            [DataMember(Name = "errors")]
            public Error[] Errors { get; set; }

            [DataContract]
            private class Error
            {
                [DataMember(Name = "index")]
                public int Index { get; set; }

                [DataMember(Name = "statusCode")]
                public int StatusCode { get; set; }

                [DataMember(Name = "message")]
                public string Message { get; set; }
            }
        }

        private static BackendResponse GetBackendResponse(string responseContent)
        {
            BackendResponse backendResponse = null;
            DataContractJsonSerializer Serializer = new DataContractJsonSerializer(typeof(BackendResponse));

            try
            {
                if (!string.IsNullOrEmpty(responseContent))
                {
                    using (MemoryStream ms = new MemoryStream(Encoding.Unicode.GetBytes(responseContent)))
                    {
                        backendResponse = Serializer.ReadObject(ms) as BackendResponse;
                    }
                }
            }
            catch
            {
                backendResponse = null;
            }

            return backendResponse;
        }
    }
```
For another example, please refer test method `TestTransmissionStatusEventWithEventsFromMultipleIKey` in `TransmissionTest.cs`

- Events Notification is initiated on `SendAsync` completion (once response is received from ingestion service) in `Transmission`. This passes `Sender` as `Microsoft.ApplicationInsights.Channel.Transmission`  and `TransmissionStatusEventArgs` with `Microsoft.ApplicationInsights.Extensibility.Implementation.HttpWebResponseWrapper` to subscriber
- Subscriber is responsible to handle event and extract information from `Transmission` and `HttpWebResponseWrapper`
- If there are no subscriber, event notification won’t be initiated

### Checklist


BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/TransmissionTest.cs
```diff
@@ -501,6 +501,31 @@ public async Task SendAsyncLogsIngestionReponseTimeAndStatusCode()
...
+
+            [TestMethod]
+            public async Task TestTransmissionStatusEventHandler()
+            {
+                var handler = new HandlerForFakeHttpClient
+                {
+                    InnerHandler = new HttpClientHandler(),
+                    OnSendAsync = (req, cancellationToken) =>
+                    {
+                        return Task.FromResult<HttpResponseMessage>(new HttpResponseMessage());
```
cijothomas, line 513: Additional test case: make this fail or throw exception or timeout.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/TransmissionTest.cs
```diff
@@ -501,6 +501,31 @@ public async Task SendAsyncLogsIngestionReponseTimeAndStatusCode()
...
+                    }
+                };
+
+                using (var fakeHttpClient = new HttpClient(handler))
+                {
+                    // Instantiate Transmission with the mock HttpClient                  
+                    Transmission transmission = new Transmission(testUri, new byte[] { 1, 2, 3, 4, 5 }, fakeHttpClient, string.Empty, string.Empty);
```
cijothomas, line 520: make a test with actual ITelemetry items being sent, each with separate ikey. Validate that the TransmissionStatusEvent receives the Transmission, and it can read the individual ikeys.

We probably need to give a wrapper to help with 206/partial success status codes. If a few items from batch is rejected, the TransmissionStatusEvent should be able to tell which ones were accepted/rejected. This requires knowledge about ingestion response format, which this SDK knows - but not the end user. So a helper/wrapper will help here.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/TransmissionTest.cs
rajkumar-rangaraj: Done.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/TransmissionTest.cs
rajkumar-rangaraj: TransmissionStatusEvent has response from ingestion endpoint and not the actual transmission. **Sender** parameter in the eventhandler method is the transmission object.

Added a new test case for this scenario. 
The sample in our todo list will cover this complete flow.

BASE/src/Microsoft.ApplicationInsights/Channel/TransmissionStatusEventArgs.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class TransmissionStatusEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TransmissionStatusEventArgs"/> class.
+        /// </summary>
+        /// <param name="response">Response from ingestion endpoint.</param>
+        public TransmissionStatusEventArgs(HttpResponseMessage response = null)
+        {
+            this.Response = response;
+        }
```
TimothyMothra, line 18: Why do you need this new class as a wrapper around the HttpResponseMessage?
This is noteworthy to me because this will be a lot of new allocations.

rajkumar-rangaraj: EventHandler requires<TEventArgs>. TransmissionStatusEventArgs class inherits EventArgs and contains HttpresponseMessage. I agree, there will be new allocation only if subscriber is listening to it. As you said this is a wrapper and hold only reference to HttpResponseMessage. When HttpResponseMessage is disposed in Transmission, Response property in this object also goes null.

In reality, this object is created and destroyed immediately as the event handler code runs synchronously.

BASE/src/ServerTelemetryChannel/ServerTelemetryChannel.cs
```diff
@@ -121,6 +121,22 @@ public string EndpointAddress
             set { this.TelemetrySerializer.EndpointAddress = new Uri(value); }
         }
 
+        /// <summary>
+        /// Gets or Sets the subscriber to an event in Transmission to track ingestion status metrics.
```
cijothomas, line 125: The comment is bit misleading - there is nothing about metrics here. Registered callback gets called with HttpResponse  - whether its used for metrics extraction or something else - this module doesn't care.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/TransmissionTest.cs
```diff
@@ -501,6 +504,224 @@ public async Task SendAsyncLogsIngestionReponseTimeAndStatusCode()
...
+            [TestMethod]
+            public async Task TestTransmissionStatusEventHandlerWithFailureTransmission()
+            {
+                // ARRANGE
+                var handler = new HandlerForFakeHttpClient
+                {
+                    InnerHandler = new HttpClientHandler(),
+                    OnSendAsync = (req, cancellationToken) =>
+                    {
+                        return Task.FromResult<HttpResponseMessage>(new HttpResponseMessage(HttpStatusCode.ServiceUnavailable));
```
cijothomas, line 546: both tests are validating scenario when httpclient actually returns a response. We need to test the scenario where HttpClient throws exception. For, eg: timeout, etc.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/TransmissionTest.cs
```diff
@@ -501,6 +504,224 @@ public async Task SendAsyncLogsIngestionReponseTimeAndStatusCode()
...
+                        Assert.AreEqual(HttpStatusCode.ServiceUnavailable, args.Response.StatusCode);
+                    };
+
+                    // ACT
+                    HttpWebResponseWrapper result = await transmission.SendAsync();
+                }
+            }
+
+            [TestMethod]
+            public async Task TestTransmissionStatusEventWithEventsFromMultipleIKey()
```
cijothomas, line 567: we need to modify the PR description with example usage. Its not clear how would one use this. In this unit test it looks like the `Transmission` is also part of the callback. But the public API doesn't make it appear so - it only exposes `HttpResponseMessage` in the callback.

ivshch:
I was looking through the PR code and got a question:
what if the request is failed with the [HttpRequestException](https://github.com/microsoft/ApplicationInsights-dotnet/blob/c3968af8e7660034df11ef7e874c447f16d7b826/BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs#L164) (in case of general network failure), it seems that the delegate is not invoked?
This probably suits the initial idea of gathering metric of successful requests, but if I'm correct it makes it unusable in our scenario #1974 - detecting the connection loss.

Would the metric be accurate if it does not include the failures?
Could it improve your solution if the delegate allows collecting failures also?

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/TransmissionTest.cs
rajkumar-rangaraj: Event handler has sender and event args. Sender object is a `Transmission` and `HttpResponseMessage` is passed as event args.

rajkumar-rangaraj:
> Left comments.
> We need sample usage code in the PR description to help users onboard to this feature as well.

Plan was to add complete working solution in samples folder. I mentioned it n `TODO` section of PR description. Should we provide complete sample in this PR description itself?

BASE/src/ServerTelemetryChannel/ServerTelemetryChannel.cs
rajkumar-rangaraj: Fixed in both `ServerTelemetryChannel` and `TelemetrySerializer`.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/TransmissionTest.cs
rajkumar-rangaraj: Set timeout to expire immediately and used cancellation token to test it.

ivshch: May I add my 2 cents:
throwing OperationCanceledException instead of returning the HttpResponseMessage from OnSendAsync will also pass the test, but the Assert.AreEqual would not be hit in that case.
Placing Assert outside the delegate would help to solve it.

ivshch: My bad, I didn't see the recent changes. The test will not pass for OperationCanceledException with the new implementation.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -209,13 +214,23 @@ public virtual async Task<HttpWebResponseWrapper> SendAsync()
...
                     }
 
+                    try
+                    {
+                        // Initiates event notification to subscriber with Transmission and TransmissionStatusEventArgs.
+                        this.TransmissionStatusEvent?.Invoke(this, new TransmissionStatusEventArgs(wrapper));
+                    }
+                    catch
+                    {
+                        // Swallow any exception here as this code is for tracing purposes only and should never throw.
```
cijothomas, line 231: we should log here and mark the event UserActionable. Because if this fails, its mostly user error and they should know it. SIlently failing is risky here.

BASE/src/Microsoft.ApplicationInsights/Channel/TransmissionStatusEventArgs.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// <summary>
+    /// Event argument to track response from ingestion endpoint.
+    /// </summary>
+    public class TransmissionStatusEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TransmissionStatusEventArgs"/> class.
+        /// </summary>
+        /// <param name="response">Response from ingestion endpoint.</param>
+        public TransmissionStatusEventArgs(HttpWebResponseWrapper response = null)
```
cijothomas, line 15: why default to null for response?

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -71,6 +71,11 @@ protected internal Transmission()
         {
         }
 
+        /// <summary>
+        /// Gets or Sets an event notification to track ingestion endpoint response.
+        /// </summary>
+        public EventHandler<TransmissionStatusEventArgs> TransmissionStatusEvent { get; set; }
```
cijothomas, line 77: Avoid public api change. Accept it inside ctor() only.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -209,12 +214,24 @@ public virtual async Task<HttpWebResponseWrapper> SendAsync()
...
                         {
                             StatusCode = (int)HttpStatusCode.RequestTimeout,
                         };
                     }
+                    finally
+                    {
+                        try
+                        {
+                            // Initiates event notification to subscriber with Transmission and TransmissionStatusEventArgs.
+                            this.TransmissionStatusEvent?.Invoke(this, new TransmissionStatusEventArgs(wrapper ?? new HttpWebResponseWrapper() { StatusCode = 0 }));
```
cijothomas, line 228: we should use the same status code we use for logging.. I think we use 999 or something?

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -209,12 +214,24 @@ public virtual async Task<HttpWebResponseWrapper> SendAsync()
...
+                        }
+                        catch (Exception ex)
+                        {
+                            CoreEventSource.Log.TransmissionStatusEventError(ex.ToInvariantString());
```
cijothomas, line 232: lets do the .ToInvariantString() only if listener is active. (non event in EventSource approach)

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/CoreEventSource.cs
```diff
@@ -652,6 +652,9 @@ public void IngestionResponseTimeEventCounter(float responseDurationInMs)
         [Event(70, Message = "Updating Exception has failed. Error: {0}", Level = EventLevel.Error)]
         public void UpdateDataFailed(string error, string appDomainName = "Incorrect") => this.WriteEvent(70, error, this.nameProvider.Name);
 
+        [Event(71, Keywords = Keywords.UserActionable, Message = "TransmissionStatusEvent has failed. Error: {0}", Level = EventLevel.Error)]
```
cijothomas, line 655: Need to tell the impact of this..
Be explicit - this will not affect actual transmission, only the notification is failing.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/TransmissionTest.cs
```diff
@@ -501,6 +504,302 @@ public async Task SendAsyncLogsIngestionReponseTimeAndStatusCode()
...
+
+                using (var fakeHttpClient = new HttpClient(handler))
+                {
+                    // Instantiate Transmission with the mock HttpClient                  
+                    Transmission transmission = new Transmission(testUri, new byte[] { 1, 2, 3, 4, 5 }, fakeHttpClient, string.Empty, string.Empty);
+
+                    // VALIDATE
+                    transmission.TransmissionStatusEvent += delegate (object sender, TransmissionStatusEventArgs args)
+                    {
+                        Assert.AreEqual((int)HttpStatusCode.OK, args.Response.StatusCode);
```
cijothomas, line 529: validate sender.
type should be transmission?

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
rajkumar-rangaraj: Yes. It is [999](https://github.com/microsoft/ApplicationInsights-dotnet/blob/7633ae849edc826a8547745b6bf9f3174715d4bd/BASE/src/ServerTelemetryChannel/Implementation/ResponseStatusCodes.cs#L8).

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
rajkumar-rangaraj: We don't have this pattern in `CoreEventSource`, Should we introduce it?

cijothomas (approved):

reyang:
@rajkumar-rangaraj if customers are using the in-memory channel, is there a way to subscribe to the transmission status update?

> Event handler TransmissionStatusEvent in Transmission is wired up through ServerTelemetryChannel, subscriber will utilize a public property in ServerTelemetryChannel to attach the event

rajkumar-rangaraj:
> @rajkumar-rangaraj if customers are using the in-memory channel, is there a way to subscribe to the transmission status update?
> 
> > Event handler TransmissionStatusEvent in Transmission is wired up through ServerTelemetryChannel, subscriber will utilize a public property in ServerTelemetryChannel to attach the event

Changes will work only for `ServerTelemetryChannel`. As there are no retry logic did not expose public property from `InMemoryChannel`. Changes would be simple to enable for `InMemoryChannel`, we just need to expose property in channel and serializer.

reyang (approved):
LGTM.

PR #1898, opened 2020-06-05
Tilee/simplify serviceoptions

TimothyMothra:
Currently when adding a property to the ApplicationInsightsServiceOptions, a develop has to remember to make a change in three places: Property declaration, Default constructor, and Copy method.
This change removes one of these three.

## Changes
- Remove default constructor with default values.
- Move all default values to be inline with the Property declaration.

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
+++ b/NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
@@ -13,14 +13,6 @@
     public class ApplicationInsightsServiceOptions
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="ApplicationInsightsServiceOptions" /> class.
-        /// Application Insights service options that controls the default behavior of application insights features.
-        /// </summary>
-        public ApplicationInsightsServiceOptions()
-        {
-        }
-
-        /// <summary>
         /// Gets or sets a value indicating whether QuickPulseTelemetryModule and QuickPulseTelemetryProcessor are registered with the configuration.
         /// Setting EnableQuickPulseMetricStream to <value>false</value>, will disable the default quick pulse metric stream. Defaults to <value>true</value>.
         /// </summary>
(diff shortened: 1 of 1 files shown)


NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
```diff
@@ -97,7 +69,7 @@ public ApplicationInsightsServiceOptions()
         /// <summary>
         /// Gets or sets the application version reported with telemetries.
         /// </summary>
-        public string ApplicationVersion { get; set; }
+        public string ApplicationVersion { get; set; } = Assembly.GetEntryAssembly()?.GetName().Version.ToString();
```
cijothomas, line 72: Does this mean `Assembly.GetEntryAssembly()?.GetName().Version.ToString()` gets executed everytime one accesses ApplicationVersion? Previously this was executed once inside ctor only.

NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
```diff
@@ -112,42 +84,42 @@ public ApplicationInsightsServiceOptions()
...
         /// </summary>
-        public bool AddAutoCollectedMetricExtractor { get; set; }
+        public bool AddAutoCollectedMetricExtractor { get; set; } = true;
 
 #if AI_ASPNETCORE_WEB
         /// <summary>
         /// Gets <see cref="RequestCollectionOptions"/> that allow to manage <see cref="RequestTrackingTelemetryModule"/>.
         /// </summary>
-        public RequestCollectionOptions RequestCollectionOptions { get; }
+        public RequestCollectionOptions RequestCollectionOptions { get; } = new RequestCollectionOptions();
```
cijothomas, line 104: same comment as applicationversion

NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs
TimothyMothra: Nope, this just defines the default values for these properties. This is only executed when a new instance is created. It will be the exact same behavior as the former default constructor.

TimothyMothra: > Auto-property initializers let you declare the initial value for an auto-property as part of the property declaration.
> https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-6#auto-property-initializers

PR #1911, opened 2020-06-12
Fix #581 (Changed TelemetryConfiguration property of TelemetryClient from internal to public.)

### LINKED ISSUE #581: Expose TelemetryClient.TelemetryConfiguration publically.
TelemetryClient brings together 3 concepts: Properties of items sent through it (Context), information about how and where to send the items (TelemetryConfiguration) and APIs to do the sending (TrackXxx(..) methods). Although both of the former aspects are required to create a client (TelemetryConfiguration.Active is used as default, but implicitly, that information is still encapsulated), only the Context is exposed outside of Core.

This is unfortunate: A consumer of a TelemetryClient may want to access information about how and where the client will deliver data. For example, metrics are aggregated across all clients within the same configuration, but separately between pipeline configurations. So, aggregation logic needs to be able to ask a client, what its telemetry configuration instance is.

The proposed change is to make the existing internal TelemetryClient.TelemetryConfiguration property public.

Expecho:
Fix Issue #581 .

## Changes
Changed TelemetryConfiguration property of TelemetryClient from internal to public.

### Checklist

### Notes for authors:

### Notes for reviewers:


BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -83,8 +83,10 @@ public string InstrumentationKey
 
         /// <summary>
         /// Gets the <see cref="TelemetryConfiguration"/> object associated with this telemetry client instance.
+        /// This property is an internal part of Application Insights infrastructure. Changes made to the configuration will not be applied in. Do not use.
         /// </summary>
-        internal TelemetryConfiguration TelemetryConfiguration
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public TelemetryConfiguration TelemetryConfiguration
```
cijothomas, line 89: I am not sure of the part - "Changes made to the configuration will not be applied in". If one modifies the config, I think it would get applied to all telemetry tracker afterwards.

Expecho: You are right, I misread [this comment](https://github.com/microsoft/ApplicationInsights-dotnet/issues/581#issuecomment-313280554). Should we instead warn that changes can affect other clients or is that not longer valid given the date of the original comment?

cijothomas: I'd expect it would affect other clients made from the same config.

btw - what is your use case for accessing Configuration from client?

cijothomas: Also yes to the question -  we need to add this warning. Deprecating .Active wouldn't impact this.

cijothomas: @TimothyMothra @rajkumar-rangaraj Do you see any issues making configuration public, but hidden to avoid misuse.?

rajkumar-rangaraj: I reviewed this PR earlier and feel it is hidden to avoid the misuse. I don't see any problem with making this public as we give control to TelemetryConfiguration in other places.

rajkumar-rangaraj:
@Expecho tests are failing and you please API file. Building the project in release mode on Visual Studio will help fix this issue.

.publicApi/Microsoft.ApplicationInsights.dll/net452/PublicAPI.Shipped.txt
```diff
@@ -880,6 +880,7 @@ Microsoft.ApplicationInsights.TelemetryClient.Initialize(Microsoft.ApplicationIn
 Microsoft.ApplicationInsights.TelemetryClient.InitializeInstrumentationKey(Microsoft.ApplicationInsights.Channel.ITelemetry telemetry) -> void
 Microsoft.ApplicationInsights.TelemetryClient.InstrumentationKey.get -> string
 Microsoft.ApplicationInsights.TelemetryClient.InstrumentationKey.set -> void
+Microsoft.ApplicationInsights.TelemetryClient.TelemetryConfiguration.get -> Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration
```
rajkumar-rangaraj, line 883: Please add this entry to `PublicAPI.UnShipped.txt`. This file needs an update to all target framework.

Expecho: Got it. Sorry for the inconvenience. For my understanding, is there any way to open those file from within Visual Studio? It seems to always point to the 452 version of the file in the solution explorer, even when changing the target framework.

TimothyMothra: Using Visual Studio, you can select to fix all occurances of an issue in the Project/Solution.
This will fix it in all frameworks.
Screenshots how to here: https://github.com/dotnet/roslyn-analyzers/issues/3322

rajkumar-rangaraj: @Expecho As I commented earlier, please move API definition from `PublicAPI.Shipped.txt` to `.publicApi/Microsoft.ApplicationInsights.dll/net452/PublicAPI.Unshipped.txt`. This line `Microsoft.ApplicationInsights.TelemetryClient.TelemetryConfiguration.get -> Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration` should move to `PublicAPI.Unshipped.txt` for all target framework.

Expecho: You're right sorry. And after reading https://www.natmarchand.fr/prevent-breaking-changes-apis-roslyn/ I do have a better understanding of the way these files work.

PR #1929, opened 2020-06-18
Cleaning AssemblyInfo.cs and *.csproj

eddynaka:
Fix Issue # .

## Changes
- removing extra tags in assemblyinfo.cs / *.csproj

### Checklist

### Notes for authors:

### Notes for reviewers:


eddynaka:
Questions:
- in some **test** projects we have `AssemblyInfo.cs` and other we don't. Should we exclude all or add for all? Example: `DiagnosticSourceListener.Tests` doesn't have and `EtwCollector.Tests` have.
- `<IsPackable>` is used in almost all test projects. Since we don't use the `Product.props` file, it's not generating the package. Should we remove as well?

TimothyMothra:
I think the `<IsPackable>false</IsPackable>` could be moved to the common Test.props file.

I'm not sure what to do with the AssemblyInfo.cs. I think it's not required for PackageReference csproj. We may need to make a case by case decision if they can be removed from the Test projects. 
Are they causing an issue or are you trying to cleanup inconsistencies?

eddynaka:
> I think the `<IsPackable>false</IsPackable>` could be moved to the common Test.props file.
> 
> I'm not sure what to do with the AssemblyInfo.cs. I think it's not required for PackageReference csproj. We may need to make a case by case decision if they can be removed from the Test projects.
> Are they causing an issue or are you trying to cleanup inconsistencies?

Ok, i will add the IsPackable to test.props
For the AssemblyInfo, just cleaning the inconsistencies.

TimothyMothra:
LGTM. Please click "Ready for review" when you're finished with these changes. :)

PR #1941, opened 2020-06-30
Improve bug template

reyang:


.github/ISSUE_TEMPLATE/dotnetcore_bug.md
```diff
@@ -0,0 +1,18 @@
+---
+name: Bug Report (.NET Core)
```
cijothomas, line 2: Distinguish between .NET Core and .NET Framework is not really actionable for us.
But its important to distinguish between ASP.NET and ASP.NET Core.

reyang: I could think the following benefits:
1. Indicating .NET / .NET Core in the issue template name would help to route JS/Python customers away and to the right place.
2. It doesn't make sense to ask people which Platform for .NET as Windows is the only one.
3. We can do auto-tagging for .NET and .NET Core, which helps us to see the trend of people moving from .NET to Core.
4. In this way we no longer need to ask whether it is ASP.NET or ASP.NET Core as it is indicated by the .NET/Core.

cijothomas: 1. I don't think we need to route JS/Python/Java customers with this template. It should be obvious this repo is not for them. We have never seen customers opening python issue here.

The only time customer misreport issue in this repo is when the artifacts produced by this repo is involved - good examples are Azure Functions, Azure Web Apps code-less attach - both uses the code/artifacts from this repo, but the support/question is NOT provided by this repo. The current template and this PR does not have anything which redirect customers directly to Azure functions for Functions issue - that was one of the key goals of updating this bug template.

2. Agree platform is always windows for .NET. By platform we don't necessarily mean OS. It could mean - are you running in a Azure Web App? Container? SF cluster etc. 

3. Don't think we need GH issues to get that telemetry.

4. As Asp.Net Core apps can be written targeting .NET Framework, we still need to ask this information. 

Knowing which nuget package produced from this repo customer is using - this is the most important information we need -(eg: Its not enough to know if an app is .NET Core or .NET Framework, we need to know if they use basesdk alone, or workerservice or something else)

reyang: > 2. Agree platform is always windows for .NET. By platform we don't necessarily mean OS. It could mean - are you running in a Azure Web App? Container? SF cluster etc.

We need to be specific, I've changed the wording, please see if it helps to bring clarity.

> 3. Don't think we need GH issues to get that telemetry.

That's okay, I've removed the second template.

> Knowing which nuget package produced from this repo customer is using - this is the most important information we need -(eg: Its not enough to know if an app is .NET Core or .NET Framework, we need to know if they use basesdk alone, or workerservice or something else)

I've made it explicit by asking customers to list the nuget packages, do you think it'll work?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,22 +1,19 @@
...
-* SDK version: 
-* .NET runtime version (.NET TargetFramework in the .csproj file):
-* Hosting Info (IIS/Azure WebApps/etc): 
-* Platform and OS version: 
+* List of [NuGet packages](https://github.com/microsoft/ApplicationInsights-dotnet#nuget-packages) and version that you are using:
+* Runtime version (e.g. net461, net48, netcoreapp2.1, netcoreapp3.1):
+* Application type (Console/ASP.NET/ASP.NET Core):
+* Hosting environment (e.g. Windows Server 2016 + IIS, Ubuntu 20.04 LTS self-hosted, Azure WebApp):
+  * If you're running the application in Azure Functions, get support [here](TBD).
+  * If you're using codeless attach, get support [here](TBD).
```
TimothyMothra, line 12: I like this change, but customers might not know the name "codeless attach". 
I think the UX calls this the ["site extension"](https://docs.microsoft.com/en-us/azure/azure-monitor/app/azure-web-apps?tabs=net#enable-agent-based-monitoring)

reyang: I need to get the right name and link from the owner.

cijothomas: https://docs.microsoft.com/en-us/azure/azure-monitor/app/codeless-overview Its here

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,22 +1,19 @@
...
-* SDK version: 
-* .NET runtime version (.NET TargetFramework in the .csproj file):
-* Hosting Info (IIS/Azure WebApps/etc): 
-* Platform and OS version: 
+* List of [NuGet packages](https://github.com/microsoft/ApplicationInsights-dotnet#nuget-packages) and version that you are using:
+* Runtime version (e.g. net461, net48, netcoreapp2.1, netcoreapp3.1):
+* Application type (Console/ASP.NET/ASP.NET Core):
+* Hosting environment (e.g. Windows Server 2016 + IIS, Ubuntu 20.04 LTS self-hosted, Azure WebApp):
+  * If you're running the application in Azure Functions, get support [here](TBD).
+  * If you're using [auto-instrumentation](https://docs.microsoft.com/en-us/azure/azure-monitor/app/codeless-overview), get support from [Azure Support Center](https://azure.microsoft.com/en-us/support/create-ticket/).
```
cijothomas, line 12: should we suggest this approach for any issue not related to SDKs?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,22 +1,19 @@
...
 labels: bug
 ---
------------------Bug Template--------------------------------------------------------------
 
-* SDK version: 
-* .NET runtime version (.NET TargetFramework in the .csproj file):
-* Hosting Info (IIS/Azure WebApps/etc): 
-* Platform and OS version: 
+* List of [NuGet packages](https://github.com/microsoft/ApplicationInsights-dotnet#nuget-packages) and version that you are using:
+* Runtime version (e.g. net461, net48, netcoreapp2.1, netcoreapp3.1):
```
cijothomas, line 8: Would it help to say where to find this info - its in the .csproj under <targetframework> ?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,22 +1,19 @@
...
+* Application type (Console/ASP.NET/ASP.NET Core):
+* Hosting environment (e.g. Windows Server 2016 + IIS, Ubuntu 20.04 LTS self-hosted, Azure WebApp):
+  * If you're running the application in Azure Functions, get support [here](TBD).
```
cijothomas, line 11: How about doing like this? https://github.com/dotnet/aspnetcore/issues/new/choose

Functions can be redirected directly to functions repo.

.github/ISSUE_TEMPLATE/bug_report.md
reyang: I think we agreed that most customers wouldn't come here for other issues?

.github/ISSUE_TEMPLATE/azure_functions_issue.md
```diff
@@ -0,0 +1,6 @@
+---
+name: Issue with Azure Functions
+about: Need help with Azure Functions
+---
+
+If you're running the application in Azure Functions, get support [here](TBD).
```
cijothomas, line 6: https://github.com/Azure/azure-webjobs-sdk
https://github.com/Azure/azure-functions-host
We can start with these.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,22 +1,17 @@
...
-* .NET runtime version (.NET TargetFramework in the .csproj file):
-* Hosting Info (IIS/Azure WebApps/etc): 
-* Platform and OS version: 
+* List of [NuGet packages](https://github.com/microsoft/ApplicationInsights-dotnet#nuget-packages) and version that you are using:
+* Runtime version (e.g. net461, net48, netcoreapp2.1, netcoreapp3.1, you can find this information from the `*.csproj` file):
+* Application type (Console/ASP.NET/ASP.NET Core):
+* Hosting environment (e.g. Windows Server 2016 + IIS, Ubuntu 20.04 LTS self-hosted, Azure WebApp):
 
 **Steps to reproduce.**
 Describe exactly how to reproduce the error. Include a code sample if applicable.
```
TimothyMothra, line 13: Could we suggest that people create a github repository with a sample project reproducing the error?
If there's a polite way to ask this, it would dramatically improve our ability to investigate reported issues.

reyang: That's too much ask IMHO.

PR #1945, opened 2020-07-01
Update nuget table and polish the README file

reyang:
1. Removed the nuget package(s) that are not supposed to be consumed directly by the customer.
2. Markdown lint.


TimothyMothra:
Can we return HostingStartup?
it's useful to use this page to verify packages have been successfully deployed.

cijothomas:
> Can we return HostingStartup?
> it's useful to use this page to verify packages have been successfully deployed.

I think we need to nuke the entire HostingStartup folder- based on my knowledge its not used and we don't support them. (Its an artifact from old Azure Web App extension world which has now switched to pre-installed extensions)

reyang:
> > Can we return HostingStartup?
> > it's useful to use this page to verify packages have been successfully deployed.
> 
> I think we need to nuke the entire HostingStartup folder- based on my knowledge its not used and we don't support them. (Its an artifact from old Azure Web App extension world which has now switched to pre-installed extensions)

If there is a clear scenario where we have end users downloading/installing this package explicitly, we probably should added it back. The download number is low though.

TimothyMothra:
> > > Can we return HostingStartup?
> > > it's useful to use this page to verify packages have been successfully deployed.
> > 
> > 
> > I think we need to nuke the entire HostingStartup folder- based on my knowledge its not used and we don't support them. (Its an artifact from old Azure Web App extension world which has now switched to pre-installed extensions)
> 
> If there is a clear scenario where we have end users downloading/installing this package explicitly, we probably should added it back. The download number is low though.

Can we stop shipping it now? Or maybe 2.15 can be the final version?

cijothomas:
> > > > Can we return HostingStartup?
> > > > it's useful to use this page to verify packages have been successfully deployed.
> > > 
> > > 
> > > I think we need to nuke the entire HostingStartup folder- based on my knowledge its not used and we don't support them. (Its an artifact from old Azure Web App extension world which has now switched to pre-installed extensions)
> > 
> > 
> > If there is a clear scenario where we have end users downloading/installing this package explicitly, we probably should added it back. The download number is low though.
> 
> Can we stop shipping it now? Or maybe 2.15 can be the final version?

Agree to ship it for 2.15 and then remove it from code completely.

PR #1958, opened 2020-07-08
Retiring Release Schedule from Readme

TimothyMothra:
Our release cadence has slowed down as we begin to shift towards OpenTelemetry.
I'm removing the Release Schedule.
We will continue to use [MILESTONES](https://github.com/microsoft/ApplicationInsights-dotnet/milestones) to coordinate releases.


Readme.md
```diff
@@ -58,36 +58,3 @@ The following packages are published from this repository:
 Nightly Builds are available on our MyGet feed:
 `https://www.myget.org/F/applicationinsights-dotnet-nightly/api/v3/index.json`
 These builds come from the develop branch. These are not signed and are not intended for production workloads.
```
cijothomas, line 60: Instead of removing the whole section, we could say the 'milestones' in github could be used to find the upcoming release or something similar.

PR #1965, opened 2020-07-14
refactor FileDiagnosticsTelemetryModule

TimothyMothra:
Refactoring common implementation from FileDiagnosticsTelemetryModule.
I want to reuse these methods in the newer file logging.

## Changes
- Move Get Identity Name to IPlatform
- Move TestDirectory to IPlatform
- cleanup FileDiagnosticsTelemetryModule

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/IPlatform.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/IPlatform.cs
@@ -1,5 +1,7 @@
 ﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation
 {
+    using System.IO;
+
     using Microsoft.ApplicationInsights.Extensibility;
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/IPlatform.cs
```diff
@@ -33,5 +35,17 @@ internal interface IPlatform
...
+        /// <summary>
+        /// Test that this process can read and write to a given directory.
+        /// </summary>
+        /// <param name="directory">The directory to be evaluated.</param>
+        void TestDirectoryPermissions(DirectoryInfo directory);
+
+        /// <summary>
+        /// Get the current process's identity.
+        /// </summary>
+        /// <returns>Returns the current identity of the process.</returns>
```
rajkumar-rangaraj, line 48: nit: Identity of the process does not change till it exit. Should be just remove current here? We could just use process instead of process's?

TimothyMothra: i think i need to remove the word "process" here.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Platform/PlatformImplementation.cs
```diff
@@ -119,6 +120,48 @@ public string GetMachineName()
             return this.hostName ?? (this.hostName = GetHostName());
         }
 
+        public void TestDirectoryPermissions(DirectoryInfo directory)
```
cijothomas, line 123: `IPlatform` is the interface this class implements. `TestDirectoryPermissions` is not part of it. Don't think it should be part of it.

TimothyMothra: I want this method in a common location so I can reuse it.
I added it to the interface. I thought that interacting with the file system would be appropriate to put in the IPlatform. 
If you disagree, I can move it to a helper class instead. :)

Issue #1298, opened 2019-11-12
First heartbeat sent after 15 minutes instead of configured interval


xavierdecoster:
It appears that the very first heartbeat is sent after `15 minutes`, which is the default built-in heartbeat interval, even when a different interval is configured. Every other subsequent heartbeat will respect the configured interval.

SDK version used: `2.10.0`

Attached some code that reproduces the issue.

Expected behavior: heartbeats start after the very first minute (as configured)
Actual behavior: heartbeats only start after 15 minutes (default interval)

[ConsoleApp1.zip](https://github.com/microsoft/ApplicationInsights-dotnet/files/3837607/ConsoleApp1.zip)

cijothomas:
Its a bug/limitation as setting interval does not alter the current timer, which is triggered upon initialize. 
https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/BASE/src/Microsoft.ApplicationInsights/Managed/Shared/Extensibility/Implementation/Tracing/HeartbeatProvider.cs#L91

Use the following workaround:
Basically, don't use .Active, instead create DiagnosticTelemetryModule by yourself, so that you can interval **before** initializing it.

static void Main(string[] args)
        {

            var tc = new TelemetryConfiguration("ikey");
            var diagModule = new DiagnosticsTelemetryModule();
            diagModule.HeartbeatInterval = TimeSpan.FromSeconds(30);
            diagModule.Initialize(tc);
            
            while(true)
            {

            }
        }

(Modify example as per your requirement, but this should give you the idea)

xavierdecoster:
@cijothomas does the call to `new TelemetryConfiguration("ikey")` still consider the `applicationinsights.config` file? Or should we rely on `TelemetryConfiguration.CreateDefault()` instead, and then set `TelemetryConfiguration.InstrumentationKey` on that instance?

cijothomas:
new TelemetryConfiguration("ikey") wont consider ai.config.

PR #1979, opened 2020-07-30
Fix Heartbeat Interval

TimothyMothra:
Fixes issue of heartbeat interval not applying immediately when changed.
#1298

## Changes
- setting the heartbeat interval will now reset the timer.

### Checklist

### Notes for authors:

### Notes for reviewers:


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/DiagnosticsTelemetryModule.cs
```diff
@@ -49,7 +49,7 @@ public bool IsHeartbeatEnabled
         }
 
         /// <summary>
-        /// Gets or sets the delay interval between heartbeats.
+        /// Gets or sets the delay interval between heartbeats. Setting this value will immediately reset the heartbeat timer.
```
cijothomas, line 52: any way to unit test this change?

TimothyMothra: I'm thinking about this now. I'm reviewing our existing unit test coverage. Unit tests are passing locally, but I wanted this to run on our build server to confirm that all our integration tests still pass.

TimothyMothra: The `System.Threading.Timer` is not easily testable. This timer doesn't expose the Interval property so can't inspect that it was changed.

We don't have proper test coverage of how heartbeats are created.
I think we would need to do some significant refactoring of this class so we can manually invoke a "timer event" and validate the telemetry that gets created.
I think that's out of scope for this PR.

I propose that we take these changes as-is and revisit this later.

PR #1992, opened 2020-08-10
Perf improvement: Stream Processing (Frugal Tip G06)

TimothyMothra:
Issue #1993 

Fix for Frugal Tip G06.
LOH allocations are coming from our Json Deserializer.
> There are quite a few copies of uncompressed large data in memory. This hurts CPU, memory allocation, and working set. When the data is large enough, it can even lead to fragmentation and OutOfMemory exception.

## Changes
- Replace MemoryStream with StreamReader
- replace string[] with List<string>

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/BASE/src/ServerTelemetryChannel/Implementation/PartialSuccessTransmissionPolicy.cs
+++ b/BASE/src/ServerTelemetryChannel/Implementation/PartialSuccessTransmissionPolicy.cs
@@ -1,6 +1,7 @@
 ﻿namespace Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation
 {
     using System;
+    using System.Collections.Generic;
     using System.Threading.Tasks;
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Channel.Implementation;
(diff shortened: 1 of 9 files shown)


BASE/src/ServerTelemetryChannel/Implementation/PartialSuccessTransmissionPolicy.cs
```diff
@@ -48,17 +49,15 @@ private static string ParsePartialSuccessResponse(Transmission initialTransmissi
             string newTransmissions = null;
             if (backendResponse.ItemsAccepted != backendResponse.ItemsReceived)
             {
-                string[] items = JsonSerializer
-                    .Deserialize(initialTransmission.Content)
-                    .Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+                List<string> items = new List<string>(JsonSerializer.DeserializeToStrings(initialTransmission.Content));
```
reyang, line 52: I might be missing something, this doesn't seem to improve any perf or reducing the allocation - as a list is always created.
Creating an enumerator and then convert it to a list might actually get it worse since we're paying additional cost of the enumerator.
Is there a test result to validate what do we actually get?

rajkumar-rangaraj: We may be able to avoid deserialization here. TelemetryItems are part of `TransmissionProcessedEventArgs` - `args.Transmission.TelemetryItems`.  I had a sample built for ingestion status, there I do something like below. 

```
List<ITelemetry> failedTransmissions = new List<ITelemetry>();

foreach (var error in backendResponse.Errors)
{
failedTransmissions.Add(transmission.TelemetryItems.ElementAt(error.Index));
}
```

stebet:
Recreating a MemoryStream for serialization every time is expensive as it allocates and resizes a byte[] array multiple times. If you can determing a max-size for each chunk, it would make more sense to Rent/Return a byte[] array to initialize the MemoryStream with to reduce the number of allocations for (de)serializing the data.

pharring:
Let's see some perf numbers on this (allocations before and after the change).

Issue #1981, opened 2020-07-31
improve Self Diagnostics and File Logging


TimothyMothra:
The motivation for this change is to improve troubleshooting the SDK.
For several customers, it's problematic to make changes to their applications for the purpose of investigating the AI SDK. Some customers have safe-deployment requirements which can delay code changes by as much as a week from being deployed into production.

To address this several SDK teams contributed to this spec: https://github.com/microsoft/ApplicationInsights-Home/blob/master/CrossSdkCollaboration/Self-Diagnostics.md
The goal is for a SysAdmin or DevOps to be able to set an environment variable and to control the SDKs file logging.

### Tasks:
- [ ] 1. Add file logging into `DiagnosticsTelemetryModule`. #1724
    - [ ] Enable Environment Variable control as defined by the spec.
    - [ ] Retire `FileDiagnosticsTelemetryModule`. We don't need two ETW subscribers. 
    - [x] Need to refactor AspNetCore's dependency on `TelemetryConfiguration.Active`. #1886
- [ ] 2. Refactor severity. Need to set independently for File and Portal logging.
- [ ] 3. File Logging on by default (Stretch Goal)
    - [ ] Feature: must be more performant and remove the dependency on lock
    - [ ] Feature: control the file size
    - [ ] Feature: must be able to remove outdated log files.

PR #1999, opened 2020-08-21
[WIP] Self Diagnostics: Part 1 Configuration

TimothyMothra:
Fix Issue #1981.

This PR is part 1 and only contains the configuration story.
If approved, will work on file writing in part 2.

## Changes
- Add self-diagnostics object to TelemetryConfiguration static constructor.
- Self Diagnostics can only be set via Environment Variable. No Public API.

### Checklist

### Notes for authors:

### Notes for reviewers:


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsProvider.cs
```diff
@@ -0,0 +1,119 @@
...
+                    return instance;
+                }
+                else
+                {
+                    return default(T);
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new Exception("Failed to parse Self-Diagnostics config string. You must fix or remove this configuration.", ex);
```
cijothomas, line 51: won't this crash app startup?

TimothyMothra: that was the initial design yes. but I don't want to merge it without having a discussion.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -64,6 +69,8 @@ static TelemetryConfiguration()
                     Activity.ForceDefaultIdFormat = true;
                 }                
             });
+
+            SelfDiagnostics = SelfDiagnosticsProvider.EvaluateSelfDiagnosticsConfig<SelfDiagnosticsFileWriter>();
```
cijothomas, line 73: nit: Method name could use a better naming - this is not just Evaluating something. Its building the actual "thing" and assigning..

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsProvider.cs
TimothyMothra: The user scenario would be... SDK isn't working as expected. Set EnVar to turn on file logging. If file logging isn't working as expected now a user would have two issues. 

I'm comfortable with this change only because it's meant for troubleshooting scenarios.
We can fail early to let the customer know that the EnVar is set incorrectly. 

If we don't do this, the customer would need something like PerfView to investigate which might not be an option in some Azure environments.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
TimothyMothra: I'm terrible at naming things so I'm open to suggestions here.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsProvider.cs
cijothomas: While it may be good/bad to prevent AppStartup due to telemetry errors, I'd stick to the existing behaviour for the SDK. We never throw, even when user provides incorrect config. I'll just the same behavior.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsFileWriter.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.SelfDiagnostics
+{
+    internal class SelfDiagnosticsFileWriter : ISelfDiagnosticsFileWriter
```
cijothomas, line 3: I am not sure but checking -- shouldn't this be IDisposable to release file handles.?

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
cijothomas, line 73: One issue with making this inside Static ctor is - there is no way to dispose.

cijothomas: probably best to move this inside the constructor.

cijothomas: and dispose the file stuff when we dispose everything else.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsProvider.cs
ThomsonTan: Probably display a warning message to say the environment variable is ignored?

---
In reply to: [475831878](https://github.com/microsoft/ApplicationInsights-dotnet/pull/1999#discussion_r475831878) [](ancestors = 475831878)

cijothomas: @ThomsonTan The question is "where to display"? :D  Server SDKs traditionally have been using EventSource to log that something went bad. But unless user is listening to EventSource using some tools like PerfView/others, there is no way they are going to know something went wrong.

ThomsonTan: I see. If there is no guaranteed way to deliver the warning/error message to users in server SDK, perhaps we could validate the environment variable as early as possible, so the app can only fail during early startup time which the user will be aware of.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
TimothyMothra: StreamWriter is created in a using and is immediately disposed.
But the EventListener that subscribes to ETW events does need to be disposed :(

I could move this to the normal ctor, but we would still only want one instance of the file logger.
This creates a new problem.. how do I properly dispose a single instance of the filelogger if 2+ TelemetryConfigurations have access to it??

TimothyMothra: in the static scenario, we can subscribe to the Process Exit and dispose then. Would look like this: https://stackoverflow.com/a/13258842/1466768

PR #2004, opened 2020-08-25
AI JS snippet changes

kryalama:
Fix Issue # .

## Changes
Update AppInsights JS snippet used in the code to latest version and refactor code to use AppInsights JS snippet as it is without escaping curly brackets.

### Checklist
- [ YES ] I ran Unit Tests locally.
- [ YES ] CHANGELOG.md updated with one line description of the fix, and a link to the original issue if available.

- [ NA ] Design discussion issue #
- [ NA ] Changes in public surface reviewed

### Notes for authors:

### Notes for reviewers:


NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/JavaScriptSnippet.cs
```diff
@@ -113,9 +113,9 @@ public string ScriptBody
                     }
                 }
 
+                Snippet = Snippet.Replace("instrumentationKey: \"INSTRUMENTATION_KEY\"",insertConfig );
                 // Return snippet
-                // Developer Note: If you recently updated the snippet and are now getting "FormatException: Input string was not in a correct format." you need to escape all the curly braces; '{' => '{{' and '}' => '}}'.
-                return string.Format(CultureInfo.InvariantCulture, Snippet, insertConfig, insertAuthUserContext);
+                return string.Format(CultureInfo.InvariantCulture, "{0}{1}" , Snippet , insertAuthUserContext);
```
TimothyMothra, line 118: you don't need a `string.format` here. can use `string.concat` instead.

```suggestion
                return string.Concat(Snippet , insertAuthUserContext);
```

TimothyMothra:
you need to get the latest changes from the `develop` branch.
this PR is against the old state of the Resources. but your earlier changes have merged already. that's creating a conflict that you need to resolve.

TimothyMothra:
@kryalama you have several build errors. These are from FxCop and StyleCop.
If you compile your project in Release mode, you'll see these locally and can fix

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/JavaScriptSnippet.cs
```diff
@@ -34,6 +31,9 @@ public class JavaScriptSnippet : IJavaScriptSnippet
 
         private readonly JavaScriptEncoder encoder;
 
+        /// <summary>JavaScript snippet.</summary>
+        private string snippet = Resources.JavaScriptSnippet;
```
TimothyMothra, line 35: I see why you changed this from `static`, so you can set the result of the string.replace on this same variable.
But I don't think this is necessary.

I believe a new instance of `JavaScriptSnippet` is going to be created for every request, so we would be unable to store the result of the string replace.
Plus you're only storing the first half (ikey) and not the second half (auth context).

I think you can change this back to the static variable `static readonly string Snippet`.
In the `ScriptBody()` method, you can store the result of string Replace in a local variable. `var snippet = this.Snippet.Replace`

kryalama: restored static readonly snippet

PR #2072, opened 2020-09-25
Create versions_and_names.md

TimothyMothra:
Migrating this doc from Home repo.
https://github.com/microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/SDK-VERSIONS.md

### DIFF
--- a/docs/versions_and_names.md
+++ b/docs/versions_and_names.md
@@ -1,0 +1,131 @@
+# SDK Version
+
+SDK version is a field you can specify on every telemetry item. This field represent the specific SDK collected this particular item. This field is used for troubleshooting.
+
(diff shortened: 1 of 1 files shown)


rajkumar-rangaraj (approved):
We may later need to revisit this document and move the other language sdk types to their repo.

TimothyMothra:
> We may later need to revisit this document and move the other language sdk types to their repo.

I agree, but for now I just want to migrate this to our repo so we don't lose it. 
However, I think an argument can be made for keeping all the names in a central location. When investigating a customer issue, I wouldn't want to visit a dozen repos to find out who one name belongs to.

PR #2092, opened 2020-10-14
updating dotnet/net versions

eddynaka:
Fix Issue # .

## Changes

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/BASE/.vsts/linux-build.yml
+++ b/BASE/.vsts/linux-build.yml
@@ -14,17 +14,17 @@ steps:
 - task: UseDotNet@2
   displayName: install dotnet core 2.1
   inputs:
-    version: 2.1.807
+    version: 2.1.811
     
 - task: UseDotNet@2
   displayName: install dotnet core 3.1
   inputs:
-    version: 3.1.301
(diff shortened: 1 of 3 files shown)


cijothomas (approved):

BASE/.vsts/linux-build.yml
```diff
@@ -14,17 +14,17 @@ steps:
 - task: UseDotNet@2
   displayName: install dotnet core 2.1
   inputs:
-    version: 2.1.807
+    version: 2.1.811
```
TimothyMothra, line 17: I thought I read that you could use a wildcard here to always use the latest version available.
something like `2.1.*`
That might save us from needing to update this file regularly. 
But then we would lose testing with specific versions.

What are your thoughts?

eddynaka: we were using. i will try to check if we still need the rsync.

so:
1. checking if changing will break (i hope it won't)
2. remove the specific version and remove rsync as well (that means 2.1.* and 3.1.*)

Issue #1963, opened 2020-07-12
Asp.Net core synthetictelemetryinitializer is not configurable.


cijothomas:
**Is your feature request related to a problem?**
The SyntheticTelemetryInitializer for Asp.Net Core (https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/SyntheticTelemetryInitializer.cs), is hard coded to detect only the Azure Availability Test traffic. And there is no option to configure it to mark additional sources.
The SyntheticTelemetryInitializer for Asp.Net (https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/WEB/Src/Web/Web/SyntheticUserAgentTelemetryInitializer.cs), is configurable, and has a default list which meets most user's requirements (https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/WEB/Src/Web/Web/net452/ApplicationInsights.config.install.xdt#L8)

**Describe the solution you'd like.**
Make Asp.Net core syntheticinitializer match the capabilities of asp.net one.

**Describe alternatives you've considered.**
Users have to author own initializer, which is not a good workaround.

fabricebauer:
I have created a separate initializer for this (to avoid impacting this one) for our own private project.
I would like to contribute, and will submit a Pull Request for it.

PR #2137, opened 2020-12-22
Add configurable SyntheticUserAgentTelemetryInitializer (#1963)

fabricebauer:
## Changes
Add equivalent of existing one in ASP.Net with default filters values and possibility
to configure (via IConfiguration injection) or customize filters (via setter).

### Checklist

### Notes for authors:

### Notes for reviewers:


cijothomas (changes requested):
Instead of adding a new initializer, we should just make the existing one customizable.

fabricebauer:
> Instead of adding a new initializer, we should just make the existing one customizable.

I thought about it indeed, but the existing one seems very specific to "Application Insights Availability Monitoring", so I thought it might be better/more flexible to have a dedicated one for "User-agent" related values. 

Also the existing one is added by default in AddAspNetCoreWebTelemetryInitializers, not for the new one, as I thought not everybody might want it by default (there is a slight performance penalty checking filters values).
 
What do you think? 
I am ok with both approaches, so if you still think it is better to have a single one, I will do the changes and submit another PR.

Thanks.

cijothomas:
>  but the existing one seems very specific to "Application Insights Availability Monitoring", so I thought it might be better/more flexible to have a dedicated one for "User-agent" related values.

This is not the case. There is nothing specific to ApplicationInsights Availability monitoring in that initializer.

cijothomas:
> lso the existing one is added by default in AddAspNetCoreWebTelemetryInitializers, not for the new one, as I thought not everybody might want it by default (there is a slight performance penalty checking filters values).

We can do similar approach as whats done in asp.net. (https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/WEB/Src/Web/Web/SyntheticUserAgentTelemetryInitializer.cs#L17)
Empty filters by default. So perf hit is only for those user's who add more filters to it.
(In case of asp.net, the default AI.config file has a bunch of filters - https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/WEB/Src/Web/Web/net452/ApplicationInsights.config.install.xdt#L8 . I think it'd be okay to have the same defaults for aspnetcore as well)

fabricebauer:
Closing this PR, submitted new PR https://github.com/microsoft/ApplicationInsights-dotnet/pull/2148

Issue #2142, opened 2021-01-02
Incorrect Parent ID


ylabade:
### Describe the bug

The parent Id seems incorrect using DependencyTrackingTelemetryModule.IncludeDiagnosticSourceActivities

### To Reproduce

```cs
public class SomeDependencyDiagnosticSource
{
    public const string DiagnosticListenerName = "Company.SomeDependency";
    public const string ActivityName = DiagnosticListenerName + ".RequestOut";
    public const string ActivityStartName = ActivityName + ".Start";
    public const string ActivityStopName = ActivityName + ".Stop";

    private static readonly DiagnosticListener _diagnostic = new DiagnosticListener(DiagnosticListenerName);

    public static SomeDependencyDiagnosticSource Instance => new SomeDependencyDiagnosticSource();

    public Activity? StartActivity()
    {
        if (_diagnostic == null)
        {
            return null;
        }

        Activity? activity = null;
        if (_diagnostic.IsEnabled() && _diagnostic.IsEnabled(DiagnosticListenerName))
        {
            activity = new Activity(ActivityName);

            // add some tags/baggages to activity.
            activity.AddTag("http.url", "https://somedependency.company.net");

            if (_diagnostic.IsEnabled(ActivityStartName))
            {
                _diagnostic.StartActivity(activity, null);
            }
            else
            {
                activity.Start();
            }
        }

        return activity;
    }

    public void StopActivity(Activity? activity)
    {
        if (activity == null)
        {
            return;
        }

        // set some tags like dependency result.

        if (_diagnostic != null && _diagnostic.IsEnabled(ActivityStopName))
        {
            _diagnostic.StopActivity(activity, null);
        }
        else
        {
            activity.Stop();
        }
    }
}
```
Usage :
```cs
public Task<DependencyResult> CallDependency()
{
    var activity = SomeDependencyDiagnosticSource.Instance.StartActivity();
    try
    {
        // Call the dependency
    }
    finally
    {
        SomeDependencyDiagnosticSource.Instance.StopActivity(activity);
    }
}
```
Application Insights configuration in ASP.NET Core startup : 

```cs
services.AddApplicationInsightsTelemetry(options =>
{
    options.InstrumentationKey = Configuration["ApplicationInsights:InstrumentationKey"];
});
services.ConfigureTelemetryModule<DependencyTrackingTelemetryModule>((m, o) =>
{
[...]

PR #2145, opened 2021-01-04
Fix: telemetry parent id when using W3C activity format

ylabade:
Fix Issue #2142 

### Checklist

### DIFF
--- a/WEB/Src/DependencyCollector/DependencyCollector/TelemetryDiagnosticSourceListener.cs
+++ b/WEB/Src/DependencyCollector/DependencyCollector/TelemetryDiagnosticSourceListener.cs
@@ -50,8 +50,23 @@ namespace Microsoft.ApplicationInsights.DependencyCollector.Implementation
             }
 
             // properly fill dependency telemetry operation context
-            telemetry.Context.Operation.Id = currentActivity.RootId;
-            telemetry.Context.Operation.ParentId = currentActivity.ParentId;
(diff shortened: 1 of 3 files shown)


cijothomas:
@ylabade Thanks for opening the bug and the PR. We'll review this soon.

WEB/Src/DependencyCollector/DependencyCollector/TelemetryDiagnosticSourceListener.cs
```diff
@@ -50,8 +50,23 @@ public void OnEvent(KeyValuePair<string, object> evnt, DiagnosticListener diagno
             }
 
             // properly fill dependency telemetry operation context
-            telemetry.Context.Operation.Id = currentActivity.RootId;
-            telemetry.Context.Operation.ParentId = currentActivity.ParentId;
+            if (currentActivity.IdFormat == ActivityIdFormat.W3C)
```
cijothomas, line 53: Could you add unit test for this? There should already be unit tests for this class,

cijothomas (commented):
Changes look good. Please add a unit test to cover the test gap.

WEB/Src/DependencyCollector/DependencyCollector/TelemetryDiagnosticSourceListener.cs
ylabade: I'll do it, thanks for the review

cijothomas (approved):

PR #2148, opened 2021-01-08
Add configurable SyntheticUserAgentTelemetryInitializer (#1963)

fabricebauer:
Add equivalent of existing one in ASP.Net with default filters values and possibility
to configure (via IConfiguration injection) or customize filters (via setter).

Fix Issue # .

## Changes

### Checklist

### Notes for authors:

### Notes for reviewers:


fabricebauer:
As discussed in https://github.com/microsoft/ApplicationInsights-dotnet/pull/2137, this is a new version with updates in the existing SyntheticTelemetryInitializer class.

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/SyntheticTelemetryInitializer.cs
```diff
@@ -4,24 +4,66 @@
...
+        private string syntheticUserAgentFilters;
+        private string[] syntheticUserAgentFilterPatterns;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="SyntheticTelemetryInitializer" /> class.
         /// </summary>
         /// <param name="httpContextAccessor">Accessor to provide HttpContext corresponding to telemetry items.</param>
-        public SyntheticTelemetryInitializer(IHttpContextAccessor httpContextAccessor)
+        /// <param name="configuration">Configuration to get synthetic user agent filters.</param>
+        public SyntheticTelemetryInitializer(IHttpContextAccessor httpContextAccessor, IConfiguration configuration)
```
cijothomas, line 31: This will be breaking change to existing customers. We can add new ctor, but need to keep old one as is., or make new params optional.

fabricebauer: Hi @cijothomas, I thought about it, indeed, but my understanding is that this would cause issues with dependency injection.
Do you mean existing customers sometimes explicitly call this constructor? (instead of relying on AddAspNetCoreWebTelemetryInitializers)
Thanks

cijothomas: > Do you mean existing customers sometimes explicitly call this constructor?

They can. Its a public API. So we have to always maintain the backward compatibility, since we follow semver guidelines.

(could you detail what is the issue DI?)

fabricebauer: I was wondering how DI would know which constructor to use and looked at https://docs.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#constructor-injection-behavior. 
I also saw https://blogs.cuttingedge.it/steven/posts/2013/di-anti-pattern-multiple-constructors/, but I am not 100% sure if this applies to .NET Core DI.

PR #2185, opened 2021-03-10
[Prototype] Support ActivityLinks

pakrym:
Currently, Azure SDKs produce a special kind of Activities that carries links in a 
``` C#
public IEnumerable<Activity> Links { get; }
```
property.

Migration to ActivitySource based activities would result in the same information being carried in the 
https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.links?view=net-5.0

This PR add support for collecting both ways of representing links.


pakrym:
cc @lmolkova

lmolkova (approved):

pakrym:
@rajkumar-rangaraj any news on this? The .NET 5 activity link support has shipped in Azure SDKs

Issue #2190, opened 2021-03-24
New Feature: Azure Active Directory


TimothyMothra:
# Azure Active Directory
Application Insights SDK has a new requirement to support AAD.

- MS Docs: [Azure AD authentication for Application Insights](https://docs.microsoft.com/en-us/azure/azure-monitor/app/azure-ad-authentication?tabs=net)
- MS Docs: https://docs.microsoft.com/azure/active-directory/fundamentals/active-directory-whatis

## Scope of Work
- Configuration
    - [x] `TelemetryConfiguration` to accept `TokenCredential` via `CredentialEnvelope`
    - [x] AspNetCore configuration - this is omitted because these settings should be configurable from a text config file. 
- Support AAD in
   - [x] `InMemoryChannel`
   - [x] `ServerTelemetryChannel`
       - [x] RetryPolicy via `AuthenticationTransmissionPolicy`
           - [x] ExponentialBackoff - this is omitted because not helpful for clientside errors.
   - [x] `QuickPulseServiceClient`
 - Token Caching 
     - [x] Azure.Identity implementation provides caching for most scenarios.
     - [ ] Our own caching - under discussion. Guidance was to rely on Azure.Identity implementation. 
- Logging
    - [x] Configuration logging
    - [x] `InMemoryChannel`
    - [x] `ServerTelemetryChannel`
    - [x] `QuickPulseServiceClient`
    - [x] `AuthenticationTransmissionPolicy`
    - [x] collect Azure.Identity logs - this is cut. The decision is to not take on the task of troubleshooting other products.
- Changelog
    - [x] review changelog for completeness
- Tests
    - [ ] investigate adding new E2E tests.
- Outstanding Todos
    - [x] Before Stable release, re-review all changes to the PublicApi
    - [ ] After Stable release, update [public doc](https://docs.microsoft.com/azure/azure-monitor/app/azure-ad-authentication?tabs=net) with stable version 
    - [ ] Transmission.CreateRequestMessage()
        - should we block transmission if token is unavailable?
        

## Requirements

- AAD should be optional for existing customers. 
- If AAD is enabled, there will be no backchannels for AAD-less telemetry. Data will be dropped!
- Proper authentication will look like an additional token in all request headers.
    - Services are responsible for validating tokens.
- Application Insights SDK will not be responsible for creating tokens. We will rely on existing work from the [Azure SDK](https://github.com/Azure/azure-sdk-for-net)
- Affected Endpoints:
[...]

PR #2191, opened 2021-03-25
AAD: Handling Azure.Core.TokenCredential

TimothyMothra:
# Adding support for Azure.Core.TokenCredential to the TelemetryConfiguration
For more information see the parent item: #2190.

## Introduction

We have a unique challenge of needing to support `Azure.Core.TokenCredential` while having incompatible Frameworks.

image

To address this I'm making the assumption that if a customer's application is able to create an instance of `TokenCredential`, then they also have the `Azure.Core` assembly in their runtime.
Going off of this assumption, we're able to use reflection to access the methods of `TokenCredential` and return types available in `Azure.Core`.
This approach is validated in unit tests.

### TelemetryConfiguration.SetCredential
We need a way for a customer to provide an implementation of `Azure.Core.TokenCredential`.
For this, I've introduced a new method `TelemetryConfiguration.SetCredential`. 

https://github.com/microsoft/ApplicationInsights-dotnet/blob/aac78edd659be65d4118813c53a44652d23184e9/BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs#L408-L412

We accept `Object` and validate that the provided object inherits `TokenCredential`.
We store this in the new property `TelemetryConfiguration.CredentialEnvelope` which is accessible by all internal classes in the SDK.

### CredentialEnvelope

We hold the instance of `TokenCredential` in a new class `ReflectionCredentialEnvelope` which as the name suggests, uses Reflection to interact with the `Object` that was provided.

https://github.com/microsoft/ApplicationInsights-dotnet/blob/aac78edd659be65d4118813c53a44652d23184e9/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs#L21-L37


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                throw new ArgumentException($"The provided {nameof(tokenCredential)} must inherit Azure.Core.TokenCredential", nameof(tokenCredential));
+            }
+        }
+
+        public object Credential { get; private set; }
+
+        public string GetToken(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var accessToken = this.getTokenMethod.Invoke(this.Credential, new object[] { this.tokenRequestContext, cancellationToken });
```
reyang, line 59: If this is intended to be called frequently, consider Expression Compile.

reyang: For example:

```csharp
            var flags = BindingFlags.Instance | BindingFlags.NonPublic;
            var ctor = typeof(Batch<T>).GetConstructor(flags, null, new Type[] { typeof(T) }, null);
            var value = Expression.Parameter(typeof(T), null);
            var lambda = Expression.Lambda<Func<T, Batch<T>>>(Expression.New(ctor, value), value);
            CreateBatch = lambda.Compile();
```

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
```diff
@@ -0,0 +1,84 @@
...
+            //var resultProperty = task.GetType().GetProperty("Result");
+            //var accessToken = resultProperty.GetValue(task);
+            //var tokenProperty = accessToken.GetType().GetProperty("Token");
+            //return (string)tokenProperty.GetValue(accessToken);
+        }
+
+        /// <summary>
+        /// Checks if the input inherits <see cref="Azure.Core.TokenCredential"/>
+        /// </summary>
+        private bool IsTokenCredential(Type inputType) => inputType.IsSubclassOf(Type.GetType("Azure.Core.TokenCredential, Azure.Core"));
```
reyang, line 82: `Type.GetType("Azure.Core.TokenCredential, Azure.Core")` needs to be precalcualted/cached so we don't end up calling reflection each time.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ICredentialEnvelope.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Authentication
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public interface ICredentialEnvelope
```
reyang, line 6: Do we need this to be public? Consider hiding it since we don't expect external components to take dependency.
Probably we don't need an interface at all?

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/TokenCredentialEnvelope.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async Task<string> GetTokenAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var accessToken = await this.tokenCredential.GetTokenAsync(requestContext: this.tokenRequestContext, cancellationToken: cancellationToken);
+            return accessToken.Token;
+        }
+    }
+}
+#endif
```
reyang, line 48: nit: put a blank line

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
TimothyMothra: This will only be called once (from the constructor) so I think we're safe to not cache this one.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ICredentialEnvelope.cs
TimothyMothra: I think this NEEDS to be public so other parts of our SDK can access the GetToken method.
Reminder that this will be accessible from the `TelemetryConfiguration`. 
This will be in the Microsoft.ApplicationInsights library, but will need to be accessed by ServerTelemetryChannel and QuickPulseTelemetryModule which are both in separate modules. Also if SnapshotDebugger and Profiler intend to leverage our work here.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/TokenCredentialEnvelope.cs
TimothyMothra: Thanks, I make this mistake a lot because Visual Studio doesn't call attention to it.
Are there any Analyzers that we could include in the project to fail on this?

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
TimothyMothra: An earlier version of this method was more complicated before I found the one-line statement (verified with unit tests). 
I could move this into the constructor now to eliminate the confusion. :)

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ICredentialEnvelope.cs
TimothyMothra: Can you comment on why you would recommend an abstract class here?
I went back and forth on this exact topic, but ultimately I couldn't find a use case for the abstract class.
Maybe you're considering something here that I've overlooked and I would like to learn :)

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// Our SDK currently targets net452, net46, and netstandard2.0.
+    /// Azure.Core.TokenCredential is only available for netstandard2.0.
+    /// I'm introducing this class as a wrapper so we can receive an instance of this class and pass it around within our SDK.
+    /// </remarks>
+    internal class ReflectionCredentialEnvelope : ICredentialEnvelope
+    {
+        private readonly object tokenRequestContext;
+        private readonly MethodInfo getTokenAsyncMethod;
+        private readonly MethodInfo getTokenMethod;
+        //private readonly Type accessTokenType;
```
ThomsonTan, line 21: What was this `accessTokenType` intended for?

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <remarks>
+        /// Must use reflection to verify that this object is an instance of Azure.Core.TokenCredential.
+        /// </remarks>
+        /// <param name="tokenCredential"></param>
+        public ReflectionCredentialEnvelope(object tokenCredential)
+        {
+            if (this.IsTokenCredential(tokenCredential.GetType()))
```
ThomsonTan, line 32: Check if `tokenCredential` is null?

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ICredentialEnvelope.cs
reyang: `public interface ICredentialEnvelope` means anyone could take dependency and implement that interface, which means we cannot add new stuff since it breaks compatibility.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
TimothyMothra: I want to use reflection to call `public async ValueTask<AccessToken> GetTokenAsync`.
In this case, `TokenCredential`, `AccessToken`, and `ValueTask<>` are all unavailable to me in older frameworks.
The example code is also commented out below at Line 68.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
reyang: If the constructor is supposed to be called once (e.g. singleton) it's fine.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/TokenCredentialEnvelope.cs
reyang: Perhaps steal from https://github.com/open-telemetry/opentelemetry-dotnet/blob/5f1acbf8ed360a497133b6ecae42f1268621dad6/build/sanitycheck.py#L25?

pharring:
This might be an opportunity to simplify dependencies and targets for net452 and net46 altogether.
Indeed, you might be able to get to the point where netstandard2.0 is the only target framework.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/CredentialEnvelope.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// </summary>
+        /// <param name="cancellationToken">The System.Threading.CancellationToken to use.</param>
+        /// <returns>A valid Azure.Core.AccessToken.</returns>
+        public abstract Task<string> GetTokenAsync(CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Get scopes for Azure Monitor as an array.
+        /// </summary>
+        /// <returns>An array of scopes.</returns>
+        protected static string[] GetScopes() => new string[] { Scope };
```
pharring, line 39: PERF: This creates a new array each time it's called. Consider using a static field -- since it's a compile-time constant.

pharring: Ah, I see that derived classes call it just once and cache the result in a TokenRequestContext.
Note that TokenRequestContext also includes a "ParentRequestId" property -- currently not being set or used here. You may find it useful to set that ID so that you have a correlation ID for AAD failures. At that point, you'll want to create a new TokenRequestContext object each time.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
```diff
@@ -0,0 +1,203 @@
...
+            /// This is a wrapper for the following method:
+            /// <code>public abstract System.Threading.Tasks.ValueTask&lt;Azure.Core.AccessToken> GetTokenAsync (Azure.Core.TokenRequestContext requestContext, System.Threading.CancellationToken cancellationToken);</code>
+            /// (https://docs.microsoft.com/dotnet/api/azure.core.tokencredential.gettokenasync).
+            /// </summary>
+            /// <returns>
+            /// The Expression Tree library cannot handle async methods.
+            /// As a workaround, this method returns two Delegates. 
+            /// First;
+            /// The first Delegate is a wrapper around GetTokenAsync which returns a ValueTask of AccessToken.
+            /// Then calls ValueTask.GetTask to convert that to a Task which is a known type for older frameworks.
```
pharring, line 143: Is this step (converting ValueTask to Task) necessary? Could you use `ValueTask<AccessToken>` directly?
i.e. Have the first delegate/expression return the `ValueTask<AccessToken>` and pull out the `.Result.Token` of that in the 2nd delegate.

TimothyMothra: I've currently got no way to `await` the `ValueTask` because that type isn't known to these older frameworks.
Also the Expression Tree api doesn't provide support for async methods.
I'm very open to suggestions here! :)

pharring: Ah, I see. The await is the problem. You need to return something "awaitable" (has a compiler-visible GetAwaiter() method). Yup. Task is probably the best way.

TimothyMothra:
Builds are currently broken because I'm working on tests. Will fix it tomorrow.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -399,6 +405,16 @@ public void Dispose()
...
 
+        /// <summary>
+        /// Set a TokenCredential for this configuration.
+        /// </summary>
+        /// <param name="tokenCredential">An instance of Azure.Core.TokenCredential.</param>
+#if NETSTANDARD2_0
+        public void SetCredential(Azure.Core.TokenCredential tokenCredential) => this.CredentialEnvelope = new TokenCredentialEnvelope(tokenCredential);
+#else
+        public void SetCredential(object tokenCredential) => this.CredentialEnvelope = new ReflectionCredentialEnvelope(tokenCredential);
+#endif
```
pharring, line 416: We have a problem with this. The profiler agent is a net472 executable so, when it references the Application Insights SDK, it picks up the net46 flavor (SetCredential types as object). However, the profiler agent also references another class library (for logging) that's built for netstandard2.0. _That_ class library references the ApplicationInsights SDK too but, since it's netstandard2.0, it picks up the netstandard2.0 flavor of App Insights SDK and, therefore, the TokenCredential flavor of this API. 

When it all comes together, there's a MissingMethodException at run time because the EXE loads the net46 version, but the logger tries to call the SetCredential(TokenCredential...) method which doesn't exist in net46.

The way it is now, in order to address this, we have to make our logging library multi-target net472 and netstandard2.0.

TimothyMothra: Managing the conflicting frameworks has been the most complicated part of all of this! :)
I'm confused why your net472 agent isn't taking the netstandard2.0 version of our SDK.

@pharring **Would it help you if we target net462?**

As of today, this sdk targets net452, net46, and netstandard2.0.
[Official guidance](https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/) is to migrate all projects off of 4.5.2, 4.6, 4.6.1 to 4.6.2 by April 26, 2022.

cijothomas: We faced similar issues in OpenTelemetry as well.

@utpilla as well who investigated this, and following up with .NET team to get an official guidance on this, if any.

cijothomas: > I'm confused why your net472 agent isn't taking the netstandard2.0 version of our SDK.

Since our project has a netstandard and netframework target, a net472 project "prefers" the netframework version. So it picks net46 as thats closes to net472, and not netstandard2.0

TimothyMothra: Thanks for the clarification @cijothomas .
In my opinion, we should include support for 4.6.2 sooner rather than later. I can start a new branch to investigate how much work this will be.
We don't need to retire older frameworks until next year.

pharring: @cijothomas answered correctly above.

@TimothyMothra, I think adding 4.6.2 support would help in the meantime.

I suspect, once you retire net45-net461, you'll probably land where I did with Snapshot Collector 1.4-pre : just ship a netstandard2.0 version (Covered [here](https://github.com/microsoft/ApplicationInsights-SnapshotCollector/discussions/5))

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -334,6 +335,11 @@ public string ConnectionString
         /// </summary>
         public TelemetrySink DefaultTelemetrySink => this.telemetrySinks.DefaultSink;
 
+        /// <summary>
+        /// Gets an envelope for Azure.Core.TokenCredential which provides an AAD Authenticated token.
+        /// </summary>
+        public CredentialEnvelope CredentialEnvelope { get; private set; }
```
cijothomas, line 341: Consider making this internal, if end users are not expected to read this back.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/CredentialEnvelope.cs
```diff
@@ -0,0 +1,44 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Authentication
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// An envelope for an instance of Azure.Core.TokenCredential.
+    /// </summary>
+    public abstract class CredentialEnvelope
```
cijothomas, line 10: Consider making internal, unless we expect end users to use this.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
reyang: Would we consider the following approach and what's the pros/cons?

Always use `SetCredential(object tokenCredential)` signature, and avoid any static dependency (package dependency and assembly reference) to the Azure SDK. Use runtime detection / reflection.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
TimothyMothra: @pharring, We're planning on making the CredentialEnvelope `internal`.
We want to get your library added to our InternalsVisibleTo [here](https://github.com/microsoft/ApplicationInsights-dotnet/blob/ddd41c578c6cfad0788bf1defa11e68792cc35c5/BASE/src/Microsoft.ApplicationInsights/Properties/AssemblyInfo.cs#L7-L11).

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
```diff
@@ -0,0 +1,205 @@
...
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// This is an implementation of <see cref="CredentialEnvelope"/> that uses reflection to interact with the Azure.Core library.
+    /// </summary>
+    /// <remarks>
+    /// Our SDK currently targets net452, net46, and netstandard2.0.
+    /// Azure.Core.TokenCredential is only available for netstandard2.0.
+    /// </remarks>
+    internal class ReflectionCredentialEnvelope : CredentialEnvelope
```
cijothomas, line 15: if this reflection based approach works good enough, is there any reason then to depend on azure.core in netstandard case?

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
pharring: ^^^ that would work for us, I think.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
pharring: @xiaomi7732 for https://github.com/microsoft/ApplicationInsights-Profiler-AspNetCore

Snapshot Collector is Microsoft.ApplicationInsights.SnapshotCollector, but note that it has a different public key.

xiaomi7732: Thanks for tagging me on this. AI.SDK and Profiler have different build cycles - and we have mixed references. **How are we going to make that work?** For example, our desktop build (no StrongName) assemblies might reference public/stable version of app insights SDK, in which case, the InternvalVisibileTo (to a strongNamed assembly) won't work. What is that I missed?

pharring: I don't think you need to worry about version numbers. InternalsVisibleTo uses just the assembly's name and public key. The question really is whether you expect to take a dependency on a version of the core SDK that has CredentialEnvelope, or would you stick with an older version and use reflection to get CredentialEnvelope.

I still haven't decided for Snapshot Collector. There's going to be some reflection anyway in order to use the CredentialEnvelope.

xiaomi7732: On this topic, I think I am asked to provide assembly names and public keys so that in Core SDK, it knows to expose the internal classes.

The deal is, during the development, our assemblies aren't going to be strongnamed - that means we won't be able to see the internal classes in the Core SDK anyway.

On the other hand, reflection will work, with or without `InternalVisibleTo`.

So, what is the point to add the `InternalVisibleTo` for our assemblies? And that is what I am asking, in case I missed something.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
TimothyMothra: Done

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
TimothyMothra: We're changing this property to `internal`. 
Optionally, if you want to utilize this without reflection you can submit a PR to add your SDK to our `InternalsVisibleTo`.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
```diff
@@ -0,0 +1,216 @@
...
+        /// Gets the TokenCredential object held by this class.
+        /// </summary>
+        public object Credential => this.tokenCredential;
+
+        /// <summary>
+        /// Gets an Azure.Core.AccessToken.
+        /// </summary>
+        /// <param name="cancellationToken">The System.Threading.CancellationToken to use.</param>
+        /// <returns>A valid Azure.Core.AccessToken.</returns>
+        public string GetToken(CancellationToken cancellationToken = default)
```
cijothomas, line 49: is it possible that this can throw an exception?

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -334,6 +335,11 @@ public string ConnectionString
         /// </summary>
         public TelemetrySink DefaultTelemetrySink => this.telemetrySinks.DefaultSink;
 
+        /// <summary>
+        /// Gets an envelope for Azure.Core.TokenCredential which provides an AAD Authenticated token.
+        /// </summary>
+        internal ReflectionCredentialEnvelope CredentialEnvelope { get; private set; }
```
cijothomas, line 341: I think you can still define an interface containing the GetToken() method, and this field be of that type. If we ever want to change the implementation from ReflectionCredentialEnvelope to something else, it'd be easy that way.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
TimothyMothra: Yes. I was leaning towards handling this insider the calling class. 
Because being unable to get a token will require behavior changes in the TelemetryChannel and QuickPulse.

Alternatively, I could handle it here and return a NULL.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
TimothyMothra: done

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
```diff
@@ -0,0 +1,244 @@
...
+        /// Gets an Azure.Core.AccessToken.
+        /// </summary>
+        /// <param name="cancellationToken">The System.Threading.CancellationToken to use.</param>
+        /// <returns>A valid Azure.Core.AccessToken.</returns>
+        public async Task<string> GetTokenAsync(CancellationToken cancellationToken = default)
+        {
+            SdkInternalOperationsMonitor.Enter();
+            try
+            {
+                return await AzureCore.InvokeGetTokenAsync(this.tokenCredential, this.tokenRequestContext, cancellationToken).ConfigureAwait(true);
```
TimothyMothra, line 79: @cijothomas , Is `ConfigureAwait(true)` the correct usage for `SdkInternalOperationsMonitor` ?

TimothyMothra: I think I answered my own question.
`ProfileServiceWrapper` calls Async methods inside an InternalOperation.
https://github.com/microsoft/ApplicationInsights-dotnet/blob/e1ff87e5e96daf4517ee9c28415f4024ec29529f/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/ApplicationId/ProfileServiceWrapper.cs#L86-L107

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelopeTests.cs
```diff
@@ -0,0 +1,210 @@
...
+    /// </remarks>
+    [TestClass]
+    [TestCategory("AAD")]
+    public class ReflectionCredentialEnvelopeTests
+    {
+        [TestMethod]
+        public void VerifyCanIdentifyValidClass()
+        {
+            var testClass2 = new TestClass2();
+            _ = new ReflectionCredentialEnvelope(testClass2);
```
cijothomas, line 34: there are no asserts? Is this test validating that no exception is throw?

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -399,6 +405,12 @@ public void Dispose()
             GC.SuppressFinalize(this);
         }
 
+        /// <summary>
+        /// Set a TokenCredential for this configuration.
+        /// </summary>
+        /// <param name="tokenCredential">An instance of Azure.Core.TokenCredential.</param>
+        public void SetCredential(object tokenCredential) => this.CredentialEnvelope = new ReflectionCredentialEnvelope(tokenCredential);
```
cijothomas, line 412: as a follow up, we could modify the method comment to be more descriptive, and can warn that this throws exception when object is not of type Azure.Core.TokenCredential.

cijothomas: an alternate name could be - SetTokenCredential or SetAzureTokenCredential, to make it more clear that this is an azure only thing.

cijothomas (approved):

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
```diff
@@ -0,0 +1,244 @@
...
+        public object Credential => this.tokenCredential;
+
+        /// <summary>
+        /// Gets an Azure.Core.AccessToken.
+        /// </summary>
+        /// <param name="cancellationToken">The System.Threading.CancellationToken to use.</param>
+        /// <returns>A valid Azure.Core.AccessToken.</returns>
+        public string GetToken(CancellationToken cancellationToken = default)
+        {
+            SdkInternalOperationsMonitor.Enter();
```
cijothomas, line 53: it may be better to leave the job of entering/exiting internal operation to the consumer (like Transmission class)

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelopeTests.cs
TimothyMothra: Good catch. yeah, this test is verifying that the reflection-based type checking is working.
I'll add a comment here to make this more obvious.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
TimothyMothra: Will investigate this in second PR.

PR #2200, opened 2021-04-02
Read Azure SDK Success status

pakrym:
Azure SDKs have more involved rules to determine whether the HTTP span  is successful (see https://github.com/Azure/azure-sdk-for-net/pull/20064)

This change delegates the decision to the activity producer instead of being hard coded in the listener logic.


pakrym:
cc @TimothyMothra

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
```diff
@@ -270,6 +271,10 @@ private static void SetHttpProperties(Activity activity, DependencyTelemetry dep
                 {
                     status = tag.Value;
                 }
+                else if (tag.Key == "otel.status_code")
```
rajkumar-rangaraj, line 274: If customer use older version of Azure SDK with latest application insights SDK (with this change). Does it break backward compatibility as `otel.status_code` does not exist with older version of Azure SDK?

pakrym: Yes, I'll think about how can we make this backward compatible.

pakrym: Done, we only ignore the default calculation if otel.status_code is available.

cijothomas (approved):

Issue #1743, opened 2020-03-24
Flush Enhancements


rajkumar-rangaraj:
## Issues with flush

Telemetry items are not send immediately, items are batched and flushed by SDK at fixed intervals (default is 30 secs) or whenever buffer is full (default is 500 items). If an application exits right after calling `Track` methods, telemetry may not be sent. To avoid telemetry loss during application shutdown, recommendation is to call `Flush` before app exit and introduce some delay for the `Flush` call to complete. The exact amount of delay isn't predictable and depends on factors like how many items or Transmission instances are in memory, how many are on disk, how many are being transmitted to the back end, and whether the channel is in the middle of exponential back-off scenarios. `Flush` method doesn't return success or failure. `Flush` functionality differs with two built-in channels and following is the difference. 

Sl.No | InMemoryChannel | ServerTelemetryChannel |
| ------------- | ------------- | ------------- |
| 1. | Flush method is synchronous  | Flush method is **not** synchronous |
| 2. | Flush returns once telemetry is sent  | Flush returns once telemetry is queued to Sender/Buffer/Storage queue  |
| 3. | OnAppExit: **No sleep/delay** is required after calling Flush  | 3.	OnAppExit: **Delay is required** after Flush so that telemetries in the queue can be sent by background tasks  |

>**Note:** Ideally, `Flush` method should be used in the shutdown activity of an application.

**Related Issues:** #407 #482 

#### InMemoryChannel
```
// Onboarded with https://www.nuget.org/packages/Microsoft.ApplicationInsights/
class Program
{
    static void Main(string[] args)
    {
        TelemetryConfiguration configuration = TelemetryConfiguration.CreateDefault();
        configuration.InstrumentationKey = "IKEY";
        var telemetryClient = new TelemetryClient(configuration);
        telemetryClient.TrackTrace("Hello World!");

        // before exit, flush the remaining data
        telemetryClient.Flush();
    }
}
```

#### ServerTelemetryChannel
```
// Onboarded with https://www.nuget.org/packages/Microsoft.ApplicationInsights.WindowsServer/
class Program
{
    static void Main(string[] args)
    {
        TelemetryConfiguration configuration = TelemetryConfiguration.CreateDefault();
        configuration.InstrumentationKey = "IKEY";
        var telemetryClient = new TelemetryClient(configuration);
        telemetryClient.TrackTrace("Hello World!");

        // before exit, flush the remaining data
        telemetryClient.Flush();

[...]

PR #2202, opened 2021-04-06
Task Based Flush API - FlushAsync - New Implementation

rajkumar-rangaraj:
Addressing #1743 

## FlushAsync

`FlushAsync` is a task based flush API for sending telemetry to application insights. 

### API
```
/// <summary>
/// Represents a task based flush for sending telemetry to application insights.
/// </summary>
public interface IAsyncFlushable : IDisposable
{
               /// <summary>
               /// Flushes the in-memory buffer asynchronously.
               /// </summary>
               /// <param name="cancellationToken">CancellationToken.</param>
               /// <returns>
               /// Returns true when telemetry data is transferred out of process (application insights server or local storage) and are emitted before the flush invocation.
               /// Returns false when transfer of telemetry data to server has failed with non-retriable http status code.
               /// </returns>
               Task<bool> FlushAsync(CancellationToken cancellationToken);
}
```

### Sample

```
class Program
{
    static void Main(string[] args)
    {
        TelemetryConfiguration configuration = TelemetryConfiguration.CreateDefault();
        configuration.InstrumentationKey = "IKEY";
        var telemetryClient = new TelemetryClient(configuration);
        telemetryClient.TrackTrace("Hello World!");

        // before exit, flush the remaining data
        var task = telemetryClient.FlushAsync(CancellationToken.None);

        try
        {
             await task;
        }
        catch(Exception ex)
        {
             throw ex;
        }
    }
}
```
### What is a transmission?

Telemetry items are batched at fixed interval (default is 30 secs) or whenever buffer is full (default is 500 items) or through manual flush. Batched items are Json serialized, compressed and converted into transmissions. These transmissions are sent to server. 

### Features
- Waits until all transmissions emitted before the flush invocation to complete
- Moves transmissions from buffer to disk
- On exception or throttle, moves transmission to disk
- Allow caller to decide when to cancel the method through `CancellationToken`

> :warning: &nbsp; Ideally, `Flush` or `FlushAsync` method should be used in the shutdown activity of an application.

## Scenarios
|Scenario| Functionality|
| ------------- | ------------- |
| Successful Transmission | <ul><li>Waits for all transmissions emitted before the flush invocation to complete</li><li>Moves items from local process buffer to disk (StorageFolder) on successful transmission</li><li>Marks `FlushAsync` task as success</li></ul>|
| Partial Transmission | <ul><li>Waits for all transmissions emitted before the flush invocation to complete</li><li> Creates new transmission from failed telemetry items within current transmission and moves it to disk</li><li> Moves local process buffer to disk</li><li>Marks `FlushAsync` task as success</li></ul>|
| Throttle | <ul><li> Transmission gets split into accepted and rejected transmission</li><li> Moves rejected transmission and local process buffer to disk</li><li> Waits for accepted transmission and all transmissions emitted before the flush invocation to complete</li><li>Marks `FlushAsync` task as success</li></ul>|
| Retriable Exception | <ul><li>Waits for all transmissions emitted before the flush invocation to complete</li><li> Moves current transmission and from local process buffer to disk</li><li>Marks `FlushAsync` task as success</li></ul>|
| Unknown Exception / Non-Whitelisted http status code | <ul><li>Waits for all transmissions emitted before the flush invocation to complete</li><li>Fails `FlushAsync` by sending return value as false</li></ul>|
| Storage issue | <ul><li>Fails `FlushAsync` by sending return value as false</li></ul>|

## Implementation

| InMemoryChannel | ServerTelemetryChannel |
| ------------- | ------------- |
| This channel offers minimal reliability guarantees and doesn't retry sending telemetry after a failure, hence `FlushAsync` method always returns true. [Learn More - Built-in telemetry channels]( https://docs.microsoft.com/en-us/azure/azure-monitor/app/telemetry-channels#built-in-telemetry-channels)   | A more advanced channel that has retry policies and the capability to store data on a local disk. Return true when telemetry data is transferred out of process (application insights server or local storage). Above explained scenarios are applicable for to this channel. |

- Both `ServerTelemetryChannel` and `InMemoryChannel` implements `IAsyncFlushable` and has implementation for `FlushAsync` API
- `TelemetrySerializer` serializes telemetries on a new task and enqueues to `Transmitter`
- `Transmission` now includes `TaskCompletionSource`. On `FlushAsync` call, `Task` is created from `TaskCompletionSource`. This `Task` reference is passed back to the caller of `FlushAsync` to await for completion. 
- Policies are updated to complete the `Transmission’s` `TaskCompletionSource` with success or failure status

## Checklist

### Notes for authors:

### Notes for reviewers:


BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -21,9 +22,17 @@ public class Transmission
...
         private static HttpClient client = new HttpClient() { Timeout = System.Threading.Timeout.InfiniteTimeSpan };
-                
+        private static long flushAsyncCounter = 0;
+
         private int isSending;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Transmission"/> class. This overload seperates telemetryitems serialization from object construction. 
+        /// </summary>
+        public Transmission()
```
reyang, line 32: If the goal is to allow other packages (e.g. telemetry channel implementations) to access this ctor, a safer approach would be https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute.

reyang (approved):
LGTM with a non-blocking suggestion.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
rajkumar-rangaraj: @cijothomas, We had offline discussion about this topic and decided not to use internals visible. Introducing internals visible will limit our public API changes, should we reconsider  `InternalsVisibleToAttribute`?

cijothomas: We can make this new fields internal, and use InternalsVisible approach now given that its intended to be accessible from ServerTelemetryChannel now.
If a need arises in the future, we can make it public, so other channels may leverage this.

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -676,6 +678,38 @@ public void Flush()
...
+        /// <returns>
+        /// Returns true when telemetry data is transferred out of process (application insights server or local storage) and are emitted before the flush invocation.
+        /// Returns false when transfer of telemetry data to server has failed with non-retriable http status.
+        /// FlushAsync on InMemoryChannel always returns true, as the channel offers minimal reliability guarantees and doesn't retry sending telemetry after a failure.
+        /// </returns>
+        public Task<bool> FlushAsync(CancellationToken cancellationToken)
+        {
+            if (this.TryGetMetricManager(out MetricManager privateMetricManager))
+            {
+                privateMetricManager.Flush(flushDownstreamPipeline: false);
```
cijothomas, line 693: shouldn't we pass the cancellationToken to metricmanager.Flush() as well?

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -139,6 +141,32 @@ public ICollection<ITelemetry> TelemetryItems
             get; private set;
         }
 
+        /// <summary>
+        /// Gets the flush async id for the transmission.
+        /// </summary>
+        /// [EditorBrowsable(EditorBrowsableState.Never)]
```
TimothyMothra, line 147: was this meant to be non-browsable?

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -139,6 +141,32 @@ public ICollection<ITelemetry> TelemetryItems
...
+        public long FlushAsyncId { get; } = Interlocked.Increment(ref flushAsyncCounter);
+
+        /// <summary>
+        /// Gets or sets a value indicating whether this transmission is created from FlushAsync API call.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public bool HasFlushTask { get; set; } = false;
```
cijothomas, line 154: nit:
The name doesn't match the summary.. `IsCreatedFromFlush` matches the summary.

(Or is the summary actually incorrect?)

BASE/src/ServerTelemetryChannel/Implementation/TelemetryBuffer.cs
```diff
@@ -205,6 +208,26 @@ IEnumerator IEnumerable.GetEnumerator()
             return this.GetEnumerator();
         }
 
+        private List<ITelemetry> MoveTelemeteryBufferForFlush()
```
cijothomas, line 211:
```suggestion
        private List<ITelemetry> GetBufferTelemetryAndResetBuffer()
```

nit.

BASE/src/ServerTelemetryChannel/Implementation/TelemetrySerializer.cs
```diff
@@ -48,9 +50,35 @@ public virtual void Serialize(ICollection<ITelemetry> items)
...
+            var transmission = new Transmission(this.EndpointAddress, items)
+                                    { TransmissionStatusEvent = this.TransmissionStatusEvent };
             this.Transmitter.Enqueue(transmission);
         }
+
+        public virtual Task<bool> SerializeAsync(ICollection<ITelemetry> items, CancellationToken cancellationToken)
+        {
+            if (this.EndpointAddress == null)
+            {
+                throw new Exception("TelemetrySerializer.EndpointAddress was not set.");
```
cijothomas, line 62: would this get propagated to customer FlushAsync or we handle it somewhere before throwing up to user?

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -139,6 +141,32 @@ public ICollection<ITelemetry> TelemetryItems
...
+
+        /// <summary>
+        /// Serializes telemetry items.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public void Serialize(Uri address, IEnumerable<ITelemetry> telemetryItems, TimeSpan timeout = default(TimeSpan))
```
cijothomas, line 160: minor point. I could be wrong as well.

feels a bit awkward for a instance method Serialize accept a list of items to be serialized...
Could we store the items in the ctor() itself, and Serialize() call simply serializes the items it already has?

BASE/src/ServerTelemetryChannel/TelemetryChannel.csproj
```diff
@@ -44,5 +44,9 @@
     <PackageReference Include="System.IO.FileSystem.AccessControl" Version="4.7.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="..\Microsoft.ApplicationInsights\Channel\TaskEx.cs" Link="TaskEx.cs" />
```
TimothyMothra, line 48: please add this to the Common project instead. Line 51.

CHANGELOG.md
```diff
@@ -68,6 +68,7 @@
 - [Fix: Disabling HeartBeats in Asp.Net Core projects causes Error traces every heart beat interval (15 minutes defualt)](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1681)
 - [Standard Metric extractor (Exception,Trace) populates all standard dimensions.](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1738)
 - [Add an explicit reference to System.Memory v4.5.4. This fixes a bug in System.Diagnostics.DiagnosticSource. We will remove this dependency when DiagnosticSource is re-released.](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1707)
+- [New Task Based Flush API - FlushAsync] (https://github.com/microsoft/ApplicationInsights-dotnet/issues/1743)
```
TimothyMothra, line 71: this needs to be under the VNext section

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
```diff
@@ -169,10 +173,84 @@ internal virtual void Enqueue(Transmission transmission)
 
             if (!this.Storage.Enqueue(transmissionGetter))
             {
+                transmission.HasFlushTask = false;
```
cijothomas, line 176: Trying to understand why does this has to be set to false here... Transmitter receives a transmission..How can it determine if HasFlushTask should be true or not, as HasFlushTask depends on if it was called from Flush..

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
```diff
@@ -169,10 +173,84 @@ internal virtual void Enqueue(Transmission transmission)
...
+            var senderStatus = this.Sender.WaitForPreviousTransmissionsToComplete(cancellationToken).ConfigureAwait(false).GetAwaiter().GetResult();
+            if (senderStatus == TaskStatus.Canceled)
+            {
+                return TaskEx.FromCanceled<bool>(cancellationToken);
+            }
+
+            return senderStatus == TaskStatus.RanToCompletion && isStorageEnqueueSuccess ? this.successTask : this.failedTask; 
+        }
+
+        internal TaskStatus MoveTransmissionsAndWaitForSender(long transmissionFlushAsyncId, CancellationToken cancellationToken)
```
cijothomas, line 240: these 2 methods can be merged/made to call other, to avoid copy-pasting.

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
```diff
@@ -169,10 +173,84 @@ internal virtual void Enqueue(Transmission transmission)
...
+        }
+
+        internal Task<bool> MoveTransmissionsAndWaitForSender(CancellationToken cancellationToken)
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return TaskEx.FromCanceled<bool>(cancellationToken);
+            }
+
+            var isStorageEnqueueSuccess = MoveTransmissions(this.Buffer.Dequeue, this.Storage.Enqueue, this.Buffer.Size);
```
cijothomas, line 229: Not sure where do we draw the line.. but this method is also a potentially long running (if high items and slow disk..), So should we pass cancellationtoken to it as well?

BASE/src/ServerTelemetryChannel/Implementation/TransmissionSender.cs
```diff
@@ -134,6 +140,41 @@ public virtual bool Enqueue(Func<Transmission> transmissionGetter)
             return enqueueSucceded;
         }
 
+        /// <summary>
+        /// Wait for inflight transmissions to complete. Honors the CancellationToken set by an application on IAsyncFlushable.FlushAsync.
+        /// </summary>
+        internal Task<TaskStatus> WaitForPreviousTransmissionsToComplete(CancellationToken cancellationToken)
+        {
+            var transmissionFlushAsyncId = this.inFlightTransmissions.LastOrDefault().Key;
```
cijothomas, line 148: what exactly does LastOrDefault does here? Gets the transmission with highest id? or the one which was added most recent?

BASE/src/ServerTelemetryChannel/Implementation/TransmissionSender.cs
```diff
@@ -155,7 +197,9 @@ private async Task StartSending(Transmission transmission)
                 try
                 {
                     TelemetryChannelEventSource.Log.TransmissionSendStarted(acceptedTransmission.Id);
-                    responseContent = await acceptedTransmission.SendAsync().ConfigureAwait(false);
+                    transmissionTask = acceptedTransmission.SendAsync();
+                    this.inFlightTransmissions.TryAdd(transmission.FlushAsyncId, transmissionTask);
```
cijothomas, line 201: adding to the dictionary can be done *before* called SendAsync() to ensure its always added to dictionary, irrespective of any exceptions.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
rajkumar-rangaraj: Yes. We are marking this properties/methods as internal. I will get this attribute removed.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
rajkumar-rangaraj: This is tricky, as we already have a constructor which takes the same number of arguments and serializes it.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
TimothyMothra: Can we make the InternalsVisible change as a separate PR, please?

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
rajkumar-rangaraj: If a move to storage fails, we set this flag. `HasFlushTask` will be honored only in case of FlushAsync call.
`Transmitter.FlushAsync` calls `Enqueue` to follow the normal process. During that attempt if a transmission object fails to move to storage that needs to be tracked hence `HasFlushTask` is used as a marker.

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
rajkumar-rangaraj: Return type of both of these methods are different and they call `WaitForPreviousTransmissionsToComplete' method with different parameters. 

Reason for me to split these functions are to avoid handling TaskCanceledException in the `Transmitter.FlushAsync` call. Felt handling exception is costlier here.

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
rajkumar-rangaraj: We have to check the if cancellationtoken within the loop, this will impact performance as everytime we loop we need to check the status of cancellationtoken. If we are fine with this overhead I will add cancellationtoken check.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionSender.cs
rajkumar-rangaraj: The one that got added most recent.
Most of the times, recent transmission should be the one with highest id.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionSender.cs
rajkumar-rangaraj: `transmissionTask` is generated by `SendAsync()`. Only awaiting on a task will throw an exception. Before awaiting a task, an entry is added to dictionary. So irrespective of exception it gets added to dictionary.

BASE/src/Microsoft.ApplicationInsights/Channel/TaskEx.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace Microsoft.ApplicationInsights.Channel
```
TimothyMothra, line 1: I see that you want to use this in multiple projects.
I recommend moving this to the "Common.Base" project at "BASE\src\Common\Common\".
You could do this as a separate PR.

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
rajkumar-rangaraj: This requires a change to metrics flush API, should we explore it in a different PR?

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
cijothomas: If its part of this PR, it'll simply the .public API file in this PR a lot.

rajkumar-rangaraj: @TimothyMothra I agree with you. This has to be done separately. I will send a new PR and get that merged before this one.

BASE/src/ServerTelemetryChannel/Implementation/TelemetrySerializer.cs
rajkumar-rangaraj: This gets propagated to customers FlushAsync. I had this test case to cover it.
```
            [TestMethod]
            public void ThrowsExceptionWhenEndpointAddressIsNull()
            {
                var serializer = new TelemetrySerializer(new StubTransmitter());
                AssertEx.Throws<Exception>(() => serializer.SerializeAsync(null, default));
            }
```
Now adding one more in TelemetryBufferTests to avoid confusion.

BASE/src/Microsoft.ApplicationInsights/Channel/TaskEx.cs
rajkumar-rangaraj: We could completely remove common project with internals visible, working this on a separate PR.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
rajkumar-rangaraj: Apt name would be `IsFlushAsyncInProgress`. Changed both summary and property name.

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
rajkumar-rangaraj: As per our offline discussion, code has to check IsCancellationRequested in every logical step. Hence added cancelationtoken check.

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
cijothomas: please put a todo marker here.

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -676,6 +678,38 @@ public void Flush()
...
+            if (pipeline != null)
+            {
+                MetricManager sharedMetricManager = pipeline.GetMetricManager(createIfNotExists: false);
+                sharedMetricManager?.Flush(flushDownstreamPipeline: false);
+
+                ITelemetryChannel channel = pipeline.TelemetryChannel;
+
+                if (channel is IAsyncFlushable asyncFlushableChannel && !cancellationToken.IsCancellationRequested)
+                {
+                    return asyncFlushableChannel.FlushAsync(cancellationToken);
```
cijothomas, line 706: Not really part of this PR, but we actually need to flush channels in *all* the sinks, as opposed to the default sink.
Please mark a TODO and we'll explore it after.

BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs
```diff
@@ -186,6 +188,25 @@ public void Flush(TimeSpan timeout)
...
 
+        /// <summary>
+        /// Will send all the telemetry items stored in the memory asynchronously.
+        /// </summary>
+        /// <param name="cancellationToken">CancellationToken.</param>
+        /// <returns>
+        /// This method is hard-coded to return true. This channel offers minimal reliability guarantees and doesn't retry sending telemetry after a failure.
+        /// </returns>
+        /// <remarks>
+        /// <a href="https://docs.microsoft.com/en-us/azure/azure-monitor/app/telemetry-channels#built-in-telemetry-channels">Learn more</a>
```
cijothomas, line 199:
```suggestion
        /// <a href="https://docs.microsoft.com/azure/azure-monitor/app/telemetry-channels#built-in-telemetry-channels">Learn more</a>
```

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -358,6 +370,19 @@ public virtual async Task<HttpWebResponseWrapper> SendAsync()
             return Tuple.Create(transmissionA, transmissionB);
         }
 
+        /// <summary>
+        /// Serializes telemetry items.
+        /// </summary>
+        internal void Serialize(Uri address, IEnumerable<ITelemetry> telemetryItems, TimeSpan timeout = default(TimeSpan))
+        {
+            this.EndpointAddress = address;
```
cijothomas, line 378: might be good to refactor (in a separate PR) this method, as this is doing more than serialization. (all the validations which were previously in ctor() is now done inside this method..)

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
```diff
@@ -169,10 +173,84 @@ internal virtual void Enqueue(Transmission transmission)
...
+        {
+            TaskStatus taskStatus = TaskStatus.Canceled;
+            if (!cancellationToken.IsCancellationRequested)
+            {
+                transmission.IsFlushAsyncInProgress = true;
+                this.Enqueue(transmission);
+
+                try
+                {
+                    this.Storage.IncrementFlushAsyncCounter();
```
cijothomas, line 191: i think this is done to prevent items being moved from Storage to Buffer/Sender, while FlushAsync is in progress... If thats the case, then this must also be done in the `MoveTransmissionsAndWaitForSender(cancellationToken)` path as well.

cijothomas (approved):
left few comments. We can address them in follow ups.
Also - please move usage examples to this PR's description.

BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs
```diff
@@ -169,10 +173,84 @@ internal virtual void Enqueue(Transmission transmission)
...
             }
         }
 
+        internal Task<bool> FlushAsync(Transmission transmission, CancellationToken cancellationToken)
+        {
+            TaskStatus taskStatus = TaskStatus.Canceled;
+            if (!cancellationToken.IsCancellationRequested)
+            {
+                transmission.IsFlushAsyncInProgress = true;
+                this.Enqueue(transmission);
```
cijothomas, line 187: There is a possibility that sender and buffer and storage are all full and data is dropped inside this method. We are blind to that possibility, and won't return false to end user in this case..

cijothomas: we can modify Enqueue to return its success status, and act acordingly. (in a future PR)

rajkumar-rangaraj: This use case is handled, if sender, buffer and storage are full IsFlushAsyncInProgress is set to false in Enqueue on line 176 `transmission.IsFlushAsyncInProgress = false;`

Line 210, in method FlushAsync there is a check to handle it.

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
rajkumar-rangaraj: Added a TODO in summary.

PR #2220, opened 2021-04-09
[POC] AAD: Configuring ServerTelemetryChannel and QuickPulseServiceClient

TimothyMothra:
# Consuming CredentialEnvelope within TelemetryChannels and QuickPulseServiceClient

For more information see the parent item: #2190.

## Introduction
This PR demonstrates an approach to handling the configuration of the CredentialEnvelope.

**TODO: Need to write unit tests to validate possible configuration scenarios.**

## TelemetryConfiguration
`TelemetryConfiguration` has a method `SetAzureTokenCredential` for a user to provide the TokenCredential.
This method could be called either before or after the user has set the `TelemetryChannel`.

To deal with this, both setting a credential and setting a TelemetryChannel will invoke `TelemetryConfiguration.SetTelemetryChannelCredentialEnvelope`.

## ISupportCredentialEnvelope
To simplify passing the CredentialEnvelope to the TelemetryChannels, I've introduced a new interface `ISupportCredentialEnvelope`.
This ensures that the `TelemetryConfiguration` can set either `InMemoryChannel` or `ServerTelemetryChannel`.

## TelemetryChannels

Each TelemetryChannel must provide the CredentialEnvelope on their respective internal classes.
TelemetryItems ultimately end up stored in a `Transmission` object. This object can be serialized for file storage. I've confirmed that the Credential object will not be serialized.

Tokens themselves will expire and therefore are not set on the Transmission object until just before invoking `Transmission.SendAsync`. If a `Transmission` needs to be retried, it will receive a new token.

### InMemoryChannel
`InMemoryChannel.CredentialEnvelope` sets `InMemoryTransmitter.CredentialEnvelope` which is used to set `Transmission.CredentialEnvelope` just before calling `Transmission.SendAsync`.

### ServerTelemetryChannel
`ServerTelemetryChannel.CredentialEnvelope` sets `Transmitter.CredentialEnvelope` and then sets `TransmissionSender.CredentialEnvelope` which is used to set `Transmission.CredentialEnvelope` just before calling `Transmission.SendAsync`.


WEB/Src/PerformanceCollector/Perf.Shared.NetFull/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -86,6 +86,7 @@ public Uri CurrentServiceUri
             DateTimeOffset timestamp,
             string configurationETag,
             string authApiKey,
+            string aadToken,
```
pharring, line 89: Just a suggestion: Call it "authToken" throughout.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/TokenCredentialEnvelope.cs
```diff
@@ -24,15 +24,31 @@ public TokenCredentialEnvelope(TokenCredential tokenCredential)
         /// <inheritdoc/>
         public override string GetToken(CancellationToken cancellationToken = default)
         {
-            var accessToken = this.tokenCredential.GetToken(requestContext: this.tokenRequestContext, cancellationToken: cancellationToken);
-            return accessToken.Token;
+            SdkInternalOperationsMonitor.Enter();
+            try
+            {
+                var accessToken = this.tokenCredential.GetToken(requestContext: this.tokenRequestContext, cancellationToken: cancellationToken);
```
pharring, line 30: We should cache the accessToken in a field because they are expensive to create. They are usually valid for 24 hours.
Assume that clients are going to use the token right away (within a few minutes).
Use the expiration time (minus 5 minutes or so) on the token to decide when to invalidate the cache.

TimothyMothra: The expectation is that the implementing Credential class would handle local caching. 
I explicitly asked this question to the Identity architect and I was told NOT to cache locally.

TimothyMothra: Are you aware of any published best-practices in this space?

pharring: I've been over the source code for ManagedIdentityCredential and there's no caching of the access token that I can see. In Azure.Core, AccessToken caching happens in a different layer -- in the BearerTokenAuthenticationPolicy that is added to an HttpPipeline. We're not using HttpPipeline here, so this code can't use that.

I imagine that the AccessTokenCache in BearerTokenAuthenticationPolicy is the best practice. It's here: https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs#L176

I find it pretty hard to read that source code because of the TaskCompletionSource and background refresh. From what I gather: First time, obviously you have to wait to call GetToken. Cache the returned value. Compute a RefreshOn time by subtracting 5 minutes from the ExpiresOn value. On subsequent calls, if DateTime.UtcNow is after ExpiresOn, obviously you need to block and get a new token. If it's close to expiry (after RefreshOn), kick off a background task to refresh it, but try and use the old one. Maybe your request will fail if you're too slow, but on the next retry, you'll block for the background refresh to complete.

Having said that, I would exercise caution adding caching directly on TokenCredentialEnvelope. Add it on the caller side. The main reason I say that is that the Linux Profiler agent is going to be accessing the TokenCredentialEnvelope too. It will call GetToken to get a token to call into the profiler endpoint. Because of the agent architecture, we need that token to be fairly long-lived (~15 minutes).

BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs
```diff
@@ -404,6 +411,19 @@ protected virtual HttpRequestMessage CreateRequestMessage(Uri address, Stream co
...
             }
 
+            if (this.CredentialEnvelope != null)
+            {
+                // TODO: NEED TO USE CACHING
+                var aadToken = this.CredentialEnvelope.GetToken();
+
+                if (aadToken == null)
+                {
+                    // TODO: DO NOT SEND. RETURN FAILURE AND LET CHANNEL DECIDE WHEN TO RETRY.
```
cijothomas, line 421: this will be considered a client side failure, and data would not be retried by default.
Its probably a complex part to decide if we want to store items for later retry, if the AAD is down..

WEB/Src/PerformanceCollector/Perf.Shared.NetFull/Implementation/QuickPulse/QuickPulseServiceClient.cs
```diff
@@ -381,6 +386,12 @@ private void AddHeaders(HttpWebRequest request, bool includeIdentityHeaders, str
                 request.Headers.Add(QuickPulseConstants.XMsQpsInvariantVersionHeaderName,
                     MonitoringDataPoint.CurrentInvariantVersion.ToString(CultureInfo.InvariantCulture));
             }
+
+            // The AAD token is an optional feature. Only include if it is provided.
+            if (!string.IsNullOrEmpty(aadToken))
```
cijothomas, line 391: is this intentionally optional for QP? elsewhere, in the regular channel, its like mandatory, if user has set tc.settokencred.

TimothyMothra: oh good catch. i need to rethink this

BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs
```diff
@@ -96,6 +95,20 @@ public TimeSpan SendingInterval
             }
         }
 
+        /// <summary>
+        /// Gets or sets the <see cref="CredentialEnvelope"/> which is used for AAD.
+        /// FOR INTERNAL USE. Customers should use <see cref="TelemetryConfiguration.SetAzureTokenCredential"/> instead.
```
pharring, line 100: We don't really need this comment anymore since ISupportCredentialEnvelope is, itself, internal.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionSender.cs
```diff
@@ -107,6 +108,15 @@ public virtual int ThrottleWindow
             }
         }
 
+        /// <summary>
+        /// Gets or sets the <see cref="CredentialEnvelope"/> which is used for AAD.
```
pharring, line 112: Instead of "AAD", suggest "authorization" or "auth-token acquisition".

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ISupportCredentialEnvelope.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Authentication
```
pharring, line 1: I know it's pedantic, but hear me out:
Authentication is what the server (Breeze) does, to check the token and decide what you're allowed to do.
Authorization is what the client (SDK) does to prove who they are.
So, I think this namespace should be `Authorization`.
It's a common-enough confusion that you'll see inconsistencies in a number of places. In some cases it's ambiguous (is the token you get an "authorization token" or an "authentication token"?), so you see people side-step and call it "auth" or "access".

PR #2233, opened 2021-04-15
examples sln

TimothyMothra:
I created a new solution: Examples.sln.

This solution will build all of our SDKS (without tests and without nuget packages).
Example project will serve as code examples of how we expect our SDK to be configured.
These example projects are ready to be configured and can be used to investigate issues in the SDK.
Because these use project references, we can use these to debug specific features within the SDK.

### DIFF
--- a/.props/Product.props
+++ b/.props/Product.props
@@ -9,7 +9,7 @@
   <Import Project=".\_AnalyzerSettings.props" />
 
   <Import Project=".\_GlobalStaticVersion.props" />
-  <Import Project=".\_Nupkg.props" />
+  <Import Project=".\_Nupkg.props"  Condition="'$(IsExamplesSolution)' != 'true'"/>
 
   <ItemGroup Condition=" $(OS) == 'Windows_NT'">
     <!--Analyzers-->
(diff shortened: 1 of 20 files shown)


.props/Product.props
```diff
@@ -9,7 +9,7 @@
   <Import Project=".\_AnalyzerSettings.props" />
 
   <Import Project=".\_GlobalStaticVersion.props" />
-  <Import Project=".\_Nupkg.props" />
+  <Import Project=".\_Nupkg.props"  Condition="'$(IsExamplesSolution)' != 'true'"/>
```
TimothyMothra, line 12: Excluding the Nupkg.props in the Examples solution will block our SDKs from building nuget packages. this will speed  up compilation when working with Examples

Everything.sln
```diff
@@ -171,7 +169,6 @@ Global
 		LOGGING\test\CommonTestShared\CommonTestShared.projitems*{7a903abd-d7fb-4610-aed8-32eb2cbaeebf}*SharedItemsImports = 5
 		LOGGING\test\Shared\Adapters.Shared.Tests.projitems*{7a903abd-d7fb-4610-aed8-32eb2cbaeebf}*SharedItemsImports = 5
 		WEB\Src\Common\Common.projitems*{7b5d95ee-50ee-4222-a03c-fae5905b3dfd}*SharedItemsImports = 5
-		BASE\src\Common\Common\Common.projitems*{936af739-4297-4016-9d70-4280042709be}*SharedItemsImports = 13
 		LOGGING\test\CommonTestShared\CommonTestShared.projitems*{93a35062-6aa4-4778-9769-428a942adcf9}*SharedItemsImports = 5
```
TimothyMothra, line 172: This shared project was recently deleted but the reference was still in this solution.

rajkumar-rangaraj (approved):

Issue #2238, opened 2021-04-20
Improve Self Diagnostics and support setting configuration in file


xiang17:
The motivation for this change is to improve troubleshooting the SDK as another attempt for #1981 in a different approach.

I'm proposing an approach for a self diagnostics module that is enabled by modifying a configuration file, rather than an environment variable. For example, the log starts flowing into a file after the SysAdmin or DevOps sets up a configuration in a certain location with contents like this:

```
{
    "LogDirectory": ".",
    "FileSize": 1024,
    "LogLevel": "Error"
}
```

There is no need of redeployment/restart to enable and disable the logs. In comparison, the environment variables are set for a process when it starts and are not updated. There are workarounds([on Windows](https://stackoverflow.com/questions/38545262/how-to-read-out-new-os-environment-variables-in-python)), but they will be OS-specific.

### Self Diagnostics Module Design Principles

1. The self diagnostics module captures and write log messages to disk, when enabled.
2. It should have minimal overhead to the current process so the user can feel confident to enable it.
3. It should be accessible to use in most environments/cases. For example, we want to use the least dependencies and/or features of the OS. It doesn't require administrator permissions. It can be enabled/disabled without restarting the process.
4. Bounded resource usage. For example, it should not consume ever-growing memory or generate an ever-growing file.

## Implementation Plan

1. Self Diagnostics Module listens to the events with EventListener.
2. Self Diagnostics Module writes the message in the event to a fix-sized local file at a certain location in a circular way. (Overwrite from beginning once write position reaches the end.)
3. Self Diagnostics Module checks a configuration file for log file location and file size periodically (for example, every 10 seconds).
4. By default, the configuration doesn't exist, Self Diagnostics Module will not write to any file. Once the file is created with valid configuration format/values, Self Diagnostics Module will start writing log file according to configuration. After the file is deleted or modified, the log file handle will be updated, either closed (if deleted or modified with invalid configuration) or closed and reopened according to the configuration file modification (if modified with valid configuration).
5. To optimize write performance, Self Diagnostics Module will utilize [Memory-Mapped Files](https://docs.microsoft.com/en-us/dotnet/standard/io/memory-mapped-files) feature instead of directly writing to file.

I'm targeting to finish the feature before the [2.18 milestone](https://github.com/microsoft/ApplicationInsights-dotnet/milestone/46)

TimothyMothra:
I have some feedback on this:
[...]

PR #2245, opened 2021-04-29
Initialization code and entry point for Self Diagnostics Module

xiang17:
Fix Issue #2238.

## Changes
Add the initialization code and entry point for Self Diagnostics Module

### Checklist
- [ ] CHANGELOG.md updated with one line description of the fix, and a link to the original issue if available: N/A. This PR is a partial implementation. More PRs to come before completion of the feature.

- [x] Design discussion issue #2238
- [ ] Changes in public surface reviewed: N/A

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
@@ -57,7 +57,6 @@
             try
             {
                 SdkInternalOperationsMonitor.Enter();
-                SelfDiagnosticsModule.EnsureInitialized();
 
                 if (modules != null && !modules.Modules.Any(module => module is DiagnosticsTelemetryModule))
                 {
--- a/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
@@ -57,6 +57,7 @@
             try
             {
                 SdkInternalOperationsMonitor.Enter();
+                SelfDiagnosticsModule.EnsureInitialized();
 
                 if (modules != null && !modules.Modules.Any(module => module is DiagnosticsTelemetryModule))
                 {
--- a/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnosticsModule.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnosticsModule.cs
@@ -6,7 +6,7 @@
     /// Self diagnostics class captures the EventSource events sent by Application Insights
     /// modules and writes them to local file for internal troubleshooting.
     /// </summary>
-    class SelfDiagnosticsModule : IDisposable
+    public class SelfDiagnosticsModule : IDisposable
     {
         /// <summary>
         /// Long-living object that hold relevant resources.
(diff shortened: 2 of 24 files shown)


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
```diff
@@ -57,6 +57,7 @@ public virtual void Initialize(TelemetryConfiguration configuration, TelemetryMo
             try
             {
                 SdkInternalOperationsMonitor.Enter();
+                SelfDiagnosticsModule.EnsureInitialized();
```
cijothomas, line 60: Would this be a safer place?
https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs#L63

xiang17: By "safer", do you mean it's a place it can be ensured to initialize? Or it can be ensured to initialize only when `System.Diagnostics.DiagnosticSource.dll` is available?

I went through the code. Only [`System.Diagnostics.Tracing`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing?view=net-5.0) namespace and `System.Diagnostics` namespace (for [`System.Diagnostics.Process`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process?view=net-5.0)) are used. It doesn't need `System.Diagnostics.DiagnosticSource.dll`.

I've moved it here but after the `TryRun` method because 1. it's a static method, 2. it's called in all code paths before `TelemetryConfigurationFactory.Initialize` anyways.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnosticsModule.cs
```diff
@@ -0,0 +1,61 @@
+﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing
+{
+    using System;
+
+    /// <summary>
+    /// Self diagnostics class captures the EventSource events sent by Application Insights
+    /// modules and writes them to local file for internal troubleshooting.
+    /// </summary>
+    public class SelfDiagnosticsModule : IDisposable
```
cijothomas, line 9: we dont need this to be public, right? The only way user is going to interact with the module is via the config.json file.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
cijothomas: Sorry for not being more specific - I meant to do this initialize inside the static ctor, to ensure its done at the very earliest possible stage.
(Has nothing to do with Sys.Diag package, so no need of doing inside TryRun..) - It looks like you already figured it out :)

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnosticsModule.cs
xiang17: You are right. I've made it internal and renamed it.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs
xiang17: I've updated it.

cijothomas (approved):

PR #2252, opened 2021-05-05
remove net452 and net46 from aspnetcore

TimothyMothra:
net452 and net46 were only included to support aspnet core 1.x which was deprecated months ago.


cijothomas (approved):
add changelog to say stopping support for net452,net46 targets

rajkumar-rangaraj (approved):

PR #2292, opened 2021-05-29
AAD: with ServerTelemetryChannel

TimothyMothra:
Parent issue #2190.
These changes are originally from #2220. 

## Changes
- Changes to `ServerTelemetryChannel` to support `TelemetryConfiguration.CredentialEnvelope`
   - `ServerTelemetryChannel.CredentialEnvelope` sets `Transmitter.CredentialEnvelope` and then sets `TransmissionSender.CredentialEnvelope` which is used to set `Transmission.CredentialEnvelope` just before calling `Transmission.SendAsync`.
- Add new interface: `ISupportCredentialEnvelope`

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Tests/TelemetryChannel.Tests.csproj
+++ b/BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Tests/TelemetryChannel.Tests.csproj
@@ -22,6 +22,9 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
+    <PackageReference Include="Azure.Core">
+      <Version>1.14.0</Version>
+    </PackageReference>
     <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="2.2.0" />
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.2.0" />
     <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="2.2.0" />
@@ -38,6 +41,12 @@
 
   <ItemGroup>
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+    <PackageReference Include="Azure.Core">
+      <Version>1.14.0</Version>
+    </PackageReference>
   </ItemGroup>
 
   <Import Project="..\..\..\Test\TestFramework\Shared\TestFramework.Shared.projitems" Label="TestFramework.Shared" />
--- a/BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs
@@ -1,10 +1,14 @@
 ﻿namespace Microsoft.ApplicationInsights.Channel
 {
     using System;
+    using System.ComponentModel;
     using System.Diagnostics;
     using System.Threading;
     using System.Threading.Tasks;
+
     using Microsoft.ApplicationInsights.Common;
+    using Microsoft.ApplicationInsights.Extensibility;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Authentication;
     using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
 
     /// <summary>
@@ -120,6 +124,20 @@
         {
             get { return this.buffer.BacklogSize; }
             set { this.buffer.BacklogSize = value; }
+        }
+
+        /// <summary>
+        /// Gets or sets the <see cref="CredentialEnvelope"/> which is used for AAD.
+        /// FOR INTERNAL USE. Customers should use <see cref="TelemetryConfiguration.SetAzureTokenCredential"/> instead.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="InMemoryChannel.CredentialEnvelope"/> sets <see cref="InMemoryTransmitter.CredentialEnvelope"/> 
+        /// which is used to set <see cref="Transmission.CredentialEnvelope"/> just before calling <see cref="Transmission.SendAsync"/>.
+        /// </remarks>
+        internal CredentialEnvelope CredentialEnvelope
+        {
+            get => this.transmitter.CredentialEnvelope;
+            set => this.transmitter.CredentialEnvelope = value;
         }
 
         internal bool IsDisposed => this.isDisposed;
--- a/BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs
@@ -1,7 +1,6 @@
 ﻿namespace Microsoft.ApplicationInsights.Channel
 {
     using System;
-    using System.ComponentModel;
     using System.Diagnostics;
     using System.Threading;
     using System.Threading.Tasks;
@@ -15,7 +14,7 @@
     /// Represents a communication channel for sending telemetry to Application Insights via HTTPS. There will be a buffer that will not be persisted, to enforce the 
     /// queued telemetry items to be sent, <see cref="ITelemetryChannel.Flush"/> should be called.    
     /// </summary>
-    public class InMemoryChannel : ITelemetryChannel, IAsyncFlushable
+    public class InMemoryChannel : ITelemetryChannel, IAsyncFlushable, ISupportCredentialEnvelope
     {
         private readonly TelemetryBuffer buffer;
         private readonly InMemoryTransmitter transmitter;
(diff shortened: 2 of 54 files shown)


BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Tests/TelemetryChannel.Tests.csproj
```diff
@@ -22,6 +22,9 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
```
cijothomas, line 24:
```suggestion
  <ItemGroup Condition="'$(TargetFramework)' != 'net452'">
```

cijothomas:
```suggestion
  <ItemGroup Condition="'$(TargetFramework)' == 'NETCORE'"> 
```

woudl this work!

BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs
```diff
@@ -96,6 +95,20 @@ public TimeSpan SendingInterval
...
 
+        /// <summary>
+        /// Gets or sets the <see cref="CredentialEnvelope"/> which is used for AAD.
+        /// FOR INTERNAL USE. Customers should use <see cref="TelemetryConfiguration.SetAzureTokenCredential"/> instead.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="ISupportCredentialEnvelope.CredentialEnvelope"/> on <see cref="InMemoryChannel"/> sets <see cref="InMemoryTransmitter.CredentialEnvelope"/> 
+        /// which is used to set <see cref="Transmission.CredentialEnvelope"/> just before calling <see cref="Transmission.SendAsync"/>.
+        /// </remarks>
+        CredentialEnvelope ISupportCredentialEnvelope.CredentialEnvelope
```
rajkumar-rangaraj, line 106: Access modifier is missing here. If the plan is to mark this method as internal, we could simplify the summary section removing "For internal use".

BASE/src/ServerTelemetryChannel/Implementation/TransmissionSender.cs
```diff
@@ -197,9 +208,11 @@ private async Task StartSending(Transmission transmission)
                 try
                 {
                     TelemetryChannelEventSource.Log.TransmissionSendStarted(acceptedTransmission.Id);
+                    acceptedTransmission.CredentialEnvelope = this.CredentialEnvelope;
```
rajkumar-rangaraj, line 211: A test is needed to ensure, both accepted and rejected transmission gets CredentialEnvelope before transmission is sent to breeze. Code looks good, this new test will give 100% confidence.

BASE/src/ServerTelemetryChannel/ServerTelemetryChannel.cs
```diff
@@ -243,6 +246,20 @@ public int LocalThrottleWindow
...
 
+        /// <summary>
+        /// Gets or sets the <see cref="CredentialEnvelope"/> which is used for AAD.
+        /// DO NOT SET DIRECTLY. Use <see cref="TelemetryConfiguration.SetAzureTokenCredential"/> instead.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="ISupportCredentialEnvelope.CredentialEnvelope"/> on <see cref="ServerTelemetryChannel"/> sets <see cref="Transmitter.CredentialEnvelope"/> and then sets <see cref="TransmissionSender.CredentialEnvelope"/> 
+        /// which is used to set <see cref="Transmission.CredentialEnvelope"/> just before calling <see cref="Transmission.SendAsync"/>.
+        /// </remarks>
+        CredentialEnvelope ISupportCredentialEnvelope.CredentialEnvelope
```
rajkumar-rangaraj, line 257: same comments as above, access modifier is missing. This syntax makes code difficult to read.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -339,7 +339,9 @@ public string ConnectionString
 
         /// <summary>
         /// Gets an envelope for Azure.Core.TokenCredential which provides an AAD Authenticated token.
+        /// FOR INTERNAL USE ONLY. To set the Credential use <see cref="SetAzureTokenCredential"/>.
         /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public CredentialEnvelope CredentialEnvelope { get; private set; }
```
rajkumar-rangaraj, line 345: I might have missed details for making this property as public. Is there a reason to keep this as public property?

BASE/src/ServerTelemetryChannel/ServerTelemetryChannel.cs
```diff
@@ -357,6 +374,8 @@ public void Initialize(TelemetryConfiguration configuration)
                 throw new ArgumentNullException(nameof(configuration));
             }
 
+            ((ISupportCredentialEnvelope)this).CredentialEnvelope = configuration.CredentialEnvelope;
```
rajkumar-rangaraj, line 377: can we just use `CredentialEnvelope=configuration.CredentialEnvelope;` instead of cast?

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
TimothyMothra: needed to change it to public so the QuickPulseTelemetryModule can read it when initialized.

BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Tests/TelemetryChannel.Tests.csproj
TimothyMothra: I haven't found a way to use wildcards in the csproj.
The `NETCOREAPP` key word is only available as a preprocessor. 
link: https://docs.microsoft.com/en-us/dotnet/core/tutorials/libraries#how-to-multitarget

At line 24 specifically, this wouldn't work because these dependencies are for netcoreapp 2.1 only. These aren't shared with netcoreapp 3.1.

BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryChannel.cs
TimothyMothra: So the reason this looks weird is because this is an [explicitly implemented interface](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation).
Explicitly Implemented interfaces cannot have access modifiers and can only be accessed via casting back to the interface.

The only way to avoid this is to make the new interface public.

BASE/src/ServerTelemetryChannel/ServerTelemetryChannel.cs
TimothyMothra: see my comment above

BASE/src/ServerTelemetryChannel/ServerTelemetryChannel.cs
TimothyMothra: see my comment above

BASE/src/ServerTelemetryChannel/Implementation/TransmissionSender.cs
TimothyMothra: I don't understand your comment here. What scenarios do you want to see tested?

BASE/src/ServerTelemetryChannel/ServerTelemetryChannel.cs
```diff
@@ -243,6 +246,21 @@ public int LocalThrottleWindow
...
+        /// <summary>
+        /// Gets or sets the <see cref="CredentialEnvelope"/> which is used for AAD.
+        /// FOR INTERNAL USE. Customers should use <see cref="TelemetryConfiguration.SetAzureTokenCredential"/> instead.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="ServerTelemetryChannel.CredentialEnvelope"/> on sets <see cref="Transmitter.CredentialEnvelope"/> and then sets <see cref="TransmissionSender.CredentialEnvelope"/> 
+        /// which is used to set <see cref="Transmission.CredentialEnvelope"/> just before calling <see cref="Transmission.SendAsync"/>.
+        /// </remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public CredentialEnvelope CredentialEnvelope
```
cijothomas, line 258: this needs to be public?

TimothyMothra: Either the Interface and Property are public.
Or the Interface is private, and the properties have to be explicitly implemented. 

I don't have a strong opinion on which implementation, but I need to be able to set this property in `ServerTelemetryChannel` from the `TelemetryConfiguration`

pharring: Keep the interface internal (with InternalsVisibleTo for the Microsoft.AI.ServerTelemetryChannel assembly) and use explicit interface implementation.
It's then impossible for user code to call it by mistake (without reflection).
You don't even need the EditorBrowsable attribute.
(You have in right in your private branch!)

rajkumar-rangaraj (approved):
LGTM

BASE/src/ServerTelemetryChannel/Implementation/ResponseStatusCodes.cs
```diff
@@ -11,6 +11,8 @@ internal class ResponseStatusCodes
         public const int InternalServerError = 500;
         public const int BadGateway = 502;
         public const int ServiceUnavailable = 503;
-        public const int GatewayTimeout = 504;
+        public const int GatewayTimeout = 504;
```
cijothomas, line 14: nit: remove this as there seems to no code using this.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionSender.cs
cijothomas: Can't find where do set CredentialEnvelope on the rejected transmission? (it may be followed up separate PR). If we miss to add credential on some transmissins, it'll lead to hard to track down data losses.

cijothomas (approved):
LGTM. Left couple of comments, tht can be followed up separately.

PR #2293, opened 2021-06-02
Fix AzureSdkDiagnosticListener from crashing user app

cijothomas:
AzureSdkDiagnosticListener component within DependencyTrackingTelemetryModule currently instantiates a new TelemetryClient, whenever a new Listener is created. (any statement like new DiagnosticListener("Azure.somename") will trigger this. All Azure sdks does this inside them)

In the event that the telemetryconfig used to initialize the module is disposed, but the DependencyTrackingTelemetryModule itself is not disposed, any attempt to create new AzureSdkListener() will result in an unhandled exception from the SDK, as AzureSdkDiagnosticListener will try to instantiate a new TelemetryClient using a disposed config. 

This violates the SDK's error handling policy of never crashing customer application after a successful initialization.

This PR modifies AzureSdk listener to create a single TelemetryClient during initialization, and re-use it for every subsequent AzureSdkListeners. If config is disposed, telemetry is not tracked (must be obvious!), but this PR also prevents the unhandled exception.

Issues like https://github.com/microsoft/ApplicationInsights-dotnet/issues/2195 will be fixed with this PR. However, the user code which is disposing telemetry config, without disposing the DependencyTrackingTelemetryModule must be fixed - but this is not a SDK bug, but user code issue and must handled by end users.

### DIFF
--- a/WEB/Src/DependencyCollector/DependencyCollector.Tests/Implementation/AzureSdkDiagnosticListenerTest.cs
+++ b/WEB/Src/DependencyCollector/DependencyCollector.Tests/Implementation/AzureSdkDiagnosticListenerTest.cs
@@ -50,6 +50,32 @@
         }
 
         [TestMethod]
+        public void AzureSdkListenerDoesNotThrowAfterInitialization()
+        {
+            using (var listener = new DiagnosticListener("Azure.SomeClient"))
+            using (var module = new DependencyTrackingTelemetryModule())
+            {
+                module.Initialize(this.configuration);
+
+                // Dispose config after initialize
+                // If AzureSdkListener attempted to create
+                // new TelemetryClient after initialize, it'd throw.
+                this.configuration.Dispose();
+
+                Activity sendActivity = new Activity("Azure.SomeClient.Send");
+                sendActivity.AddTag("kind", "client");
+
+                listener.StartActivity(sendActivity, null);
+                listener.StopActivity(sendActivity, null);
+
+                var listenerAnother = new DiagnosticListener("Azure.AnotherClient");
+                var listenerYetAnother = new DiagnosticListener("Azure.YetAnotherClient");
+
+                Assert.AreEqual(0, this.sentItems.Count);
+            }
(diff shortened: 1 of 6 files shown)


WEB/Src/DependencyCollector/DependencyCollector.Tests/Implementation/AzureSdkDiagnosticListenerTest.cs
```diff
@@ -49,6 +49,32 @@ public void CleanUp()
...
+                // new TelemetryClient after initialize, it'd throw.
+                this.configuration.Dispose();
+
+                Activity sendActivity = new Activity("Azure.SomeClient.Send");
+                sendActivity.AddTag("kind", "client");
+
+                listener.StartActivity(sendActivity, null);
+                listener.StopActivity(sendActivity, null);
+
+                var listenerAnother = new DiagnosticListener("Azure.AnotherClient");
```
cijothomas, line 71: these statements represent customer code (typically when they use some Azure sdks) which would have thrown unhandled exception, without the fix in this PR.
This would have resulted in exception, only if the DiagnosticListener was created *after* SDK initialization, because any DiagnosticListener created before SDK initialization will get callback within the subscribe code itself, which is try..catch..wrapped. (https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/WEB/Src/DependencyCollector/DependencyCollector/Implementation/DiagnosticSourceListenerBase.cs#L52-L58). 

Any listener created afterwards goes here (https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/WEB/Src/DependencyCollector/DependencyCollector/Implementation/DiagnosticSourceListenerBase.cs#L72-L97) which has no try..catch. Will do a follow up to make sure SDK is protected in all those scenarios as well.

cijothomas:
clsoing as this got lot of accidnetal changes. new one https://github.com/microsoft/ApplicationInsights-dotnet/pull/2294

PR #2305, opened 2021-06-17
enable self-diagnostics in example app

TimothyMothra:
The self-diagnostics feature has been invaluable while testing.
I would like to enable this in our example app.

I confirmed that the *.log files are excluded from git.

### DIFF
--- a/examples/WebApp.AspNetCore/ApplicationInsightsDiagnostics.json
+++ b/examples/WebApp.AspNetCore/ApplicationInsightsDiagnostics.json
@@ -1,5 +1,5 @@
 ﻿{
   "LogDirectory": ".",
   "FileSize": 1024,
-  "LogLevel": "Verbose"
+  "LogLevel": "Warning"
 }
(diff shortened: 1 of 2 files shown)


examples/WebApp.AspNetCore/WebApp.AspNetCore.csproj
```diff
@@ -8,4 +8,11 @@
     <ProjectReference Include="..\..\NETCORE\src\Microsoft.ApplicationInsights.AspNetCore\Microsoft.ApplicationInsights.AspNetCore.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <!--This file enables self-diagnostics-->
```
cijothomas, line 12: perhaps add a link to self-diagnostics doc here.

examples/WebApp.AspNetCore/ApplicationInsightsDiagnostics.json
```diff
@@ -0,0 +1,5 @@
+﻿{
+  "LogDirectory": ".",
+  "FileSize": 1024,
+  "LogLevel": "Verbose"
```
cijothomas, line 4: Warning would be sufficient for most cases, unless investigating something very specific.

rajkumar-rangaraj (approved):

cijothomas (approved):
LGTM.
Do we have asp.net sample apps and worker/console as well?

TimothyMothra:
> LGTM.
> Do we have asp.net sample apps and worker/console as well?

@cijothomas not yet, but this is something I want to add! :)

PR #2312, opened 2021-06-23
AAD: new AuthenticationTransmissionPolicy for retry scenarios.

TimothyMothra:
Issue #2190 

## Changes
- new class: `AuthenticationTransmissionPolicy` with retry rules for AAD scenarios. Off by default.
- change to `TransmissionPolicyCollection` to enable `AuthenticationTransmissionPolicy`
- refactor Unit Test `ThrottlingTransmissionPolicyTest`. My new tests are based on this.
- new Unit Tests that verify `AuthenticationTransmissionPolicy` triggers throttling for only specific HTTP StatusCodes.

Note: Exponential backoff is removed from scope of this PR and will be addressed in a following PR.

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
+++ b/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
@@ -8,6 +8,13 @@
     using Microsoft.ApplicationInsights.Channel.Implementation;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
 
+    /// <summary>
+    /// This class defines how the SDK will behave when it receives Response Codes 
+    /// from the Ingestion Service related to Authentication (AAD) scenarios.
+    /// </summary>
+    /// <remarks>
+    /// This class is disabled by default and expected to be enabled only when AAD has been configured in the AI SDK.
+    /// </remarks>
     internal class AuthenticationTransmissionPolicy : TransmissionPolicy, IDisposable
     {
         private BackoffLogicManager backoffLogicManager;
@@ -34,30 +41,36 @@
             GC.SuppressFinalize(this);
         }
 
+        /// <summary>
+        /// This method subscribes to the <see cref="Transmitter.TransmissionSent"/> event.
+        /// This encapsulates all <see cref="HttpWebResponseWrapper.StatusCode"/> related to Authentication (AAD) scenarios.
+        /// </summary>
+        /// <remarks>
+        /// AN EXPLANATION OF THE STATUS CODES:
+        /// - <see cref="ResponseStatusCodes.BadRequest"/>
+        /// "HTTP/1.1 400 Incorrect API was used - v2 API does not support authentication".
+        /// This indicates that the AI resource was configured for AAD, but SDK was not.
+        /// This is a configuration issue and is not recoverable from the client side. 
+        /// If the customer chooses to disable AAD in the AI Resource, the SDK could resume sending.
+        /// - <see cref="ResponseStatusCodes.Unauthorized"/>
+        /// "HTTP/1.1 401 Unauthorized - please provide the valid authorization token".
+        /// This indicates that the authorization token was either absent, invalid, or expired.
+        /// The root cause is not known and we should throttle retries.
+        /// - <see cref="ResponseStatusCodes.Forbidden"/>
+        /// "HTTP/1.1 403 Forbidden - provided credentials do not grant the access to ingest the telemetry into the component".
+        /// This indicates the configured identity does not have permissions to publish to this resource.
+        /// This is a configuration issue and is not recoverable from the client side. 
+        /// This can be recovered if the user changes the AI Resource's configured Access Control.
+        /// </remarks>
         private void HandleTransmissionSentEvent(object sender, TransmissionProcessedEventArgs e)
         {
             if (this.Enabled && e.Response != null)
             {
-                switch (e.Response.StatusCode) // HttpWebResponseWrapper
+                switch (e.Response.StatusCode)
                 {
                     case ResponseStatusCodes.BadRequest:
-                        // "HTTP/1.1 400 Incorrect API was used - v2 API does not support authentication".
-                        // This indicates that the AI resource was configured for AAD, but SDK was not.
-                        // This is a configuration issue and is not recoverable from the client side. 
-                        // If the customer chooses to disable AAD in the AI Resource, the SDK could resume sending.
-                        this.ApplyThrottlePolicy(e);
-                        break;
                     case ResponseStatusCodes.Unauthorized:
-                        // "HTTP/1.1 401 Unauthorized - please provide the valid authorization token".
-                        // This indicates that the authorization token was either absent, invalid, or expired.
-                        // The root cause is not known and we should throttle retries.
-                        this.ApplyThrottlePolicy(e);
-                        break;
                     case ResponseStatusCodes.Forbidden:
-                        // "HTTP/1.1 403 Forbidden - provided credentials do not grant the access to ingest the telemetry into the component".
-                        // This indicates the configured identity does not have permissions to publish to this resource.
-                        // This is a configuration issue and is not recoverable from the client side. 
-                        // This can be recovered if the user changes the AI Resource's configured Access Control.
                         this.ApplyThrottlePolicy(e);
                         break;
                 }
--- a/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
+++ b/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
@@ -16,7 +16,7 @@
     internal class AuthenticationTransmissionPolicy : TransmissionPolicy, IDisposable
     {
         private BackoffLogicManager backoffLogicManager;
-        private TaskTimerInternal pauseTimer = new TaskTimerInternal { Delay = TimeSpan.FromSeconds(BackoffLogicManager.SlotDelayInSeconds) };
+        private TaskTimerInternal pauseTimer = new TaskTimerInternal { Delay = TimeSpan.FromMinutes(1) };
 
         public bool Enabled { get; set; } = false;
 
@@ -88,7 +88,7 @@
             this.Transmitter.Enqueue(e.Transmission);
 
             // Ingestion service does not provide a retry value for these scenarios.
-            // this.pauseTimer is initialized using the value from BackoffLogicManager.SlotDelayInSeconds
+            // Check this.pauseTimer above for the configured wait time.
             this.pauseTimer.Start(() =>
                 {
                     this.ResetPolicy();
--- a/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
+++ b/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
@@ -51,7 +51,8 @@
         /// "HTTP/1.1 400 Incorrect API was used - v2 API does not support authentication".
         /// This indicates that the AI resource was configured for AAD, but SDK was not.
         /// This is a configuration issue and is not recoverable from the client side. 
-        /// If the customer chooses to disable AAD in the AI Resource, the SDK could resume sending.
+        /// IMPORTANT: We ignore this error and allow telemetry to be dropped.
+        /// Ingestion also uses HTTP 400 for invalid JSON and is indistinguishable from AAD errors.
         /// - <see cref="ResponseStatusCodes.Unauthorized"/>
         /// "HTTP/1.1 401 Unauthorized - please provide the valid authorization token".
         /// This indicates that the authorization token was either absent, invalid, or expired.
@@ -68,7 +69,6 @@
             {
                 switch (e.Response.StatusCode)
                 {
-                    case ResponseStatusCodes.BadRequest:
                     case ResponseStatusCodes.Unauthorized:
                     case ResponseStatusCodes.Forbidden:
                         this.ApplyThrottlePolicy(e);
@@ -89,7 +89,6 @@
             this.backoffLogicManager.ReportBackoffEnabled(e.Response.StatusCode);
             this.Transmitter.Enqueue(e.Transmission);
 
-            // Ingestion service does not provide a retry value for these scenarios.
             // Check this.pauseTimer above for the configured wait time.
             this.PauseTimer.Start(() =>
                 {
--- a/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
+++ b/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
@@ -1,10 +1,8 @@
 ﻿namespace Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation.TransmissionPolicy
 {
     using System;
-    using System.Globalization;
     using System.Threading.Tasks;
 
-    using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.Channel.Implementation;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
 
@@ -90,7 +88,7 @@
             this.Transmitter.Enqueue(e.Transmission);
 
             // Ingestion service does not provide a retry value. We use our own here.
-            this.pauseTimer.Delay = TimeSpan.FromSeconds(30);
+            //this.pauseTimer.Delay = TimeSpan.FromSeconds(30);
             this.pauseTimer.Start(() =>
                 {
                     this.ResetPolicy();
--- a/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
+++ b/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
@@ -36,47 +36,32 @@
 
         private void HandleTransmissionSentEvent(object sender, TransmissionProcessedEventArgs e)
         {
-            if (this.Enabled)
+            if (this.Enabled && e.Response != null)
             {
-                HttpWebResponseWrapper httpWebResponseWrapper = e.Response;
-                if (httpWebResponseWrapper != null)
+                switch (e.Response.StatusCode) // HttpWebResponseWrapper
                 {
-                    if (httpWebResponseWrapper.StatusCode == ResponseStatusCodes.BadRequest)
-                    {
-                        // HTTP/1.1 400 Incorrect API was used - v2 API does not support authentication.
+                    case ResponseStatusCodes.BadRequest:
+                        // "HTTP/1.1 400 Incorrect API was used - v2 API does not support authentication".
                         // This indicates that the AI resource was configured for AAD, but SDK was not.
-                        // This is a configuration issue and is not recoverable. 
-                        this.ApplyHaltPolicy(e);
-                    }
-                    else if (httpWebResponseWrapper.StatusCode == ResponseStatusCodes.Unauthorized)
-                    {
-                        // HTTP/1.1 401 Unauthorized - please provide the valid authorization token
+                        // This is a configuration issue and is not recoverable from the client side. 
+                        // If the customer chooses to disable AAD in the AI Resource, the SDK could resume sending.
+                        this.ApplyThrottlePolicy(e);
+                        break;
+                    case ResponseStatusCodes.Unauthorized:
+                        // "HTTP/1.1 401 Unauthorized - please provide the valid authorization token".
                         // This indicates that the authorization token was either absent, invalid, or expired.
                         // The root cause is not known and we should throttle retries.
                         this.ApplyThrottlePolicy(e);
-                    }
-                    else if (httpWebResponseWrapper.StatusCode == ResponseStatusCodes.Forbidden)
-                    {
-                        // HTTP/1.1 403 Forbidden - provided credentials do not grant the access to ingest the telemetry into the component
+                        break;
+                    case ResponseStatusCodes.Forbidden:
+                        // "HTTP/1.1 403 Forbidden - provided credentials do not grant the access to ingest the telemetry into the component".
                         // This indicates the configured identity does not have permissions to publish to this resource.
-                        // This can be recovered if the user changes the AI Resource configuration.
+                        // This is a configuration issue and is not recoverable from the client side. 
+                        // This can be recovered if the user changes the AI Resource's configured Access Control.
                         this.ApplyThrottlePolicy(e);
-                    }
+                        break;
                 }
             }
-        }
-
-        private void ApplyHaltPolicy(TransmissionProcessedEventArgs e)
-        {
-            this.MaxSenderCapacity = 0;
-            this.MaxBufferCapacity = 0;
-            this.MaxStorageCapacity = null;
-
-            this.LogCapacityChanged();
-            this.Apply();
-
-            this.backoffLogicManager.ReportBackoffEnabled(e.Response.StatusCode);
-            this.Transmitter.Enqueue(e.Transmission);
         }
 
         private void ApplyThrottlePolicy(TransmissionProcessedEventArgs e)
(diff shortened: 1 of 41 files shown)


BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        private void HandleTransmissionSentEvent(object sender, TransmissionProcessedEventArgs e)
+        {
+            if (this.Enabled)
+            {
+                HttpWebResponseWrapper httpWebResponseWrapper = e.Response;
+                if (httpWebResponseWrapper != null)
+                {
+                    if (httpWebResponseWrapper.StatusCode == ResponseStatusCodes.BadRequest)
```
pharring, line 44: Suggest: `switch (httpWebResponseWrapper.StatusCode)`

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            this.MaxSenderCapacity = 0;
+            this.MaxBufferCapacity = 0;
+            this.MaxStorageCapacity = null;
+
+            this.LogCapacityChanged();
+            this.Apply();
+
+            this.backoffLogicManager.ReportBackoffEnabled(e.Response.StatusCode);
+            this.Transmitter.Enqueue(e.Transmission);
```
cijothomas, line 79: why enqueue again?

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            // Ingestion service does not provide a retry value. We use our own here.
+            this.pauseTimer.Delay = TimeSpan.FromSeconds(30);
```
cijothomas, line 95: i think we should do exponential backoffs, similar to how its used in other existing policies.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,124 @@
...
+                    }
+                }
+            }
+        }
+
+        private void ApplyHaltPolicy(TransmissionProcessedEventArgs e)
+        {
+            this.MaxSenderCapacity = 0;
+            this.MaxBufferCapacity = 0;
+            this.MaxStorageCapacity = null;
```
cijothomas, line 73: Is this intentional that we send items to `Storage`, to be reattempted if app restarts with a valid config?

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
TimothyMothra: `Transmitter.Enqueue()` is responsible for sending telemetry to either Ingestion or Buffer or Storage:
https://github.com/microsoft/ApplicationInsights-dotnet/blob/4af30fd1b37ebf0d10e3c162752a7d2a7f0cc3e7/BASE/src/ServerTelemetryChannel/Implementation/Transmitter.cs#L158-L194


All of our TransmissionPolicies work this way. Here's `ThrottlingTransmissionPolicy` for example:
https://github.com/microsoft/ApplicationInsights-dotnet/blob/4af30fd1b37ebf0d10e3c162752a7d2a7f0cc3e7/BASE/src/ServerTelemetryChannel/Implementation/ThrottlingTransmissionPolicy.cs#L32-L74

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,124 @@
...
+                        // HTTP/1.1 403 Forbidden - provided credentials do not grant the access to ingest the telemetry into the component
+                        // This indicates the configured identity does not have permissions to publish to this resource.
+                        // This can be recovered if the user changes the AI Resource configuration.
+                        this.ApplyThrottlePolicy(e);
+                    }
+                }
+            }
+        }
+
+        private void ApplyHaltPolicy(TransmissionProcessedEventArgs e)
```
rajkumar-rangaraj, line 69: If we receive HTTP 400 from the ingestion service, This method will block sender/buffer for the lifetime of an app. Storage will max out in short span of time as not transmissions are not sent to the service. Why are we planning to store transmissions offline in this scenario anyways we will lose data at later point.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
rajkumar-rangaraj: Do we have a recommendation from Azure Identity on the retry timeframe?

pharring: See: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/RetryOptions.cs

Exponential back-off. Maximum of 3 retries. Initial delay 0.8s; maximum delay 1 minute.

```csharp
        /// <summary>
        /// The maximum number of retry attempts before giving up.
        /// </summary>
        public int MaxRetries { get; set; } = 3;

        /// <summary>
        /// The delay between retry attempts for a fixed approach or the delay
        /// on which to base calculations for a backoff-based approach.
        /// </summary>
        public TimeSpan Delay { get; set; } = TimeSpan.FromSeconds(0.8);

        /// <summary>
        /// The maximum permissible delay between retry attempts.
        /// </summary>
        public TimeSpan MaxDelay { get; set; } = TimeSpan.FromMinutes(1);

        /// <summary>
        /// The approach to use for calculating retry delays.
        /// </summary>
        public RetryMode Mode { get; set; } = RetryMode.Exponential;

        /// <summary>
        /// The timeout applied to an individual network operations.
        /// </summary>
        public TimeSpan NetworkTimeout { get; set; } = TimeSpan.FromSeconds(100);
```

Also, the [following](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/Internal/RetryPolicy.cs#L207) is relevant (note the randomness applied). 

```csharp
    private TimeSpan CalculateExponentialDelay(int attempted)
    {
        return TimeSpan.FromMilliseconds(
            Math.Min(
                (1 << (attempted - 1)) * _random.Next((int)(_delay.TotalMilliseconds * 0.8), (int)(_delay.TotalMilliseconds * 1.2)),
                _maxDelay.TotalMilliseconds));
    }
```

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
TimothyMothra: I think data loss is inevitable in each of these scenarios.

Thinking about 400 specifically, I think it is recoverable if the customer disables AAD on their AI Resource.
For that reason, I think storage-only is inappropriate here.

TimothyMothra:
> Do we have a recommendation from Azure Identity on the retry timeframe?

We do not. In each of these cases, Azure Identity is working as expected. 
If a customer reaches one of these cases, this means that the configuration between Azure & SDK are not in sync.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -36,49 +36,34 @@ public void Dispose()
...
-                        this.ApplyHaltPolicy(e);
-                    }
-                    else if (httpWebResponseWrapper.StatusCode == ResponseStatusCodes.Unauthorized)
-                    {
-                        // HTTP/1.1 401 Unauthorized - please provide the valid authorization token
+                        // This is a configuration issue and is not recoverable from the client side. 
+                        // If the customer chooses to disable AAD in the AI Resource, the SDK could resume sending.
+                        this.ApplyThrottlePolicy(e);
+                        break;
+                    case ResponseStatusCodes.Unauthorized:
```
rajkumar-rangaraj, line 50: shall we write using below syntax, it makes code easier to read.
```
case ResponseStatusCodes.Unauthorized:
case ResponseStatusCodes.Forbidden:
       this.ApplyThrottlePolicy(e);
       break;
```
Ref: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/switch#the-switch-section, like case 2 and 3 combined from doc sample.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
TimothyMothra: I think the `BackoffLogicManager` handles the exponential retries.
https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/BASE/src/ServerTelemetryChannel/Implementation/BackoffLogicManager.cs

Let me spend some time understanding how this works. 
In my local testing the retry is a fixed interval, so maybe I've configured something incorrectly.

TimothyMothra: The `BackoffLogicManager` is not documented so I'm still trying to understand how this works.

It seems that the exponential backoff is triggered when it parses the retry value from the response header in `GetBackOffTimeInterval()`.
https://github.com/microsoft/ApplicationInsights-dotnet/blob/ce50b28c693fbb88dd749065039d9c1c57443ac9/BASE/src/ServerTelemetryChannel/Implementation/BackoffLogicManager.cs#L142-L175

THIS won't work for us because in the AAD scenarios, our Ingestion Service does not provide a retry header.

The only example I've found of this is in the `PartialSuccessTransmissionPolicy`:
https://github.com/microsoft/ApplicationInsights-dotnet/blob/ce50b28c693fbb88dd749065039d9c1c57443ac9/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/PartialSuccessTransmissionPolicy.cs#L144-L155

I need more time to find a different way to use the existing code.

----

@cijothomas would it be possible to merge this as-is, without the exponential backoff? 
If so, this would unblock Beta3 and give me more time to work on a solution.

pharring: You may be overthinking this. 

If it's just about AAD auth, really all you need to be concerned with is this:
If you get a response that indicates the token is invalid (401 Unauthorized), then you should go back to the TokenCredential and ask for a new token and _immediately_ retry. No need for any delay. If you get another 401, then fail the call (your credential is bad).

You can probably see why caching inside the TokenCredential is a hard to cope with. If you ask for a new token, but you get back the _exact_ same token (because it was cached and it isn't close enough to expiration), you're hosed. How do you deal with that? I think you just have to fail on the spot. 

For 403s (forbidden), the situation is subtle. That indicates that, while the token was good, you just don't have access (Metrics Publisher role is missing). It could be that you just added the role and it hasn't propagated to all regions yet. So, buffering on disk and retrying later might be the right thing -- but that's a long-cycle retry loop. Minutes, not seconds. It would be great if the SDK did that, but I wouldn't grumble if telemetry got dropped in this situation.

TimothyMothra: Thanks, @pharring.
In this case, I'm advocating for avoiding the exponential backoff. And I'll change the retry to at least a minute.

I agree, I think data loss is going to be inevitable in most of these situations.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,120 @@
+﻿namespace Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation.TransmissionPolicy
+{
+    using System;
+    using System.Threading.Tasks;
+
+    using Microsoft.ApplicationInsights.Channel.Implementation;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+
+    /// <summary>
+    /// This class defines how the SDK will behave when it receives Response Codes
```
cijothomas, line 10:
```suggestion
    /// This class defines how the ServerTelemetryChannel will behave when it receives Response Codes 
```

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,120 @@
...
+
+    using Microsoft.ApplicationInsights.Channel.Implementation;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation;
+
+    /// <summary>
+    /// This class defines how the SDK will behave when it receives Response Codes 
+    /// from the Ingestion Service related to Authentication (AAD) scenarios.
+    /// </summary>
+    /// <remarks>
+    /// This class is disabled by default and expected to be enabled only when AAD has been configured in the AI SDK.
```
cijothomas, line 14: "in the AI SDK" - i'd avoid these terms, and be more specific, even though these are internal code..

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// This is a configuration issue and is not recoverable from the client side. 
+        /// This can be recovered if the user changes the AI Resource's configured Access Control.
+        /// </remarks>
+        private void HandleTransmissionSentEvent(object sender, TransmissionProcessedEventArgs e)
+        {
+            if (this.Enabled && e.Response != null)
+            {
+                switch (e.Response.StatusCode)
+                {
+                    case ResponseStatusCodes.BadRequest:
```
cijothomas, line 69: could you confirm if BadRequest is not thrown by ingestion for anything else like a invalid json, or json exceeding size etc?

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,120 @@
...
+            this.MaxBufferCapacity = null;
+            this.MaxStorageCapacity = null;
+
+            this.LogCapacityChanged();
+            this.Apply();
+
+            this.backoffLogicManager.ReportBackoffEnabled(e.Response.StatusCode);
+            this.Transmitter.Enqueue(e.Transmission);
+
+            // Ingestion service does not provide a retry value for these scenarios.
```
cijothomas, line 90: you can still use the BackOffLogicManager - its smart enough to deal with the situation where the ingestion service does not provide a 'retry-after" value, and picks a default, and then do the exponential back offs.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,120 @@
...
+        private void ApplyThrottlePolicy(TransmissionProcessedEventArgs e)
+        {
+            this.MaxSenderCapacity = 0;
+            this.MaxBufferCapacity = null;
+            this.MaxStorageCapacity = null;
+
+            this.LogCapacityChanged();
+            this.Apply();
+
+            this.backoffLogicManager.ReportBackoffEnabled(e.Response.StatusCode);
```
cijothomas, line 87: it doesn't look like the backofflogicmanager is used anywhere but in this line.. which seems incorrect.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,120 @@
...
+                        break;
+                }
+            }
+        }
+
+        private void ApplyThrottlePolicy(TransmissionProcessedEventArgs e)
+        {
+            this.MaxSenderCapacity = 0;
+            this.MaxBufferCapacity = null;
+            this.MaxStorageCapacity = null;
```
cijothomas, line 82: not clear why set this to null?

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,120 @@
...
+                        this.ApplyThrottlePolicy(e);
+                        break;
+                }
+            }
+        }
+
+        private void ApplyThrottlePolicy(TransmissionProcessedEventArgs e)
+        {
+            this.MaxSenderCapacity = 0;
+            this.MaxBufferCapacity = null;
```
cijothomas, line 81: not clear here either..

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
TimothyMothra: null is the default value. it means that there is no limit.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
TimothyMothra: null is the default value. it means that there is no limit.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,120 @@
...
+                    case ResponseStatusCodes.Forbidden:
+                        this.ApplyThrottlePolicy(e);
+                        break;
+                }
+            }
+        }
+
+        private void ApplyThrottlePolicy(TransmissionProcessedEventArgs e)
+        {
+            this.MaxSenderCapacity = 0;
```
cijothomas, line 80: if the intention is to temporarily send every new transmissions (including this one), directly to disk, then I think its best to follow what ErrorHandlingPoilcy currently does.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
TimothyMothra: We have multiple threads now about BackoffLogicManager... , so i'm sorry if i've lost the current thread. This may be easier to discuss on a call.

I based my implementation off of `ThrottlingTransmissionPolicy`
https://github.com/microsoft/ApplicationInsights-dotnet/blob/ce50b28c693fbb88dd749065039d9c1c57443ac9/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/ThrottlingTransmissionPolicy.cs#L62-L72

At line 65, `ThrottlingTransmissionPolicy` calls `this.backoffLogicManager.GetBackOffTimeInterval` and provides the value from the retry header.
This triggers the backofflogicmanager to begin the exponential backoff. 
I tried to discribe this above.

I can probably fake this by providing a custom formatted value. But I'll need to do more testing around this specifically.

TimothyMothra: What's strange to me is that in the entire SDK, we only have one implementation of TransmissionPolicy that is correctly resetting the backoff:

`PartialSuccessTransmissionPolicy` is the only class that calls `this.backoffLogicManager.ResetConsecutiveErrors();`
https://github.com/microsoft/ApplicationInsights-dotnet/blob/ce50b28c693fbb88dd749065039d9c1c57443ac9/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/PartialSuccessTransmissionPolicy.cs#L105-L112

I don't have an explanation for this and would need more time to fully investigate

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
cijothomas: sorry my question was - why does **this** policy have to set it to null?

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
TimothyMothra: ErrorHandling Policy sets Sender to 0, Buffer to 0, and leaves the Storage as default (null).

https://github.com/microsoft/ApplicationInsights-dotnet/blob/ce50b28c693fbb88dd749065039d9c1c57443ac9/BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/ErrorHandlingTransmissionPolicy.cs#L71-L101

I'm okay with this! :)
My rationale here was that we could utilize both the Buffer and Storage and might be able to recover more telemetry.

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
TimothyMothra: I confirmed that none of our other Policies are handling 400.
I tested using invalid json and received;
`{"itemsReceived":1,"itemsAccepted":0,"errors":[{"index":0,"statusCode":400,"message":"Unknown domain type 'undefined'"}]}`

cijothomas: Does this mean that, if we sent a bad json, ingestion responds with 400, and AuthenticationTransmissionPolicy would queue this item to be retried...400...retries..400...retries. (and forever, as the bad json will be forever a bad json)?

(The current behavior is - we drop the item if Ingestion responds with 400, and rightly so, as its pointless to retry a invalid json).

cijothomas: AuthenticationTransmissionPolicy can ignore 400 as well and drop the data. If user disables AAD in portal, telemetry could be sent, but there is no way to distinguish this vs invalid jsob. So it might be better off not supporting retry for 400

BASE/src/ServerTelemetryChannel/Implementation/TransmissionPolicy/AuthenticationTransmissionPolicy.cs
```diff
@@ -0,0 +1,121 @@
...
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// This method subscribes to the <see cref="Transmitter.TransmissionSent"/> event.
+        /// This encapsulates all <see cref="HttpWebResponseWrapper.StatusCode"/> related to Authentication (AAD) scenarios.
+        /// </summary>
+        /// <remarks>
+        /// AN EXPLANATION OF THE STATUS CODES:
+        /// - <see cref="ResponseStatusCodes.BadRequest"/>
```
rajkumar-rangaraj, line 50: This policy does not handle BadRequest, let us remove it from the comment.

rajkumar-rangaraj (approved):
This PR looks good. We could take exponential backoff on a different issue/conversation.

Currently, AI SDK enables exponential backoff for server(ingestion) side issues. All AAD error codes sent by ingestion service is for client issue. I don't think we should be enabling exponential backoff for client errors, rather trying to reach out the service at a constant time would be a good approach.

PR #2317, opened 2021-07-01
AAD: Misc changes

TimothyMothra:
Fix Issue #2190

## Changes
- Remove old TODO (Transmission.cs).
- Remove unnecessary comment (AuthenticationTransmissionPolicy.cs).
- Update changelog regarding previous PR's breaking change

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
@@ -33,9 +33,6 @@
             if (IsValidType(tokenCredential))
             {
                 this.tokenRequestContext = AzureCore.MakeTokenRequestContext(scopes: AuthConstants.GetScopes());
-
-                // Call GetToken to initialize the internal cache.
-                _ = this.GetToken();
             }
             else
             {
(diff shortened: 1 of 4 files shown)


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
```diff
@@ -33,6 +33,9 @@ public ReflectionCredentialEnvelope(object tokenCredential)
             if (IsValidType(tokenCredential))
             {
                 this.tokenRequestContext = AzureCore.MakeTokenRequestContext(scopes: AuthConstants.GetScopes());
+
+                // Call GetToken to initialize the internal cache.
+                _ = this.GetToken();
```
pharring, line 38: This results in (probably) synchronous network I/O in the constructor. Are you sure you want to do that? Doesn't this get called as soon as you set the AzureTokenCredential on TelemetryConfiguration?

TimothyMothra: Yes, this gets called when setting the Credential on TelmeteryConfiguration.
It was my intention to have this call during app startup, but I wasn't considering the worst-case scenario.

If you think it would be better to remove it I can do that.

pharring: Unless we're relying on side-effects, yes I would definitely remove that because it will have startup performance impact.

Everything's lazy in the Azure SDKs I've used. You can construct a client with a TokenCredential cheaply, but no actual token fetching is done until you actually need it (via BearerTokenPipelinePolicy)

rajkumar-rangaraj (approved):

pharring (approved):

PR #2336, opened 2021-07-19
Prune public api (prep 2.18.0 stable release)

TimothyMothra:
## Cleaning up the Public API.
- Remove `CredentialEnvelope`
- Remove `AuthToken`
- Add Perf library to the Base SDK's internals-visible-to. (QuickPulseServiceClient)

The `CredentialEnvelope` and `AuthToken` are our wrappers around the `Azure.Identity` libraries.
First, because we don't control these libraries, we don't want to provide a contract of their values.
Second, these classes are not intended for use by our customers. 

## For Partners
These changes will be available in the nightly build after this PR merges
Please ping me if you need access to these changes sooner.

Partners are invited to submit a separate PR to add their libraries to our internals-visible-to.


TimothyMothra:
@renatosalas

BASE/src/Microsoft.ApplicationInsights/Properties/AssemblyInfo.cs
```diff
@@ -9,6 +9,7 @@
 [assembly: InternalsVisibleTo("Microsoft.ApplicationInsights.AspNetCore.Tests, PublicKey=" + AssemblyInfo.PublicKey)]
 [assembly: InternalsVisibleTo("Microsoft.ApplicationInsights.WorkerService.Tests, PublicKey=" + AssemblyInfo.PublicKey)]
 [assembly: InternalsVisibleTo("Microsoft.AI.ServerTelemetryChannel, PublicKey=" + AssemblyInfo.PublicKey)]
+[assembly: InternalsVisibleTo("Microsoft.AI.PerfCounterCollector, PublicKey=" + AssemblyInfo.PublicKey)]
```
cijothomas, line 12: you want to add serviceprofiler/snapshot to this list?

TimothyMothra: I don't have their assembly names nor their public keys. 
i think it's best of those product owners submit a separate PR for this change.

cijothomas (approved):

cijothomas:
@pharring Please take a look. Thanks!

pharring:
For SnapshotCollector, we use reflection for CredentialEnvelope and AuthToken (it's a "light-up" feature), so making them internal is OK as long as we understand that they can't change in the future.
For the Windows profiler agent, I don't think this is a problem. We set the CredentialEnvelope, but we neither read or use it.
For the Linux profiler, I think we'll want to add an InternalsVisibleTo entry for Microsoft.ApplicationInsights.Profiler.AspNetCore. @xiaomi7732, what do you think?

xiaomi7732:
> For SnapshotCollector, we use reflection for CredentialEnvelope and AuthToken (it's a "light-up" feature), so making them internal is OK as long as we understand that they can't change in the future.
> For the Windows profiler agent, I don't think this is a problem. We set the CredentialEnvelope, but we neither read nor use it.
> For the Linux profiler, I think we'll want to add an InternalsVisibleTo entry for Microsoft.ApplicationInsights.Profiler.AspNetCore. @xiaomi7732, what do you think?

Thanks for loop me in. Linux Profiler will use reflection for CredentialEnvelop too. No need for `InternalsVisibleTo` there either.

Issue #2357, opened 2021-08-10
AzureInstanceMetadata: replace tests


TimothyMothra:
Attempting to switch to FrameworkReferences (#2355), and I found a problem with some tests for AzureInstanceMetadada.

### unnecessary local server

First issue, the tests for **AzureInstanceMetadata** run a local server to intercept HttpRequests.
There were breaking changes to `WebHostBuilder` in the change from netcore2.x -> netcore3.x.
https://github.com/microsoft/ApplicationInsights-dotnet/blob/a90eb8fa1013b5aaf484bc740b97d22a2df8141d/WEB/Src/WindowsServer/WindowsServer.Tests/Mock/AzureInstanceMetadataServiceMock.cs#L27-L36

We do not need a local server to test AzureInstanceMetadata, we only need to mock the HTTP Response.
This is a good candidate to switch to use HttpClient mocking.
https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.-ctor?view=net-5.0#System_Net_Http_HttpClient__ctor_System_Net_Http_HttpMessageHandler_

### invalid tests

Second issue, tests were replacing the HTTP call with a Func<> that returns a parsed object.
The existing tests seem to verify the behavior of this mock class, and not the HttpClient.

While rewriting tests, I've found that the HttpClient successfully handles the test condition.

----

## Affected product code:

`AzureMetadataRequestor` ctor accepts a Func<>.
https://github.com/microsoft/ApplicationInsights-dotnet/blob/a90eb8fa1013b5aaf484bc740b97d22a2df8141d/WEB/Src/WindowsServer/WindowsServer/Implementation/AzureMetadataRequestor.cs#L34-L37

When making a request, it can switch to use the Func instead.
This is used in unit tests to invoke a local server.

https://github.com/microsoft/ApplicationInsights-dotnet/blob/a90eb8fa1013b5aaf484bc740b97d22a2df8141d/WEB/Src/WindowsServer/WindowsServer/Implementation/AzureMetadataRequestor.cs#L59-L86

## Affected Unit Tests:

Some of these tests are testing a condition that is only reproducible using the mock code.
I'm summarizing the affected tests here and will replace the tests where appropriate.

- `AzureInstanceMetadataEndToEndTests`
    - `SpoofedResponseFromAzureIMSDoesntCrash`
	tests that a request sucessfully parses the json response
	- `AzureImsResponseTooLargeStopsCollection`
[...]

PR #2358, opened 2021-08-10
modify AzureMetadataRequestor to accept HttpClient in ctor. rewrite AzureInstanceMetadata tests to mock HttpClient instead of using in-proc server.

TimothyMothra:
When switching to FrameworkReferences, discovered a problem with some tests for AzureInstanceMetadada (see full write up in #2357).

## Changes
- Replace local inproc server with HttpClient and mock HttpMessageHandler.
- Remove test `AzureImsResponseTooLargeStopsCollection`. HttpClient correctly handles and parses the larger response.
- Replace test `AzureImsResponseTimesOut` with `AzureImsResponseHandlesException`. Could not force a timeout in a test context. HttpClient throws an exception when timeout. I changed this test to throw an exception and verify that this is handled correctly.

### Checklist

### Notes for authors:

### Notes for reviewers:


WEB/Src/WindowsServer/WindowsServer/Implementation/AzureMetadataRequestor.cs
```diff
@@ -58,52 +52,40 @@ public Task<AzureInstanceComputeMetadata> GetAzureComputeMetadataAsync()
...
             AzureInstanceComputeMetadata azureIms = null;
             DataContractJsonSerializer deserializer = new DataContractJsonSerializer(typeof(AzureInstanceComputeMetadata));
 
-            using (var azureImsClient = new HttpClient())
-            {
-                azureImsClient.MaxResponseContentBufferSize = AzureMetadataRequestor.AzureImsMaxResponseBufferSize;
-                azureImsClient.DefaultRequestHeaders.Add("Metadata", "True");
-                azureImsClient.Timeout = this.AzureImsRequestTimeout;
+            this.httpClient.MaxResponseContentBufferSize = AzureMetadataRequestor.AzureImsMaxResponseBufferSize;
+            this.httpClient.DefaultRequestHeaders.Add("Metadata", "True");
```
reyang, line 79: Would this end up adding multiple headers to the `httpClient`?

WEB/Src/WindowsServer/WindowsServer/Implementation/AzureMetadataRequestor.cs
```diff
@@ -58,52 +52,40 @@ public Task<AzureInstanceComputeMetadata> GetAzureComputeMetadataAsync()
...
+            this.httpClient.Timeout = this.AzureImsRequestTimeout;
```
reyang, line 80: Do we need to change Timeout each time?

reyang (changes requested):
Please find my comments.

reyang:
The change seems to touch things beyond test cases. Either we need to change the PR title/description or scope it down to test cases.

WEB/Src/WindowsServer/WindowsServer/Implementation/AzureMetadataRequestor.cs
TimothyMothra: moved to constructor

WEB/Src/WindowsServer/WindowsServer/Implementation/AzureMetadataRequestor.cs
```diff
@@ -20,15 +20,24 @@ internal class AzureMetadataRequestor : IAzureMetadataRequestor
         internal const int AzureImsMaxResponseBufferSize = 512;
 
         /// <summary>
-        /// Default timeout for the web requests made to obtain Azure IMS data. Internal to expose to tests.
+        /// Default timeout for the web requests made to obtain Azure IMS data.
         /// </summary>
-        internal TimeSpan AzureImsRequestTimeout = TimeSpan.FromSeconds(10);
+        private TimeSpan AzureImsRequestTimeout = TimeSpan.FromSeconds(10);
 
         private readonly HttpClient httpClient;
```
reyang: Who owns the lifecycle management for this http client? Who is responsible for `Dispose`?

What is the thread safety guarantee? Do we allow concurrent invocations?

TimothyMothra: this is a miss on my part. I need to implement the correct Dispose pattern.

This code is only ever called once, when the SDK initializes TelemetryModules. There are no retries.

https://github.com/microsoft/ApplicationInsights-dotnet/blob/a90eb8fa1013b5aaf484bc740b97d22a2df8141d/WEB/Src/WindowsServer/WindowsServer/AzureInstanceMetadataTelemetryModule.cs#L60-L89

https://github.com/microsoft/ApplicationInsights-dotnet/blob/a90eb8fa1013b5aaf484bc740b97d22a2df8141d/WEB/Src/WindowsServer/WindowsServer/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs#L59-L103

WEB/Src/WindowsServer/WindowsServer/Implementation/IAzureMetadataRequestor.cs
```diff
@@ -1,10 +1,11 @@
 ﻿namespace Microsoft.ApplicationInsights.WindowsServer.Implementation
 {
+    using System;
     using System.Collections.Generic;
     using System.Threading.Tasks;
     using Microsoft.ApplicationInsights.WindowsServer.Implementation.DataContracts;
 
-    internal interface IAzureMetadataRequestor
+    internal interface IAzureMetadataRequestor : IDisposable
```
reyang, line 8: Do we need this?

WEB/Src/WindowsServer/WindowsServer/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
```diff
@@ -101,5 +103,25 @@ public async Task<bool> SetDefaultPayloadAsync(IHeartbeatPropertyManager provide
 
             return hasSetFields;
         }
+
+        public void Dispose()
```
reyang, line 107: Who will call this method?

TimothyMothra: The call stack here is: TelemetryModule > Provider > Requestor > HttpClient.

Inside the TelemetryModule, I moved the Provider into a using to ensure it gets disposed.

WEB/Src/WindowsServer/WindowsServer/Implementation/IAzureMetadataRequestor.cs
TimothyMothra: The call stack here is: TelemetryModule > Provider > Requestor > HttpClient.

We don't NEED this interface to be disposable. 
Alternatively, the Provider can check if it's private requestor implements IDisposable.
This introduces a little more complexity and it's not as obvious that the Requestor needs to be disposed. (The Analyzers can't catch this scenario).

reyang: If we don't need this, let's try not to add it.

reyang: FYI this might be a good approach https://github.com/open-telemetry/opentelemetry-dotnet/blob/95893457080a772eb75e7885e640d2c055b2a0e1/src/OpenTelemetry/Trace/TracerProviderSdk.cs#L304

WEB/Src/WindowsServer/WindowsServer/Implementation/AzureComputeMetadataHeartbeatPropertyProvider.cs
reyang: Stepping back, if the whole purpose is to allow mock in the unit test, would reflection (setting some private member instead of having to change any production code) work?

TimothyMothra: I'm not sure that a private member helps us here.

The original class doesn't have a private HttpClient. It creates the client just-in-time to make a request.
This is how it avoided the responsibility of Disposing.
https://github.com/microsoft/ApplicationInsights-dotnet/blob/a90eb8fa1013b5aaf484bc740b97d22a2df8141d/WEB/Src/WindowsServer/WindowsServer/Implementation/AzureMetadataRequestor.cs#L88-L110



I can mock any HTTP response if I can provide a custom HttpMessageHandler into the HttpClient.
```
var mockHttpMessageHandler = new Mock<HttpMessageHandler>();
new HttpClient(mockHttpMessageHandler.Object)
```

TimothyMothra: I came up with an alternate approach that works.
Tomorrow i'll clean it up and create a new PR.

TimothyMothra:
superceded by #2360

Issue #2251, opened 2021-05-05
End of Life for .NET Core 2.1 (August 21, 2021)


TimothyMothra:
## .NET Core 2.1 is officially scheduled to reach end of life on [August 21, 2021](https://dotnet.microsoft.com/platform/support/policy/dotnet-core#:~:text=august%2021%2C%202021)

- Official Statement: https://dotnet.microsoft.com/platform/support/policy/dotnet-core

Using this Issue to track planned changes.

- [x] Remove this framework from all projects
- [x] Remove this framework from all build definitions
- [x] Individual package references should be replaced with framework references wherever possible. (https://docs.microsoft.com/en-us/aspnet/core/fundamentals/target-aspnetcore?view=aspnetcore-5.0&tabs=visual-studio)
    - [x] `WindowsServer.Tests.csproj`: 
     `AzureInstanceMetadataEndToEndTests` depends on `AzureInstanceMetadataServiceMock` which creates an in-proc server to handle HTTP requests. 
     This is being addressed in a separate issue #2357 
    - [ ] `Microsoft.ApplicationInsights.AspNetCore.Tests.csproj`
    See also: https://weblog.west-wind.com/posts/2020/Feb/26/Working-with-IWebHostEnvironment-and-IHostingEnvironment-in-dual-targeted-NET-Core-Projects
    > 'IHostingEnvironment' is obsolete: 'This type is obsolete and will be removed in a future version. 
         The recommended alternative is Microsoft.AspNetCore.Hosting.IWebHostEnvironment.'
         
- [ ] Consolidate extra dependencies in AspNetCore taken to address security issue: #2199 #2243
- [x] Verify all projects have test coverage for all supported versions of dotnet.
  This is being addressed in a separate issue #2379
- [ ] Update changelog

HamsterExAstris:
Just to make sure I understand: is this issue only deprecating the .NET Core 2.1 _runtime_, or was support for ASP.NET Core 2.1 on the .NET Framework runtime (which is [still under support](https://dotnet.microsoft.com/platform/support/policy/aspnet#dotnet-core)) dropped too?

PR #2363, opened 2021-08-12
cleanup test projects

TimothyMothra:
As a part of #2251, I'm reviewing all test projects and trying to get them on the same target frameworks.
This PR is pulling some of my changes out of #2361.

I'm cleaning up test projects by removing unused properties:
- `<Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />` was a bug fix for VS2017 and no longer necessary. ([link](https://github.com/microsoft/vstest/issues/472))
- Some of the extra properties in the AspNetCore projects were preventing VS and some CLI from running those tests (FIXED!!)

- Preparing to move the `<TargetFrameworks>` to the Test.props file. We few exceptions, we should have uniform `TargetFrameworks` across all our projects. 
  I'll convert individual test projects to use the centralized TargetFrameworks in follow up PRs because this is requiring some other changes in those individual projects (preprocessors and conditional PackageReferences)

### DIFF
--- a/.props/Test.props
+++ b/.props/Test.props
@@ -9,6 +9,22 @@
   <Import Project=".\_AnalyzerSettings.props" />
 
   <PropertyGroup>
+    <!-- Our test matrix includes every currently supported version of .NET
+          - net4.5.2 (EoL April 2022)
+          - net4.6.0 (EoL April 2022)
+          - net4.6.1 (EoL April 2022)
+          - net4.6.2
+          - net4.7.2
+          - net4.8.0
+          - netcoreapp3.1 (EoL Dec 2022)
+          - net5.0 (EoL Feb 2022)
+          - net6.0 (GA Nov 2021)
+    -->
+    <TargetFrameworks>net462;net472;net480;netcoreapp3.1;net5.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(OS) != 'Windows_NT'">netcoreapp3.1;net5.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <PropertyGroup>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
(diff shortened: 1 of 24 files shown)


.props/Test.props
```diff
@@ -8,6 +8,22 @@
...
+          - net4.6.2
+          - net4.7.2
+          - net4.8.0
+          - netcoreapp3.1 (EoL Dec 2022)
+          - net5.0 (EoL Feb 2022)
+          - net6.0 (GA Nov 2021)
+    -->
+    <TargetFrameworks>net462;net472;net480;netcoreapp3.1;net5.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(OS) != 'Windows_NT'">netcoreapp3.1;net5.0</TargetFrameworks>
+  </PropertyGroup>
```
TimothyMothra, line 25: FYI: although 452, 460, and 461 are still supported today, i don't see any value in onboarding any new projects to these frameworks.
Any test project that currently targets these can continue to target them like this:

`<TargetFrameworks>net452,net46,net461;$(TargetFrameworks)</TargetFrameworks>`

rajkumar-rangaraj: Every test projects below has an entry for `TargetFrameworks`, why do we need to define this in `test.prop`

rajkumar-rangaraj (approved):

.props/Test.props
```diff
@@ -8,6 +8,22 @@
...
+          - net4.6.0 (EoL April 2022)
+          - net4.6.1 (EoL April 2022)
+          - net4.6.2
+          - net4.7.2
+          - net4.8.0
+          - netcoreapp3.1 (EoL Dec 2022)
+          - net5.0 (EoL Feb 2022)
+          - net6.0 (GA Nov 2021)
+    -->
+    <TargetFrameworks>net462;net472;net480;netcoreapp3.1;net5.0</TargetFrameworks>
```
cijothomas, line 23: net48 vs net480?

TimothyMothra:
#2379

PR #2365, opened 2021-08-13
Enable SDK to create package for auto-instrumentation

rajkumar-rangaraj:
Thanks @TimothyMothra for the [PoC](https://github.com/microsoft/ApplicationInsights-dotnet/pull/2348)

## Changes

- Downgraded DiagnosticSource
- Added conditional compilation field `REDFIELD` which is used as preprocessor to decide the event source name
- Enabled `redfield` compilation switch to build and generate package for auto-instrumentation


Directory.Build.props
```diff
@@ -7,8 +7,22 @@
 
     <!-- This is used to disable some build properties. -->
     <IsExamplesSolution Condition="'$(SolutionName)' == 'Examples' ">true</IsExamplesSolution>
+
+    <!-- Remove below entry before merge, using it to build and test Redfield -->
+	<Redfield>true</Redfield>
```
TimothyMothra, line 12: TODO: don't forget to remove this before merge :)

rajkumar-rangaraj: All tests with `Redfield=true` ran success. Remove this entry, will rely on our pipeline to introduce flag.

.props/_GlobalStaticVersion.props
```diff
@@ -14,6 +14,7 @@
     <SemanticVersionMinor>18</SemanticVersionMinor> <!-- If changing the Minor version, also update the Date value. -->
     <SemanticVersionPatch>0</SemanticVersionPatch>
     <PreReleaseMilestone></PreReleaseMilestone> <!--Valid values: beta1, beta2, EMPTY for stable -->
+	<PreReleaseMilestone Condition="'$(Redfield)' == 'True'">redfield</PreReleaseMilestone>
```
reyang, line 17:
```suggestion
    <PreReleaseMilestone Condition="'$(Redfield)' == 'True'">redfield</PreReleaseMilestone>
```

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/DiagnosticsModule/DiagnosticsEventListener.cs
```diff
@@ -86,6 +86,34 @@ protected override void OnEventSourceCreated(EventSource eventSource)
...
+                    case "Redfield-Microsoft-ApplicationInsights-Extensibility-EventCounterCollector":
+                    case "Redfield-Microsoft-ApplicationInsights-Extensibility-PerformanceCollector":
+                    case "Redfield-Microsoft-ApplicationInsights-Extensibility-PerformanceCollector-QuickPulse":
+                    case "Redfield-Microsoft-ApplicationInsights-Extensibility-Web":
+                    case "Redfield-Microsoft-ApplicationInsights-Extensibility-WindowsServer":
+                    case "Redfield-Microsoft-ApplicationInsights-WindowsServer-Core":
+                    case "Redfield-Microsoft-ApplicationInsights-Extensibility-EventSourceListener":
+                    case "Redfield-Microsoft-ApplicationInsights-AspNetCore":
+                    case "Redfield-Microsoft-ApplicationInsights-LoggerProvider":
+                    case "Microsoft-AspNet-Telemetry-Correlation":
```
reyang, line 110: It is possible to hit this condition at all?

reyang: Curious, what's the performance of this switch statement vs. a dictionary look up?

rajkumar-rangaraj: Good catch. This needs to be fixed.

BASE/src/Microsoft.ApplicationInsights/Microsoft.ApplicationInsights.csproj
```diff
@@ -23,9 +23,21 @@
     <PackageReference Include="Microsoft.Diagnostics.Tracing.EventRegister" Version="1.1.28" Condition="$(OS) == 'Windows_NT'">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="5.0.0" />
   </ItemGroup>
 
+  <Choose>
+    <When Condition="'$(Redfield)' == 'True'">
+      <ItemGroup>
```
reyang, line 30: Is it possible to use something simpler like:

```
<ItemGroup Condition="'$(Redfield)' == 'True'">
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.7.0" />
</ItemGroup>
```

or even:

```
<PackageReference Include="System.Diagnostics.DiagnosticSource" Version="5.0.0" Condition="'$(Redfield)' != 'True'"/>
```

reyang: or perhaps extract the version number, and set the version based on conditions in the common file (e.g. where `$(RedField)` is being defined:

```
<PackageReference Include="System.Diagnostics.DiagnosticSource" Version="$(SystemDiagnosticsDiagnosticSourcePkgVer)" />
```

reyang (approved):
LGTM with some suggestions.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/FileDiagnosticsModule/TraceSourceForEventSource.cs
```diff
@@ -13,7 +13,12 @@
     internal class TraceSourceForEventSource : TraceSource, IEventListener, IDisposable
     {
         private const long AllKeyword = -1;
+
+#if REDFIELD
```
vishweshbankwar, line 17: Do we need to change this?

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/DiagnosticsModule/DiagnosticsEventListener.cs
rajkumar-rangaraj: Performance needs to be validated. Definitely, dictionary might be faster as we check 15 cases every time.

BASE/src/Microsoft.ApplicationInsights/Microsoft.ApplicationInsights.csproj
rajkumar-rangaraj: How about this one?

```
  <PropertyGroup Label="Package version for System.Diagnostics.DiagnosticSource">
    <SystemDiagnosticsDiagnosticSourcePkgVer>5.0.0</SystemDiagnosticsDiagnosticSourcePkgVer>
    <SystemDiagnosticsDiagnosticSourcePkgVer Condition="'$(Redfield)' == 'True'">4.7.0</SystemDiagnosticsDiagnosticSourcePkgVer>
  </PropertyGroup>

  <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="$(SystemDiagnosticsDiagnosticSourcePkgVer)" />
```

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/FileDiagnosticsModule/TraceSourceForEventSource.cs
rajkumar-rangaraj: This might not be needed, but when we capture data using `FileDiagnostics` we will get a source name with `RedField` prefixed. We might not rely on this feature anymore, we should use `SelfDiagnostics` feature going forward.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/DiagnosticsModule/DiagnosticsEventListener.cs
TimothyMothra: I'm fuzzy on the exact details, I thought after a certain number of cases, a switch gets compiled as a hashmap?

BASE/src/Microsoft.ApplicationInsights/Microsoft.ApplicationInsights.csproj
TimothyMothra: I liked the explicitness of the `Choose` statement. To me, this communicates that these settings are coupled and not trivial to change.

But I don't have a strong opinion. Raj's suggestion is definitely cleaner!

WEB/Src/DependencyCollector/DependencyCollector/DependencyTrackingTelemetryModule.cs
```diff
@@ -282,10 +282,9 @@ protected virtual void Dispose(bool disposing)
         [MethodImpl(MethodImplOptions.NoOptimization | MethodImplOptions.NoInlining)]
         private static void PrepareFirstActivity()
         {
-            using (var activity = new Activity("Microsoft.ApplicationInights.Init"))
-            {
-                activity.Start();
-            }
+            var activity = new Activity("Microsoft.ApplicationInsights.Init");
+            activity.Start();
+            activity.Stop();
```
xiang17, line 287: Is this extra line of code related to the downgrade of DiagnosticSource?

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/CoreEventSourceTest.cs
```diff
@@ -25,5 +26,30 @@ private static EventAttribute GetEventAttribute(string methodName)
...
+
+#if REDFIELD
+        /// <summary>
+        /// This is a sanitiy check.
+        /// The 'Redfield' compilation flag should switch the name of EventSource class.
+        /// Devs can review the test log and confirm that this test runs and passes.
+        /// This will serve as a verification that the Redfield compilation flag worked as expected.
+        /// 
+        /// To run this test:
+        /// dotnet build /p:Redfield=True ".\dotnet\BASE\Microsoft.ApplicationInsights.sln"
```
xiang17, line 38: Will this test be covered by GitHub Action?

WEB/Src/DependencyCollector/DependencyCollector/DependencyTrackingTelemetryModule.cs
rajkumar-rangaraj: Yes. using dispose pattern is not supported in earlier version of DiagnosticSource.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/CoreEventSourceTest.cs
rajkumar-rangaraj: Yes. We run a package against pipeline with compilation flag and that validates it. 
I validated this test in this PR itself. If you look at the commit `Remove hard-coded Redfield property` the tests ran successfully before that commit should include this test.

cijothomas:
Note: Won't be able to support https://github.com/microsoft/ApplicationInsights-dotnet/pull/2185/files with this change

Directory.Build.props
```diff
@@ -7,8 +7,25 @@
 
     <!-- This is used to disable some build properties. -->
     <IsExamplesSolution Condition="'$(SolutionName)' == 'Examples' ">true</IsExamplesSolution>
+
+    <!-- This is used to change EventSource names. -->
+    <DefineConstants Condition="'$(Redfield)' == 'True'">$(DefineConstants);REDFIELD</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup Label="Package versions for System.Diagnostics.DiagnosticSource">
+    <SystemDiagnosticsDiagnosticSourcePkgVer>5.0.0</SystemDiagnosticsDiagnosticSourcePkgVer>
```
cijothomas, line 16: 5.0.1 would be ideal as it has critical timestamp accuracy fix.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/CoreEventSourceTest.cs
TimothyMothra: now that i've thought about it, i think Sean is right. lets introduce a simple github action to sanity check this compilation. @rajkumar-rangaraj ping me on Monday, i have another branch with a github compatible yml that we can repurpose for this.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/DiagnosticsModule/DiagnosticsEventListener.cs
rajkumar-rangaraj: Performed some research with IL code and then hit an article which speaks about this scenario - https://stackoverflow.com/questions/11617091/in-a-switch-vs-dictionary-for-a-value-of-func-which-is-faster-and-why

Directory.Build.props
rajkumar-rangaraj: This is out of the scope for this PR as we are not changing any packages related to the actual flow. Do you want to change the version in PR?

rajkumar-rangaraj:
> Note: Won't be able to support https://github.com/microsoft/ApplicationInsights-dotnet/pull/2185/files with this change

We added a redfied sanity check in GitHub actions. This ensures builds are not broken with new changes.

This is a temporary workaround until we resolve the DiagnosticSource versioning issue. We will remove these changes once the DiagnosticSource issue is resolved.

TimothyMothra (approved):

.github/workflows/redfield-sanity-check.yml
```diff
@@ -0,0 +1,39 @@
...
+    - name: restore (Base sln)
+      run: dotnet restore ./BASE/Microsoft.ApplicationInsights.sln
+
+    - name: Build
+      run: dotnet build -p:Redfield=True ./BASE/Microsoft.ApplicationInsights.sln --configuration Release --no-restore
+
+    - name: Test
+      id: test1
+      continue-on-error: true
+      run: dotnet test ./BASE/Microsoft.ApplicationInsights.sln --framework ${{ matrix.framework }} --configuration Release --no-build --logger:"console;verbosity=detailed" ${{ matrix.args }}
```
TimothyMothra, line 39: Because all the other builds run the full tests, I think for this build we can run only the Redfield tests.
We can filter by test name or introduce a new attribute.

rajkumar-rangaraj: There are changes in diagnostic listeners and few tests to include Redfield event source. It would be good to run it against the complete project. For example, Please check changes to tests `EventCounterListener.cs` in this PR.

.github/workflows/redfield-sanity-check.yml
```diff
@@ -0,0 +1,39 @@
...
+              args: "--filter TestCategory!=WindowsOnly"
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: restore (Base sln)
+      run: dotnet restore ./BASE/Microsoft.ApplicationInsights.sln
+
+    - name: Build
+      run: dotnet build -p:Redfield=True ./BASE/Microsoft.ApplicationInsights.sln --configuration Release --no-restore
```
cijothomas, line 34: only building Base sdk?

BASE/src/Microsoft.ApplicationInsights/ActivityExtensions.cs
```diff
@@ -52,7 +52,11 @@ private static bool Initialize()
         {
             try
             {
+#if REDFIELD
```
cijothomas, line 55: Can you open an issue to completely remove this class. it should not be required.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/RichPayloadEventSource.cs
```diff
@@ -21,7 +21,11 @@ internal sealed partial class RichPayloadEventSource : IDisposable
         internal readonly EventSource EventSourceInternal;
 
         /// <summary>Event provider name.</summary>
+#if REDFIELD
+        private const string EventProviderName = "Redfield-Microsoft-ApplicationInsights-Data";
```
cijothomas, line 25: Partners like Azure profiler team needs to be aware of this?

cijothomas: Todd mentioned some azure app service diagnostics also depend on the provider name.

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AppMapCorrelationEventSource.cs
```diff
@@ -7,7 +7,11 @@
     /// <summary>
     /// ETW EventSource tracing class.
     /// </summary>
+#if REDFIELD
```
cijothomas, line 10: for a separate PR , this file seems doing nothing and could be removed!

cijothomas:
> > Note: Won't be able to support https://github.com/microsoft/ApplicationInsights-dotnet/pull/2185/files with this change
> 
> We added a redfied sanity check in GitHub actions. This ensures builds are not broken with new changes.
> 
> This is a temporary workaround until we resolve the DiagnosticSource versioning issue. We will remove these changes once the DiagnosticSource issue is resolved.

I think we can support PRs like the above, by doing it inside IF !REDFIELD check, as we'll have 5.0.0 of DS when not-redfield.

cijothomas (approved):

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/RichPayloadEventSource.cs
vishweshbankwar: IIRC, profiler team has already updated their side last year to listen to this eventsource. But we can double check with Paul.

rajkumar-rangaraj: Good call out. Documents needs to be updated to collect ETW traces from both SDK and Redfield. We cannot expect customer to switch to appropriate set of events for Nuget and auto-instrumentation approach. Will track this in a separate internal item and get docs updated when auto-instrumentation goes live with new package.

.github/workflows/redfield-sanity-check.yml
TimothyMothra: Running our full test suite here will fail.
We have too many flaky tests and Github Actions does not have a mechanism for retries. 
(I'm investigating this in #2345)
If we want to run more test projects here, we must filter those tests.

BASE/src/Microsoft.ApplicationInsights/ActivityExtensions.cs
rajkumar-rangaraj: Investigated a little and found methods from this class is called in several places.
For example Tryrun is called in `OperationCorrelationTelemetryInitializer.cs`. 

Did I understand your comment incorrectly?

cijothomas: yea all the places where this ActivityExtensions is used can be modified to directly use Activity API, instead of doing it via this class.

PR #2406, opened 2021-09-08
Auto-Instrumentation - Enable AI SDK to honor repacked Azure.Identity assembly name

rajkumar-rangaraj:
Fix Issue # .

- `Azure.Core` assembly name is hard-coded in `ReflectionCredentialEnvelope` class
- This changes will enable auto-instrumentation to use loaded IL Repacked Azure.Identity library
- This approach will avoid conflict with customer libraries

## Changes
Introduce a local to decide between redfelid and normal 

### Checklist

### DIFF
--- a/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
@@ -20,9 +20,9 @@
     internal class ReflectionCredentialEnvelope : CredentialEnvelope
     {
 #if REDFIELD
-        private static readonly string AzureCoreAssemblyName = "Azure.Identity.ILRepack";
+        private static volatile string AzureCoreAssemblyName = "Azure.Identity.ILRepack";
 #else
-        private static readonly string AzureCoreAssemblyName = "Azure.Core";
+        private static volatile string AzureCoreAssemblyName = "Azure.Core";
 #endif
 
         private readonly object tokenCredential;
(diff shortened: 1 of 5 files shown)


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
```diff
@@ -19,6 +19,12 @@
     /// </remarks>
     internal class ReflectionCredentialEnvelope : CredentialEnvelope
     {
+#if REDFIELD
+        private static readonly string AzureCoreAssemblyName = "Azure.Identity.ILRepack";
```
reyang, line 23: Can this be a private const?

reyang (approved):
LGTM.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Authentication/ReflectionCredentialEnvelope.cs
rajkumar-rangaraj: Kept it intentionally as readonly to use reflection if there be a need in this area. We had to prepare this change as there was no option for reflection. Please correct me if this approach is not the right thing to do.

reyang: Got it, consider removing the "readonly" part to emphasize the semantic (and also prevent the compiler from optimizing too much).

reyang: https://fluentprogrammer.com/volatile-keyword-in-c-and-c-plus-plus/

PR #2411, opened 2021-09-14
stale

TimothyMothra:
This repo has hundreds of issues that are not being commented or worked on.
I'm introducing Github's Stale Action (https://github.com/actions/stale) to automatically close any issues that are inactive.

Beginning with these params: `days-before-stale: 300` `days-before-close: 7`.
This is to quickly remove the oldest issues in this repo.

Later, I'm proposing that these params be changed to `100` and `14` to keep open issues relevant to our latest versions.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,27 @@
+# Syntax: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
+# Github Actions Stale: https://github.com/actions/stale
+# Description: The purpose of this build is to compile and test with redfield flag.
+
+name: Stale Issues and PRs
+on:
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,27 @@
+# Syntax: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
+# Github Actions Stale: https://github.com/actions/stale
+# Description: The purpose of this build is to compile and test with redfield flag.
```
reyang, line 3: What does this description mean?

reyang (approved):

.github/workflows/stale.yml
TimothyMothra: copy paste 🤦‍♂️

Issue #2327, opened 2021-07-09
Support Http Redirect from Ingestion Service


TimothyMothra:
The Application Insights Ingestion Service has implemented a feature to redirect client SDKs from the global service endpoint to a regional endpoint.

Currently, the ingestion service is responding with [Http 307](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_redirection).

The AI SDK needs to cache this endpoint for future POSTs.

## Code Path

Endpoints are recommended to be set on the `TelemetryConfiguration.ConnectionString`.
This value is then set on the `TelemetryChannel`.

### TelemetryConfiguration

Note: TelemetrySink ctor will initialize InMemoryChannel w/ TelemetryConfiguration.

#### InMemoryChannel.Flush() 
- InMemoryTransmitter.Flush() > InMemoryTransmitter.DequeueAndSend() > InMemoryTransmitter.Send() 
- Transmission.SendAsync()

#### ServerTelemetryChannel.Initialize(TelemetryConfiguration)
- ServerTelemetryChannel.Flush() 
- TelemetryBuffer.FlushAsync() 
- TelemetrySerializer.Serialize() 
- Transmitter.Enqueue()  
- TransmissionSender.StartSending() 
- Transmission.SendAsync()

### Transmission.SendAsync

The `HttpClient` used by both `TelemetryChannels` is here:

https://github.com/microsoft/ApplicationInsights-dotnet/blob/99b87a135cc2dd6d208c3e965ab84c0e14074c97/BASE/src/Microsoft.ApplicationInsights/Channel/Transmission.cs#L166-L200

PR #2419, opened 2021-09-20
[WIP] redirect

TimothyMothra:
Fix Issue #2327

work in progress. can run unit tests here faster than on my dev box

## Changes

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/RedirectHttpHandlerTests.cs
+++ b/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/RedirectHttpHandlerTests.cs
@@ -1,0 +1,220 @@
+﻿#if NETCOREAPP
+namespace Microsoft.ApplicationInsights.TestFramework.Channel
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Net.Http;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    using Microsoft.ApplicationInsights.Channel;
+    using Microsoft.AspNetCore.Builder;
+    using Microsoft.AspNetCore.Hosting;
+    using Microsoft.AspNetCore.Http;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    [TestClass]
+    public class RedirectHttpHandlerTests
+    {
+        private const string helloString = "Hello World!";
+
+        private const string Localurl1 = "http://localhost:1111";
+        private const string Localurl2 = "http://localhost:2222";
+
+        /// <summary>
+        /// Verify behavior of HttpClient without <see cref="RedirectHttpHandler"/>.
+        /// Setup two local servers, where server #1 will redirect requests to #2.
+        /// </summary>
+        [TestMethod]
+        public async Task DefaultUseCase()
+        {
+            int counter1 = 0;
+            int counter2 = 0;
+
+            using var localServer1 = new LocalInProcHttpServer(Localurl1)
+            {
+                ServerLogic = async (httpContext) =>
+                {
+                    counter1++;
+                    httpContext.Response.StatusCode = StatusCodes.Status308PermanentRedirect;//.Status307TemporaryRedirect;
+                    httpContext.Response.Headers.Add("Location", Localurl2);
+                    await httpContext.Response.WriteAsync("redirect");
+                },
+            };
+
+            using var localServer2 = new LocalInProcHttpServer(Localurl2)
+            {
+                ServerLogic = async (httpContext) =>
+                {
+                    counter2++;
+                    await httpContext.Response.WriteAsync(helloString);
+                },
+            };
+
+            var client = new MyCustomClient(url: Localurl1);
+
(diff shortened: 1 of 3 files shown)


BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Channel/RedirectHttpHandlerTests.cs
```diff
@@ -0,0 +1,193 @@
...
+            {
+                this.uri = new Uri(url);
+                this.httpClient = (httpMessageHandler == null)
+                    ? new HttpClient()
+                    : new HttpClient(httpMessageHandler);
+            }
+
+            public async Task<string> GetAsync()
+            {
+                var result = await this.httpClient.SendAsync(new HttpRequestMessage(HttpMethod.Get, this.uri));
```
pharring, line 187: This is all very well, but Breeze uses POST (see Transmission.CreateRequestMessage). Don't you think we should test that too?
In particular, Breeze's use of cache-control on a POST is non-standard.

Also, the point of writing our own RedirectHttpHandler is to ensure that auth headers are preserved on redirect. We should have a test for that too.

BASE/Test/TestFramework/Shared/LocalInProcHttpServer.cs
```diff
@@ -25,13 +32,25 @@ public LocalInProcHttpServer(string url)
...
         }
 
+        private Task Server(HttpContext context)
+        {
+            this.RequestCounter++;
+            
+            if (this.ServerSideAsserts != null)
+            {
+                this.ServerSideAsserts(context);
+            }
```
pharring, line 49:
```suggestion
            this.ServerSideAsserts?.Invoke(context);
```

TimothyMothra: i tried that and it gave me an error! let me check again...

BASE/src/Microsoft.ApplicationInsights/Channel/RedirectHttpHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            private readonly object lockObj = new object();
+
+            private T cachedValue = default;
+
+            private DateTimeOffset expiration = DateTimeOffset.MinValue;
+
+            public bool TryRead(out T cachedValue)
+            {
+                if (DateTimeOffset.Now < this.expiration)
```
pharring, line 108:
```suggestion
                if (DateTimeOffset.UtcNow < this.expiration)
```

BASE/src/Microsoft.ApplicationInsights/Channel/RedirectHttpHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+                    return false;
+                }
+            }
+        
+            public void Set(T cachingValue, TimeSpan expire)
+            {
+                lock (this.lockObj)
+                {
+                    this.cachedValue = cachingValue;
+                    this.expiration = DateTimeOffset.Now.Add(expire);
```
pharring, line 125:
```suggestion
                    this.expiration = DateTimeOffset.UtcNow.Add(expire);
```

TimothyMothra: good catch, thanks

BASE/Test/TestFramework/Shared/LocalInProcHttpServer.cs
```diff
@@ -25,13 +32,22 @@ public LocalInProcHttpServer(string url)
...
+                    app.Run(Server);
                 })
                 .Build();
 
             Task.Run(() => this.host.RunAsync(this.cts.Token));
         }
 
+        private Task Server(HttpContext context)
+        {
+            this.RequestCounter++;
```
pharring, line 44:
```suggestion
            Interlocked.Increment(ref this.RequestCounter);
```

BASE/src/Microsoft.ApplicationInsights/Channel/RedirectHttpHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+                    return false;
+            }
+        }
+
+        private static bool TryGetRedirectVars(HttpResponseMessage httpResponseMessage, out Uri redirectUri, out TimeSpan expire)
+        {
+            expire = httpResponseMessage?.Headers?.CacheControl?.MaxAge ?? default;
+            redirectUri = httpResponseMessage?.Headers?.Location;
+
+            return expire != default && redirectUri != null && redirectUri.IsAbsoluteUri;
```
pharring, line 62: I just want to point out that setting the CacheControl on a POST is... ambiguous at best. I recognize it's what Breeze does, but I think it's non-standard. 

My understanding is that cache-control relates to the content returned, not to the Location header.
Of course, Breeze is free to use cache-control this way, if they want.

The upshot is that this class is tuned to work only with Breeze -- as long as they continue to set cache-control on POST. Specifically, though, if any other endpoint returns 308, setting the location header but not cache_control, then we won't actually follow the redirect.

Perhaps you could drop the `expire != default` clause, making the cache-control optional. You'd still follow the redirect, but wouldn't cache the resulting

PR #2425, opened 2021-09-25
[WIP] Control Document Quotas for QuickPulse

rambhatt-msft:
Fix Issue # .

## Changes

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/WEB/Src/PerformanceCollector/PerformanceCollector/CollectionConfiguration.cs
+++ b/WEB/Src/PerformanceCollector/PerformanceCollector/CollectionConfiguration.cs
@@ -67,7 +67,7 @@
             this.CreateMetadata();
 
             // create document streams based on description in info
-            this.CreateDocumentStreams(out CollectionConfigurationError[] documentStreamErrors, timeProvider, previousDocumentStreams ?? ArrayExtensions.Empty<DocumentStream>());
+            this.CreateDocumentStreams(out CollectionConfigurationError[] documentStreamErrors, timeProvider, info, previousDocumentStreams ?? ArrayExtensions.Empty<DocumentStream>());
 
             // create performance counters
             this.CreatePerformanceCounters(out CollectionConfigurationError[] performanceCounterErrors);
@@ -169,6 +169,7 @@
         private void CreateDocumentStreams(
             out CollectionConfigurationError[] errors,
             Clock timeProvider,
+            CollectionConfigurationInfo info,
             IEnumerable<DocumentStream> previousDocumentStreams)
         {
             var errorList = new List<CollectionConfigurationError>();
@@ -186,6 +187,17 @@
                         documentStream.ExceptionQuotaTracker.CurrentQuota,
                         documentStream.EventQuotaTracker.CurrentQuota,
                         documentStream.TraceQuotaTracker.CurrentQuota));
+
+            Dictionary<string, Tuple<float, float, float, float, float>> previousMaxQuotasByStreamId =
+                previousDocumentStreams.ToDictionary(
+                    documentStream => documentStream.Id,
+                    documentStream =>
+                    Tuple.Create(
+                        documentStream.RequestQuotaTracker.MaxQuota,
+                        documentStream.DependencyQuotaTracker.MaxQuota,
+                        documentStream.ExceptionQuotaTracker.MaxQuota,
+                        documentStream.EventQuotaTracker.MaxQuota,
+                        documentStream.TraceQuotaTracker.MaxQuota));
 
             if (this.info.DocumentStreams != null)
             {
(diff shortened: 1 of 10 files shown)


WEB/Src/PerformanceCollector/PerformanceCollector/CollectionConfiguration.cs
```diff
@@ -207,18 +231,24 @@ private void CreatePerformanceCounters(out CollectionConfigurationError[] errors
...
+                            initialDependencyQuota: info.InitialQuota ?? initialQuotas?.Item2,
+                            initialExceptionQuota: info.InitialQuota ?? initialQuotas?.Item3,
+                            initialEventQuota: info.InitialQuota ?? initialQuotas?.Item4,
+                            initialTraceQuota: info.InitialQuota ?? initialQuotas?.Item5,
+                            maxRequestQuota: maxQuota ?? maxQuotas?.Item1,
+                            maxDependencyQuota: maxQuota ?? maxQuotas?.Item2,
+                            maxExceptionQuota: maxQuota ?? maxQuotas?.Item3,
+                            maxEventQuota: maxQuota ?? maxQuotas?.Item4,
+                            maxTraceQuota: maxQuota ?? maxQuotas?.Item5,
+                            quotaAccrualRatePerSec: quotaAccrualRatePerSec);
```
tokaplan, line 251: What if _quotaAccrualRatePerSec_ is _null_? We have 3 values being passed in regarding quota, are all combinations valid? I suggest you validate the values somewhere at the top of this method, and make sure any error is added to the _errors_ array so that it's reported back to the user who configured it.

WEB/Src/PerformanceCollector/PerformanceCollector/QuickPulseTelemetryProcessor.cs
```diff
@@ -126,6 +129,18 @@ public void Initialize(TelemetryConfiguration configuration)
             this.RegisterSelfWithQuickPulseTelemetryModule();
         }
 
+        void IQuickPulseTelemetryProcessor.UpdateGlobalQuotas(Clock timeProvider, float? maxGlobalTelemetryQuota, float? initialGlobalTelemetryQuota, float? quotaAccrualRatePerSec)
+        {
+            if (maxGlobalTelemetryQuota != null || initialGlobalTelemetryQuota != null || quotaAccrualRatePerSec != null)
```
tokaplan, line 134: So if all of them are _null_, we just silently don't do anything? Wouldn't that be very confusing?

WEB/Src/PerformanceCollector/PerformanceCollector/QuickPulseTelemetryProcessor.cs
```diff
@@ -126,6 +129,18 @@ public void Initialize(TelemetryConfiguration configuration)
...
+            {
+                this.globalQuotaTracker = new QuickPulseQuotaTracker(
```
tokaplan, line 136: Just checking - you are aware that this.globalQuotaTracker is used concurrently by multiple threads potentially thousands of times every second, correct?

tokaplan (changes requested):
Let's talk through this and do another round of changes.

WEB/Src/PerformanceCollector/PerformanceCollector/Implementation/Service contract/CollectionConfigurationInfo.cs
```diff
@@ -12,6 +12,15 @@ internal class CollectionConfigurationInfo
         [DataMember]
         public string ETag { get; set; }
 
+        [DataMember(IsRequired = false)]
```
tokaplan, line 15: New DTO here:
- InitialQuota is optional. If it's set, we hard-reset _current_ quota when creating the new configuration. If it's not set, we use _current_ quota from the previous state of things to use as _initial_ for the new configuration. If this is the first configuration, then use a default.
- Max and AccrualRate are mandatory for the new DTO.
- The new DTO is optional. If the new DTO is not specified, then all values fall back to the default.

WEB/Src/PerformanceCollector/PerformanceCollector/QuickPulseTelemetryModule.cs
```diff
@@ -702,6 +702,18 @@ private CollectionConfigurationError[] OnUpdatedConfiguration(CollectionConfigur
         {
             // we need to preserve the current quota for each document stream that still exists in the new configuration
             CollectionConfigurationError[] errorsConfig;
+
+            lock (this.telemetryProcessorsLock)
+            {
+                foreach (var telemetryProcessor in this.TelemetryProcessors)
+                {
+                    if (telemetryProcessor is IQuickPulseTelemetryProcessor)
```
TimothyMothra, line 710: to avoid doing the parsing twice, use something like 

```
if (telemetryProcessor is IQuickPulseTelemetryProcessor qptp)
{
      qptp.UpdateGlobalQuotas()
}
```

rambhatt-msft: ah! neat!

PR #2492, opened 2021-12-09
script to unlist old beta packages

TimothyMothra:
This was an ask from Carter.

This script will query package versions and `unlist` any prerelease versions older than the latest stable.


cijothomas:
Why do we need it?

TimothyMothra:
@cijothomas I'm working with Carter to unlist all the old betas. I finished this yesterday but we'll need to run this script regularly to continue to unlist betas.
This script will probably get added to our release pipeline as a cleanup step.

rajkumar-rangaraj (approved):

PR #2512, opened 2022-01-10
Update Readme with package policy

TimothyMothra:
Update readme with support policy and package deprecation policy


Readme.md
```diff
@@ -67,6 +67,8 @@ The following packages are published from this repository:
...
 
+Customers are encouraged to use the latest stable version as that is the version that will get fixes and updates. Beta packages are not recommended for use in production & support is limited. Upon release of the new stable version, the old Beta packages will be unlisted & the old minor version will be marked as deprecated. Application Insights follows the [Azure SDK Lifecycle & support policy](https://azure.github.io/azure-sdk/policies_support.html).
```
cijothomas, line 70: please share an example of the deprecation poliucy..
if we release, says, 2.20.0, do we immediately deprecate 2.19.0?

TimothyMothra: How about...

> Example; When 2.20.0 is released, 2.20.0-beta1 will be unlisted and 2.19.0 will be deprecated.

reyang (approved):

Readme.md
```diff
@@ -67,6 +67,7 @@ The following packages are published from this repository:
...
 
+Customers are encouraged to use the latest stable version as that is the version that will get fixes and updates. Beta packages are not recommended for use in production & support is limited. Upon release of the new stable version, the old Beta packages will be unlisted & the old minor version will be marked as deprecated. Application Insights follows the [Azure SDK Lifecycle & support policy](https://azure.github.io/azure-sdk/policies_support.html). (For example: When 2.20.0 is released, 2.20.0-beta1 will be unlisted and 2.19.0 will be deprecated.)
```
reyang, line 70: Consider wrapping this line.

Readme.md
cijothomas: I am okay with the example. But I don't see such mentions in the linked Azure SDK lifecycle doc. Could you point to the relevant section in the Azure SDK policy? And Azure SDKs don't seem to be following that rule eg: https://www.nuget.org/packages/Azure.Storage.Blobs

rajkumar-rangaraj (approved):

Readme.md
cartersocha: The Azure SDK deprecation policy is a bit confusing. Attaching a [working doc](https://microsoft.sharepoint.com/:w:/r/teams/azureecosystem/servicerings/_layouts/15/Doc.aspx?sourcedoc=%7Be732bbe5-2902-47f9-8267-c7ba82f66ed9%7D&action=edit&wdPreviousSession=8bfe1f14-309d-45a7-9540-bb7f552b6a56&cid=00F256AE-68A1-47CD-824F-0B54ED157FA4) that was shared with me previously

A couple choice quotes:
"Beta – A new SDK that is available for early access and feedback purposes and is not recommended for use in production. The beta version support is limited to GitHub issues and response time is not guaranteed. Beta releases live typically for less than 1 year, after which they are either retired or released to GA." Emphasis on the last line

"Released Azure SDK packages will remain available in package managers and open source repositories indefinitely; therefore, internal retirement policy is not applicable for marking Azure SDK packages deprecated in the package manager. "  --------- Beta packages aren't considered releases per the terminology table. Minor version can NOT be deleted.

I also spoke to Elena Raikhman from the Azure Retirement Reviewer org. Pasting the summary from our meeting below:
•	Per Azure policy only the most recent minor version of an SDK is supported
o	Older versions can be marked as deprecated in the package manager
•	Deprecation announcements are only needed for breaking changes / major version upgrades
•	Beta SDKs are not officially supported by Microsoft but a team can choose to provide support depending on the preview timeline
o	Old beta packages can be deleted from the package manager once the preview has ended or the version is GA.

The Azure retirement DL is APRSupport@microsoft.com if any further clarifications are needed. 
https://azure.github.io/azure-sdk/policies_support.html

cijothomas: > Per Azure policy only the most recent minor version of an SDK is supported

Feels a bit aggressive.. Forces customers to update their packages every 2-3 months (as we typically release a stable 2-3 months).

reyang: My interpretation is a bit different. I think customers should be able to use older version, and as long as they are using the backend service (e.g. Azure), the support tickets are still valid. It's just that the bug fix will only go to the latest version, so we don't patch old version.

Issue #1186, opened 2019-08-01
TelemetryClient.Flush deadlocks


protointent:
If you are reporting bug/issue, please provide detailed Repro instructions.

## Repro Steps
1. netcoreapp2.2 console app 
2. telemetry configuration:
- ServerTelemetryChannel
- DependencyTrackingTelemetryModule and UnobservedExceptionTelemetryModule
3. call TelemetryClient.Flush on workload completion (async Task<int> Main)

##  Actual Behavior
The app deadlocks during Flush operation.
Occurrence is reliably about 1 in 8000 runs in a consistent hardware and deployment environment.

## Expected Behavior
not deadlock

## Version Info 
SDK Version : appinsights 2.10.0
.NET Version : netcore 2.2
How Application was onboarded with SDK(VisualStudio/StatusMonitor/Azure Extension) :
OS : win10-x64
Hosting Info (IIS/Azure WebApps/ etc) : console

Two stacks seem relevant for this:
Thread1
ntdll.dll!NtWaitForMultipleObjects()	Unknown
KERNELBASE.dll!WaitForMultipleObjectsEx()	Unknown
[Managed to Native Transition]	
Microsoft.ApplicationInsights.dll!Microsoft.ApplicationInsights.Metrics.Extensibility.MetricSeriesAggregatorBase<double>.UpdateAggregate(Microsoft.ApplicationInsights.Metrics.Extensibility.MetricValuesBufferBase<double> buffer)	Unknown
Microsoft.ApplicationInsights.dll!Microsoft.ApplicationInsights.Metrics.Extensibility.MetricSeriesAggregatorBase<double>.CompleteAggregation(System.DateTimeOffset periodEnd)	Unknown
Microsoft.ApplicationInsights.dll!Microsoft.ApplicationInsights.Metrics.MetricAggregationManager.GetNonpersistentAggregations(System.DateTimeOffset tactTimestamp, Microsoft.ApplicationInsights.Metrics.MetricAggregationManager.AggregatorCollection aggregators)	Unknown
[...]

PR #2595, opened 2022-05-07
fix Flush deadlock by implementing SpinWait pattern using Interlocked.CompareExchange

TimothyMothra:
Fix Issue #1186.

Metrics Flushing has a deadlock which occurs here: 
https://github.com/microsoft/ApplicationInsights-dotnet/blob/5aed71ee55973e20808b06bd06d67d96157e787b/BASE/src/Microsoft.ApplicationInsights/Metrics/Extensibility/MetricSeriesAggregatorBase.cs#L398-L411

This PR removes the lock and implements a SpinWait pattern using Interlocked.CompareExchange().
To make this work, Interlocked.CompareExchange needs access to the index variable in the class `MetricValuesBuffer`. Because of this, I've had to do some refactoring.

## Changes
- removed `lock` from the `MetricSeriesAggregatorBase` class.
- added new method `TryGetFlushIndexes()` into the `MetricValuesBuffer` class.
  This new method uses Interlocked.CompareExchange to try to advance the index variable.

## Alternatives considered
I had a PoC that replaced the lock with Monitor.TryEnter. #2594
This more closely matches the original author's intent using a lock, and adds a timeout to mitigate the deadlock.
This was abandoned in favor of removing the lock entirely.

### Checklist

### Notes for authors:

### Notes for reviewers:


TimothyMothra:
GitHub won't let Kennedy comment. He shared his comments offline:

> Won't let me leave a comment on that PR...
> But taking a look, I'm curious why a while(true) and not a for loop with a break and return false at the end?
> continue at the bottom is also not needed
> Otherwise looks good

TimothyMothra:
GitHub won't let Kennedy comment. He shared his comments offline:

> Won't let me leave a comment on that PR...
> But taking a look, I'm curious why a while(true) and not a for loop with a break and return false at the end?
> continue at the bottom is also not needed
> Otherwise looks good

PR #2612, opened 2022-06-16
fix Metric livelock by replacing potential infinate loop in MetricValuesBuffer.GetAndResetValue

TimothyMothra:
Fix Issue #1186.

This PR mitigates the potential infinite while loop in `MetricValuesBuffer` by adding an exit condition.

## Changes
- added an exit condition to `MetricValuesBuffer.GetAndResetValue()`.

## Explanation

`MetricValuesBuffer` has a section that could get stuck in an infinate loop here:
https://github.com/microsoft/ApplicationInsights-dotnet/blob/92766878547151706353cbe5a916724e53b47199/BASE/src/Microsoft.ApplicationInsights/Metrics/Extensibility/MetricValuesBuffer.cs#L90-L103

`GetAndResetValueOnce` will read a value from the buffer at an index and set reset that to `Double.NaN`.
This loop has no way to breakout if another thread has already reset this value.

This would also affect the `lock` in `MetricSeriesAggregatorBase`: 
https://github.com/microsoft/ApplicationInsights-dotnet/blob/5aed71ee55973e20808b06bd06d67d96157e787b/BASE/src/Microsoft.ApplicationInsights/Metrics/Extensibility/MetricSeriesAggregatorBase.cs#L398-L411

## Alternatives considered
My previous PR replaced the lock in MetricSeriesAggregatorBase. #2595.

However, when considering what happens when we break out... 
Breaking out of `MetricSeriesAggregatorBase` may cause the SDK to lose a complete batch of metrics.
Instead, we should only drop a single metric if breaking out of `MetricValuesBuffer`.

### Checklist

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/BASE/src/Microsoft.ApplicationInsights/Metrics/Extensibility/MetricValuesBuffer.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Metrics/Extensibility/MetricValuesBuffer.cs
@@ -81,29 +81,36 @@
 
         public TValue GetAndResetValue(int index)
         {
-            for (int spinCountdown = 10000; spinCountdown > 0; spinCountdown--)
-            {
-                TValue value = this.GetAndResetValueOnce(this.values, index);
+            TValue value = this.GetAndResetValueOnce(this.values, index);
 
-                if (this.IsInvalidValue(value))
+            if (this.IsInvalidValue(value))
+            {
+#pragma warning disable SA1129 // Do not use default value type constructor
+                var spinWait = new SpinWait();
+#pragma warning restore SA1129 // Do not use default value type constructor
+
+                value = this.GetAndResetValueOnce(this.values, index);
+                while (this.IsInvalidValue(value))
                 {
-                    if (spinCountdown % 100 == 0)
+                    spinWait.SpinOnce();
+
+                    if (spinWait.Count % 100 == 0)
                     {
-                        // In tests (including stress tests) we always finished waiting before 100 cycles.
+                        // In tests (including stress tests) we always finished wating before 100 cycles.
                         // However, this is a protection against en extreme case on a slow machine. 
                         Task.Delay(10).ConfigureAwait(continueOnCapturedContext: false).GetAwaiter().GetResult();
                     }
+                    else if (spinWait.Count > 10000)
+                    {
+                        // exceeded maximum spin count. Break out to avoid infinite loop.
+                        return this.DefaultValue;
+                    }
 
-                    continue;
-                }
-                else
-                {
-                    return value;
+                    value = this.GetAndResetValueOnce(this.values, index);
                 }
             }
 
-            // exceeded maximum spin count
-            return this.DefaultValue;
+            return value;
         }
 
         protected abstract void ResetValues(TValue[] values);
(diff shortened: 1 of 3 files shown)


BASE/src/Microsoft.ApplicationInsights/Metrics/Extensibility/MetricValuesBuffer.cs
```diff
@@ -79,31 +81,29 @@ public void ResetIndicesAndData()
 
         public TValue GetAndResetValue(int index)
         {
-            TValue value = this.GetAndResetValueOnce(this.values, index);
-
-            if (this.IsInvalidValue(value))
+            for (int spinCountdown = 10000; spinCountdown > 0; spinCountdown--)
```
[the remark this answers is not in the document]
TimothyMothra: I was trying to remove the SpinWait and while loop with a simpler for loop and a counter.

SpinWait has it's own counter which can be used here and this is a less invasive change.

BASE/src/Microsoft.ApplicationInsights/Metrics/Extensibility/MetricValuesBuffer.cs
```diff
@@ -91,13 +93,18 @@ public TValue GetAndResetValue(int index)
...
                     if (spinWait.Count % 100 == 0)
                     {
                         // In tests (including stress tests) we always finished wating before 100 cycles.
                         // However, this is a protection against en extreme case on a slow machine. 
                         Task.Delay(10).ConfigureAwait(continueOnCapturedContext: false).GetAwaiter().GetResult();
                     }
+                    else if (spinWait.Count > 10000)
+                    {
+                        // exceeded maximum spin count. Break out to avoid infinite loop.
+                        return this.InvalidValue;
```
cijothomas, line 106: can we simply return `value`?

TimothyMothra: yes, because it's already an invalid value. 🤦‍♂️
good catch!

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - [Extension methods to retrive specific operation details.](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1350)
 - [Mark Instrumentation Key based APIs as Obsolete](https://github.com/microsoft/ApplicationInsights-dotnet/issues/2560).
   - See also: https://docs.microsoft.com/azure/azure-monitor/app/migrate-from-instrumentation-keys-to-connection-strings
+- [Fix: Livelock in MetricValuesBuffer.](https://github.com/microsoft/ApplicationInsights-dotnet/pull/2612)
```
cijothomas, line 12: Link to https://github.com/microsoft/ApplicationInsights-dotnet/issues/1186 would be nice as well in the changelog

BASE/src/Microsoft.ApplicationInsights/Metrics/Extensibility/MetricValuesBuffer.cs
```diff
@@ -91,13 +91,18 @@ public TValue GetAndResetValue(int index)
...
+
                     if (spinWait.Count % 100 == 0)
                     {
                         // In tests (including stress tests) we always finished wating before 100 cycles.
                         // However, this is a protection against en extreme case on a slow machine. 
                         Task.Delay(10).ConfigureAwait(continueOnCapturedContext: false).GetAwaiter().GetResult();
                     }
+                    else if (spinWait.Count > 10000)
+                    {
+                        // exceeded maximum spin count. Break out to avoid infinite loop.
```
cijothomas, line 103: Log with Warning severity?

cijothomas (approved):
LGTM. Need some validation before stable releasing this.

rajkumar-rangaraj (approved):

PR #2634, opened 2022-07-21
fix policheck issues (SDL)

TimothyMothra:
## Changes
- remove all instances of "whitelist"/"blacklist" from source.
   These terms may have a perceived association with racial bias.
   For additional reading see: https://www.acm.org/diversity-inclusion/words-matter

### DIFF
--- a/BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Tests/Implementation/TransmissionPolicy/ErrorHandlingTransmissionPolicyTest.cs
+++ b/BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Tests/Implementation/TransmissionPolicy/ErrorHandlingTransmissionPolicyTest.cs
@@ -256,7 +256,7 @@ namespace Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implement
             [TestMethod]
             public void LogsWarningWhenDataLossIntentional()
             {
-                // ErrorHandlingTransmissionPolicy does retry only for a subset of status codes. For 
+                // ErrorHandlingTransmissionPolicy does retry only for a selected subset of status codes. For 
                 // others telemetry is dropped. This test is to validate that those are logged.
                 using (var listener = new TestEventListener())
                 {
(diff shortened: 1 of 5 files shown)


BASE/src/ServerTelemetryChannel/Implementation/TelemetryChannelEventSource.cs
```diff
@@ -581,6 +570,17 @@ public void UnexpectedBreezeResponseErrorIndexWarning(int index, string appDomai
             this.WriteEvent(79, index, this.ApplicationName);
         }
 
+        [Event(80, Message = "TransmissionDataLossError. Telemetry items are being lost here as the response code is not in the subset of retriable codes." +
```
rajkumar-rangaraj, line 573: Can't we just modify the method definition?

TimothyMothra: Are we allowed to change the method name in an EventSource class?
I'll follow up with you offline

BASE/Test/ServerTelemetryChannel.Test/TelemetryChannel.Tests/Implementation/TransmissionPolicy/ErrorHandlingTransmissionPolicyTest.cs
```diff
@@ -248,15 +248,15 @@ public void LogsDataLossEventsWhenExceptionisNull()
...
+                    // Sets flush task to failure on ignored status code
                     Assert.IsFalse(failedTransmission.IsFlushAsyncInProgress);
                 }
             }
 
             [TestMethod]
             public void LogsWarningWhenDataLossIntentional()
             {
-                // ErrorHandlingTransmissionPolicy does retry only for a whitelisted set of status codes. For 
+                // ErrorHandlingTransmissionPolicy does retry only for a subset of status codes. For
```
cijothomas, line 259:
```suggestion
                // ErrorHandlingTransmissionPolicy does retry only for a selected subset of status codes. For 
```

BASE/src/ServerTelemetryChannel/Implementation/TelemetryChannelEventSource.cs
cijothomas: we can change EventSource method names. Its not part of public API contract.

TimothyMothra: It's not the Public API but I thought it was part of the EventSource Manifest?

TimothyMothra: This is the only information I've been able to find regarding versioning:

> Best Practices
> 7. Attempt to keep EventSource object back compatible and version them appropriately. The default version for an event is 0. The version can be changed by setting [EventAttribute.Version](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventattribute.version). Change the version of an event whenever you change the data that is serialized with it. Always add new serialized data to the end of the event declaration, that is, at the end of the list of method parameters. If this isn't possible, create a new event with a new ID to replace the old one.

https://docs.microsoft.com/en-us/dotnet/core/diagnostics/eventsource-instrumentation#best-practices

PR #2635, opened 2022-07-22
Support CosmosDb dependency tracking

lmolkova:
Cosmos DB SDK ([Microsoft.Azure.Cosmos v3.TODO](https://www.nuget.org/packages/Microsoft.Azure.Cosmos) now supports dependency tracking, even when used in `Direct` (TCP) mode.

You should see `DependencyTelemetry` reported for all public API calls (that involve network communication with the service). If you use `Gateway` (HTTP) mode, you would also see dependencies for underlying HTTP calls. For `Direct` calls, you might see a few nested HTTP dependencies too, but low-level TCP calls are not traced.

Attributes reported by Cosmos SDK can be found here: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3058. They are populated as custom dimensions.

New dependencies tracing public API calls should have:
- `Target` following `{CosmosDB resource name} | {database name}` pattern, e.g. `cosmos-tracing-demo | my-store`. 
- `Name` follows `{container name} | {operation}` pattern, e.g. `orders | Read`
- `Type` matches Resource Provider namespace (`Microsoft.DocumentDB`)
- And `sdkversion` matches `rdddsaz`.

You can enable distributed tracing for Cosmos with the following API: TODO, add a link to cosmos docs.

In addition to dependency calls, Cosmos DB reports extensive diagnostics when an operation fails or takes too long. You can configure it with the following settings on the Cosmos DB client - TODO.
This diagnostics is sent as a log (`TraceTelemetry`) with `Warning` severity having `CosmosDiagnostics` as JSON in the message. You can identify it using `EventName` (in custom dimensions) with the `RecordDiagnosticsForRequests` value.
Such `TraceTelemetry` should be a child of Cosmos DB dependency telemetry described above.

## Changes

CosmosDB direct (TCP mode) dependency tracking (coming in future versions of Cosmos DB v3 SDK). 
Uses attributes outlined [here](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3058) and additional cosmos diagnostics is coming from the specific event source and is tracked as `TraceTelemetry`. 

EventSource verbosity is expected to be low - it will be controlled by CosmosDb customers on CosmosDb side. Cosmos would write it when an operation takes longer than a configurable threshold or when an error happens.

### Checklist

### Notes for reviewers:

### DIFF
--- a/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
+++ b/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
@@ -15,7 +15,7 @@
         public AzureSdkDiagnosticListenerSubscriber(TelemetryConfiguration configuration) : base(configuration)
         {
             // listen to Cosmos EventSource only - other logs can be sent using ILogger
-            this.logsListener = new AzureSdkEventListener(this.Client, EventLevel.Warning, "Azure.Cosmos");
+            this.logsListener = new AzureSdkEventListener(this.Client, EventLevel.Informational, "Azure.Cosmos");
             this.Client.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("rdd" + RddSource.DiagnosticSourceListenerAzure + ":");
         }
 
--- a/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
+++ b/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
@@ -3,16 +3,26 @@
     using System;
     using System.Diagnostics;
     using Microsoft.ApplicationInsights.Common;
+    using System.Diagnostics.Tracing;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
 
-    internal sealed class AzureSdkDiagnosticListenerSubscriber : DiagnosticSourceListenerBase<object>
+    internal sealed class AzureSdkDiagnosticListenerSubscriber : DiagnosticSourceListenerBase<object>, IDisposable
     {
         public const string DiagnosticListenerName = "Azure.";
+        private readonly IDisposable logsListener;
 
         public AzureSdkDiagnosticListenerSubscriber(TelemetryConfiguration configuration) : base(configuration)
         {
+            // listen to Cosmos EventSource only - other logs can be sent using ILogger
+            this.logsListener = new AzureSdkEventListener(this.Client, EventLevel.Warning, "Azure.Cosmos");
             this.Client.Context.GetInternalContext().SdkVersion = SdkVersionUtils.GetSdkVersion("rdd" + RddSource.DiagnosticSourceListenerAzure + ":");
+        }
+
+        public override void Dispose()
+        {
+            this.logsListener?.Dispose();
+            base.Dispose();
         }
 
         internal override bool IsSourceEnabled(DiagnosticListener diagnosticListener)
--- a/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
+++ b/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
@@ -2,8 +2,8 @@
 {
     using System;
     using System.Diagnostics;
-    using Microsoft.ApplicationInsights.Common;
     using System.Diagnostics.Tracing;
+    using Microsoft.ApplicationInsights.Common;
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
 
--- a/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
+++ b/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
@@ -15,6 +15,8 @@
 
     internal class AzureSdkDiagnosticsEventHandler : DiagnosticsEventHandlerBase
     {
+        // Microsoft.DocumentDB is an Azure Resource Provider namespace. We use it as a dependency span as-is
+        // and portal will take care about visualizing it properly.
         private const string CosmosDBResourceProviderNs = "Microsoft.DocumentDB";
 #if NET452
         private static readonly DateTimeOffset EpochStart = new DateTime(1970, 1, 1, 0, 0, 0, 0, System.DateTimeKind.Utc);
@@ -372,8 +374,23 @@
             }
 
             // similar to SqlClientDiagnosticSourceListener
-            telemetry.Target = string.Join(" | ", dbAccount, dbName);
-            telemetry.Name = string.Join(" | ", dbContainer, dbOperation);
+            telemetry.Target = JoinNullable(dbAccount, dbName);
+            telemetry.Name = JoinNullable(dbContainer, dbOperation);
+        }
+
+        private static string JoinNullable(string first, string second)
+        {
+            if (first == null)
+            {
+                return second ?? String.Empty;
+            }
+
+            if (second == null)
+            {
+                return first;
+            }
+
+            return String.Concat(first, " | ", second);
         }
 
         private static void SetMessagingProperties(Activity activity, OperationTelemetry telemetry)
--- a/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
+++ b/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
@@ -361,22 +361,22 @@
                 else if (tag.Key == "db.cosmosdb.container")
                 {
                     dbContainer = tag.Value;
-                    telemetry.Properties[tag.Key] = dbContainer;
                 }
                 else if (tag.Key == "db.cosmosdb.status_code")
                 {
                     telemetry.ResultCode = tag.Value;
                 }
-                else if (tag.Key.StartsWith("db.cosmosdb.", StringComparison.Ordinal))
+                else if (!tag.Key.StartsWith("db.cosmosdb.", StringComparison.Ordinal))
                 {
-                    telemetry.Properties[tag.Key] = tag.Value;
+                    continue;
                 }
+                
+                telemetry.Properties[tag.Key] = tag.Value;
             }
 
             // similar to SqlClientDiagnosticSourceListener
             telemetry.Target = string.Join(" | ", dbAccount, dbName);
             telemetry.Name = string.Join(" | ", dbContainer, dbOperation);
-            telemetry.Data = dbOperation;
         }
 
         private static void SetMessagingProperties(Activity activity, OperationTelemetry telemetry)
--- a/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
+++ b/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
@@ -16,6 +16,7 @@
 
     internal class AzureSdkDiagnosticsEventHandler : DiagnosticsEventHandlerBase
     {
+        private const string CosmosDBResourceProviderNs = "Microsoft.DocumentDB";
 #if NET452
         private static readonly DateTimeOffset EpochStart = new DateTime(1970, 1, 1, 0, 0, 0, 0, System.DateTimeKind.Utc);
 #endif
@@ -100,7 +101,7 @@
                                 dependency.SetOperationDetail(evnt.Value.GetType().FullName, evnt.Value);
                             }
                         }
-                        else if (dependency.Type.EndsWith(RemoteDependencyConstants.AzureCosmosDb, StringComparison.Ordinal))
+                        else if (dependency.Type == CosmosDBResourceProviderNs)
                         {
                             SetCosmosDbProperties(currentActivity, dependency);
                         }
@@ -254,10 +255,6 @@
             {
                 component = RemoteDependencyConstants.AzureServiceBus;
             } 
-            else if (component == "Microsoft.DocumentDB") 
-            {
-                component = RemoteDependencyConstants.AzureCosmosDb;
-            }
 
             if (component != null)
             {
--- a/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
+++ b/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
@@ -12,7 +12,6 @@
     using Microsoft.ApplicationInsights.Extensibility;
     using Microsoft.ApplicationInsights.Extensibility.Implementation;
     using Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing;
-    using static Microsoft.ApplicationInsights.MetricDimensionNames.TelemetryContext;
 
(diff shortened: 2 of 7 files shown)


WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
```diff
@@ -2,14 +2,24 @@
...
 
-    internal sealed class AzureSdkDiagnosticListenerSubscriber : DiagnosticSourceListenerBase<object>
+    internal sealed class AzureSdkDiagnosticListenerSubscriber : DiagnosticSourceListenerBase<object>, IDisposable
     {
         public const string DiagnosticListenerName = "Azure.";
+        private readonly IDisposable logsListener;
 
         public AzureSdkDiagnosticListenerSubscriber(TelemetryConfiguration configuration) : base(configuration)
         {
+            // listen to Cosmos EventSource only - other logs can be sent using ILogger
```
cijothomas, line 15: I understand this is draft PR, but can we just ask customers to leverage existing EventSource adapter to minimize changes here?
 
https://docs.microsoft.com/en-us/azure/azure-monitor/app/asp-net-trace-logs#use-eventsource-events

lmolkova: Good point! I'd love to, but it'd mean users would need to install and configure `Microsoft.ApplicationInsights.EventSourceListener`, while we believe the events coming through this source are essential for users to debug any cosmos-related issues. 
Adding a dependency to `Microsoft.ApplicationInsights.EventSourceListener` from DependencyCollector would be an overkill.

sourabh1007:
Why PR headline talk about Cosmos DB "Direct Mode" only?

WEB/Src/DependencyCollector/DependencyCollector/Implementation/RemoteDependencyConstants.cs
```diff
@@ -10,6 +10,7 @@ internal static class RemoteDependencyConstants
         public const string AzureTable = "Azure table";
         public const string AzureQueue = "Azure queue";
         public const string AzureDocumentDb = "Azure DocumentDB";
+        public const string AzureCosmosDb = "Azure CosmosDB";
```
lmolkova, line 13: DocumentDb is a legacy name and it'd be great to use CosmosDb instead.

TODO: need to run it by Portal team as well.

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
```diff
@@ -325,6 +337,49 @@ private static bool IsMessagingDependency(string dependencyType)
...
+                else if (!tag.Key.StartsWith("db.cosmosdb.", StringComparison.Ordinal))
+                {
+                    continue;
+                }
+                
+                telemetry.Properties[tag.Key] = tag.Value;
+            }
+
+            // similar to SqlClientDiagnosticSourceListener
+            telemetry.Target = string.Join(" | ", dbAccount, dbName);
```
sourabh1007, line 379: Anyhow Appinsight is removing documents.azure.com. We would need it

image

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
```diff
@@ -2,19 +2,29 @@
...
+            this.logsListener = new AzureSdkEventListener(this.Client, EventLevel.Warning, "Azure.Cosmos");
```
sourabh1007, line 18: Can we make Event level user configurable?

sourabh1007:
Events are loosing message while listened by appinsight..below is the log I got

_Application Insights Telemetry: {"name":"AppTraces","time":"2022-09-23T17:42:26.1395222Z","iKey":"2fabff39-6a32-42da-9e8f-9fcff7d99c6b","tags":{"ai.application.ver":"1.0.0.0","ai.cloud.roleInstance":"DESKTOP-I8MO142","ai.operation.id":"2239023b88bb44b60eab840e1983cded","ai.operation.parentId":"80155111011166ad","ai.internal.sdkVersion":"dotnetc:2.22.0-14","ai.internal.nodeName":"DESKTOP-I8MO142"},"data":{"baseType":"MessageData","baseData":{"ver":2,**"message":"n/a"**,"severityLevel":"Warning","properties":{"EventName":"WriteWarningEvent","EventId":"2","AspNetCoreEnvironment":"Development","DeveloperMode":"true","CategoryName":"Azure.Cosmos"}}}}_

Cosmosdb SDK doesn't send n/a.

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
lmolkova: yes, we can, it might still be a bit difficult for users to configure it. It seems we also have a configuration on the Cosmos SDK side and it'd be nice to have all control on one SDK side and avoid inconsistent config, what do you think?

lmolkova:
> Events are loosing diagnostics message while listened by appinsight..below is the log I got

Fixed, please check

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
sourabh1007: There is value to have this information on UI....Since we have limited space on UI..let me take the team's feedback and come back to you on this.

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
sourabh1007: At SDK side customer cannot control which type of event needs to be listened. 
SDK will emit events on the basis of listener's configuration. So, customer has to configure at AI SDK which kind of events they want to listen, and SDK will emit accordingly.
From SDK customer can only control the feature enablement.

sourabh1007:
with latest changes, this are looking good except one minor thing i.e. I am seeing cosmosdb text instead of cosmosdb icon (like we were getting previously)

image

lmolkova:
> with latest changes, things are looking good except one minor thing i.e. I am seeing cosmosdb text instead of cosmosdb icon (like we were getting previously) 

for this we need portal team to support Azure CosmosDB (in addition to DocumentDB)  - there is nothing I can do on SDK side

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
lmolkova: > At SDK side customer cannot control which type of event needs to be listened.
But they control which events are emitted with enablement flag AND thresholds, no? 

Can you please explain the scenario that you want to enable?
customer can't configure Cosmos SDK, but can  configure AI SDK?

AI SDK enablement tends to be one line of code, most customers will struggle to configure things beyond that, especially when they use codeless attach, I'd rather avoid adding extra configuration if there is no good scenario or need.

sourabh1007:
> > with latest changes, things are looking good except one minor thing i.e. I am seeing cosmosdb text instead of cosmosdb icon (like we were getting previously)
> 
> for this we need portal team to support Azure CosmosDB (in addition to DocumentDB) - there is nothing I can do on SDK side

but previously this icon was coming. in this latest package only it is not showing up

WEB/Src/DependencyCollector/DependencyCollector/Implementation/RemoteDependencyConstants.cs
lmolkova: Checked with @OsvaldoRosado - we should populate resource provider namespace - `Microsoft.DocumentDB`

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
sourabh1007: if I am not mistaken, here you are capturing only Event Level as Warning? What will happen if SDK emits events in Event level as Information? will those get captured?

sourabh1007: another question about this hardcoded "Azure.Cosmos" I am wondering after this PR we won't be able to change this name? I think we should be fine with this as we won't be able to change this name in CosmosDb SDK also.

lmolkova: Discussed with @sourabh1007 offline:
- AI SDK should listen to Info events by default. No config necessary
- Cosmos SDK owns tracing configuration (enablement, threshold to record diagnostics, etc), so it should also allow customers to enable info events.

This way all cosmos tracing configurations will stay in one place and would apply to AI SDK, OTel and custom listeners.

sourabh1007 (approved):
Looks fine to me. Thanks :)

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkEventListener.cs
```diff
@@ -0,0 +1,117 @@
...
+#else
+        private static readonly object[] EmptyArray = Array.Empty<object>();
+#endif
+
+        private readonly List<EventSource> eventSources = new List<EventSource>();
+        private readonly TelemetryClient telemetryClient;
+        private readonly EventLevel level;
+        private readonly string traitName;
+
+        public AzureSdkEventListener(TelemetryClient telemetryClient, EventLevel level, string traitName)
```
reyang, line 24: Minor, maybe s/traitName/prefix?

reyang (approved):
LGTM.

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
```diff
@@ -2,19 +2,29 @@
...
-    internal sealed class AzureSdkDiagnosticListenerSubscriber : DiagnosticSourceListenerBase<object>
+    internal sealed class AzureSdkDiagnosticListenerSubscriber : DiagnosticSourceListenerBase<object>, IDisposable
     {
         public const string DiagnosticListenerName = "Azure.";
+        private readonly IDisposable logsListener;
 
         public AzureSdkDiagnosticListenerSubscriber(TelemetryConfiguration configuration) : base(configuration)
         {
+            // listen to Cosmos EventSource only - other logs can be sent using ILogger
+            this.logsListener = new AzureSdkEventListener(this.Client, EventLevel.Informational, "Azure.Cosmos");
```
cijothomas, line 18: Are these logs fired very rarely? or they can be verbose? Since this is going to be enabled by default, we'll need ensure it wont be too noisy.

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
```diff
@@ -325,6 +333,49 @@ private static bool IsMessagingDependency(string dependencyType)
...
```
cijothomas, line 375: ever need to account for the possibility of either of this being null?

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
```diff
@@ -325,6 +333,49 @@ private static bool IsMessagingDependency(string dependencyType)
...
+            telemetry.Name = string.Join(" | ", dbContainer, dbOperation);
```
cijothomas, line 376: same as above - do we need to anticipate this being null?

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
```diff
@@ -325,6 +333,49 @@ private static bool IsMessagingDependency(string dependencyType)
...
+                {
+                    telemetry.ResultCode = tag.Value;
+                    continue;
+                }
+                else if (!tag.Key.StartsWith("db.cosmosdb.", StringComparison.Ordinal))
+                {
+                    continue;
+                }
+                
+                telemetry.Properties[tag.Key] = tag.Value;
```
cijothomas, line 371: if we are already populating db.name, and other things into target/name fields, do we need to store them again into property bag as well?

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticListenerSubscriber.cs
cijothomas: just saw that PR description answers this qn already..

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
```diff
@@ -15,6 +15,7 @@
 
     internal class AzureSdkDiagnosticsEventHandler : DiagnosticsEventHandlerBase
     {
+        private const string CosmosDBResourceProviderNs = "Microsoft.DocumentDB";
```
cijothomas, line 18: Is DocumentDB instead of CosmosDB intentionally done for UI to show icons? If yes, please add a small comment above.

cijothomas (approved):
LGTM, left few minor comments.
Given we do not have true e2e tests, would you update the PR description (or a new issue), to act as the documentation for this enhancement? To answer potential customer/support questions like:
What is the min version of CosmosDb package required?
Whether this works for TCP mode as well? Would we see duplicate if Http mode is used?
Also add a note that TraceTelemetry is produced for certain scenarios (this is unusual compared to other dependencies collected automatically)
How are target/name populated?

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
lmolkova: Since `Target` and `Name` contain concatenated values optimized for portal UX and specific for AppInsights, we want to keep the original attributes as is for custom queries - e.g. we want users to be able to query by operation across containers or across all databases in account. This way Cosmos can document their telemetry once and it'd work for any backend: Azure Monitor, Jaeger, or Splunk.

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
lmolkova: Yes, this is intentional, comment added

WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
lmolkova: it should not be possible, but agree that we should be a bit more cautions here. fixed.

lmolkova:
Thank you @reyang and @cijothomas for the review! I believe I addressed all the feedback in the code. PTAL.

As for documentation, I updated the description on this PR, just need to add a few details.

@sourabh1007, can you please help us clarify the following:
-  Exact Cosmos DB version that supports tracing
- Is there an example on how to enable and configure tracing? Can you share it or provide a link to existing samples/docs, anything? 

Thanks!

PR #2647, opened 2022-08-02
Update CONTRIBUTING.md

TimothyMothra:
### Changes
- add 460 and 461 to versions list
- add instructions for using preview versions of .NET with Visual Studio

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -21,6 +21,8 @@ To successfully build the sources on your machine, make sure you've installed th
 - Visual Studio 2022 Community, Professional or Enterprise
 - .NET SDKs (https://dotnet.microsoft.com/download)
     - .NET Framework 4.5.2
+    - .NET Framework 4.6.0
+    - .NET Framework 4.6.1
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -28,6 +28,10 @@ To successfully build the sources on your machine, make sure you've installed th
     - .NET 6
     - .NET 7
 
+Note: .NET has an annual release cycle and we include the preview version in our test matrix.
```
cijothomas, line 31: if you are here, can you also update the pre-requisites above? we need to have all .NET Frameworks . net452,nrt46,net461,net471 .. we dont list all of them now!

TimothyMothra: net452 is already included! :)

I'm discussing with Raj if we can remove net46 and net461 from this repo.

I can't find any references to net471.

cijothomas: image
These 4 I had to install to build the solution locally. Yes there is no net471

TimothyMothra: Done

PR #2675, opened 2022-09-23
Remove python related entries from sdkname

lzchen:
Removing python values as they are outdated and this document should only contain things related to dot net.


cijothomas (approved):
thanks!

rajkumar-rangaraj (approved):

PR #2684, opened 2022-09-29
Remove SDK versioning doc

reyang:
This doc seems to be outdated with some broken links and missing information.
I guess if there is still value to keep this public, it should be moved to docs.microsoft.com.


cijothomas (changes requested):
Will need to decide the new home before nuking this.

PR #2704, opened 2022-10-31
address CVE-2021-24112

sstassen-wowcorp:
## Changes
Include System.Drawing.Common 4.7.2 since we can't update System.Diagnostics.PerformanceCounter to address [CVE-2021-24112](https://github.com/advisories/GHSA-rxg9-xrhp-64gj)

This would be temporary until the issue can be addressed as discussed [here](https://github.com/dotnet/runtime/issues/64592)

### Checklist
- [ ] I ran Unit Tests locally.
- [x] CHANGELOG.md updated with one line description of the fix, and a link to the original issue if available.

For significant contributions please make sure you have completed the following items:

- [ ] Design discussion issue #
- [ ] Changes in public surface reviewed

The PR will trigger build, unit tests, and functional tests automatically. Please follow [these](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/.github/CONTRIBUTING.md) instructions to build and test locally.

### Notes for authors:
- FxCop and other analyzers will fail the build. To see these errors yourself, compile localy using the Release configuration.

### Notes for reviewers:

- We support [comment build triggers](https://docs.microsoft.com/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#comment-triggers)
  - `/AzurePipelines run` will queue all builds
  - `/AzurePipelines run <pipeline-name>` will queue a specific build

PR #2707, opened 2022-11-04
mitigate CVE-2021-24112

TimothyMothra:
Supersedes: #2704 

- https://portal.msrc.microsoft.com/en-US/security-guidance/advisory/CVE-2021-24112

Our SDK has an implicit reference to `System.Drawing.Common`:
https://github.com/microsoft/ApplicationInsights-dotnet/blob/35b1a303f0fa07b9d93f178920ec65d400f729c8/WEB/Src/PerformanceCollector/PerformanceCollector/Perf.csproj#L30-L32

- Microsoft.ApplicationInsights.PerfCounterCollector
  - System.Diagnostics.PerformanceCounter https://www.nuget.org/packages/System.Diagnostics.PerformanceCounter/4.7.0
    - System.Configuration.ConfigurationManager https://www.nuget.org/packages/System.Configuration.ConfigurationManager/4.7.0
      - System.Security.Permissions https://www.nuget.org/packages/System.Security.Permissions/4.7.0
        - System.Windows.Extensions https://www.nuget.org/packages/System.Windows.Extensions/4.7.0
          - System.Drawing.Common https://www.nuget.org/packages/System.Drawing.Common/4.7.0
					

We could mitigate this by taking a dependency on `System.Drawing.Common` v4.7.2.

Instead, we're going to upgrade `System.Diagnostics.PerformanceCounter` from v4.7.0 to v6.0.0 to get a version of `System.Drawing.Common` without a vulnerability.

### DIFF
--- a/WEB/Src/PerformanceCollector/PerformanceCollector/Perf.csproj
+++ b/WEB/Src/PerformanceCollector/PerformanceCollector/Perf.csproj
@@ -30,6 +30,17 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="1.0.0" />
     <PackageReference Include="System.Diagnostics.PerformanceCounter" Version="4.7.0" />
+    
+    <!--
+    System.Drawing.Common 4.7.0 has a vulnerability https://msrc.microsoft.com/update-guide/en-US/vulnerability/CVE-2021-24112
+    
+    This is an implicit dependency from System.Diagnostics.PerformanceCounter.
+    (System.Diagnostics.PerformanceCounter 4.7.0 > System.Configuration.ConfigurationManager 4.7.0 > System.Security.Permissions 4.7.0 > System.Windows.Extensions 4.7.0 > System.Drawing.Common 4.7.0)
+    
+    The recommended mitigation is to take an explicit dependency on System.Drawing.Common 4.7.2.
+    -->
+    <PackageReference Include="System.Drawing.Common" Version="4.7.2" />
+    
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


WEB/Src/PerformanceCollector/PerformanceCollector/Perf.csproj
```diff
@@ -30,6 +30,17 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="1.0.0" />
     <PackageReference Include="System.Diagnostics.PerformanceCounter" Version="4.7.0" />
+    
+    <!--
+    System.Drawing.Common 4.7.0 has a vulnerability https://msrc.microsoft.com/update-guide/en-US/vulnerability/CVE-2021-24112
```
cijothomas, line 35: System.Diagnostics.PerformanceCounter --> Is it possible to pick a new version of this package instead?

TimothyMothra: Yes, see my alternatives considered.

TimothyMothra: I think it's very odd for our sdk to have a "Drawing" dependency.

If we upgrade System.Diagnostics.PerformanceCounter, we would have to jump two major versions to get a version of `System.Drawing.Common` without the vulnerability. I already confirmed that this supports NetStandard2.0 which is what we need.



v4.7.0
image


v6.0.0
image

cijothomas (approved):

ryan-shelby-byte:
Is this going to be in a beta release?

TimothyMothra:
Hi @ryan-shelby-byte 
Yes, this will be in 2.22-Beta2, but we don't have a release date planned yet.

As a temporary mitigation, you can take an explicit dependency on a newer version of `System.Drawing.Common`.

PR #2712, opened 2022-11-14
Support internal and client spans from cosmos db

lmolkova:
Addressing portal team feedback on new Cosmos instrumentation:
- in Direct (TCP mode), Cosmos will set Client kind on activities. Even though Cosmos uses TCP, some rare HTTP requests will be made under these activities, so there will be a few occasional client->client spans. 
- In Gateway (HTTP mode), Cosmos will set Internal kind on activities. Users are expected to get both internal (logical) activities and nested HTTP calls.

### DIFF
--- a/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
+++ b/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
@@ -43,15 +43,15 @@
         {
             try
             {
+                if (SdkInternalOperationsMonitor.IsEntered())
+                {
+                    // Because we support AAD, we must to check if an internal operation is being caught here (type = "InProc | Microsoft.AAD").
+                    return;
+                }
+
                 var currentActivity = Activity.Current;
                 if (evnt.Key.EndsWith(".Start", StringComparison.Ordinal))
                 {
-                    if (SdkInternalOperationsMonitor.IsEntered())
-                    {
-                        // Because we support AAD, we must to check if an internal operation is being caught here (type = "InProc | Microsoft.AAD").
-                        return;
-                    }
-
                     OperationTelemetry telemetry = null;
 
                     foreach (var tag in currentActivity.Tags)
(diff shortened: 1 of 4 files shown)


WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs
```diff
@@ -102,8 +104,11 @@ public override void OnEvent(KeyValuePair<string, object> evnt, DiagnosticListen
                                 dependency.SetOperationDetail(evnt.Value.GetType().FullName, evnt.Value);
                             }
                         }
-                        else if (dependency.Type == CosmosDBResourceProviderNs)
+                        else if (dependency.Type == ClientCosmosDbDependencyType || dependency.Type == InternalCosmosDbDependencyType)
```
sourabh1007, line 107: shouldn't we have condition on _CosmosDBResourceProviderNs_. Something like...if resourcetype is DocumentDB then always set cosmosdb properties..no matter what is dependency type.

lmolkova: why? we only have those two possible and they represent a combination of kind and az.namespace (CosmosDBResourceProviderNs). So currently it covers all possible cases and if we'd ever have more cases, we'll need to change it as well. We have tests that check it.

sourabh1007: `Kind` is not deciding factor here...we are not setting anything specific to a `kind`. That's why I think it makes sense to have this condition on the basis of `az.namespace` only.

lmolkova: cosmos does not depend on kind, but code is not specific to cosmos. `depednency.Type` here is calculated based on `az.namespace`. If you believe this code has some problem, I'm happy to fix it. Otherwise, I'd prefer to keep it like this and follow the same patterns we have for EventHubs or ServiceBus.

sourabh1007: As discussed offline, It is intentional to keep this very restrictive condition to allow specific "kind"s  to set comsodb properties.
 In future, if we want to introduce a new type then appinsight sdk needs to be changed in order to support this.

sourabh1007 (approved):
looks good!!

reyang (approved):
LGTM with a changelog entry.

lmolkova:
> LGTM with a changelog entry.

thank you @reyang!  I updated the changelog.

lmolkova:
thank you @TimothyMothra !

Issue #2539, opened 2022-02-11
Use of AAD authentication with Microsoft.Extensions.Azure can cause continuous trace loop due to lack of token caching


maskati:
NuGet packages:
- Microsoft.ApplicationInsights.AspNetCore 2.20.0
- Microsoft.Extensions.Azure 1.1.1 (brings in Azure.Core and Azure.Identity)
Runtime version: net6.0
Hosting environment: Azure Web App with managed identity and associated Application Insights

### Describe the bug

Application Insights trace logs are filling up with the following four traces when using AAD authenticated configuration in combination with with [Microsoft.Extensions.Azure for dependency injection of Azure SDK clients](https://docs.microsoft.com/fi-fi/dotnet/azure/sdk/dependency-injection#register-client):
1. Azure.Identity: ManagedIdentityCredential.GetToken invoked. Scopes: [ https://monitor.azure.com//.default ]...
2. Azure.Core: Request [00000000-0000-0000-0000-000000000000] GET http://127.0.0.1:41252/MSI/token/...
3. Azure.Core: Response [00000000-0000-0000-0000-000000000000] 200 OK (00.0s)...
4. Azure.Identity: ManagedIdentityCredential.GetToken succeeded. Scopes: [ https://monitor.azure.com//.default ]...

It seems that the Application Insights SDK is actually requesting a new token from `TokenCredential` for every telemetry transmission, and is also logging the logs related to requesting this token. This is resulting in a continuous loop of transmitting token acquisition trace logs related to the previous transmission of token acquisition trace logs. This is caused by `AzureEventSourceLogForwarder` in Microsoft.Extensions.Azure [forwarding Azure SDK logs to .NET logs and Application Insights](https://docs.microsoft.com/en-us/dotnet/azure/sdk/logging#map-to-aspnet-core-logging).

A related but somewhat separate issue is that the current implementation does no token caching, which places unnecessary pressure on the application and managed identity infrastructure. The original feature request #2190 states:
> Azure.Identity implementation provides caching for most scenarios.
> Our own caching - under discussion. Guidance was to rely on Azure.Identity implementation.

However the most common scenario of using managed identities does not implement caching. See for example https://github.com/Azure/azure-sdk-for-net/issues/24046 or [ManagedIdentityCredential.GetToken](https://docs.microsoft.com/en-us/dotnet/api/azure.identity.managedidentitycredential.gettoken):
[...]

PR #2720, opened 2022-12-07
Address #2539 by caching auth token

pharring:
Fix Issue #2539.

## Changes
Cache the Authentication Token provided by CredentialEnvelope to improve performance and eliminate a continuous trace loop.

Introduce a CachedReflectionCredentialEnvelope class, derived from ReflectionCredentialEnvelope and use it when constructing the credential envelope in TelemetryConfiguration. The cached version will check if the cached auth token still has enough remaining lifetime (at least 5 minutes, by default) to be re-used.

### Checklist


pharring:
@TimothyMothra PTAL

pharring:
@TimothyMothra, PTAL

PR #2723, opened 2022-12-19
Added support to parse self-diagnostics settings from environment variables.

Yun-Ting:
Fix Issue https://github.com/microsoft/ApplicationInsights-dotnet/issues/2724 .

## Changes
Add support to parse self-diagnostics settings from environment variables.

### Checklist

- [x] Design discussion issue #

### Notes for authors:

### Notes for reviewers:

### DIFF
--- a/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
+++ b/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
@@ -1,5 +1,6 @@
 ﻿namespace Microsoft.ApplicationInsights.Extensibility.Implementation.Tracing.SelfDiagnostics
 {
+    using System;
     using System.Diagnostics;
     using System.IO;
     using System.Text;
@@ -71,6 +72,32 @@
             finally
             {
                 CleanupConfigFile();
+            }
+        }
+
+        [TestMethod]
+        public void SelfDiagnosticsConfigRefresh_ReadFromEnviornmentVariables()
+        {
+            Environment.SetEnvironmentVariable("LogDirectory", ".");
+            Environment.SetEnvironmentVariable("LogLevel", "Error");
+
+            using (var configRefresher = new SelfDiagnosticsConfigRefresher())
+            {
+                // Emitting event of EventLevel.Error
+                CoreEventSource.Log.InvalidOperationToStopError();
+
+                var filePath = configRefresher.CurrentFilePath;
+
+                int bufferSize = 512;
+                byte[] actualBytes = ReadFile(filePath, bufferSize);
+                string logText = Encoding.UTF8.GetString(actualBytes);
+                Assert.IsTrue(logText.StartsWith(MessageOnNewFileString));
+
+                // The event was captured
+                string logLine = logText.Substring(MessageOnNewFileString.Length);
+                string logMessage = ParseLogMessage(logLine);
+                string expectedMessage = "Operation to stop does not match the current operation. Telemetry is not tracked.";
+                Assert.IsTrue(logMessage.StartsWith(expectedMessage));
             }
         }
 
--- a/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
+++ b/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
@@ -18,7 +18,7 @@
             @"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{7}Z:", RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
         [TestMethod]
-        public void SelfDiagnosticsConfigRefresher_OmitAsConfigured()
+        public void SelfDiagnosticsConfigRefresher_ReadFromJson_OmitAsConfigured()
         {
             try
             {
@@ -45,7 +45,7 @@
         }
 
         [TestMethod]
-        public void SelfDiagnosticsConfigRefresher_CaptureAsConfigured()
+        public void SelfDiagnosticsConfigRefresher_ReadFromJson_CaptureAsConfigured()
         {
             try
             {
@@ -76,7 +76,25 @@
         }
 
         [TestMethod]
-        public void SelfDiagnosticsConfigRefresh_ReadFromEnviornmentVariables()
+        public void SelfDiagnosticsConfigRefresher_ReadFromEnviornmentVars_OnlyLogDirectoryWasSet()
+        {
+            Environment.SetEnvironmentVariable("LogDirectory", ".");
+
+            using (var configRefresher = new SelfDiagnosticsConfigRefresher())
+            {
+                // Emitting event of EventLevel.Error
+                CoreEventSource.Log.InvalidOperationToStopError();
+                var filePath = configRefresher.CurrentFilePath;
+
+                int bufferSize = 512;
+                byte[] actualBytes = ReadFile(filePath, bufferSize);
+                string logText = Encoding.UTF8.GetString(actualBytes);
+                Assert.IsTrue(logText.StartsWith(MessageOnNewFileString));
+            }
+        }
+
+        [TestMethod]
+        public void SelfDiagnosticsConfigRefresher_ReadFromEnviornmentVars()
         {
             Environment.SetEnvironmentVariable("LogDirectory", ".");
             Environment.SetEnvironmentVariable("LogLevel", "Error");
--- a/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
+++ b/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
@@ -76,28 +76,12 @@
         }
 
         [TestMethod]
-        public void SelfDiagnosticsConfigRefresher_ReadFromEnviornmentVars_OnlyLogDirectoryWasSet()
-        {
-            Environment.SetEnvironmentVariable("LogDirectory", ".");
-
-            using (var configRefresher = new SelfDiagnosticsConfigRefresher())
-            {
-                // Emitting event of EventLevel.Error
-                CoreEventSource.Log.InvalidOperationToStopError();
-                var filePath = configRefresher.CurrentFilePath;
-
-                int bufferSize = 512;
-                byte[] actualBytes = ReadFile(filePath, bufferSize);
-                string logText = Encoding.UTF8.GetString(actualBytes);
-                Assert.IsTrue(logText.StartsWith(MessageOnNewFileString));
-            }
-        }
-
-        [TestMethod]
         public void SelfDiagnosticsConfigRefresher_ReadFromEnviornmentVars()
         {
-            Environment.SetEnvironmentVariable("LogDirectory", ".");
-            Environment.SetEnvironmentVariable("LogLevel", "Error");
+            var key = "APPLICATIONINSIGHTS_LOG_DIAGNOSTICS";
+            //var value = "LogDirectory=C:\\home\\LogFiles\\SelfDiagnostics, FileSize=2048, LogLevel=Error";
+            var value = "LogDirectory=C:\\home\\LogFiles\\SelfDiagnostics";
+            Environment.SetEnvironmentVariable(key, value);
 
             using (var configRefresher = new SelfDiagnosticsConfigRefresher())
             {
@@ -112,10 +96,10 @@
                 Assert.IsTrue(logText.StartsWith(MessageOnNewFileString));
 
                 // The event was captured
-                string logLine = logText.Substring(MessageOnNewFileString.Length);
-                string logMessage = ParseLogMessage(logLine);
-                string expectedMessage = "Operation to stop does not match the current operation. Telemetry is not tracked.";
-                Assert.IsTrue(logMessage.StartsWith(expectedMessage));
+                //string logLine = logText.Substring(MessageOnNewFileString.Length);
+                //string logMessage = ParseLogMessage(logLine);
+                //string expectedMessage = "Operation to stop does not match the current operation. Telemetry is not tracked.";
+                //Assert.IsTrue(logMessage.StartsWith(expectedMessage));
             }
         }
 
--- a/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
+++ b/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
@@ -5,6 +5,7 @@
     using System.IO;
     using System.Text;
     using System.Text.RegularExpressions;
+    using Microsoft.ApplicationInsights.Extensibility.Implementation.Platform;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
 
     [TestClass]
@@ -76,30 +77,69 @@
         }
 
         [TestMethod]
-        public void SelfDiagnosticsConfigRefresher_ReadFromEnviornmentVars()
+        public void SelfDiagnosticsConfigRefresher_ReadFromEnviornmentVar()
         {
+
             var key = "APPLICATIONINSIGHTS_LOG_DIAGNOSTICS";
-            //var value = "LogDirectory=C:\\home\\LogFiles\\SelfDiagnostics, FileSize=2048, LogLevel=Error";
-            var value = "LogDirectory=C:\\home\\LogFiles\\SelfDiagnostics";
+            var value = "LogDirectory=C:\\home\\LogFiles\\SelfDiagnostics, FileSize=2048, LogLevel=Error";
             Environment.SetEnvironmentVariable(key, value);
 
-            using (var configRefresher = new SelfDiagnosticsConfigRefresher())
+            try
             {
-                // Emitting event of EventLevel.Error
-                CoreEventSource.Log.InvalidOperationToStopError();
+                using (var configRefresher = new SelfDiagnosticsConfigRefresher())
+                {
+                    // Emitting event of EventLevel.Error
+                    CoreEventSource.Log.InvalidOperationToStopError();
+                    var filePath = configRefresher.CurrentFilePath;
 
-                var filePath = configRefresher.CurrentFilePath;
+                    int bufferSize = 512;
+                    byte[] actualBytes = ReadFile(filePath, bufferSize);
+                    string logText = Encoding.UTF8.GetString(actualBytes);
+                    Assert.IsTrue(logText.StartsWith(MessageOnNewFileString));
 
-                int bufferSize = 512;
-                byte[] actualBytes = ReadFile(filePath, bufferSize);
-                string logText = Encoding.UTF8.GetString(actualBytes);
-                Assert.IsTrue(logText.StartsWith(MessageOnNewFileString));
+                    // The event was captured
+                    string logLine = logText.Substring(MessageOnNewFileString.Length);
+                    string logMessage = ParseLogMessage(logLine);
+                    string expectedMessage = "Operation to stop does not match the current operation. Telemetry is not tracked.";
+                    Assert.IsTrue(logMessage.StartsWith(expectedMessage));
+                }
+            }
+            finally
+            {
+                Environment.SetEnvironmentVariable(key, null);
+                PlatformSingleton.Current = null; // Force reinitialization in future tests so that new environment variables will be loaded.
+            }
+        }
 
-                // The event was captured
-                //string logLine = logText.Substring(MessageOnNewFileString.Length);
-                //string logMessage = ParseLogMessage(logLine);
-                //string expectedMessage = "Operation to stop does not match the current operation. Telemetry is not tracked.";
-                //Assert.IsTrue(logMessage.StartsWith(expectedMessage));
+        [TestMethod]
+        public void SelfDiagnosticsConfigRefresher_ConfigSetByEnviornmentVar_Wins_Over_ConfigSetByJson()
+        {
+            var key = "APPLICATIONINSIGHTS_LOG_DIAGNOSTICS";
+            var value = "LogDirectory=C:\\home\\LogFiles\\SelfDiagnostics";
+            Environment.SetEnvironmentVariable(key, value);
+            
+            CreateConfigFile();
+
+            try
+            {
+                using (var configRefresher = new SelfDiagnosticsConfigRefresher())
+                {
+                    // Emitting event of EventLevel.Error
+                    CoreEventSource.Log.InvalidOperationToStopError();
+                    var filePath = configRefresher.CurrentFilePath;
+                    StringAssert.Equals(Directory.GetParent(filePath), value);
+
+                    int bufferSize = 512;
+                    byte[] actualBytes = ReadFile(filePath, bufferSize);
+                    string logText = Encoding.UTF8.GetString(actualBytes);
+                    Assert.IsTrue(logText.StartsWith(MessageOnNewFileString));
+                }
+            }
+            finally
+            {
+                Environment.SetEnvironmentVariable(key, null);
+                PlatformSingleton.Current = null; // Force reinitialization in future tests so that new environment variables will be loaded.
+                CleanupConfigFile();
             }
         }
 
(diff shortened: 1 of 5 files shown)


BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
```diff
@@ -31,11 +36,149 @@ internal class SelfDiagnosticsConfigParser
...
+            }
+
+            return this.TryGetConfigFromJsonFile(ref logDirectory, ref fileSizeInKB, ref logLevel);
+        }
+
+        internal static bool TryGetConfigFromEnvrionmentVariable(out string logDirectory, out int fileSizeInKB, out EventLevel logLevel)
         {
             logDirectory = null;
             fileSizeInKB = 0;
             logLevel = EventLevel.LogAlways;
```
xiang17, line 71: These are not necessary, as you are going to reset each value in each `TryParse*` method.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
```diff
@@ -31,11 +36,149 @@ internal class SelfDiagnosticsConfigParser
...
+
+            /// <summary>
+            /// Parse self-diagnostics settings from tje JSON file.
+            /// </summary>
+            ConfigJson,
+        }
+
         public bool TryGetConfiguration(out string logDirectory, out int fileSizeInKB, out EventLevel logLevel)
+        {
+            if (TryGetConfigFromEnvrionmentVariable(out logDirectory, out fileSizeInKB, out logLevel))
```
xiang17, line 57: Could this method ever throw exception? What would happen if it does?

troubleshooting/ETW/Readme.md
```diff
@@ -127,43 +127,69 @@ This file will not exceed the configured max size and will be circularly overwri
...
 #### Configuration
 
-Configuration is controlled by a file named `ApplicationInsightsDiagnostics.json`.
+Configuration is controlled by a file named `ApplicationInsightsDiagnostics.json` or environment variables.
 The configuration file must be no more than 4 KiB, otherwise only the first 4 KiB of content will be read.
 
-**To enable self-diagnostics**, go to the [current working directory](https://en.wikipedia.org/wiki/Working_directory) of your process and create a configuration file.
+#### Two ways to enable self-diagnostics
+
+1. go to the [current working directory](https://en.wikipedia.org/wiki/Working_directory) of your process and create a configuration file.
```
xiang17, line 135:
```suggestion
1. Go to the [current working directory](https://en.wikipedia.org/wiki/Working_directory) of your process and create a configuration file.
```

troubleshooting/ETW/Readme.md
```diff
@@ -127,43 +127,69 @@ This file will not exceed the configured max size and will be circularly overwri
...
 
-1. `LogDirectory` is the directory where the output log file will be stored. 
+1. `LogDirectory` is the directory where the output log file will be stored.
 It can be an absolute path or a relative path to the current directory.
 
 2. `FileSize` is a positive integer, which specifies the log file size in [KiB](https://en.wikipedia.org/wiki/Kibibyte).
 This value must be between 1 MiB and 128 MiB (inclusive), or it will be rounded to the closest upper or lower limit.
-The log file will never exceed this configured size, and will be circularly rewriten.
+The log file will never exceed this configured size, and will be circularly rewriten. For the case that configuration
+was passed in via enviornment variables, the default value is `1 MiB` if this parameter was not set.
```
xiang17, line 187: Is there the code for setting default value if missing?

troubleshooting/ETW/Readme.md
```diff
@@ -127,43 +127,69 @@ This file will not exceed the configured max size and will be circularly overwri
...
 
-3. `LogLevel` is the lowest level of the events to be captured. 
+3. `LogLevel` is the lowest level of the events to be captured.
 This value must match one of the [fields](https://docs.microsoft.com/dotnet/api/system.diagnostics.tracing.eventlevel#fields) of the `EventLevel` enum.
 Lower severity levels encompass higher severity levels (e.g. `Warning` includes the `Error` and `Critical` levels).
+For the case that configuration was passed in via enviornment variables, the default value is `LogAlways` if this parameter was not set.
```
xiang17, line 192: Same question of default value for missing parameter.

troubleshooting/ETW/Readme.md
```diff
@@ -127,43 +127,69 @@ This file will not exceed the configured max size and will be circularly overwri
 
 #### Configuration
 
-Configuration is controlled by a file named `ApplicationInsightsDiagnostics.json`.
+Configuration is controlled by a file named `ApplicationInsightsDiagnostics.json` or environment variables.
 The configuration file must be no more than 4 KiB, otherwise only the first 4 KiB of content will be read.
 
-**To enable self-diagnostics**, go to the [current working directory](https://en.wikipedia.org/wiki/Working_directory) of your process and create a configuration file.
+#### Two ways to enable self-diagnostics
```
xiang17, line 133: Should this be at the same level as the `#### Configuration` above?

troubleshooting/ETW/Readme.md
```diff
@@ -127,43 +127,69 @@ This file will not exceed the configured max size and will be circularly overwri
 
 #### Configuration
 
-Configuration is controlled by a file named `ApplicationInsightsDiagnostics.json`.
+Configuration is controlled by a file named `ApplicationInsightsDiagnostics.json` or environment variables.
```
xiang17, line 130: What would happen if both are set? Or what should the customer expect to happen?

And what if one of them is of incorrect value/format?

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
```diff
@@ -74,6 +75,50 @@ public void SelfDiagnosticsConfigRefresher_CaptureAsConfigured()
             }
         }
 
+        [TestMethod]
+        public void SelfDiagnosticsConfigRefresher_ReadFromEnviornmentVars_OnlyLogDirectoryWasSet()
+        {
+            Environment.SetEnvironmentVariable("LogDirectory", ".");
```
xiang17, line 81: Do you need to unset the variable(s) once each test case is finished? Will it affect other test cases?

troubleshooting/ETW/Readme.md
```diff
@@ -127,43 +127,69 @@ This file will not exceed the configured max size and will be circularly overwri
...
+
+        // "LogLevel" is an optional parameter, i.e. if "LogDirectory" was set, the default value for LogLevel is LogAlways.
+        Environment.SetEnvironmentVariable("LogLevel", "LogAlways");
+    }
+    ```
+
+#### To disable self-diagnostics
+
+Depending on how self-diagnostcis was enabled, delete the configuration file
+or delete the enviornment variable(s).
```
xiang17, line 174: Might be always plural if you don't assign default value for missing parameters.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
```diff
@@ -9,8 +9,13 @@
     internal class SelfDiagnosticsConfigParser
     {
         public const string ConfigFileName = "ApplicationInsightsDiagnostics.json";
+
         private const int FileSizeLowerLimit = 1024;  // Lower limit for log file size in KB: 1MB
         private const int FileSizeUpperLimit = 128 * 1024;  // Upper limit for log file size in KB: 128MB
+        
+        private const string LogDirectory = "LogDirectory";
+        private const string FileSize = "FileSize";
+        private const string LogLevel = "LogLevel";
```
TimothyMothra, line 18: Are these the environment variables?

I would make these ALL CAPS to differentiate them.
Please use the prefix "APPLICATIONINSIGHTS_" or better yet "APPLICATIONINSIGHTS_LOG_"

Following the precedent we set with the connection string: https://learn.microsoft.com/en-us/azure/azure-monitor/app/sdk-connection-string?tabs=net#environment-variable

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
```diff
@@ -31,11 +36,149 @@ internal class SelfDiagnosticsConfigParser
...
+            return this.TryGetConfigFromJsonFile(ref logDirectory, ref fileSizeInKB, ref logLevel);
+        }
+
+        internal static bool TryGetConfigFromEnvrionmentVariable(out string logDirectory, out int fileSizeInKB, out EventLevel logLevel)
         {
             logDirectory = null;
             fileSizeInKB = 0;
             logLevel = EventLevel.LogAlways;
+
+            if (!TryParseLogDirectory(ParseLocation.EnviornmentVariable, Environment.GetEnvironmentVariable(LogDirectory), out logDirectory))
```
TimothyMothra, line 73: There's technically nothing wrong with `Environment.GetEnvironmentVariable`.
But this repo uses an abstraction: `PlatformSingleton.Current.TryGetEnvironmentVariable`

See this example:
https://github.com/microsoft/ApplicationInsights-dotnet/blob/a99b7d646b903df17cb6c8f6814a8bef5bc934d0/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs#L472

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
```diff
@@ -31,11 +36,149 @@ internal class SelfDiagnosticsConfigParser
...
+
+            if (!TryParseLogLevel(ParseLocation.EnviornmentVariable, Environment.GetEnvironmentVariable(LogLevel), out logLevel))
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        internal static bool TryParseLogDirectory(ParseLocation location, string val, out string logDirectory)
```
rajkumar-rangaraj, line 93: Do we need an enum `ParseLocation`?
You could call this method, if env variable is not null. This will reduce the changes to these existing methods.

rajkumar-rangaraj: Other option is to use overloaded methods.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
```diff
@@ -31,11 +36,149 @@ internal class SelfDiagnosticsConfigParser
...
+        }
+
+        internal static bool TryParseLogLevel(ParseLocation location, string val, out EventLevel logLevel)
+        {
+            logLevel = EventLevel.LogAlways;
+            if (location == ParseLocation.EnviornmentVariable)
+            {
+                if (!String.IsNullOrEmpty(val))
+                {
+                    logLevel = (EventLevel)Enum.Parse(typeof(EventLevel), val);
```
rajkumar-rangaraj, line 159: What will happen if `val` has invalid value?
Should we consider using `TryParse`?

troubleshooting/ETW/Readme.md
rajkumar-rangaraj, line 130: Specify the list of environment variables.

troubleshooting/ETW/Readme.md
```diff
@@ -127,43 +127,69 @@ This file will not exceed the configured max size and will be circularly overwri
...
-```
+    ```json
+    {
+        "LogDirectory": ".",
+        "FileSize": 1024,
+        "LogLevel": "Error"
+    }
+    ```
+
+2. For the case that users are not allowed to create the `ApplicationInsightsDiagnostics.json` config file
```
rajkumar-rangaraj, line 153: We could just say to collect logs/self-diagnostics using environment variables follow below steps. No need to call out "For the case that users are not allowed to create the `ApplicationInsightsDiagnostics.json` config file"

troubleshooting/ETW/Readme.md
```diff
@@ -127,43 +127,69 @@ This file will not exceed the configured max size and will be circularly overwri
...
+at the above mentioning locations due to hosting enviornment limiations, configuration can be set in via enviornment variables.
+
+    Example:
+
+    ```csharp
```
rajkumar-rangaraj, line 158: This example is misleading, it not mandate to set the environment variable through code.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
TimothyMothra: should run these tests in a try/catch/finally and set the vars to null

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
Yun-Ting: Yeah, I agree that the design of parsing/try-catching logic could be better but now sure how yet.

For the current design, if we don't assign the out parameters a default value, when control reaches line 75, compiler would complain that the out parameters must be assigned before control leaves the current method.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
rajkumar-rangaraj: One other option is to control the behavior through a new property or field instead of passing the hard-coded value as first parameter and validating within the method.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
```diff
@@ -74,6 +75,50 @@ public void SelfDiagnosticsConfigRefresher_CaptureAsConfigured()
...
+
+                int bufferSize = 512;
+                byte[] actualBytes = ReadFile(filePath, bufferSize);
+                string logText = Encoding.UTF8.GetString(actualBytes);
+                Assert.IsTrue(logText.StartsWith(MessageOnNewFileString));
+            }
+        }
+
+        [TestMethod]
+        public void SelfDiagnosticsConfigRefresher_ReadFromEnviornmentVars()
```
TimothyMothra, line 97: this test could be flaky.
tests often run in parallel, and this could capture a different message.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
Yun-Ting: The only place I can see now exception might happen is here:

``
logLevel = (EventLevel)Enum.Parse(typeof(EventLevel), val);
``

Shall we do try/catch in the scope of `TryGetConfigFromEnvrionmentVariable` to accommodate the exceptions that we can't foresee, or try/catch the above mentioning Enum.Parse part should be sufficient?

troubleshooting/ETW/Readme.md
Yun-Ting: yup, it's here:

https://github.com/microsoft/ApplicationInsights-dotnet/blob/5b4fd3eccc58a8dd450ea58aa0c538302f31da04/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs#L118

troubleshooting/ETW/Readme.md
Yun-Ting: And here:

https://github.com/microsoft/ApplicationInsights-dotnet/blob/5b4fd3eccc58a8dd450ea58aa0c538302f31da04/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs#LL154-L154C45

troubleshooting/ETW/Readme.md
TimothyMothra: Are we requiring that users should set all three environment variables?
I wonder if it would be easier for users if there was only one environment variable consisting of key/value pairs?
@rajkumar-rangaraj is there any recommendation/standard that we should follow?

rajkumar-rangaraj: I agree. Tt is easier to manage if it is one environment variable with key/value pairs.

Yun-Ting: Current design is as follows:

Read config from environment variables, if exception occurs while parsing the variables, `TryGetConfiguration` returns false. If environment variables are successfully parsed, i.e. `TryGetConfigFromEnvrionmentVariable` returns true, `TryGetConfiguration` returns true without executing `TryGetConfigFromJsonFile`.

For the corner case that users set config via environment variables *and* JSON file:
1. If config were set *correctly both* via environment variables and the JSON file, SDK takes the configs set by environment variables.
2. If the config were set *incorrectly* via environment variables, SDK takes the configs set by JSON file.

I intentionally leave out the description regarding the above behavior in README because my assumption is that users initially will only try one path or the other regarding config setting. The situation of configs were set via both environment variables and JSON file most likely will only happen if the users use one way, but didn't see the diagnostics file being written.

I am thinking instead of having this corner case documented in README, throwing early in code when configs are not successfully set as expected might be a more direct way to communicate with the users.

How do you feel about it?

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
Yun-Ting: Thank you! `TryGetEnviornmentVariable` is better as it also checks if the variable is empty.
Will update accordingly.

troubleshooting/ETW/Readme.md
Yun-Ting: key/value pairs sounds better!!
I am not sure what should be a proper name for holding the pairs.
"SelfDiagnosticsConfig"?

TimothyMothra: The name of the config file is "ApplicationInsightsDiagnostics".
So to keep it similar, I would vote for "APPLICATIONINSIGHTS_DIAGNOSTICS".

xiang17: The design is good. I meant to say it's better to have a description in README regarding which take precedence. It could be a short note like this in the [Application Insights guide](https://learn.microsoft.com/en-us/azure/azure-monitor/app/asp-net-core?tabs=netcore6):

> An connection string specified in code wins over the environment variable APPLICATIONINSIGHTS_CONNECTION_STRING, which wins over other options.

Another example [here](https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-config#connection-string).

It could be in the `#### Two ways to enable self-diagnostics` section or wherever you see fit.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
xiang17: By the way, `Environment.GetEnvironmentVariable` could also throw exceptions.

Yun-Ting: I've added try/catch to `TryGetConfigFromEnvrionmentVariable`.
Currently, unlike in `TryGetConfigFromJsonFile`, I use `throw` when there is an exception because I think it would be nice to get users attention when configs set via environment variables are not being parsed correctly.

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
Yun-Ting: I've added try/finally to reset the environment variables for the tests; thank you

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
Yun-Ting: updated as suggested

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
xiang17: Be careful of in what scenarios this happens. It could be in production code. For the local file configuration approach, if the user accidentally has a bad config, their app will crash because of it. I'm not sure about the environment variable.

The question boils down to: "can the app be crashed while running because of user/admin modified an environment variable?"

Yun-Ting: Ah I see what you mean! 
Maybe we can write some logs for this case instead of throwing.
I'm just not sure where to write the exception logs...

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
Yun-Ting: I've removed the Enum. Please checkout the current design. Thank you!

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
toddfoust: Hey team, one question I have.... can we toggle environment variable values of a running process? Say customer has a production app, I want to simply start collecting self-diagnostic logs, then find some way to set that environment variable for the host environment, will the tracing start logging within 10 seconds? Then when I want to stop that tracing, can I just adjust the env var and see tracing stop within 10 seconds? This behavior works when we rely on reading these settings from Json file, but I'm not certain if a "refresh" of environment variables requires a process restart each time? My original goal was to have env var that deteremines what path the SDK will poll every 10 seconds for the json file with those settings. Does this direct environment variable settings, where you logDirectory, file size, loglevel still give us an opportunity to turn on/off tracing of this production running app without restarts?

troubleshooting/ETW/Readme.md
Yun-Ting: I've added an extra `#` to lower the level of the title.

troubleshooting/ETW/Readme.md
Yun-Ting: I've added a note regarding your suggestion. Thank you for the examples.

troubleshooting/ETW/Readme.md
Yun-Ting: updated

troubleshooting/ETW/Readme.md
Yun-Ting: updated

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigRefresherTest.cs
Yun-Ting: I've added logic to clear the environment variable and force initialize of PlatformSingleton to address the concern. Thanks!

troubleshooting/ETW/Readme.md
Yun-Ting: I've updated to use a single environment variable to control the config. Thanks!

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Tracing/SelfDiagnostics/SelfDiagnosticsConfigParser.cs
Yun-Ting: I will collaborate with Rohit in testing the above scenarios and update here.

Issue #2850, opened 2024-02-28
Dependencies to unsupported and end of life nuget packages and frameworks


antymon4o:
Application Insights projects target frameworks net452 and net46 that are now not supported any more.
List of currently supported framework versions can be seen at: https://dotnet.microsoft.com/en-us/learn/dotnet/what-is-dotnet-framework.

Additionally, the project Microsoft.ApplicationInsights.AspNetCore has a dependency on packages Microsoft.AspNetCore.Http and Microsoft.AspNetCore.Hosting that are deprecated. Not only these packages are deprecated, but the packages that these packages depend are also deprecated. 
This brings the implication that when some project depends on Microsoft.ApplicationInsights.AspNetCore version 2.22.0, then this project also gets the dependencies to packages that are deprecated and are no longer maintained.

The proposed solution for the first issue is to remove the unsupported targeted frameworks.
For the deprecated AspNetCore packages, the proposal is to migrate the reference from nuget package to framework reference. This also requires a pop-up in the targeted framework from netstandard2.0 to net6.0.

MichaCo:
Totally agree, but this is probably a duplicate of #2811

antymon4o:
Yes, it seems so.
I have already proposed a solution for these issues. The PR is waiting to be reviewed.
I hope someone from the team will be able to look at it and after if some tunings are required, the issues can be resolved.

zhimin-z:
This issue is stale because it has been open 300 days with no activity. Remove stale label or this will be closed in 7 days. Commenting will instruct the bot to automatically remove the label.

Issue #2811, opened 2023-09-07
Microsoft.ApplicationInsights.AspNetCore used deprecated NuGet packages


MarcoK80:
Is your feature request related to a problem? Please describe the problem.
Dotnet list package --outdated / --deprecated shows for the packages

Microsoft.ApplicationInsights.AspNetCore 2.21.0
Microsoft.ApplicationInsights.Profiler.AspNetCore 2.5.3
Microsoft.ApplicationInsights.Profiler.Core 2.5.3

outdated and if we update them deprecated packages
Microsoft.AspNetCore.Hosting
Microsoft.AspNetCore.Http

Is there any plan to migrate these packages to full .net core 7 or .net core 8 support?

Describe the solution you'd like
Support native .net core 7 and .net core 8

Additional context
No response

agehrke:
> Taking a dependency on Microsoft.AspNetCore.Hosting v2.2.0 would resolve this issue, but would also break support for NetCore v2.1.
  Instead I'm taking a direct dependency on the fixed version Microsoft.AspNetCore.Http.
  We can remove this when NetCore v2.1 reaches EOL on August 21, 2021.

[A comment inside Microsoft.ApplicationInsights.AspNetCore.csproj](https://github.com/microsoft/ApplicationInsights-dotnet/blob/c420e04562876791f27a61644760d7b7512832d9/NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj#L42). Seems like the comment was long forgotten.

patelriki13:
Hi @TimothyMothra

Any updates on this? Any ETA?

because `Microsoft.AspNetCore.Http` package has vulnerability [CVE-2020-1045](https://dtrack.mm-software.com/vulnerabilities/NVD/CVE-2020-1045)

https://github.com/microsoft/ApplicationInsights-dotnet/issues/2199

`We can remove this when NetCore v2.1 reaches EOL on August 21, 2021.`

So any ETA?

MichaCo:
Any update / plans on this @TimothyMothra ?
I would expect the library to multi target Net6 and Net8 by now to include the proper AspNetCore framework dependencies instead of totally outdated Nuget refrences.

My team is using App insights in .NET Core for microservices for years now and with the lasted changes how Net6/8 references framework dependencies, it is really painful to include this Nuget package with all the old/outdated dlls which all end up in every service's bin / publish dir... 
[...]

PR #2854, opened 2024-03-05
Remove target frameworks net452;net46; because they are no longer sup…

antymon4o:
…ported #2850

use frameworkreference for Microsoft.AspNetCore.App instead of icrosoft.AspNetCore.* nuget packages, because they are now deprecated and end of life.

Fix Issue #2850.

## Changes
Remove target frameworks net452;net46; and netcoreapp3.1; because they are no longer supported.
Affected projects are: DependencyCollector, TelemetryChannel, Perf, WindowsServer, Microsoft.ApplicationInsights.
Use frameworkreference for Microsoft.AspNetCore.App instead of icrosoft.AspNetCore.* nuget packages, because they are now deprecated and end of life.
Microsoft.ApplicationInsights.AspNetCore targets net6 to enable framework reference to Microsoft.AspNetCore.App.

### Checklist

### Notes for authors:

### DIFF
--- a/.props/Test.props
+++ b/.props/Test.props
@@ -20,7 +20,7 @@
           - net7.0 (GA Nov 2022)
     -->
 
-    <SupportedFrameworks_NetCore>net8.0;net7.0;net6.0;netcoreapp3.1;</SupportedFrameworks_NetCore>
+    <SupportedFrameworks_NetCore>net8.0;net7.0;net6.0;</SupportedFrameworks_NetCore>
     <SupportedFrameworks_NetFx Condition="$(OS) == 'Windows_NT'">net481;net480;net472;net462;</SupportedFrameworks_NetFx>
     <LegacyFrameworks_NetFx Condition="$(OS) == 'Windows_NT'">net46;net452;</LegacyFrameworks_NetFx>
     
--- a/.props/_GlobalStaticVersion.props
+++ b/.props/_GlobalStaticVersion.props
@@ -11,9 +11,9 @@
       Update for every public release.
     -->
     <SemanticVersionMajor>2</SemanticVersionMajor>
-    <SemanticVersionMinor>22</SemanticVersionMinor> <!-- If changing the Minor version, also update the Date value. -->
+    <SemanticVersionMinor>23</SemanticVersionMinor> <!-- If changing the Minor version, also update the Date value. -->
     <SemanticVersionPatch>0</SemanticVersionPatch>
-    <PreReleaseMilestone></PreReleaseMilestone> <!--Valid values: beta1, beta2, EMPTY for stable -->
+    <PreReleaseMilestone>beta1</PreReleaseMilestone> <!--Valid values: beta1, beta2, EMPTY for stable -->
     <PreReleaseMilestone Condition="'$(Redfield)' == 'True'">redfield</PreReleaseMilestone>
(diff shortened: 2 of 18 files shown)


TimothyMothra (changes requested):
Although I would love to see this change, we have a requirement to continue supporting these frameworks for same legacy systems.

antymon4o:
Thank you for you comment, @TimothyMothra.
I perfectly understand that dropping support for specific framework version is not an easy decision to make.
In the company I work, we have been using .NET Framework since version 1.1 for the systems we are developing. And as you can imagine, we have tons of legacy code. This is pretty normal for enterprise software that is evolving for decades. During the years we have been upgrading to the proper supported version of .NET Framework and .NET Core.
But there is a difference between having legacy code and running it on unsupported platform. The former is normal, but the latter is unacceptable if only for security and compliance reasons.
Of course, all the versions of ApplicationInsights packages up to 2.22 will always support net452 and net46. And thus if any legacy systems need them, they can use them.
As we cannot tolerate dragging dependencies of unsupported and legacy frameworks and packages, the need is for a version of ApplicationInsights that is dependent on current, supported packages and frameworks.
Given that there is no code change, and thus no behavior change in the library, what is the problem to release this new version?

MichaCo:
> Although I would love to see this change, we have a requirement to continue supporting these frameworks for same legacy systems.

@TimothyMothra I'd suggest a major version release across the board (3.0).

There is a huge difference between AspNetCore 2.x/3.x and the latest versions under NET6/8 in terms of what nugets to reference vs using Framework dependencies...

Everything prior NET6 is long out of LTS anyways, so, why not just clean up the mess and publish a clean version for the new .NET 6/8 versions?

antymon4o:
In Microsoft.ApplicationInsights project, there are a lot of comments like:

> // .NET 4.5.2 have a custom implementation of RichPayloadEventSource

And in `RichPayloadEventSource.cs`  there is a conditional compilation `#if !NET452`.

The question is: 
Is this implementation specific to NET452 only or is it specific to all framework versions in general?

If this implementation should be used in all framework versions, then this condition should be changed.

pharring:
@antymon4o I believe the conditional in RichPayloadEventSource.cs is correct. That's because support for anonymous types in EventSource events was introduced in .NET 4.6. There's a comment starting on line 393 of that file explaining it.

antymon4o:
@TimothyMothra can you please have another look at the PR and tell if the support for the deprecated and LTS out frameworks is critical.
Is there any chance the proposed changes to be merged in a new major version 3.0 of ApplicationInsigts?

cijothomas:
> Is there any chance the proposed changes to be merged in a new major version 3.0 of ApplicationInsigts?

Sorry, there are no plans for a 3.0 Application Insights! However, dropping target framework can be done without major version bump! I am not suggesting we merge this PR and do normal release - I'll defer to @TimothyMothra to decide if this is acceptable or not.

It maybe worth exploring/fixing `Microsoft.ApplicationInsights.AspNetCore` package alone to fix the original issue described in https://github.com/microsoft/ApplicationInsights-dotnet/issues/2850, more precisely raised in https://github.com/microsoft/ApplicationInsights-dotnet/issues/2811

antymon4o:
It is possible to fix only ApplicationInsights.AspNetCore package, but it would be a somewhat partial fix.
Actually, the fix for this package is part of this PR. If it is useful I can post only them as a part of a new PR. 
I just need to know if it is definite that this PR is rejected? Who can confirm this and the accepted/approved approach to fix the issues?

antymon4o:
I have created e new PR #2860 with only the changes to `Microsoft.ApplicationInsights.AspNetCore` package and linked it to #2811.

Issue #1066, opened 2019-01-19
Exception serialization does not handler Empty exception message.


cijothomas:
Oringally reported:https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/823
Application Insights Rejecting Exceptions with empty Message

Repro:
 TelemetryClient tc = new TelemetryClient();
            tc.Context.InstrumentationKey = "validikey";
            tc.TrackException(new Exception(""));
            tc.Flush();

Response from backend:
 Field 'message' on type 'ExceptionDetails' is required but missing or empty. Expected: string, Actual: undefined

// The following gets mesage automatically populated, but passing "" or string.empty breaks it.
var ex = new Exception();
tc.TrackException(ex); 

Fix would be to modify Exception sanitization to handle this properly.

kristofferjalen:
This bug is still there.

NeilMountford:
This is still an issue, I've created a PR with a unit test: #2857 that will hopefully solve it.

PR #2697, opened 2022-10-20
Populate required string value for property Message

kristofferjalen:
Fix Issue #1066.

**Note: This change requires careful review. Also, I have not run the tests, since it's [not possible to build the projects locally](https://github.com/microsoft/ApplicationInsights-dotnet/issues/2178).**

## Changes
Application Insights [is rejecting exceptions](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1066) with `Exception.Message` property being empty or null, as illustrated by an example:

```c#
var configuration = TelemetryConfiguration.CreateDefault();
configuration.ConnectionString = "InstrumentationKey=...";
var telemetryClient = new TelemetryClient(configuration);
telemetryClient.TrackException(new Exception("")); // ignored
telemetryClient.TrackException(new Exception("foo")); // logged
telemetryClient.Flush();
```
Response:
```json
{
    "itemsReceived": 1,
    "itemsAccepted": 0,
    "errors":
    [
        {
            "index": 0,
            "statusCode": 400,
            "message": "109: Field 'message' on type 'ExceptionDetails' is required but missing or empty. Expected: string, Actual: undefined"
        }
    ]
}
```

### Checklist
- [ ] I ran Unit Tests locally.
- [ ] CHANGELOG.md updated with one line description of the fix, and a link to the original issue if available.

For significant contributions please make sure you have completed the following items:

- [ ] Design discussion issue #
- [ ] Changes in public surface reviewed

The PR will trigger build, unit tests, and functional tests automatically. Please follow [these](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/.github/CONTRIBUTING.md) instructions to build and test locally.

### Notes for authors:
- FxCop and other analyzers will fail the build. To see these errors yourself, compile localy using the Release configuration.

### Notes for reviewers:

- We support [comment build triggers](https://docs.microsoft.com/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#comment-triggers)
  - `/AzurePipelines run` will queue all builds
  - `/AzurePipelines run <pipeline-name>` will queue a specific build

PR #2857, opened 2024-03-13
Populate required field Message with n/a if it is empty

NeilMountford:
Fix Issue #1066.

## Changes
Application Insights rejects exceptions that have empty messages. This means that some exceptions are lost, making investigation of runtime issues difficult. I noticed this when investigating an issue in a .NET 8 API, where a `CosmosException` for a conflict occurred, was logged on the console but wasn't logged in App Insights.

There was previously a PR #2697 that never got merged due to lack of tests.

Original example from the other PR:
```csharp
var configuration = TelemetryConfiguration.CreateDefault();
configuration.ConnectionString = "InstrumentationKey=...";
var telemetryClient = new TelemetryClient(configuration);
telemetryClient.TrackException(new Exception("")); // ignored
telemetryClient.TrackException(new Exception("foo")); // logged
telemetryClient.Flush();
```

Original response from the other PR:
```json
{
    "itemsReceived": 1,
    "itemsAccepted": 0,
    "errors":
    [
        {
            "index": 0,
            "statusCode": 400,
            "message": "109: Field 'message' on type 'ExceptionDetails' is required but missing or empty. Expected: string, Actual: undefined"
        }
    ]
}
```

### Checklist
- [✅] I ran Unit Tests locally.*
- [✅] CHANGELOG.md updated with one line description of the fix, and a link to the original issue if available.

* I had some test failures on a fresh clone, but the tests for the relevant part of the code were passing, so I'm hoping it's just an issue with some dependencies on my machine.

- [N/A] Design discussion issue #
- [N/A] Changes in public surface reviewed

### Notes for authors:

### DIFF
--- a/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/ExceptionConverterTest.cs
+++ b/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/ExceptionConverterTest.cs
@@ -191,7 +191,7 @@
         [DataRow(" ")]
         public void ProvidesDefaultMessage(string message)
         {
-            var exp = new Exception(string.Empty);
+            var exp = new ExceptionWithMessageOverride(message);
 
             ExceptionDetails expDetails = ExceptionConverter.ConvertToExceptionDetails(exp, null);
 
@@ -224,6 +224,21 @@
             }
 
             throw new AggregateException("exception message");
+        }
+        
+        /// <summary>
+        /// Overrides the Message property of the base exception, so that null messages are
+        /// returned as null, rather than a default, e.g. "Exception of type 'System.Exception' was thrown".
+        /// </summary>
+        private class ExceptionWithMessageOverride : Exception
+        {
+            public ExceptionWithMessageOverride(string message)
+                : base(message)
+            {
+                Message = message;
+            }
+            
+            public override string Message { get; }
         }
     }
 }
--- a/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/ExceptionConverterTest.cs
+++ b/BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/ExceptionConverterTest.cs
@@ -185,6 +185,19 @@
             Assert.AreEqual(5, expDetails.message.Length);
         }
 
+        [TestMethod]
+        [DataRow(null)]
+        [DataRow("")]
+        [DataRow(" ")]
+        public void ProvidesDefaultMessage(string message)
+        {
+            var exp = new Exception(string.Empty);
+
+            ExceptionDetails expDetails = ExceptionConverter.ConvertToExceptionDetails(exp, null);
+
+            Assert.AreEqual("n/a", expDetails.message);
+        }
+
         [MethodImplAttribute(MethodImplOptions.NoInlining)]
         private Exception CreateException(int numberOfStackpoints)
         {
--- a/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/External/ExceptionDetailsImplementation.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/External/ExceptionDetailsImplementation.cs
@@ -1,8 +1,7 @@
-﻿using Microsoft.ApplicationInsights.DataContracts;
-
-namespace Microsoft.ApplicationInsights.Extensibility.Implementation.External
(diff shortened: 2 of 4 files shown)


NeilMountford:
@microsoft-github-policy-service agree

cijothomas (approved):
Thanks, this looks good!

Will ask one more owner to review before merge, as we are adding limited additions (including fixes) to this repo.

NeilMountford:
Thanks @cijothomas, I've pushed a further change to move `using Microsoft.ApplicationInsights.DataContracts;` into the namespace block to make the build happy for the net4* builds - apologies for not catching this locally, I don't have access to a Windows machine currently. Could you approve the workflows again please? 🙂

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/ExceptionConverterTest.cs
```diff
@@ -185,6 +185,19 @@ public void DoesNotTrimShortExceptionMessages()
             Assert.AreEqual(5, expDetails.message.Length);
         }
 
+        [TestMethod]
+        [DataRow(null)]
+        [DataRow("")]
+        [DataRow(" ")]
+        public void ProvidesDefaultMessage(string message)
```
TimothyMothra, line 192: What's the intention of this parameter? I don't see it used.

rajkumar-rangaraj, line 192: `message` parameter is not used within this method.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/External/ExceptionDetailsImplementation.cs
```diff
@@ -21,7 +22,7 @@ internal static ExceptionDetails CreateWithoutStackInfo(Exception exception, Exc
             {
                 id = exception.GetHashCode(),
                 typeName = exception.GetType().FullName,
-                message = exception.Message,
+                message = Utils.PopulateRequiredStringValue(exception.Message, "message", typeof(ExceptionTelemetry).FullName),
```
TimothyMothra, line 25: nit: note that this helper method is using `string.IsNullOrEmpty`. 
https://github.com/microsoft/ApplicationInsights-dotnet/blob/bd8b0c25712e6f3c5275ae06b0f0e9331340fee4/BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/Utils.cs#L44-L49

I see in your test cases you passed in a string with a space `" "`. Note that this helper method won't catch this.

TimothyMothra (changes requested):
please address comments

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/Extensibility/Implementation/ExceptionConverterTest.cs
TimothyMothra: my comment is a dupe of https://github.com/microsoft/ApplicationInsights-dotnet/pull/2857#discussion_r1526645511

NeilMountford: My bad, started with just an empty string test then decided to add `DataRow`s to cover null and whitespace too. I've fixed the test to use the parameter now.

BASE/src/Microsoft.ApplicationInsights/Extensibility/Implementation/External/ExceptionDetailsImplementation.cs
NeilMountford: I fixed the test to use the parameter and it failed for the null and whitespace examples. 

- The null failure was due to the `Message` property on `Exception` returning a default if the `_message` field is null. I've added a custom exception for the test that overrides Message and returns the provided message as is.
- The empty string failure was down to the `.IsNullOrEmpty` so I've added a new, similar helper method that instead uses the provided `IsNullOrWhitespace` extension. In cases where it is null/whitespace I've used the same logging method, but can add another slightly different one with a different EventId if you prefer?

TimothyMothra (approved):
LGTM

NeilMountford:
@TimothyMothra What are the next steps to get this merged/released? Let me know if you need anything from me 🙂

jiilaa:
So, half a year since this fix was merged (and even longer since the latest release for this library), maybe a good time for a release? :)

asos-AdamCox:
@TimothyMothra / @cijothomas are chance of a release? We've just spent an entire day chasing this bug down only to find it was found in 2019, fixed 7 months ago and still not released...

PR #2860, opened 2024-03-28
Target Microsoft.ApplicationInsights.AspNetCore to netcoreapp3.1 and remove package references to deprecated nugets Microsoft.AspNetCore.*

antymon4o:
Fix Issue #2811 .

## Changes
- Target Microsoft.ApplicationInsights.AspNetCore to netcoreapp3.1
- Remove package references to nugets Microsoft.AspNetCore.*
- For AspNetCore use framework reference to Microsoft.AspNetCore.App

### Checklist
- [ X ] I ran Unit Tests locally.
- [ X ] CHANGELOG.md updated with one line description of the fix, and a link to the original issue if available.

### DIFF
--- a/NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
+++ b/NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
@@ -3,9 +3,9 @@
 
   <PropertyGroup>
     <AssemblyName>Microsoft.ApplicationInsights.AspNetCore</AssemblyName>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
 
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netstandard2.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netcoreapp3.1</TargetFrameworks>
     <DefineConstants>$(DefineConstants);AI_ASPNETCORE_WEB;</DefineConstants>
   </PropertyGroup>
 
@@ -32,32 +32,10 @@
     <ProjectReference Include="..\..\..\WEB\Src\WindowsServer\WindowsServer\WindowsServer.csproj" />
     <ProjectReference Include="..\..\..\WEB\Src\EventCounterCollector\EventCounterCollector\EventCounterCollector.csproj" />
     <ProjectReference Include="..\..\..\LOGGING\src\ILogger\ILogger.csproj" />
-    
-    <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="2.1.1" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.0" />
   </ItemGroup>
 
   <ItemGroup>
-    <!--
-    Microsoft.AspNetCore.Http has a vulnerability https://msrc.microsoft.com/update-guide/vulnerability/CVE-2020-1045
-    System.Text.Encodings.Web has a vulnerability https://msrc.microsoft.com/update-guide/vulnerability/CVE-2021-26701 
-    
-    These are both implicit dependencies from Microsoft.AspNetCore.Hosting.
-    (Microsoft.AspNetCore.Hosting > Microsoft.AspNetCore.Http)
-    (Microsoft.AspNetCore.Hosting > Microsoft.AspNetCore.Hosting.Abstractions > Microsoft.AspNetCore.Http.Abstractions > System.Text.Encodings.Web)
-    -->
-    
-    <!--
-    Taking a dependency on Microsoft.AspNetCore.Hosting v2.2.0 would resolve this issue, but would also break support for NetCore v2.1.
-    Instead I'm taking a direct dependency on the fixed version Microsoft.AspNetCore.Http.
-    We can remove this when NetCore v2.1 reaches EOL on August 21, 2021.
-    -->
-    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.1.22" />
-
-    <!-- 
-    We must take a temporary dependency on this newer version until Microsoft.AspNetCore.Hosting updates their dependencies.
-    -->
-    <PackageReference Include="System.Text.Encodings.Web" Version="4.7.2" />
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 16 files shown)


antymon4o:
@cijothomas @TimothyMothra, could you please take a look at these changes and say if they are OK with you considerations about still supporting legacy frameworks?

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
```diff
@@ -3,9 +3,9 @@
 
   <PropertyGroup>
     <AssemblyName>Microsoft.ApplicationInsights.AspNetCore</AssemblyName>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
```
cijothomas, line 6: won't removing netstandard2.0 lose support for asp.net core 2.1 a on .NET framework?

https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#servicing
image

or is that already unsupported?

NETCORE/src/Microsoft.ApplicationInsights.WorkerService/Microsoft.ApplicationInsights.WorkerService.csproj
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup>
     <AssemblyName>Microsoft.ApplicationInsights.WorkerService</AssemblyName>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net462;netstandard2.0</TargetFrameworks>
```
cijothomas, line 6: this, is unrelated?

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
antymon4o: ASP.NET Core 2.1 is the de facto version that is already deprecated and no longer supported. 
Removing dependency to deprecated ASP.NET Core 2.1 nuget packages is the goal of this PR.

NETCORE/src/Microsoft.ApplicationInsights.WorkerService/Microsoft.ApplicationInsights.WorkerService.csproj
antymon4o: It seems that there isn't, but with the changes to AI.AspNetCore project some tests for AI.WorkerService.Tests for framework 4.* failed. 
There isn't direct dependency between AI.AspNetCore and AI.WorkerService, But...
There was runtime error when executing the tests with this exception:

`System.IO.FileLoadException: 'Could not load file or assembly 'System.Security.Principal.Windows, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)'
   at Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Shared.Implementation.WindowsIdentityProvider.Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Shared.Implementation.IIdentityProvider.GetName() in X:\github\ApplicationInsights-dotnet\BASE\src\ServerTelemetryChannel\Managed\Shared\Implementation\WindowsIdentityProvider.cs:line 15
   at Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.Implementation.ApplicationFolderProvider.CreateTelemetrySubdirectory(DirectoryInfo root) in X:\github\ApplicationInsights-dotnet\BASE\src\ServerTelemetryChannel\Implementation\ApplicationFolderProvider.cs:line 271`

To fix the test, adding net462 to AI.WorkerService was required.
And it seems logical to have this specific target for AI.WorkerService, because AI.WorkerService depends on TelemetryChannel which has two target frameworks net452, netstandard2.0 and has conditional compilation statements dependent on NET452/NETFRAMEWORK.

I cannot give an explanation why the tests were passing before. Maybe it was coincidence because this assembly System.Security.Principal.Windows was copied from other dependencies (AspNet.Core) in test output folder without direct dependency, even though the code required them.

NETCORE/src/Microsoft.ApplicationInsights.AspNetCore/Microsoft.ApplicationInsights.AspNetCore.csproj
cijothomas: its a bit more complex... ASP.NET Core 2.1 is deprecated, however, ASP.NET Core 2.1 on .NET Framework is not...
I believe this is still a supported version in this repo, so won't be possible to remove it.

Also, given almost all of new investments are occurring only in OpenTelemetry, it may not be desired to spend much time on this repo. Of course, if there are security issues, it should be promptly addressed.

antymon4o: You are correct. ASP.NET Core 2.1.x is supported on .NET Framework.
Given that MS split the support for ASP.NET Core on .NET Core and .NET Framework, maybe a better approach is to also split the targets of AI.AspNetCore - for net462 and netcoreapp3.1:

- For net462, AI.AspNetCore will use nuget packages for ASP.NET Core 2.1.x
- For netcoreapp3.1, AI.AspNetCore will use framework reference to  Microsoft.AspNetCore.App.

This way AI.AspNetCore will continue to support ASP.NET Core 2.1.x on .NET Framework through the nuget package.

How does this sound to you?

cijothomas: That might be correct. Unfortunately, I do not have enough cycles to think through all the implications of that. 

@TimothyMothra can make a final call if we want to proceed with this.

@antymon4o Apologies in advance, if we chose to not proceed with this. Almost all investments are occurring in OpenTelemetry based solutions, given it is the future for instrumentation.

antymon4o: @cijothomas thank you for you comments!
I understand that the focus and the future is in the OpenTelemetry.
But since the changes in this PR are only on targeted frameworks and nuget packages, and there is no behavior change, I can not understand what resources or investments are required to review and complete this PR. The work is completed. It is not needed for someone else to do/code something more.

Is it possible at least for the PR checks to be run just to see if the builds and the test are passing on GitHub too?

agehrke:
Any chance this PR and the related issue could get prioritized?

PR #2896, opened 2024-08-12
[SDL] split codeql workflows

TimothyMothra:
This PR splits the current CodeQL workflow into two separate workflows for each language; csharp and javascript.


TimothyMothra:
Good news, both csharp and javascript are running in this PR.
There is one error about this not running on the main branch, but that will be resolved after this merges :)

rajkumar-rangaraj (approved):

PR #2909, opened 2024-10-23
add net9 to test framework

TimothyMothra:


WEB/Src/WindowsServer/WindowsServer.Tests/AppServicesHeartbeatTelemetryModuleTests.cs
```diff
@@ -143,7 +143,7 @@ private void RemoveTestEnvVarsAssociatedToModule(AppServicesHeartbeatTelemetryMo
         {
             foreach (var kvp in appServicesHbeatModule.WebHeartbeatPropertyNameEnvVarMap)
             {
-                Environment.SetEnvironmentVariable(kvp.Value, string.Empty);
+                Environment.SetEnvironmentVariable(kvp.Value, null);
```
TimothyMothra, line 146: seems like there was a runtime change.
NET8 and prior expected this call to delete the Environment Variable.
One of the tests was asserting that the Environment Variable was empty.

Starting in NET9, this was setting the Environment Variable to string.Empty, causing the same test to fail.

rajkumar-rangaraj (approved):

PR #2958, opened 2025-02-18
fix compiler issue re: `Reverse()` method in Test project.

TimothyMothra:
Discovered a new issue affecting test projects today. 

> CS0023	Operator '.' cannot be applied to operand of type 'void'

The problems appears to be with the use of the `Reverse()` method.
In each of the use cases, the `Reverse()` method is being called on an array of `TelemetryDocument` objects.
Each of the use cases are expecting to use the [Linq Reverse method](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.reverse?view=net-9.0), which returns an IEnumerable and can be used to chain multiple methods.
Instead, the compiler has started selecting the [Array Reverse method](https://learn.microsoft.com/en-us/dotnet/api/system.array.reverse?view=net-9.0) which returns a void which causes the above error.

I don't know what has caused the sudden breaking change. The last successful build was 2 weeks ago and was using .NET 8.0.12. Today's CI is using 8.0.13.

This PR fixes the issue by changing the order in which Reverse() is called to ensure it's being called on an Enumerable instead of on the Array type.

### DIFF
--- a/WEB/Src/PerformanceCollector/Perf.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
+++ b/WEB/Src/PerformanceCollector/Perf.Tests/QuickPulse/QuickPulseTelemetryProcessorTests.cs
@@ -643,7 +643,7 @@
             telemetryProcessor.Process(trace);
 
             // ASSERT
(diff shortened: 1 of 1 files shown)


rajkumar-rangaraj (approved):
Please check copilot comments before the merge.

PR #2998, opened 2025-10-06
Track Trace

harsimar:
Implementation for TrackTrace.

### DIFF
--- a/BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
@@ -4,10 +4,9 @@
     using System.Collections.Generic;
     using System.ComponentModel;
     using System.Diagnostics;
-    using System.Globalization;
+    using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
-    using Azure.Core;
     using Microsoft.ApplicationInsights.Channel;
     using Microsoft.ApplicationInsights.DataContracts;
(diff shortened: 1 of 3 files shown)


BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -202,7 +204,10 @@ public void TrackTrace(string message, SeverityLevel severityLevel, IDictionary<
         {
             if (properties != null && properties.Count > 0)
             {
-                using (this.logger.BeginScope(properties))
+                List<KeyValuePair<string, object>> scope = properties
```
rajkumar-rangaraj: Is there a perf-efficient way to cover this one instead of using LINQ?

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -878,29 +872,70 @@ internal void TrackPageView(PageViewTelemetry telemetry)
...
+        }
+
+        private readonly struct DictionaryLogState : IReadOnlyList<KeyValuePair<string, object>>
+        {
+            public readonly string Message;
+            private readonly IDictionary<string, string> properties;
+
+            public DictionaryLogState(IDictionary<string, string> properties, string message)
+            {
+                this.properties = properties;
```
rajkumar-rangaraj:
```suggestion
               this.properties = properties ?? new Dictionary<string, string>();
```

PR #3004, opened 2025-10-23
TrackException

harsimar:


BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -323,19 +323,19 @@ public void TrackException(Exception exception, IDictionary<string, string> prop
...
             }
 
-            var telemetry = new ExceptionTelemetry(exception);
-
-            if (properties != null && properties.Count > 0)
+            IDictionary<string, string> allProperties = new Dictionary<string, string>();
+            if (properties != null)
             {
-                // Utils.CopyDictionary(properties, telemetry.Properties);
+                Utils.CopyDictionary(properties, allProperties);
```
rajkumar-rangaraj: Is it possible to convert directly to DictionaryLogState, rather than doing it here and then passing it again to DictionaryLogState?

examples/BasicConsoleApp/Program.cs
```diff
@@ -15,7 +15,7 @@ static void Main(string[] args)
         {
             var telemetryConfig = new TelemetryConfiguration
             {
```
rajkumar-rangaraj, line 17: Please remove this change and delete your resource.

PR #3006, opened 2025-10-24
TrackEvent

harsimar:


BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -167,7 +166,7 @@ public void TrackEvent(EventTelemetry telemetry)
                 telemetry = new EventTelemetry();
```
rajkumar-rangaraj: We should remove this instance creation of EventTelemetry()

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
```diff
@@ -134,22 +134,21 @@ public bool IsEnabled()
...
-        public void TrackEvent(string eventName, IDictionary<string, string> properties = null, IDictionary<string, double> metrics = null)
+        public void TrackEvent(string eventName, IDictionary<string, string> properties = null)
         {
-            var telemetry = new EventTelemetry(eventName);
-
-            if (properties != null && properties.Count > 0)
+            if (string.IsNullOrEmpty(eventName))
             {
-                Utils.CopyDictionary(properties, telemetry.Properties);
+                eventName = "CustomEventDefaultName";
```
rajkumar-rangaraj: The AI SDK has never permitted an empty event name, so we don’t need to account for this scenario. If eventName is null or empty, we can log it to the event source and return. 

Since the event source isn’t fully set up yet, you can add a TODO to log to the event source.

BASE/src/Microsoft.ApplicationInsights/TelemetryClient.cs
harsimar: Do you mean to populate that instance creation with an event name, or to remove the if statement and the code inside it entirely?

rajkumar-rangaraj: same comment as above https://github.com/microsoft/ApplicationInsights-dotnet/pull/3006/files#r2461573984 . If null log to event source and return. No need to create new instance of EventTelemetry.

PR #3005, opened 2025-10-24
NLog ApplicationInsightsTarget FlushAsync

snakefoot:
## Changes
- Improves application-shutdown behavior by changing from synchronous Flush() to the recommended asynchronous FlushAsync()

Followup to #3001

PR #3013, opened 2025-11-09
NLog ApplicationInsightsTarget FlushAsync

snakefoot:
## Changes
- Improves application-shutdown behavior by changing from synchronous Flush() to the recommended asynchronous FlushAsync()

Followup to #3001 - Restores #3005 that was reverted by #3009

### DIFF
--- a/LOGGING/src/NLogTarget/ApplicationInsightsTarget.cs
+++ b/LOGGING/src/NLogTarget/ApplicationInsightsTarget.cs
@@ -29,9 +29,7 @@ namespace Microsoft.ApplicationInsights.NLogTarget
 
         private TelemetryClient telemetryClient;
         private TelemetryConfiguration telemetryConfiguration;
-        private DateTime lastLogEventTime;
         private NLog.Layouts.Layout connectionStringLayout = string.Empty;
-        private bool initializationFailed;
 
         /// <summary>
         /// Initializers a new instance of ApplicationInsightsTarget type.
@@ -68,7 +66,6 @@ namespace Microsoft.ApplicationInsights.NLogTarget
         internal void BuildPropertyBag(LogEventInfo logEvent, ITelemetry trace)
         {
             trace.Timestamp = logEvent.TimeStamp;
-            trace.Sequence = logEvent.SequenceID.ToString(CultureInfo.InvariantCulture);
 
             IDictionary<string, string> propertyBag;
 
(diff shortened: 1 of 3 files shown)


snakefoot:
@rajkumar-rangaraj + @harsimar Ready for review and merge.

LOGGING/src/NLogTarget/ApplicationInsightsTarget.cs
```diff
@@ -201,25 +193,13 @@ protected override void FlushAsync(AsyncContinuation asyncContinuation)
...
                 {
-                    // Nothing has been written, so nothing to wait for
                     asyncContinuation(null);
                 }
                 else
                 {
-                    // Documentation says it is important to wait after flush, else nothing will happen
-                    // https://docs.microsoft.com/azure/application-insights/app-insights-api-custom-events-metrics#flushing-data
-                    System.Threading.Tasks.Task.Delay(TimeSpan.FromMilliseconds(500)).ContinueWith((task) => asyncContinuation(null));
+                    this.telemetryClient.FlushAsync(System.Threading.CancellationToken.None).ContinueWith(t => asyncContinuation(t.Exception));
```
rajkumar-rangaraj, line 202: FlushAsync is not ready yet to take this change

snakefoot: Can you explain why [TelemetryClient.FlushAsync](https://learn.microsoft.com/dotnet/api/microsoft.applicationinsights.telemetryclient.flushasync) doesn't work and should be avoided? Maybe the API docs should be updated ?

rajkumar-rangaraj: The shimmed version does not include the updated FlushAsync yet. We need to address FlushAsync first before adding this feature. I'm not sure if you've tested the changes with a sample app. Please use NLogConsoleApp from the examples folder (Develop branch) to confirm it works, as it currently disrupts the data flow. https://github.com/microsoft/ApplicationInsights-dotnet/tree/develop/examples/NLogConsoleApp

rajkumar-rangaraj: Use https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/SHIM/Microsoft.ApplicationInsights.Shim.sln to load the projects

snakefoot: If `FlushAsync()` has no valid implementation in the SHIM, then I guess it doesn't make sense to use it. But the console application runs without issue when calling `FlushAsync`.

snakefoot:
Now see what you mean:
```csharp
        public void Flush()
        {
            CoreEventSource.Log.TelemetlyClientFlush();

            // Force flush all providers
            this.sdk.TracerProvider?.ForceFlush();
            this.sdk.MeterProvider?.ForceFlush();
            this.sdk.LoggerProvider?.ForceFlush();
        }

        public Task<bool> FlushAsync(CancellationToken cancellationToken)
        {
            return cancellationToken.IsCancellationRequested ? TaskEx.FromCanceled<bool>(cancellationToken) : Task.FromResult(false);
        }
```
Maybe consider updating `FlushAsync` from being no-operation to Task.Run that calls `Flush()` in the SHIM-implementation.

PR #3017, opened 2025-11-12
[Shim] TrackMetric, GetMetric & TrackValue with tests

rajkumar-rangaraj:
* Removed public api's from `MetricTelemetry` that does not have OpenTelemetry equivalent
* Shimmed TrackMetric, GetMetric & TrackValue with tests
* Updated example app

### DIFF
--- a/BASE/src/Microsoft.ApplicationInsights/Metric.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Metric.cs
@@ -76,8 +76,8 @@
         /// <param name="metricValue">The value to be aggregated.</param>
         /// <param name="dimension1Value">The value of the 1st dimension.</param>
         /// <returns><c>True</c> if the specified value was added to the <c>MetricSeries</c> indicated by the specified dimension name;
-        /// <c>False</c> if the indicated series could not be created because a dimension cap or a metric series cap was reached.</returns>
-        /// <exception cref="ArgumentException">If the number of specified dimension names does not match the dimensionality of this <c>Metric</c>.</exception>
+        /// <c>False</c> if the indicated series could not be created because a dimension cap or a metric series cap was reached,
+        /// or if the number of specified dimension values does not match the dimensionality of this <c>Metric</c>.</returns>
         public bool TrackValue(double metricValue, string dimension1Value)
         {
             if (this.dimensionNames.Length != 1)
@@ -107,8 +107,8 @@
         /// <param name="metricValue">The value to be aggregated.</param>
         /// <param name="dimension1Value">The value of the 1st dimension.</param>
         /// <returns><c>True</c> if the specified value was added to the <c>MetricSeries</c> indicated by the specified dimension name;
-        /// <c>False</c> if the indicated series could not be created because a dimension cap or a metric series cap was reached.</returns>
-        /// <exception cref="ArgumentException">If the number of specified dimension names does not match the dimensionality of this <c>Metric</c>.</exception>
+        /// <c>False</c> if the indicated series could not be created because a dimension cap or a metric series cap was reached,
+        /// or if the number of specified dimension values does not match the dimensionality of this <c>Metric</c>.</returns>
         public bool TrackValue(object metricValue, string dimension1Value)
         {
             if (metricValue == null)
@@ -133,8 +133,8 @@
         /// <param name="dimension1Value">The value of the 1st dimension.</param>
         /// <param name="dimension2Value">The value of the 2nd dimension.</param>
         /// <returns><c>True</c> if the specified value was added to the <c>MetricSeries</c> indicated by the specified dimension name;
-        /// <c>False</c> if the indicated series could not be created because a dimension cap or a metric series cap was reached.</returns>
-        /// <exception cref="ArgumentException">If the number of specified dimension names does not match the dimensionality of this <c>Metric</c>.</exception>
+        /// <c>False</c> if the indicated series could not be created because a dimension cap or a metric series cap was reached,
+        /// or if the number of specified dimension values does not match the dimensionality of this <c>Metric</c>.</returns>
         public bool TrackValue(double metricValue, string dimension1Value, string dimension2Value)
         {
             if (this.dimensionNames.Length != 2)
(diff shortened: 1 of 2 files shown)


BASE/src/Microsoft.ApplicationInsights/Metric.cs
```diff
@@ -0,0 +1,877 @@
...
+        /// <param name="dimension1Value">The value of the 1st dimension.</param>
+        /// <returns><c>True</c> if the specified value was added to the <c>MetricSeries</c> indicated by the specified dimension name;
+        /// <c>False</c> if the indicated series could not be created because a dimension cap or a metric series cap was reached.</returns>
+        /// <exception cref="ArgumentException">If the number of specified dimension names does not match the dimensionality of this <c>Metric</c>.</exception>
+        public bool TrackValue(double metricValue, string dimension1Value)
+        {
+            if (this.dimensionNames.Length != 1)
+            {
+                // TODO: Log exception
+                return false;
```
xiang17, line 86: Should throw an exception?

BASE/src/Microsoft.ApplicationInsights/Metric.cs
```diff
@@ -0,0 +1,877 @@
...
+        /// <summary>
+        /// Tracks the specified value using the <c>MetricSeries</c> associated with the specified dimension value.<br />
+        /// An aggregate representing tracked values will be automatically sent to the cloud ingestion endpoint at the end of each aggregation period.
+        /// This overload may only be used with 1-dimensional metrics. Use other overloads to specify a matching number of dimension values for this metric.
+        /// </summary>
+        /// <param name="metricValue">The value to be aggregated.</param>
+        /// <param name="dimension1Value">The value of the 1st dimension.</param>
+        /// <returns><c>True</c> if the specified value was added to the <c>MetricSeries</c> indicated by the specified dimension name;
+        /// <c>False</c> if the indicated series could not be created because a dimension cap or a metric series cap was reached.</returns>
+        /// <exception cref="ArgumentException">If the number of specified dimension names does not match the dimensionality of this <c>Metric</c>.</exception>
```
xiang17: sync comment with code

BASE/src/Microsoft.ApplicationInsights/Metric.cs
```diff
@@ -0,0 +1,877 @@
...
```
xiang17: same as above for here and rest of the file

BASE/src/Microsoft.ApplicationInsights/Metric.cs
```diff
@@ -0,0 +1,877 @@
...
+        /// <param name="dimension1Value">The value of the 1st dimension.</param>
+        /// <param name="dimension2Value">The value of the 2nd dimension.</param>
+        /// <returns><c>True</c> if the specified value was added to the <c>MetricSeries</c> indicated by the specified dimension name;
+        /// <c>False</c> if the indicated series could not be created because a dimension cap or a metric series cap was reached.</returns>
+        /// <exception cref="ArgumentException">If the number of specified dimension names does not match the dimensionality of this <c>Metric</c>.</exception>
+        public bool TrackValue(double metricValue, string dimension1Value, string dimension2Value)
+        {
+            if (this.dimensionNames.Length != 2)
+            {
+                // throw new ArgumentException("This metric expects 2 dimension values.");
```
xiang17, line 142: Why is it commented out?

BASE/src/Microsoft.ApplicationInsights/Metrics/MetricsManager.cs
```diff
@@ -0,0 +1,52 @@
...
+            return this.histograms.GetOrAdd(key, _ => 
+                this.meter.CreateHistogram<double>(key));
+        }
+
+        /// <summary>
+        /// Disposes the meter and all associated instruments.
+        /// </summary>
+        public void Dispose()
+        {
+            this.meter?.Dispose();
```
xiang17:
```suggestion
            this.meter.Dispose();
```

Looks like this is never null.

BASE/src/Microsoft.ApplicationInsights/Metrics/MetricIdentifier.cs
```diff
@@ -0,0 +1,667 @@
...
+            if (partValue == null)
+            {
+                throw new ArgumentNullException(partName);
+            }
+
+            if (allowEmpty)
+            {
+                if (partValue.Length > 0 && String.IsNullOrWhiteSpace(partValue))
+                {
+                    throw new ArgumentException(Invariant($"{partName} may not be non-empty, but whitespace-only."));
```
xiang17, line 57:
```suggestion
                    throw new ArgumentException(Invariant($"{partName} may not be non-empty and whitespace-only."));
```

BASE/src/Microsoft.ApplicationInsights/Metric.cs
rajkumar-rangaraj: Telemetry SDKs should not throw exceptions after initialization.

BASE/src/Microsoft.ApplicationInsights/Metrics/MetricsManager.cs
rajkumar-rangaraj: we need to enable nullable and fix all of this pattern seperately.

BASE/src/Microsoft.ApplicationInsights/Metrics/MetricIdentifier.cs
rajkumar-rangaraj: pre-existing code, did not change anything in it.

BASE/src/Microsoft.ApplicationInsights/Metric.cs
rajkumar-rangaraj: check original implementation - https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/BASE/src/Microsoft.ApplicationInsights/Metric.cs

BASE/src/Microsoft.ApplicationInsights/Metric.cs
rajkumar-rangaraj: to move the message to log.

BASE/src/Microsoft.ApplicationInsights/Metric.cs
rajkumar-rangaraj: will check this carefully

BASE/src/Microsoft.ApplicationInsights/Metric.cs
xiang17: Every method in original implementation calls `TryGetDataSeries` which throws exception https://github.com/microsoft/ApplicationInsights-dotnet/blob/98bc6c540e2dcccc78e4b356cd70e03d146a01ad/BASE/src/Microsoft.ApplicationInsights/Metric.cs#L448 (on latest main branch)

BASE/src/Microsoft.ApplicationInsights/Metrics/MetricsManager.cs
xiang17: A few lines above used `meter.CreateHistogram`. At least keep these two in sync.

Even if nullable is enabled, it looks like this should never be null. (`meter` is readonly)

BASE/src/Microsoft.ApplicationInsights/Metric.cs
rajkumar-rangaraj: This is because both the metric collector and exporter belong to the same project. In our situation, the shim relies on the OpenTelemetry implementation, so throwing isn't an option.

BASE/src/Microsoft.ApplicationInsights/Metric.cs
rajkumar-rangaraj: Fixed, please check

xiang17 (approved):

PR #3018, opened 2025-11-17
[Shim] .NET Framework - Support ConnStr from ApplicationInsights.config

rajkumar-rangaraj:


WEB/Src/Web/Web/Implementation/ApplicationInsightsConfigurationReader.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <param name="configPath">The path to the config file.</param>
+        /// <returns>The connection string if found; otherwise, null.</returns>
+        private static string ReadConnectionStringFromConfig(string configPath)
+        {
+            XDocument xml = XDocument.Load(configPath);
+            XElement root = xml.Element(XmlNamespace + "ApplicationInsights");
+            
+            if (root != null)
+            {
+                XElement connectionStringElement = root.Element(XmlNamespace + "ConnectionString");
```
xiang17, line 97: Is it tested?

rajkumar-rangaraj: Yes, in a local app. We don't have tests for Web.csproj yet.

PR #3022, opened 2025-11-17
[Shim] TrackTrace and TrackEvent tests

rajkumar-rangaraj:


BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/TelemetryClientTest.cs
```diff
@@ -60,9 +393,6 @@ public void TrackMetricWithNameAndValue()
             this.telemetryClient.Flush();
             
             // Wait briefly for metrics to be exported
-            System.Threading.Thread.Sleep(200);
```
xiang17, line 63: nit: looks like the comment above should be removed, instead of the one below

xiang17 (approved):

PR #3034, opened 2025-12-02
[Shim] Cleanup unused usings, address logging todo's

rajkumar-rangaraj:


BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -191,7 +190,7 @@ public void Dispose()
         /// <exception cref="ArgumentException">An ArgumentException is thrown if the provided object does not inherit Azure.Core.TokenCredential.</exception>
 #pragma warning disable CA1822 // Mark members as static
 #pragma warning disable CA1801 // Review unused parameters
-        public void SetAzureTokenCredential(object tokenCredential)
+        internal void SetAzureTokenCredential(object tokenCredential)
```
harsimar, line 193: Is there an intent to set this to public in a future iteration?

rajkumar-rangaraj: correct, the shape of the api will change to  `SetAzureTokenCredential(TokenCredential tokenCredential)`

rajkumar-rangaraj:
> Generally looks good, seems like there are a lot of warnings regarding the System.Exception being too broad - if this is intentional maybe we could look into silencing these kinds of warnings.

That message keeps showing up in PRs. We should consider adding pragma disable to remove it. It's not related to this PR, but we can address it separately.

Issue #3058, opened 2025-12-19
Exporter Options Configuration for Microsoft.ApplicationInsights module 3.x


harsimar:
See the parent issue. In the base module, configuration for exporter properties is likely to be added in TelemetryConfiguration.cs.

PR #3065, opened 2025-12-31
Exporter options for base module

harsimar:
This adds some public apis to configure AzureMonitorExporterOptions from  @Microsoft.ApplicationInsights 3.x

See [issue](https://github.com/microsoft/ApplicationInsights-dotnet/issues/3058)

### DIFF
--- a/BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
+++ b/BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
@@ -43,7 +43,7 @@
         // Exporter options
         private float? samplingRatio;
         private double? tracesPerSecond;
-        private string? storageDirectory;
+        private string storageDirectory;
         private bool? disableOfflineStorage;
         private bool? enableLiveMetrics;
         private bool? enableTraceBasedLogsSampler;
@@ -150,7 +150,7 @@
         /// Gets or sets the directory for offline telemetry storage.
         /// When null, the Azure Monitor Exporter default (system default location) is used.
         /// </summary>
-        public string? StorageDirectory
+        public string StorageDirectory
         {
             get => this.storageDirectory;
             set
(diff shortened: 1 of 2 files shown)


BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/TelemetryConfigurationSetExporterOptionsTests.cs
```diff
@@ -0,0 +1,350 @@
...
+    /// <summary>
+    /// Tests for Azure Monitor Exporter configuration methods in TelemetryConfiguration.
+    /// </summary>
+    public class TelemetryConfigurationSetExporterOptionsTests : IDisposable
+    {
+        private TelemetryConfiguration telemetryConfiguration;
+
+        public TelemetryConfigurationSetExporterOptionsTests()
+        {
+            this.telemetryConfiguration = new TelemetryConfiguration();
```
rajkumar-rangaraj: we should use `TelemetryConfiguration.CreateDefault()`

BASE/Test/Microsoft.ApplicationInsights.Test/Microsoft.ApplicationInsights.Tests/TelemetryConfigurationSetExporterOptionsTests.cs
```diff
@@ -0,0 +1,350 @@
...
+        #region SetSamplingRatio Tests
+
+        [Fact]
+        public void SetSamplingRatio_WithValidRatio_SetsSamplingRatioInExporterOptions()
+        {
+            // Arrange
+            this.telemetryConfiguration.ConnectionString = "InstrumentationKey=test-ikey";
+
+            // Act
+            this.telemetryConfiguration.SetSamplingRatio(0.5F);
```
rajkumar-rangaraj: Could you set all six public APIs (introduced in this PR) in a single test and verify that their values are correctly applied? We need to make sure that the builder.Services.Configure<AzureMonitorExporterOptions> call does not overwrite them with default values.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -40,6 +40,14 @@ public sealed class TelemetryConfiguration : IDisposable
         private bool isBuilt = false;
         private bool isDisposed = false;
 
+        // Exporter options
+        private float samplingRatio = 1.0f;
```
rajkumar-rangaraj: We shouldn't manage defaults here. Instead, we could define these as nullable, and if they are null, we simply ignore mapping those properties to AzMon properties.

BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs
```diff
@@ -109,6 +117,91 @@ public string ConnectionString
             }
         }
 
+        /// <summary>
+        /// Gets or sets the sampling ratio for traces (0.0 to 1.0).
+        /// A value of 1.0 means all telemetry is sent, 0.5 means 50% is sent.
+        /// Default is 1.0 (all telemetry sent).
+        /// </summary>
+        public float SamplingRatio
```
rajkumar-rangaraj:
```suggestion
        public float? SamplingRatio
        {
            get => this.samplingRatio;
            set
            {
                this.ThrowIfBuilt();
                this.samplingRatio = value;
            }
        }
```

PR #3068, opened 2026-01-06
Integrate Application Insights Service Options / Cleanup

rajkumar-rangaraj:
This PR integrates Application Insights Service Options with the Azure Monitor Exporter (OpenTelemetry-based) telemetry collection and performs general cleanup of legacy properties. The primary goal is to ensure that the `EnableDependencyTrackingTelemetryModule` and `EnableRequestTrackingTelemetryModule` settings in `ApplicationInsightsServiceOptions` are respected when using the OpenTelemetry exporter.

**Key Changes**
Introduced ActivityFilterProcessor: A new [OpenTelemetry processor]() that filters activities based on service configuration. It suppresses export for:

Client, Internal, and Producer activities when dependency tracking is disabled.

Server and Consumer activities when request tracking is disabled.

Reflection-based Configuration: Added TrySetInternalProperty to safely configure standard metrics and performance counters in the [Azure Monitor Exporter]() without breaking compatibility.

API Cleanup: Removed properties from ApplicationInsightsServiceOptions, `EndpointAddress` and `EnableActiveTelemetryConfigurationSetup`.

### DIFF
--- a/BASE/README.md
+++ b/BASE/README.md
@@ -59,7 +59,7 @@ Select the option that best describes your situation:
   - [Tracking Page Views](#tracking-page-views)
 - [Configuration](#configuration)
   - [TelemetryConfiguration](#telemetryconfiguration)
-  - [Setting Context Properties](#setting-context-properties)
+- [Setting Context Properties](#setting-context-properties)
   - [Dependency Injection](#dependency-injection)
 - [Advanced Scenarios](#advanced-scenarios)
   - [Enriching Telemetry with Activity Processors](#enriching-telemetry-with-activity-processors)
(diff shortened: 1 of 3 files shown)


NETCORE/test/Shared/ActivityFilterProcessorTests.cs
```diff
@@ -0,0 +1,280 @@
...
+                Assert.NotNull(activity);
+            }
+
+            // ASSERT
+            Assert.True(activity.IsAllDataRequested, "Client activity should NOT be filtered when EnableDependencyTrackingTelemetryModule is true");
+        }
+
+#if AI_ASPNETCORE_WEB
+        [Fact]
+        public void WhenBothEnabled_DoesNotFilterServerActivities()
```
harsimar: Naming of this test is confusing me a little. Seems it is testing the request tracking but mentions "both" in name.

rajkumar-rangaraj: Fixed.

PR #3084, opened 2026-01-23
Add DisableTelemetry support using OpenTelemetry OTEL_SDK_DISABLED

rajkumar-rangaraj:
## Summary
Adds support for disabling telemetry via `TelemetryConfiguration.DisableTelemetry` property by leveraging OpenTelemetry SDK 1.15.0's `OTEL_SDK_DISABLED` environment variable.

### Core Implementation
- **TelemetryConfiguration.Build()**: When `DisableTelemetry` is `true`, sets `OTEL_SDK_DISABLED=true` environment variable before creating the OpenTelemetry SDK, causing it to return no-op implementations

### DI Support (AspNetCore/WorkerService)
- Updated `AddTelemetryConfigAndClient()` to apply `IConfigureOptions<TelemetryConfiguration>` and `IPostConfigureOptions<TelemetryConfiguration>` callbacks, enabling `services.Configure<TelemetryConfiguration>()` pattern
- Added `TelemetryConfiguration` injection in `AzureMonitorExporterOptions.Configure` to set `OTEL_SDK_DISABLED` in IConfiguration when `DisableTelemetry` is true

### Tests
- Added `TelemetryConfigurationOtelSdkDisabledTests` with 2 tests validating the feature for non-DI scenarios
- Tests use `[Collection("TelemetryClientTests")]` to prevent parallel execution issues with environment variable manipulation

### Documentation
- Updated `docs/concepts.md` with usage examples for both direct and DI scenarios
- Updated `CHANGELOG.md`

## Usage

**Non-DI:**
```csharp
var configuration = new TelemetryConfiguration
{
    ConnectionString = "...",
    DisableTelemetry = true
};
var tc = new TelemetryClient(configuration);
```

**DI (ASP.NET Core / Worker Service):**
```
// Must be called BEFORE AddApplicationInsightsTelemetry()
services.Configure<TelemetryConfiguration>(tc => tc.DisableTelemetry = true);
services.AddApplicationInsightsTelemetry();
```

**Notes**
* Requires OpenTelemetry SDK 1.15.0+ which supports OTEL_SDK_DISABLED
* For DI scenarios, DisableTelemetry must be configured before AddApplicationInsightsTelemetry() is called

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -65,11 +65,9 @@ Please review our full guide on [Sampling in Application Insights](https://docs.
 You can disable all telemetry collection using the `DisableTelemetry` property:
 
 ```C#
-var configuration = new TelemetryConfiguration
-{
-    ConnectionString = "InstrumentationKey=00000000-0000-0000-0000-000000000000",
-    DisableTelemetry = true
-};
+var configuration = TelemetryConfiguration.CreateDefault();
+configuration.ConnectionString = "InstrumentationKey=00000000-0000-0000-0000-000000000000";
+configuration.DisableTelemetry = true;
 var tc = new TelemetryClient(configuration);
 ```
 


docs/concepts.md
```diff
@@ -58,3 +58,38 @@ Please review our full guide on [Telemetry correlation in Application Insights](
...
 
 Please review our full guide on [Sampling in Application Insights](https://docs.microsoft.com/azure/azure-monitor/app/sampling).
+
+
+## Disabling Telemetry
+
+You can disable all telemetry collection using the `DisableTelemetry` property:
+
+```C#
+var configuration = new TelemetryConfiguration
```
harsimar: TelemetryConfiguration.CreateDefault()

NETCORE/src/Shared/Extensions/ApplicationInsightsExtensions.cs
```diff
@@ -105,6 +106,19 @@ private static void AddTelemetryConfigAndClient(IServiceCollection services)
                 {
                     configuration.ConnectionString = options.ConnectionString;
                 }
+
+                // Apply any Configure<TelemetryConfiguration> callbacks
+                var configureOptions = provider.GetServices<IConfigureOptions<TelemetryConfiguration>>();
```
harsimar, line 111: It seems like this would also make it possible to set other TelemetryConfiguration options (such as sampling etc) - do you think it would make things more complex? Does it make sense to add a DisableTelemetry to ApplicationInsightsServiceOptions if not already present and not add the extra provider.getservices for TelemetryConfig?

docs/concepts.md
rajkumar-rangaraj: Fixed

NETCORE/src/Shared/Extensions/ApplicationInsightsExtensions.cs
rajkumar-rangaraj: Lets keep simple by introducing those properties in ApplicationInsightsServiceOptions

PR #3091, opened 2026-01-28
Remove netstandard2.0 target framework from core packages

rajkumar-rangaraj:
Removes netstandard2.0 from the following packages:

* Microsoft.ApplicationInsights
* Microsoft.ApplicationInsights.AspNetCore
* Microsoft.ApplicationInsights.WorkerService

These packages now target net8.0, net9.0, net10.0 (and net462 for the core SDK).


harsimar (approved):

Saibamen:
Why it is better to target direct versions instead of netstandard?

rajkumar-rangaraj:
> Why it is better to target direct versions instead of netstandard?

please review - https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0#net-standard-problems

That being said, given the customer request and the fact that many netstandard apps have referenced it, we will include netstandard 2.0 in the next release in a couple of weeks.

Issue #3111, opened 2026-02-10
ApplicationVersion is ignored in v3


ArturDorochowicz:
* List of [NuGet packages](https://github.com/microsoft/ApplicationInsights-dotnet#nuget-packages) and version that you are using:
  * Microsoft.ApplicationInsights.AspNetCore 3.0.0
* Runtime version (e.g. net461, net48, netcoreapp2.1, netcoreapp3.1, etc. You can find this information from the `*.csproj` file):
  * net10.0 
* Hosting environment (e.g. Azure Web App, App Service on Linux, Windows, Ubuntu, etc.):
  - Windows

### Describe the bug

In v3, the value set on `ApplicationInsightsServiceOptions.ApplicationVersion` is ignored and not sent with telemetry.

### To Reproduce

Just do `services.AddApplicationInsightsTelemetry()`. This should be typically sending "1.0.0" with the telemetry. 

https://github.com/microsoft/ApplicationInsights-dotnet/blob/3845c4079a07175ee57800d16d98c80693a76101/NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs#L48

The version is not sent. This can be verified in the Azure Portal.

rajkumar-rangaraj:
Thanks for reporting this! You're absolutely right - this needs to be fixed.

PR #3124, opened 2026-02-20
Add ApplicationVersionResourceDetector for service.version

rajkumar-rangaraj:
Fix Issue #3111 .

## Changes

Introduced ApplicationVersionResourceDetector to set the application version as the service.version resource attribute in OpenTelemetry. Registered the detector in both ASP.NET Core and WorkerService ApplicationInsights extension flows. Added unit tests to verify correct behavior for valid, null, and empty version values.

### Checklist

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -2,24 +2,44 @@
 # Github Actions Stale: https://github.com/actions/stale
 # Description: The purpose of this action is to close stale issues and PRs.
 
-name: Stale Issues and PRs
+name: "Manage stale issues and pull requests"
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 0 * * *' # End of every day
+    - cron: "37 5 * * *" # 5:37 AM UTC daily, arbitrary time to avoid GitHub API load spikes
+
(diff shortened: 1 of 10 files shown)


rajkumar-rangaraj:
> What happens if the customer does not explicitly set the version?

https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/NETCORE/src/Shared/Extensions/ApplicationInsightsServiceOptions.cs#L48

We have default set from the code using  `public string ApplicationVersion { get; set; } = Assembly.GetEntryAssembly()?.GetName().Version.ToString();` value of it will be 1.0.0.

> Please double check in the breaking changes & migration guidance docs that any references to ApplicationVersion or ComponentVersion are up to date with this change.

Will check this out.

rajkumar-rangaraj:
> Please double check in the breaking changes & migration guidance docs that any references to ApplicationVersion or ComponentVersion are up to date with this change.

A minor adjustment is required for ComponentVersion from the DI perspective. However, we should plan to address it after we resolve the TelemetryContext issue.

harsimar (approved):

Issue #3115, opened 2026-02-16
OpenTelemetrySdk not defined with TelemetryClient using DI when calling flushAsync


andrkn:
* List of [NuGet packages](https://github.com/microsoft/ApplicationInsights-dotnet#nuget-packages) and version that you are using:
```
<PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="3.0.0" />
```
* Runtime version (e.g. net461, net48, netcoreapp2.1, netcoreapp3.1, etc. You can find this information from the `*.csproj` file):
* Hosting environment (e.g. Azure Web App, App Service on Linux, Windows, Ubuntu, etc.):

net10.0, windows

### Describe the bug

In my WorkerService I have an issue when trying to use telemetryClient.Flush/FlushAsync when doing shutdown of the service. It crashes as _private OpenTelemetrySdk sdk;_ is null within my TelemetryClient.
Exception from TelemetryClient: System.NullReferenceException: 'Object reference not set to an instance of an object.', 

### To Reproduce

Attached code should reproduce my issue with using TelemetryClient.FlushAsync

[TestWorker.zip](https://github.com/user-attachments/files/25343208/TestWorker.zip)

rajkumar-rangaraj:
@andrkn Thanks for reporting this.

We’ve identified the root cause of the issue and confirmed the behavior you’re seeing. We’ll address this with a fix in an upcoming update.

PR #3125, opened 2026-02-20
Fix TelemetryClient.Flush/FlushAsync NullReferenceException in DI scenarios 

rajkumar-rangaraj:
Fix Issue #3115 .

## Changes

`TelemetryClient.Flush()` and `FlushAsync()` throw `System.NullReferenceException` when called from DI scenarios because `this.sdk` is null.

In DI scenarios, `TelemetryClient` is constructed with `isFromDependencyInjection: true`, which intentionally skips calling `TelemetryConfiguration.Build()` — the OTel SDK is owned by `AddOpenTelemetry()` in the DI container, not by `TelemetryConfiguration`. However, both `Flush()` and `FlushAsync()` unconditionally access `this.sdk.TracerProvider`, causing:

```
System.NullReferenceException: Object reference not set to an instance of an object.
at Microsoft.ApplicationInsights.TelemetryClient.<>c__DisplayClass36_0.<FlushAsync>b__0()
```
- Added an `IServiceProvider` field and a new internal constructor that accepts it
- Updated the DI `TelemetryClient` factory in `AddTelemetryConfigAndClient` to pass the `IServiceProvider`
- `Flush()` and `FlushAsync()` now check: if `this.sdk` is available (non-DI), flush via the SDK; otherwise if `this.serviceProvider` is available (DI), resolve `TracerProvider`/`MeterProvider`/`LoggerProvider` from the DI container and flush those

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -2,24 +2,44 @@
 # Github Actions Stale: https://github.com/actions/stale
 # Description: The purpose of this action is to close stale issues and PRs.
 
-name: Stale Issues and PRs
+name: "Manage stale issues and pull requests"
 on:
   workflow_dispatch:
   schedule:
-    - cron: '0 0 * * *' # End of every day
+    - cron: "37 5 * * *" # 5:37 AM UTC daily, arbitrary time to avoid GitHub API load spikes
+
+permissions:
+  contents: read
 
 jobs:
   stale:
+    if: github.event.repository.fork == false
+    permissions:
+      issues: write # for actions/stale to close stale issues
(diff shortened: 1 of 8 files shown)


rajkumar-rangaraj:
> How was this tested? Consider adding some automated testing

We don't have an automated infra built yet. It's outside the scope of the PR. We might need to consider building one.
I tested using the repro provided by customer.

maarten-kieft:
I'm running in this issue; what is the expected release date of this fix?

rajkumar-rangaraj:
> I'm running in this issue; what is the expected release date of this fix?

Within 2 weeks.

PR #3164, opened 2026-04-06
Fix Web package initialization issues

harsimar:
Post the 3.1.0 release, a classic aspnet application won't start up properly due to the prepend method throwing an exception. 
This PR removes that prepend method, but also refactors code for context processors to be set in a way that the exception doesn't throw.

Side considerations - 
- Connection string is settable via env var and config
- avoid null exceptions from internal feature tracking
- Testing to ensure this doesn't cause regression in other packages.


harsimar:
Closing as I will be opening another PR that uses a different approach to solve the issue.

